Issue #16, opened 2013-12-20
Remove Intel MKL dependency


rodrigob:
It is mentioned in the install instruction that this is work in progress.
While at ICCV I quickly implemented a branch where I remplace matrix operations with Eigen3 calls, and random generators by Boost::random generators.
I hope this is not redundant with ongoing work on private branches.

The branch can be found at
https://github.com/rodrigob/caffe

I got things to compile, however I noticed that some tests fails (thanks for creating a non-trivial set of unit tests !).
I have not been able to compile a version with MKL to compare, but I can only assume that tests should not fail.

Current fails are

[  FAILED  ] FlattenLayerTest/1.TestCPUGradient, where TypeParam = double
[  FAILED  ] StochasticPoolingLayerTest/0.TestGradientGPU, where TypeParam = float
[  FAILED  ] StochasticPoolingLayerTest/1.TestGradientGPU, where TypeParam = double
[  FAILED  ] MultinomialLogisticLossLayerTest/1.TestGradientCPU, where TypeParam = double

which all sounds nasty (gradient computation errors in neural networks, big no no).

I will spend some time inspecting to see what goes wrong there, but any suggestion/comment/idea is welcome.

PR #80, opened 2014-02-07
Add steps to install multi-threaded OpenBLAS on Ubuntu

### LINKED ISSUE #79: Support multithreading in the CPU mode of Solver::Solve
In each iteration of Solver::Solve, there are four chances to accelerate the computation.
The first opportunity is the most complex one since Net::ForwardBackward invokes the Forward and Backward of all the layers that comprise a net.

``` cpp
      Dtype loss = net_->ForwardBackward(bottom_vec);
```

The second chance is more straightforward. An OpenMP directive is enough to parallelize the independent computation for each param_id.

``` cpp
      ComputeUpdateValue();
```

The only extra trick that is needed to deal with the next occasion is to distinguish CPU and GPU mode.

``` cpp
      net_->Update();
```

The last one involves a plain old OpenMP friendly nested for loop.

``` cpp
      Test();
```

kloudkl:
Multi-threaded OpenBLAS makes a huge performance difference. The benchmarks with and without it in comments to #16 demonstrated more than 5 times speed-up for boost-eigen and MKL on a machine with 4 Hyper-Threading CPU cores (supporting 8 threads).

This fixes #79.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,6 +23,7 @@ We have also successfully compiled caffe on OS X 10.8.5, using homebrew-installe
 * CUDA
 * Boost
 * MKL (but see the boost-eigen branch for a boost/Eigen3 port)
+* A multi-threaded BLAS library such as OpenBLAS
 * OpenCV
 * glog, gflags, protobuf, leveldb, snappy
 * For the Python wrapper: python, numpy (>= 1.7 preferred), and boost_python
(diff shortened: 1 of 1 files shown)


jeffhammond:
This statement is categorically false: "it is known that ATLAS is inherently single-threaded."  ATLAS has been threaded for 5+ years

http://math-atlas.sourceforge.net/faq.html#tnum
http://math-atlas.sourceforge.net/timing/newThr395/index.html

PR #216, opened 2014-03-15
Implement the DeepFace net architecture achieving near human level face recognition performance

kloudkl:
Facebook claimed near human level face verification performance on the standard benchmark dataset LFW[1]. It took only 3 days to train on more than 4.4 million face images for about 15 epochs. In comparison, it took Caffe 9 to 10 days to train on the 1.2 million ILSVRC images for about 90 epochs. The former was about two times faster. It is not clear whether the so called GPU-based engine used to train the net consists of a single GPU or multiple GPUs. 

It used only two computationally expensive convolutional layers and three relatively cheaper locally connected layers. Unlike the convolutional layer, the locally connected layer used kernels without weight sharing. The details are not very clear in [1]. Its references [2] and [3] may give more clues.

The net parameters still have rough edges and the locally connected layer has yet to be implemented.

[1] Yaniv Taigman (Facebook), Ming Yang (Facebook), Marc'Aurelio Ranzato (Facebook), Lior Wolf (Tel Aviv University). DeepFace: Closing the Gap to Human-Level Performance in Face Verification. CVPR 2014, Columbus, Ohio.
[2] K. Gregor and Y. LeCun. Emergence of complex-like cells in a temporal product network with local receptive ﬁelds. arXiv:1006.0448, 2010.
[3] G. B. Huang, H. Lee, and E. Learned-Miller. Learning hierachical representations for face verifiation with convolutional deep bebief networks. In CVPR, 2012.

### DIFF
--- a/examples/deepface/deepface.prototxt
+++ b/examples/deepface/deepface.prototxt
@@ -1,0 +1,262 @@
+name: "DeepFaceNet"
+input: "data"
+input_dim: 10
+input_dim: 3
+input_dim: 227
+input_dim: 227
+layers {
+  layer {
+    name: "conv1"
+    type: "conv"
+    num_output: 32
+    kernelsize: 11
+    stride: 4
+    weight_filler {
+      type: "gaussian"
+      std: 0.01
+    }
+    bias_filler {
+      type: "constant"
+      value: 0.
+    }
+    blobs_lr: 1.
+    blobs_lr: 2.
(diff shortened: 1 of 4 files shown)


bamos:
If anybody's looking through caffe's issues for DeepFace-related content,
note that the network in this PR has some major differences with
the DeepFace architecture.

PR #258, opened 2014-03-25
Implement regularizers

kloudkl:
This PR replaces #113 to set the merging target to be the dev branch. Please refer to #113 for the discussions about the design decisions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,6 +220,10 @@ $(BUILD_DIR)/src/gtest/%.o: src/gtest/%.cpp
 	$(CXX) $< $(CXXFLAGS) -c -o $@
 	@echo
 
+$(BUILD_DIR)/src/$(PROJECT)/%.cuo: src/$(PROJECT)/%.cu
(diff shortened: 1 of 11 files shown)


longjon:
Closing as this is now out-of-date/abandoned and can be executed through less intrusive means, e.g., explicit loss layers or per-param regularization options.

zhaogengyan:
Hello @longjon , I have been searching for how to involve both L1 and L2 norm regularization on W in the cost function for the whole night, but got no answer. There are many people asking the same question in the User group, but there is nobody answering. I'm super confused on how to do the 

> explicit loss layers or per-param regularization options

mentioned by you.

Can Caffe involve both L1 and L2 regularization at the same time? And can Caffe involve layer-wise different regularizers (some layers get L1 and others get L2)? Thank you very much.

robmosh:
plus one interested here for a regularization layer

foolwood:
plus one interested here for a regularization layer

PR #308, opened 2014-04-08
DenseNet feature pyramid computation

moskewcz:
### explanatory tech report:
[DenseNet: Implementing Efficient ConvNet Descriptor Pyramids](http://arxiv.org/abs/1404.1869)
Forrest Iandola, Matt Moskewicz, Sergey Karayev, Ross Girshick, Trevor Darrell, and Kurt Keutzer

### additional integration notes (see main notes below inlined from the DENSENET_MERGE_TODO file)
- last rebase was a while ago now; there were some delays getting our tech report polished up
- in the mean time, some work on replacements for the GPL'd code is in progress; see:
 - BLF: https://github.com/moskewcz/boda/blob/master/src/blf_pack.cc
 - CNN padding/support/sizes calcs: https://github.com/moskewcz/boda/blob/master/src/conv_util.cc
 - hosting repo / framework: https://github.com/moskewcz/boda/

### inlined from: DENSENET_MERGE_TODO
list of the issues blocking the merge of the DenseNet feature branch:

- critical
 - replacement of GPL'd code (including removal from history)
 - update build process / Makefile to match current practice 
 - tests
 - trivial: remove DenseNet README.md header
 - trivial: remove this todo file

- unclear neccessity, semantics, and/or priority
 - general cleanup of commit sequence (probably mostly squashing)
 - input interface changes (i.e. jpeg filname as input -> ?)
 - output interface changes (?, but probably something: image support size, multiple layer output, alignment, etc.)
 - if still reading image files after any iface changes and removal of GPL code, use XX instead of YY


shelhamer:
Thanks Forrest @forresti and Matt @moskewcz for posting the code for your tech report. Since the PR has been made, others can find this snapshot of the implementation.

However, this cannot be merged. There are the steps helpfully listed by Matt, but more fundamentally there was and still is the goodness of fit. As implemented this was essentially done independent of the framework and then attached. This shows in the 5000+ line diff.

Caffe is capable of dense inference learning -- blobs can take the necessary shapes, convolution is convolution and matrix multiplications can be cast to convolution, and losses can take different input and truth shapes. (The [editing model parameters](http://nbviewer.ipython.org/github/BVLC/caffe/blob/master/examples/net_surgery.ipynb) example shows how to make a model fully convolutional for inference.)

The contribution of this PR is to assemble pyramids and show how to do extraction. This seems like it could be done the most clearly and effectively as an example instead of through further code. For instance (1)  the stitching approach could be turned into a data layer or (2) each level of the pyramid could be an input to a weight-shared pyramid net. At its simplest this could be a Python data layer that need not even be slow if one does multi-processing for prefetching.

As a pointer to dense models and matrix output follow #1698. Closing.

PR #330, opened 2014-04-16
MNIST autoencoder example

jeffdonahue:
This PR moves `examples/lenet/` to `examples/mnist/` and adds the necessary files to train an autoencoder on MNIST with the architecture of Hinton & Salakhutdinov [1] using SGD with no pre-training (e.g. via RBM).  It uses a sparse Gaussian initialization (added to `filler.hpp`) as suggested by [2] as a strategy for training autoencoders via SGD\* without pretraining.  It uses a fixed LR of 0.0001 which could probably be greatly improved upon, but I haven't played with it much.

After 2 million iterations (which took a few hours on the GPU -- I didn't originally intend to train it this long but this is what it was at when I came back to it) the test L2 reconstruction error was around 1.5-1.6.   (For an idea of what this means, a reconstruction with 2 out of 784 pixels flipped from perfectly white to perfectly black (or vice versa) would have an L2 reconstruction error of 2.0)

*actually [2] used Nesterov's accelerated gradient, but this is not currently implemented in Caffe and SGD seems to be fairly effective.

[1] http://www.cs.toronto.edu/~hinton/science.pdf

[2] http://jmlr.org/proceedings/papers/v28/sutskever13.pdf

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -106,7 +106,7 @@ message LayerParameter {
   // line above the enum. Update the next available ID when you add a new
   // LayerType.
   //
-  // LayerType next available ID: 27
+  // LayerType next available ID: 28
   enum LayerType {
     // "NONE" layer type is 0th enum element so that we don't cause confusion
     // by defaulting to an existent LayerType (instead, should usually error if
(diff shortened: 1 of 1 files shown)


moi90:
I'd be very happy about some comments in mnist_autoencoder.prototxt or a accompanying readme about how exactly the autoencoder works. In particular, the roles of the two distinct loss layers (and why the one operates on `flatdata` and `decode1` and the other on `flatdata` and `decode1neuron`) are not clear to me.

liuruoze:
I agree with @moi90 , a readme and tutorials would be very nice to one who wants to used them.
If @jeffdonahue  could give some advice, that would be great.

PR #355, opened 2014-04-24
Wrap up SyncedMem resize from @kloudkl; make train/test nets share data blobs

jeffdonahue:
This PR wraps up the SyncedMem changes from @kloudkl in #250 (thanks for doing most of the work on this @kloudkl!).  I ended up removing the thrust host/device_vector dependency as I couldn't get them working at all (almost all tests segfaulted), and instead kept the `void*` pointers that we were already using to store the data (and still syncing using explicit `cudaMemcpy` calls).  I added private fields `cpu_capacity_` and `gpu_capacity_` which keep track of the current space allocation, and methods `cpu_resize()` and `gpu_resize()` allocate additional memory (copying any current memory) if the current allocation is insufficient.

The final commit makes use of this new functionality and my `ShareData` and `ShareDiff` methods from #332 to share all the *data* blobs between the training and test net (#332 only shared weight blobs).  Because of this change I've expanded the imagenet_val batchsize from 50 to 250 (as close to the train batch size of 256 as possible while still evenly dividing 50K...) for slightly faster test time with little memory cost.

I ran some tests of the ImageNet model on a Tesla K40 and found little to no difference in training/test time (not counting the test time improvement due to the expanded batch size), while memory use after the first test iteration improves significantly:

```
test batch size 50:

    @ dev:
    before first test iteration: 3014 MB (GPU memory use based on 'nvidia-smi')
     after first test iteration: 3255 MB
    -> 241 MB additional memory used by test net

    @ capacity-aware-memory:
    before first test iteration: 3014 MB
     after first test iteration: 3053 MB
    -> 39 MB additional memory used by test net, 84% reduction vs. dev


test batch size 250:

    @ dev:
    before first test iteration: 3014 MB
     after first test iteration: 4143 MB
    -> 1129 MB additional memory used by test net

    @ capacity-aware-memory:
    before first test iteration: 3014 MB
     after first test iteration: 3141 MB
    -> 127 MB additional memory used by test net, 89% reduction vs. dev
```

Note that the memory use of the test net is non-zero, because additional memory is still allocated for the test net in at least these three cases:

(1) blobs used inside of layers to store intermediate computations (the solver has no way of knowing about these; would have to change each layer's code to share these),
(2) blobs in the test net which do not share a name with any blob in the train net (e.g. the `accuracy` top blob in the ImageNet test net -- which isn't a great example because it's only 8 bytes, but you get the idea)
(3) blobs in the test net which are larger (in terms of `count`) than correspondingly named blobs in the train net (there aren't any of these cases in any of Caffe's sample models, but it would obviously happen if you, e.g., used a larger batch size in the test net, or did something weird like swapping the names of `conv1` and `conv5` between the train and test net, etc.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -331,7 +331,7 @@ $(GTEST_OBJ): $(GTEST_SRC) | $(GTEST_BUILD_DIR)
 	$(CXX) $< $(CXXFLAGS) -c -o $@
 	@ echo
 
-$(BUILD_DIR)/src/$(PROJECT)/%.cuo: src/$(PROJECT)/%.cu
+$(OBJ_BUILD_DIR)/%.cuo: src/$(PROJECT)/%.cu | $(OBJ_BUILD_DIR)
 	$(CUDA_DIR)/bin/nvcc $(NVCCFLAGS) $(CUDA_ARCH) -c $< -o $@
 	@ echo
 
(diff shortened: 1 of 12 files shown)


jeffdonahue:
Closing this as I've abandoned it; @tnarihi feel free to cherry-pick or reference for #1985 if useful.

PR #398, opened 2014-05-07
Add L2 hinge loss

sguada:
This PR extends #303 adding C param to the loss and by adding a L2 norm to hinge_loss. These should help to implement L1 (L2) -SVMs with L1 (L2) regularization.

@longjon could you take a look, and let me know if your examples are still working.

By default it uses `C=1.0` and `hinge_norm=L1`, so it behaves as hinge loss.


jainanshul:
I have been looking for a prototxt file myself explaining how to use caffe to train linear SVM. Does anyone have an example that I could use?

PR #400, opened 2014-05-08
modify im2col.cpp into a multi-thread version

luotao1:
When training ImageNet model on CPU, `ConvolutionLayer<Dtype>::Forward_cpu` and `ConvolutionLayer<Dtype>::Backward_cpu` occupy about 80% time, and `im2col_cpu` + `col2im_cpu` occupy 30%~50% time of ConvolutionLayer, thus, we modify im2col.cpp into a multi-thread version.

We test out modified version on our 4-physical-core machine, if we create 4 pthreads, we can decrease 50% time of `im2col_cpu` + `col2im_cpu` . And you can define your **THREAD_NUM** based on the physical core number of your machine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ LIBRARY_DIRS += $(BLAS_LIB)
 
 # Complete build flags.
 COMMON_FLAGS += $(foreach includedir,$(INCLUDE_DIRS),-I$(includedir))
-CXXFLAGS += -pthread -fPIC $(COMMON_FLAGS)
+CXXFLAGS += -pthread -fPIC -D THREADS=$(PHTREAD_NUM) $(COMMON_FLAGS)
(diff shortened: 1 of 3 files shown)


jeffdonahue:
Closing this -- the 1.25x speedup with 4x CPU usage doesn't seem significant enough to justify the additional complexity here.

DevilGragon:
Hi, @luotao1 
   I try to use your method to improve forward pass performance only. So I replaced the original im2col_cpu with your im2col_cpu multithreads function and conpiled successfully. But the problem is that the test result in the same model is different from the original result, and the accuracy is even 0.1 to the original 0.7. I don't know why. Could you give me some advice? Thanks

luotao1:
@DevilGragon My code is finished in 2014, maybe is conflict with the current code. And I don't test it now. So if you just replaced the im2col_cpu easily, there may exist some problems.

Issue #408, opened 2014-05-12
Add support for opencl


bhack:
Hi Theano is adding support to opencl[1] trought CLBLAS[2]. 
Can caffe use this kind of solution to add opencl (more vendor neutral) support?

[1]https://github.com/Theano/libgpuarray
[2]https://github.com/clMathLibraries/clBLAS

gururise:
Any progress on OpenCL support?  Its a shame that Caffe is locked to a single vendor for GPU support.

jramapuram:
Alternatively you can use something like : 
 - https://github.com/viennacl/viennacl-dev or 
 - https://github.com/arrayfire/arrayfire

bhack:
Another effort started at https://github.com/BVLC/caffe/pull/2195

lemberger:
anybody tested this https://github.com/BVLC/caffe/pull/2610 https://github.com/naibaf7/caffe ?

shelhamer:
Done by the official `opencl` branch of BVLC Caffe.

Issue #382, opened 2014-05-03
Be aware of the competing fast GPU neural network library CXXNET


kloudkl:
Since this February, there have been a "(convolutional) neural network toolkit" [CXXNET ](https://github.com/antinucleon/cxxnet) based on the "Lightweight CPU/GPU Matrix/Tensor Template Library in C++/CUDA" [mshadow](https://github.com/tqchen/mshadow). The toolkit is able to classify 400 images per second, i.e. about 35 million per day, on a GTX 780 GPU. It seems to be faster than Caffe which can process 20 million per day on a K20 and 40 million per day on a K40. 
Since CXXNET is using the tensor library, its code is also much more concise than Caffe's.

PR #415, opened 2014-05-14
Unify the CPU, CUDA and OpenCL math functions API in the device wrapper classes

kloudkl:
This PR wraps the math functions as suggested by both [Yangqing](https://github.com/BVLC/caffe/issues/382#issuecomment-42465152) in #382 and [jeffdonahue](https://github.com/BVLC/caffe/issues/408#issuecomment-43035934) in #408 to abstract the device type from the algorithms.


hughperkins:
Reading through this thread, and 408, is it not entirely unreasonable to assume the migrating caffe to work with OpenCL would be both considerable amount of work, potentially conflict with several CUDA optimizations, and generally go 'against the flow' of what other Caffe contributors are looking to achieve?  I think this is the case, and that's why I'm rather writing an OpenCL convolutional network library 'from the ground up', at https://github.com/hughperkins/ClConvolve , but just touching base, in case my current approach is a bit too 'not invented here'?

momer:
@hughperkins Looks like an interesting project coming together. 

All: given that @kloudkl will no longer be contributing, is this PR still under review, or is this specific effort of bringing some abstraction / opencl support to the project and ready to be closed?

hughperkins:
@momer: Thank-you momer :-)

shelhamer:
@hughperkins @momer this PR is still open to remind us about the effort to abstract devices. Until this effort is revived it'll stay here as a placeholder and example of one approach. Once the abstraction to host both CUDA and OpenCL implementations arrives then @hughperkins layer implementations could be helpful!

hughperkins:
@shelhamer, yes that's not a bad idea.  The kernel implementations and specifically: debugging such implementations, is the most time-consuming part, so re-using the existing OpenCL kernels from ClConvolve could be an appropriate way forward.  Then, someone just needs to add an abstraction layer to Caffe, which might be a much simpler task, given the presence of existing kernels?  Maybe the existing kernels are neither complete nor optimal, but they could make a good start, and then completeness and optimization can be added over time.

I've added a brief architectural summary to the ClConvolve readme at https://github.com/hughperkins/ClConvolve#architecture , which details the 3 levels of abstraction available, ie NeuralNet, Propagate vs BackpropErrorsv2 vs BackpropWeights, and then the individual kernels/implementations, eg Propagate1, Propagate2, Propagate3.

An appropriate level of re-use for Caffe might be the individual kernels/implementations, eg [Propagate3.cpp](https://github.com/hughperkins/ClConvolve/blob/master/src/Propagate3.cpp) , as you say.

shelhamer:
Closing as this has been carried on in #610 which will itself be replaced by a master edition for resurrection someday.

PR #439, opened 2014-05-22
Openmp

borisgin:
Parallel version of caffe for CPU based on OpenMP. Significant speed-up on CPU .Scales well with number of cores (3x for 4 cores , 10x - for 16 cores). Modified files: constitutional , pulling and relu layers , im2col and col2im.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -272,8 +272,11 @@ ifeq ($(DEBUG), 1)
 	COMMON_FLAGS += -DDEBUG -g -O0
 	NVCCFLAGS += -G
 else
-	COMMON_FLAGS += -DNDEBUG -O2
+	COMMON_FLAGS += -DNDEBUG -O2 
 endif
+
+# AVX2 flags
+# CXXFLAGS += -march=core-avx2 -ffast-math -mfma
 
 # cuDNN acceleration configuration.
 ifeq ($(USE_CUDNN), 1)
@@ -303,6 +306,8 @@ ifeq ($(BLAS), mkl)
 else ifeq ($(BLAS), open)
 	# OpenBLAS
 	LIBRARIES += openblas
+        BLAS_INCLUDE ?= /opt/OpenBLAS/include
+        BLAS_LIB ?= /opt/OpenBLAS/lib
 else
 	# ATLAS
 	ifeq ($(LINUX), 1)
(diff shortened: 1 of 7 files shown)


xiaodaxia:
@borisgin    Hello, I feel a bit confused by the following benchmark.
Fixed Makefile to support openmp with MKL. Current benchmark results:
cifar10/ train_quick):
desktop Core i7-3770K CPU @ 3.50GHz × 4 cores x no HT): 100 sec
K20: 90 sec
Though I only have gtx980, but the train_quick demo finished in more than 250s. So I think
there is a misunderstanding? I think the only difference is that I used openblas/atlas instead of MKL,
but in my understanding the mkl will not affect gpu performance that much?
Does the train_quick demo always run 5000 iterations as it does now?

forresti:
Is this PR (or something similar) going to be merged soon?

When I checked not too long ago, CPU Caffe is unnecessarily slow without OpenMP. I'm debating tacking my own OpenMP things together... but merging this PR with present-day Caffe would be optimal.

talda:
I wouldn't hold my breath. This PR is more than a year old and it doesn't seem the maintainers think it should be merged. (probably better to close it)

shelhamer:
While CPU execution can be further optimized this PR is closed since it is against the deprecated dev branch. This branch was not merged at the time due to concerns about further complexity and dependencies.

bhack:
@naibaf7 It is working also on CPU at https://github.com/BVLC/caffe/pull/2610

naibaf7:
@shelhamer @bhack 
#2610 Uses OpenCL kernels and a CPU (MKL, OpenBLAS, Atlas) BLAS.
While OpenCL might not be the best for CPUs, the Alexnet runs twice as fast on Hybrid-OpenCL (that's what I call it) than the "legacy" singlethreaded CPU branch on a quadcore CPU.

Crefeda:
Hi, I was try to build your openmp version, on a CentOS 6.5 computer and I got a protobuf version error
my current one is libprotoc 2.4.1
could you let me know which version was compatible for you?

borisgin:
Hi Crefeda,
Can you post a detailed error, please?
Did caffe-master branch build passed without issues?
Thanks, Boris

Yangqing:
@Crefeda If I recall correctly, caffe relies on some features that are introduced in protobuf 2.5, so I think 2.5 and above should work.

PR #484, opened 2014-06-09
Java Bindings

kloudkl:
We need to use Caffe in a pure Java environment like Hadoop. There are two popular candidates JNI and JNA to bridge C/C++ to Java. While JNA is much easier to develop, [JNI is superior in performance](http://developers.opengamma.com/blog/2012/05/25/jna-jni-and-raw-java-performance). But it is quite an overhead to directly write JNI generator scripts like [the generator for OpenCV](https://github.com/Itseez/opencv/tree/master/modules/java/generator). 

SWIG is a tool used by many scripting languages to simplify porting C/C++ libraries and can automatically generate the JNI codes from the C/C++ headers files.

The HDF5_DATA_{DATASET/LABEL}_NAME macros were duplicated when the vision_layers.hpp was split into several smaller files. They are not allowed by the SWIG compiler. One thing that must be noted is that SWIG conflicts with the Ubuntu package protobuf-compiler. Protocol Buffers has to be installed from source. Although the pretty hacky generator scripts in this PR have eliminated most of the warnings and errors, there are still many left that potentially break the generated API.

TODO: tests.

### DIFF
--- a/java/generator/gen_java.py
+++ b/java/generator/gen_java.py
@@ -1,0 +1,966 @@
+#!/usr/bin/env python
+
+import sys, re, os.path
+from string import Template
+
+try:
+    from cStringIO import StringIO
+except:
+    from StringIO import StringIO
+
+class_ignore_list = (
+)
+
+const_ignore_list = (
+)
+
+const_private_list = (
+)
+
+# { Module : { public : [[name, val],...], private : [[]...] } }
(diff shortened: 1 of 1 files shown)


futurely:
https://github.com/BVLC/caffe/issues/1823#issuecomment-93659994

PR #505, opened 2014-06-16
Non-square Filters and Separated Stride and Padding

shelhamer:
- [x] add non-square kernel size, padding, and stride fields
- [x] check these parameters
- [x] add rectangular im2col test
- [x] fix checks -- see #614 
- [x] fix GPU rectangular im2col (thanks @ejaz-izy!)
- [x] add separable filter test case for convolution layer.

Accept pairs of height/Y and width/X values for kernel size, stride, and pad in lieu of a single, shared value.

Of course the square/equal case still works and the old defaults (stride = 1 and padding = 0) are kept.

Open to comments on implementation and style.


amitjslearn2:
Does non square kernels/filters work in caffe now?

Issue #440, opened 2014-05-22
Cannot use Caffe on another GPU when GPU 0 has full memory


jamt9000:
When GPU 0 is out of action, Cublas and Curand fail to initialise

```
E0522 16:04:44.114862 27133 common.cpp:31] Cannot create Cublas handle. Cublas won't be available.
E0522 16:04:44.427568 27133 common.cpp:38] Cannot create Curand generator. Curand won't be available.
```

Then, even if a different device is chosen, training or testing a net fails:

```
F0522 16:07:19.198012 27133 math_functions.cpp:50] Check failed: status == CUBLAS_STATUS_SUCCESS (13 vs. 0)  CUBLAS_STATUS_EXECUTION_FAILED
```

A workaround is to manually do cudaSetDevice(x) before starting Caffe, however this is not possible with the Python API.

argman:
Just a tip, I find the reason for this is that I have some cuda libraries in some corner before I reinstall a new one, that caused caffe use the old libraries, so you maye try "locate libcuda*" and remove all the old ones, then installl a new cuda library.

Best.

asa008:
F1030 16:15:43.990840 1937 math_functions.cu:28] Check failed: status == CUBLAS_STATUS_SUCCESS (13 vs. 0) CUBLAS_STATUS_EXECUTION_FAILED

The above error will occur when installing CUDA9.0
By installing [patchesPatch 2](https://developer.nvidia.com/cuda-90-download-archive?target_os=Windows&target_arch=x86_64&target_version=10&target_type=exelocal) (Released Mar 5, 2018)solve

PR #507, opened 2014-06-16
Fix Caffe::SetDevice to avoid initializing on default device

longjon:
This PR addresses #440 by moving the call to `cudaSetDevice` before any calls to `Get`, which may perform GPU memory allocation.

Despite what is written in #440, this _can_ be used in Python without any further changes, provided (somewhat counterintuitively) one calls `Net.set_device` before `Net.set_mode_gpu`.

AFAICT there is no need for the `cu*Destroy` cleanup to come before `cudaSetDevice`; let me know if you think otherwise.

### DIFF
--- a/src/caffe/common.cpp
+++ b/src/caffe/common.cpp
@@ -69,11 +69,13 @@ void Caffe::SetDevice(const int device_id) {
   if (current_device == device_id) {
     return;
   }
+  // The call to cudaSetDevice must come before any calls to Get, which
+  // may perform initialization using the GPU.
+  CUDA_CHECK(cudaSetDevice(device_id));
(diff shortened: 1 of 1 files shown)


howtobeahacker:
Hi developers,
I believe this problem should be fixed by now. However, when I tried example [00-classification.ipynb](https://github.com/BVLC/caffe/blob/master/examples/00-classification.ipynb), this error occur again.
When I set my device into second gpu because my first GPU is busy as following:
```
# GPU mode
caffe.set_device(1)
caffe.set_mode_gpu()
net.forward()  # call once for allocation
%timeit net.forward()
```

Then the kernel of the notebook is died unexpectedly and I cannot execute code anymore. Thank you and hope to have an message from you guys.
@An

Issue #389, opened 2014-05-05
Explicitly disable backward propagation of a layer for controlled fine tuning


kloudkl:
The Google video classification CNN explored four transfer learning methods training from scratch, fine-tuning top layer (classifier), fine-tuning top 3 layers, and fine-tuning all layers [1]. Fine-tuning specific layers keeps the generic features of the other layers untouched during training. They found that fine-tuning top 3 layers performed best.

It is not very straightforward to reason about whether the backward propagation of a layer is disabled or not in Caffe as shown in #100 and #103. So it would be nice to be able explicitly disable that.

[1] Andrej Karpathy, George Toderici, Sanketh Shetty, Thomas Leung, Rahul Sukthankar, Li Fei-Fei. Large-Scale Video Classification with Convolutional Neural Networks. CVPR 2014.

PR #509, opened 2014-06-17
No backward in the feature extraction network

kloudkl:
Because the learning rate is default to be 1 and backward is default to be true (#389), even though the phase is set to `Caffe::TEST`, the feature extraction network still performs backward propagation unnecessarily.


Noiredd:
If no backward propagation is not explicitly called, are the gradients still being calculated during inference? I'm optimizing a network for inference speed, and for this particular application I'm not interested in backprop - `caffe time` still includes it in measurements though. Ideally, I'd like to be able to train a network and then disable all unnecessary computations in order to benchmark it. Do I have to modify the entire proto, or is there some known best way to approach this?

PR #522, opened 2014-06-20
Accuracy without loss

sguada:
This PR would separate the loss from the accuracy. This will allow using different losses while computing the accuracy.

Once the `AccuracyLayer` doesn't compute the loss then one need to add a `LossLayer` if want to compute it.

There are two options for `AccuracyLayer`, now it is doing a ArgMax on the data to compute the predicted label and then compared with expected label. That could be removed and let the `ArgMaxLayer` do that.

Another option will be to create the `ArgMaxLayer`, and maybe a `LossLayer` within the `AccuracyLayer` then connect them.


baimin1:
how to dowload these 20 files?

PR #546, opened 2014-06-26
Weight Sharing

shelhamer:
[Original PR notes copied from #500. This PR replaces it.]

This adds the ability to share parameters between layers, which has a number of applications, the canonical one perhaps being recurrent neural network (RNN) training.

To share weights between two or more layers with parameters (currently just `InnerProductLayer`s and `ConvolutionLayer`s), specify the same `param` for all of these layers.  (You can also name the biases with a second `param`, as in the `blobs_lr` and `weight_decay` parameters.)  You can see a very simple example of this in `src/caffe/test/test_net.cpp`: see the unit test named `InitDiffDataSharedWeightsNet`:

```
layers: {
  name: 'innerproduct1'
  type: INNER_PRODUCT
  inner_product_param {
    num_output: 10
    bias_term: false
    weight_filler {
      type: 'gaussian'
      std: 10
    }
  }
  param: 'sharedweights'
  bottom: 'data'
  top: 'innerproduct1'
}
layers: {
  name: 'innerproduct2'
  type: INNER_PRODUCT
  inner_product_param {
    num_output: 10
    bias_term: false
    weight_filler {
      type: 'gaussian'
      std: 10
    }
  }
  param: 'sharedweights'
  bottom: 'data'
  top: 'innerproduct2'
}
```

This means layers innerproduct1 and innerproduct2 are sharing the same set of weights as they've both specified `param: 'sharedweights'`.  And in this case they also take the same bottom blob, (`data`), so their outputs, top blobs `innerproduct1` and `innerproduct2`, should be identical (so this is not actually something you'd ever want to do; I do it there just for testing purposes).

Note that in this case we specify only one blob name because we've set `bias_term: false`; if we didn't have `bias_term: false` we'd need to specify two `param`s, but probably the second one should be empty unless we actually want to share biases.  (Specifying the empty string as a `param` is equivalent to not specifying a `param` in my implementation.)

```
param: 'sharedweights'
param: ''
```

The entire implementation is in `Net::Init`, `Net::AppendParam`, and `Net::Update`.  `Init` figures out which layer will actually "own" the shared param (the first one to list its `param`), and `Update` adds the non-owned layers' computed diffs into the diff of the owner blob, then only actually performs updates on owned blobs.  Memory-wise, all shared blobs actually point to the same memory location for the parameter's data, but still have separately allocated diff blobs, as the logic to handle learning rate, weight decay, etc is still handled by the Solver (which is blissfully unaware that parameters can be shared).

Open to hearing feedback on the interface, implementation, etc.  I'm not sure I'm happy with `param` as the name of the field, I think it would be less ambiguous to use `param_name` or something, but would be inconsistent with the other per-parameter field `blobs_lr` (and actually to be consistent with that it should be `blobs_name`, but I strongly prefer the singular here..).

### DIFF
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -242,7 +242,7 @@ void Net<Dtype>::AppendParam(const NetParameter& param, const int layer_id,
   }
   const int net_param_id = params_.size();
   params_.push_back(layers_[layer_id]->blobs()[param_id]);
-  param_net_indices_.push_back(make_pair(layer_id, param_id));
+  layer_param_indices_.push_back(make_pair(layer_id, param_id));
   if (!param_size || !param_name.size() || (param_name.size() &&
       param_names_index_.find(param_name) == param_names_index_.end())) {
     // This layer "owns" this parameter blob -- it is either anonymous
@@ -257,7 +257,7 @@ void Net<Dtype>::AppendParam(const NetParameter& param, const int layer_id,
     const int owner_net_param_id = param_names_index_[param_name];
     param_owners_.push_back(owner_net_param_id);
     const pair<int, int>& owner_index =
-        param_net_indices_[owner_net_param_id];
+        layer_param_indices_[owner_net_param_id];
     const int owner_layer_id = owner_index.first;
     const int owner_param_id = owner_index.second;
     LOG(INFO) << "Sharing parameters '" << param_name << "' owned by "
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -144,7 +144,7 @@ class Net {
   vector<vector<Blob<Dtype>*> > top_vecs_;
   vector<vector<int> > top_id_vecs_;
   vector<int> param_owners_;
-  vector<pair<int, int> > param_net_indices_;
+  vector<pair<int, int> > layer_param_indices_;
   map<string, int> param_names_index_;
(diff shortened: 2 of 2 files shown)


src/caffe/net.cpp
```diff
@@ -403,8 +472,36 @@ void Net<Dtype>::ToProto(NetParameter* param, bool write_diff) {
...
+    case Caffe::GPU:
+      this_diff = params_[i]->gpu_diff();
+      owner_diff = params_[param_owners_[i]]->mutable_gpu_diff();
+      caffe_gpu_add(count, this_diff, owner_diff, owner_diff);
+      break;
+    default:
+      LOG(FATAL) << "Unknown caffe mode: " << Caffe::mode();
+    }
+  }
+  // Now, update the owned parameters.
```
abhi2610: @ashafaei Was this fixed in the current version?

shelhamer: @ashafaei @abhi2610 This was never actually an issue although it was worth raising since you need to know how the `Net`, `Solver` and `Blob`s cooperate. The loss / backward logic only decides whether backward is computed for a layer or not. `Net::Update()` is always called by the solver, all of the shared weight params will accumulate their diff with the owner in the loop at 478, and then `Blob::Update()` will always be called for every weight owner as in the loop at 501. Thanks @jeffdonahue for discussion.

Issue #150, opened 2014-02-24
Continuous Integration / Build Monitoring


shelhamer:
Configure a buildbot to check `master` and `dev` on push and report test status and the like.

alexbw:
Curious if anyone found out a GPU CI solution?

PR #561, opened 2014-06-30
CPU-only build

shelhamer:
**Note: the CUDA libraries are still required for now.**
- [x] add `CPU_ONLY` flag to Makefile to skip GPU compilation
- [x] stub GPU methods to die and complain if their use is attempted in CPU-only build
- [x] hide GPU code from CPU-only build by preprocessor guards
- [x] collect CUDA includes into a single single header that does actual CUDA or CUDA alternate stubs
- [x] define CPU-only Caffe singleton and SyncedMemory
- [x] separate all CUDA code by guards for CUDA-less CPU-only build

This adds a CPU-only switch to the `Makefile.config` that builds Caffe without compiling any GPU code. When this flag is uncommented
- `make` builds libcaffe.so and libcaffe.a without any GPU compilation at all (not a single call to nvcc).
- `make test` builds the tests as usual.
- `make runtest` only runs the CPU tests.

Mostly I was curious what this would take since it partially addresses #150. With this build we should be able to configure a Travis-CI hook for CPU testing of PRs and such.

Not that we should necessarily merge this as-is or at all however, especially with the grossness of 81791ec, and the macro pollution I added in 871b0ce.

I imagine the #415, BVLC:device-abstraction work would make carrying out the CPU / CUDA separation at compilation simpler. Perhaps this is useful all the same?

What does everybody think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ before_install:
 install:
   - wget https://google-glog.googlecode.com/files/glog-0.3.3.tar.gz -O /tmp/glog-0.3.3.tar.gz && tar -C /tmp -xzvf /tmp/glog-0.3.3.tar.gz && rm /tmp/glog-0.3.3.tar.gz
   - cd /tmp/glog-0.3.3 && ./configure && make && sudo make install && cd -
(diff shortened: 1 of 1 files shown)


rajhlinux:
How do I compile it for FreeBSD 13.1 without GPU?

PR #569, opened 2014-06-30
Transform layers [DON'T MERGE]

sguada:
This is a tentative approach to add transformation layers that would allow to crop_mirror, center_scale the blobs.

The first use would be to help Data source and pre-processing separation. The source would produce a vector of blobs, one per image, the transformation layers and finally concat all the blobs into one.

The concept is similar to do `map(transformation, bottom_blobs, top_blobs)` 

@Yangqing What do you think?

### DIFF
--- a/include/caffe/transform_layers.hpp
+++ b/include/caffe/transform_layers.hpp
@@ -1,0 +1,126 @@
+// Copyright 2014 BVLC and contributors.
+
+#ifndef CAFFE_TRANSFORM_LAYERS_HPP_
+#define CAFFE_TRANSFORM_LAYERS_HPP_
+
+#include <string>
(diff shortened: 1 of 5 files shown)


jeffdonahue:
Closing as abandoned.  We agree with the motivation (data transformers becoming layers of their own rather than applied by each data layer) but this needs a bit more thought (prefetching etc., see discussion above) and a rebase.

whjxnyzh:
@sguada  are there RGB jittering in caffe？Thank you

sguada:
I don't think so, but you could implement them easily using other layers.

Sergio

2015-06-01 0:30 GMT-07:00 whjxnyzh <notifications@github.com>:

Issue #316, opened 2014-04-11
How to model a Siamese Network?


zgxiangyang:
which is usefull to solve a two classification problems.

[8] S. Chopra, R. Hadsell, and Y. LeCun. Learning a similarity metric discriminatively, with application to face veriﬁcation. In CVPR, 2005. 2, 5

anxiaoxi45:
@sunbaigui hi,Have you solved the problem? I also want to know how write the image list file, did it work like this format (img1 img2 label) ? How did you do?

sunbaigui:
@anxiaoxi45 I havn't done this yet, sorry for reply so late, kind of busy recently, have you done it yet?

gfl94:
@anxiaoxi45 @sunbaigui Have you solved the  (img1 img2 label) problem?

szahid86:
@gfl94 : I have tried to solve the (img1 img2 label) problem using following approach but failed. Please note that I am very newbie to python and caffe. I followed a very pingpong approach.

1- I decided to use the Siamese example file (mnist_siamese_train_test.prototxt), and make some changes in it. The example mnist_siamese_train_test.prototxt requires the input train and test data to be in lmdb format.

2- So I tried to convert my train and test data into lmdb format using convert_mnist_siamese_data.cpp but I was unable to convert my data because of some weird errors.

3- Next I found some code, that shows the possibility to use .jpg files at data layer, rather than converting to lmdb format. That code uses two train1.txt and train2.txt files (with the format, path/to/image/1.jpg label). Label value is '1' if the image files in the corresponding lines of the two .txt files are of the same person, and '0' otherwise.

Sample Data layers for your reference:

layers {
  name: "data"
  type: IMAGE_DATA
  top: "data"
  top: "label"
  image_data_param {
    source: "/home/cvip/caffe/examples/siamese/My_DB/TrainLabels_1.txt"
    batch_size: 128
  }
  transform_param {
 # crop_size: 40
  mirror: true
  }
  include: { phase: TRAIN }
}
layers {
  name: "data"
  type: IMAGE_DATA
  top: "data"
  top: "label"
  image_data_param {
    source: "/home/cvip/caffe/examples/siamese/lfw_funneled/TestImageLabels_1.txt"
    batch_size: 100
  }
  transform_param {
 # crop_size: 40
  mirror: true

  }
  include: { phase: TEST }
}

layers {
  name: "data_p"
  type: IMAGE_DATA
  top: "data_p"
  top: "label_p"
[...]

PR #639, opened 2014-07-07
learn a distance function suitable for use with hinge loss in siamese network

jackculpepper:
This layer takes two inputs (x0 and x1) and learns a weights vector (w) and scalar bias (b).

y = b + \sum_i w_i f( x0_i - x1_i )

where f(x) = x^2 or f(x) = |x|.

This is useful for learning siamese networks, which has been talked about in other issues (https://github.com/BVLC/caffe/issues/316, https://github.com/BVLC/caffe/pull/546).

Weights sharing has already been merged into dev, but a layer of this type is also necessary.

### DIFF
--- a/include/caffe/vision_layers.hpp
+++ b/include/caffe/vision_layers.hpp
@@ -130,6 +130,39 @@ class ConvolutionLayer : public Layer<Dtype> {
   int N_;
 };
 
+/* DistanceLayer
+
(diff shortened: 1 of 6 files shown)


leiucky:
@jackculpepper you said that we can calculate distance of two vectors using eltwise layer above. I wonder that how to sum all elements of a vector? For example in #2438  , how to calculate Euclidean distance? I have perform the subtraction and square step.

PR #643, opened 2014-07-08
support windows 7 64bit sytem for developing

OpenHero:
Download those packages and install in the path
1. HDF5: D:\HDF_Group\HDF5\1.8.13\include;
2. leveldb D:\leveldb-master\include;
3. pthread D:\pthreads\Pre-built.2\include;
4. opencv D:\opencv\build\include;
5. protobuf D:\protobuf-2.5.0\src;
6. glog D:\glog-0.3.3\src\windows;
7. boost D:\boost_1_55_x64\boost_1_55_0;
8. mkl(education evaluation) C:\Program Files %28x86%29\Intel\Composer XE 2013\mkl\include;
9. F:\project\dl\caffe_gpu\src\caffe\include;
10. CUDA C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v6.0\include

### DIFF
--- a/examples/mnist/convert_mnist_data.cpp
+++ b/examples/mnist/convert_mnist_data.cpp
@@ -17,6 +17,10 @@
 
 #include "caffe/proto/caffe.pb.h"
 
+#if _MSC_VER
+# define snprintf _snprintf
+#endif
+
 uint32_t swap_endian(uint32_t val) {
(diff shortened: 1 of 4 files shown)


happynear:
@memoryemon , I have solved the problem by adding all the cpp files to the exe project. Then the compile procedure would be slower than using a static library. I haven't found any better solution. Have you fixed it now?

Issue #677, opened 2014-07-11
Issues while doing "make matcaffe"


superhans:
Hi Guys,

While running "make matcaffe" on my machine (Ubuntu 12.04, Matlab R2013b), I got a mountain of errors. Trying "mex matcaffe.cpp" gave me the same mountain.

I'm going to copy paste this below. I can't figure out what the problem is, so I thought I could post here. 
Thanks.

```
matcaffe.o: In function `set_device(int, mxArray_tag**, int, mxArray_tag const**)':
matcaffe.cpp:(.text+0x104): undefined reference to `google::LogMessage::LogMessage(char const*, int, int)'
matcaffe.cpp:(.text+0x10d): undefined reference to `google::LogMessage::stream()'
matcaffe.cpp:(.text+0x14b): undefined reference to `google::LogMessage::~LogMessage()'
matcaffe.cpp:(.text+0x169): undefined reference to `caffe::Caffe::SetDevice(int)'
matcaffe.cpp:(.text+0x183): undefined reference to `google::LogMessage::~LogMessage()'
matcaffe.o: In function `init(int, mxArray_tag**, int, mxArray_tag const**)':
matcaffe.cpp:(.text+0x1c7): undefined reference to `google::LogMessage::LogMessage(char const*, int, int)'
matcaffe.cpp:(.text+0x1d0): undefined reference to `google::LogMessage::stream()'
matcaffe.cpp:(.text+0x20f): undefined reference to `google::LogMessage::~LogMessage()'
matcaffe.cpp:(.text+0x25b): undefined reference to `caffe::Net<float>::Net(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
matcaffe.cpp:(.text+0x359): undefined reference to `caffe::Net<float>::CopyTrainedLayersFrom(std::basic_string<char, std::char_traits<char>, std::allocator<char> >)'
matcaffe.cpp:(.text+0x3d1): undefined reference to `google::LogMessage::~LogMessage()'
matcaffe.o: In function `reset(int, mxArray_tag**, int, mxArray_tag const**)':
matcaffe.cpp:(.text+0x4a8): undefined reference to `google::LogMessage::LogMessage(char const*, int)'
matcaffe.cpp:(.text+0x4b1): undefined reference to `google::LogMessage::stream()'
matcaffe.cpp:(.text+0x4ce): undefined reference to `google::LogMessage::~LogMessage()'
[...]

Issue #605, opened 2014-07-03
Cannot build Matlab binding for Matlab R2014a


Felix-neko:
I'm trying to build Matlab extension for Matlab R2014a. I have CUDA 6.0, GCC 4.7 and MKL as BLAS library.

I perform `make all` (it's OK). Then I try

`make matcaffe`

It gives me the following error log:

```
/path/to/MATLAB/bin/mex matlab/caffe/matcaffe.cpp build/lib/libcaffe.a \
        CXXFLAGS="\$CXXFLAGS -pthread -fPIC -DNDEBUG -O2 -DUSE_MKL -I/usr/local/include/python2.7 -I/usr/local/lib/python2.7/dist-packages/numpy/core/include -I/usr/local/include -Ibuild/src -I./src -I./include -I/usr/local/cuda/include -I/opt/intel/mkl/include -Wall" \
        CXXLIBS="\$CXXLIBS -L/usr/local/lib -L/usr/local/lib -L/usr/lib -L/usr/local/cuda/lib64 -L/usr/local/cuda/lib -L/opt/intel/mkl/lib/intel64 -lcudart -lcublas -lcurand -lpthread -lglog -lprotobuf -lleveldb -lsnappy -lboost_system -lhdf5_hl -lhdf5 -lopencv_core -lopencv_highgui -lopencv_imgproc -lmkl_rt" -output matlab/caffe/caffe.mexa64
Building with 'g++'.
/home/my_name/Projects/caffe/matlab/caffe/matcaffe.cpp: In function 'mxArray* do_get_weights()':
/home/my_name/Projects/caffe/matlab/caffe/matcaffe.cpp:188:50: warning: narrowing conversion of '(& layer_blobs)->std::vector<_Tp, _Alloc>::size<boost::shared_ptr<caffe::Blob<float> >, std::allocator<boost::shared_ptr<caffe::Blob<float> > > >()' from 'std::vector<boost::shared_ptr<caffe::Blob<float> >, std::allocator<boost::shared_ptr<caffe::Blob<float> > > >::size_type {aka long unsigned int}' to 'const mwSize {aka const int}' inside { } is ill-formed in C++11 [-Wnarrowing]
build/lib/libcaffe.a(math_functions.o): In function `void caffe::caffe_cpu_gemm<float>(CBLAS_TRANSPOSE, CBLAS_TRANSPOSE, int, int, int, float, float const*, float const*, float, float*)':
[...]

PR #696, opened 2014-07-15
Fix bug that breaks MATLAB 2014a compilation

rbgirshick:
Change how mex is called to fix broken compilation with MATLAB 2014a. I've tested this with 2014a and 2012b (2012b was working previous to this change and continues to work with this fix). This PR may address issues that people have been reporting (e.g., #605, #677).


black3391:
It seems that I fixed this problom by modifing the "STATIC_NAME" to "NAME", in makefile in line 266 an line 272
screenshot from 2015-09-10 12 59 46

cslxiao:
black3391's method works well, thanks.

kisseternity:
It works for me. I've been confused with this problem several days, thanks very much!!!!!!

PR #755, opened 2014-07-22
Finishing touches on @bhack's SliceLayer (#680)

jeffdonahue:
Finishes #680. Thanks for contributing this layer @bhack, and thanks for your review @sguada.  Will merge once Travis passes.

### DIFF
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -179,8 +179,8 @@ class SplitLayer : public Layer<Dtype> {
 };
 
 /* SliceLayer
-  Takes one blobs and slice it along either num or
-  channel dim, outputting the result.
+  Takes one blobs and slices it along either num or channel dim,
+  outputting the result into as many top blobs as needed.
 */
 template <typename Dtype>
 class SliceLayer : public Layer<Dtype> {
@@ -213,7 +213,7 @@ class SliceLayer : public Layer<Dtype> {
   int height_;
   int width_;
   int slice_dim_;
-  google::protobuf::RepeatedField<google::uint32> slice_point_;
+  vector<int> slice_point_;
 };
 
 }  // namespace caffe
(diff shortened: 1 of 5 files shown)


include/caffe/common_layers.hpp
```diff
@@ -178,6 +178,44 @@ class SplitLayer : public Layer<Dtype> {
...
+  explicit SliceLayer(const LayerParameter& param)
+      : Layer<Dtype>(param) {}
+  virtual void SetUp(const vector<Blob<Dtype>*>& bottom,
+      vector<Blob<Dtype>*>* top);
+
+  virtual inline LayerParameter_LayerType type() const {
+    return LayerParameter_LayerType_SLICE;
+  }
+  virtual inline int ExactNumBottomBlobs() const { return 1; }
+  virtual inline int MinTopBlobs() const { return 2; }
```
ronghanghu: I am sort of confused here when I use this layer, and traced back to here... Why `MinTopBlobs()` returns 2? (why not 1?)

bhack: @ronghanghu Slice is a knife for a blob. A division by 1 exist in math. But is it operative useful when you are cooking?

ronghanghu: You are right :)

I was trying to use slice layer as a component in another layer to automatically slice T x ... blob into T blobs of 1x ..., and run into this error when T == 1. I'll handle the special case myself.

seanbell: Why not allow for 1 top blob?  I understand that it doesn't make sense to use a `SliceLayer` for that purpose, but if a user asks for it, the layer could act as a pass-through, doing nothing at all.  This makes the layer more general.

bhack: If there are some derived reuse of this layer then could be extended to hande the Top 1 case in a way that not impact performance (cause it is a useless computation) so the logic doesn't need to be reproduced everywhere.

PR #784, opened 2014-07-25
CMake CPU-Only build

akosiorek:
- `CPU-ONLY` CMake build option to facilitate #561
- CPU-Only build added to Travis-CI
- Tests of the CMake CPU-Only build under Travis-CI

In order to build CPU-Only version of Caffe you have to run:
`cmake -DCPU-ONLY=ON path/to/CMakeLists.txt`

When `CPU-ONLY` is enabled no CUDA libraries are looked for or linked to any target, nor are any `*.cu` files compiled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,11 +37,11 @@ script:
   # CMake build.
   - mkdir build
   - cd build
-    ## CPU-GPU build
+  ## CPU-GPU: build only.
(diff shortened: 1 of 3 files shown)


ipazc:
You meant `DCPU_ONLY` instead of `DCPU-ONLY`.

PR #792, opened 2014-07-25
Update blob.hpp for checking blob indices

dutran:
Correct index checking for blob


shelhamer:

Issue #876, opened 2014-08-07
Multi-GPU operation and data / model Parallelism


shelhamer:
Multi-GPU operation and data / model / hybrid parallelism are planned and in development for Caffe. The purpose of the thread is to focus the conversation, since this has been asked here, there, and everywhere. There are several ways to approach parallelization, so feel free to discuss your own work to this end here.

Note that Caffe does work with multiple GPUs in a standalone fashion right now: you can train on one GPU while extracting features on another and so on.

naibaf7:
For forward processing, I'm currently trying to implement SplitNet, which adds the possibility to split up a network DAG on multiple devices and process the network forwarding asynchronously across multiple devices.
This is necessary because our dense labeling network (for microscopy images) does not fit well into the memory of one GPU anymore.

futurely:
While #2114 and a series of related PRs https://github.com/BVLC/caffe/issues/1535#issuecomment-117018020 have solved the data parallelism, there isn't yet a pull request for model parallelism. [Here](https://github.com/facebook/fbcunn/blob/master/fbcunn/ModelParallel.lua) is Facebook's implementation just for reference.

PhoenixDai:
Looks like Nvidia has figured out how to train model on Caffe with multiple GPUs. This page https://developer.nvidia.com/digits showed performance comparison on training with different number of GPUs. In the video on the page, it showed how to use multiple GPUs with DIGITS. Is there any plan to have this feature in Caffe?

choosehappy:
@PhoenixDai 's comment is related to the digits github issue: https://github.com/NVIDIA/DIGITS/issues/92. It seems they have forked their own caffe version which supports multiple gpus?

thatguymike:
The Nvidia branch uses #2114

shaibagon:
Linking to SO [related question](http://stackoverflow.com/q/33726333/1714410).

futurely:
The master branch supports multi-GPU training. Please refer to the latest documents.
https://github.com/BVLC/caffe/blob/master/docs/multigpu.md
https://github.com/BVLC/caffe/blob/master/docs/tutorial/interfaces.md

xiaoxiongli:
thank you for your great job!   Now I am training googlenet in my K80, as you know , K80 has 2 core, and I enable these 2 core by "-gpu 0,1", the training speed is faster! 
[...]

PR #884, opened 2014-08-08
(WIP) Overlap the CUDA data transfers with computations with cudaMemcpyAsync

kloudkl:
This is the first step tried to enable copying data among CPU and multiple GPUs which is required by #876. Of course, even there is only a single GPU, the non-blocking memory copying can still accelerate the speed.

The detailed theories under the hood are explained in "[How to Overlap Data Transfers in CUDA C/C++](http://devblogs.nvidia.com/parallelforall/how-overlap-data-transfers-cuda-cc/)".

Not tested on GPU yet.

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -148,6 +148,11 @@ class Caffe {
   static void SetDevice(const int device_id);
   // Prints the current GPU status.
   static void DeviceQuery();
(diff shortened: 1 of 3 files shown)


shelhamer:
This has been handled separately by parallel branch(es). Expect PRs when everything is tuned and checks out.

Issue #432, opened 2014-05-21
Extracting features with Matlab script "matcaffe_demo.m"


caffecuda:
Hi,

I'm planning to use a pretrained ilsvrc12 model as a feature extractor for a new image dataset, and train svm classifiers for the new set. Is this sensible at all?

The "layers" output of the Matlab script "matcaffe_demo.m" has the following fields for
the 8 layers, which of them should I use for my purpose? Many thanks.

```
[ 4-D single]
[96x1 single]

[  4-D single]
[256x1 single]

[  4-D single]
[384x1 single]

[  4-D single]
[384x1 single]

[  4-D single]
[256x1 single]

[9216x4096 single]
[4096x1    single]

[4096x4096 single]
[4096x1    single]

[4096x1000 single]
[1000x1    single]
```

PR #931, opened 2014-08-15
Matcaffe feature extractor

dgolden1:
Create `matcaffe_getscores.m` which can get both imagenet scores (and plot results) for an input image, or return 4096 features from the penultimate layer per the procedure (somewhat vaguely) described at https://github.com/BVLC/caffe/issues/432.

### DIFF
--- a/matlab/caffe/matcaffe_getscores.m
+++ b/matlab/caffe/matcaffe_getscores.m
@@ -10,6 +10,8 @@ function varargout = matcaffe_getscores(im, varargin)
 % b_force_init: force initialization of model (e.g., if using a new model definition)
 %  (default: false)
 % prediction_type: one of 'imagenet' (default) or 'features' for feature extractor
+% b_print_results: print top image matches (only valid if prediction_type = 'imagenet')
+%  (default: true)
 %
 % OUTPUTS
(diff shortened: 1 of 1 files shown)


mgarbade:
Seems like your program "matcaffe_getscores.m" is only working for Matlab R2015a ??? I'm using R2013a and it gives me an error at 
p = inputParser;
p.addParameter
Error:
"No appropriate method, property, or field addParameter for class inputParser."

I'm still trying to extract the 4096  dimensional feature vector for an image using the Matlab wrapper. Is there any common solution?
I didn't really understand the solution given by https://github.com/BVLC/caffe/issues/432
I tried to modify the file 
caffe-master/models/bvlc_reference_caffenet/deploy.prototxt
by deleting the last few layers up to 
layer {
  name: "drop6"
Still "scores = matcaffe_demo(im, 1);" gives me a 1000 dimensional vector of scores :(.
Any help would really be appreciated

mgarbade:
I found the solution: Modifying "caffe-master/models/bvlc_reference_caffenet/deploy.prototxt" as explained in #432 was the solution. I had to restart Matlab after modifying the file such that Matlab reloads it.

PR #937, opened 2014-08-16
Sparse Data Support

alemagnani:
This PR has been replaced by #2364

This adds some basic support for sparse data in CSR format. The main thing is a SparseBlob to store the sparse data and an extension to InnerProduct that handles both dense and sparse depending on what is presented at the input. A new data layer is added to read data from LevelDB.
Some more details:
* subclass of Blob that handles matrices in CSR format (SparseBlob)
* subclass of InnerProductLayer to deal with a sparse blob as input (it handle dense at the same time)
* a new data layer (DataLayerSparseInput) that reads for LevelDB sparse vector and creates SparseBlob
* support for CuSparse
* new sparse math operations on both GPU and CPU
* small change to Syncedmem to make it possible to own/not own GPU data.  

* new Proto object SparseDatum to store sparse vectors.
* changed the way Net creates Blobs to support sparse blobs. Basically the bottom layer gets to decide what Blob to create. This way the DataLayerSparseInput produces SparseBlobs (see the new method in layer_factory.cpp). There could be other ways to achieve the same results.
* exstensive test for the new math function and the new layers
* 2 different implementations of GPU sparse kernels (one is commented out). This is because I found out that CUSparse has very poor performance at least in the way I used it.

If this is of interest to people I also have locally a data layer for sparse data that uses a  block of memory and an extension to Pycaffe to make use of it and deal in python with sparse matrices.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ INCLUDE_DIRS += $(BUILD_INCLUDE_DIR) ./src ./include
 ifneq ($(CPU_ONLY), 1)
 	INCLUDE_DIRS += $(CUDA_INCLUDE_DIR)
 	LIBRARY_DIRS += $(CUDA_LIB_DIR)
-	LIBRARIES := cudart cublas curand
+	LIBRARIES := cudart cublas cusparse curand
 endif
 LIBRARIES += glog gflags protobuf leveldb snappy \
 	lmdb boost_system hdf5_hl hdf5 m \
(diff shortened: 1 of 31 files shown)


alemagnani:
I have a rebase ready on top of current master. Should I create a  new PR on master?

shelhamer:
That would be great. Thanks @alemagnani!

jeffdonahue:
I'll close this as it's been replaced by #2364.  Thanks for updating @alemagnani!

PR #959, opened 2014-08-21
Contrastive loss layer for training siamese nets

nickcarlevaris:
Hi All,

I've started work on a contrastive loss layer that, when combined with weight sharing #546, can be used to train siamese nets. 

The layer implements:
loss = 1/2 (y d + (1-y) \max(margin-d, 0))
d = \sum_i (a_i - b_i)^2
where d is the distance between two features a and b, and y is binary, indicating if the two features are similar or dissimilar.

This loss function was proposed in:
Raia Hadsell, Sumit Chopra, Yann LeCun "Dimensionality Reduction by Learning an Invariant Mapping"

I still need to implement the GPU version and flesh out the tests, but before I go too far, is this something that you would be interested in merging? I can also add a small example based on MNIST for documentation if that is of interest.

Thanks, 
Nick


xiaoyong:
@okn2020 You may check out the DeepID2 paper:
Yi Sun, Xiaogang Wang, Xiaoou Tang. "Deep Learning Face Representation by Joint Identification-Verification".

Issue #684, opened 2014-07-14
how to develop new layers


wshenx:
Hi everyone. I'm new to caffe. I want to develop new layers but I don't know to start. I checked the development guide on http://caffe.berkeleyvision.org/development.html. But I still have no clues. Is there anyone who can give a small tutorial step by step on how to develop a new layer in caffe. That will be of great value. Thanks.

kuaitoukid:
It is not a must, but safe to write a test file

zisetiankong324:
@longjon  Hi,I followed your step and I have added my new layer,however,when I run caffe,I encountered an error:segmentation fault(core denied),would you please tell me what happend?

gucasdongzi:
@longjon good~

antran89:
@wshenx You can use QtCreator + Cmake combination that I used in Linux.

artiit:
If I want to use GPU to train net, and I only write the cpp code, no .cu file, Is it OK?

antran89:
@artiit Then your Caffe only runs on CPU. By default, Caffe has a safe fall-back mechanism when you do not have GPU CUDA implementation.

williford:
For future reference, the wiki page has apparently moved to https://github.com/BVLC/caffe/wiki/Development

LinHungShi:
@williford I follow the step, but when I train the model, it says unknown layer type.. Do I have to remake caffe after I write the new layer?

misaka-10032:
It takes so long to test all the layers. Is there a way to only test a specific layer?

matthieudelaro:
Here is the complete command I use to recompile caffe, recompile the tests, and run the tests of the layer I'm working on:
```
make all && make test && build/test/test_all.testbin --gtest_filter='*LocalLayerTest*'"
```

cuijianaaa:
use Code::Blocks in linux，just like visual studio in windows，very easy

PhilippeCb:
@cuijianaaa I am trying to use code::block to speed up the testing but cannot manage to start a project with existing files, and I am not confident enough to work directly in the caffe home directory (caffe is finaly working and I would like to keep it that way :) ) 
could you tell me how to create a project from a copy of the caffe directory?
or if this is absolutely the wrong way to do it please tell me I am new to programming

PR #1020, opened 2014-09-01
Python layer for rapidly writing nets in Python

longjon:
~~This PR depends on #1014, and shouldn't be merged before that one.~~

Caffe is fast, but adding new layers is a multistep process (see #684), and is subject to all of the pitfalls of development in a low-level, compiled language. For quickly trying new ideas, speed of development may be more important than runtime speed.

This PR addresses this gap by allowing layers to be written in Python. One adds layers to the net prototxt that look like the following:
```
layers {
  name: "python"
  type: PYTHON
  bottom: "data"
  top: "output"
  python_param {
    module: "my_module"
    layer: "MyLayer"
  }
}
```

Then, one implements layers in Python with more or less the same interface as in C++, as below.
```python
class MyLayer(object):
    """Simple layer that multiplies input by ten."""

    def setup(self, bottom, top):
        pass

        top[0].reshape(bottom[0].num, bottom[0].channels,
                bottom[0].height, bottom[0].width)

        top[0].data[...] = 10 * bottom[0].data

        if propagate_down[0]:
            bottom[0].diff[...] = 10 * top[0].diff
```

In order to make this work, the main caffe binaries need to be linked against the Python libraries. This is solved by adding a build option, `WITH_PYTHON_LAYER`, as well as a couple `ifdef`s. These changes are probably not quite ready for merge; in particular, this should break the cmake build system, and the build option is not being tested in Travis yet. (@akosiorek or others, if you're eager to make this work with cmake, PRs against this branch are welcome.)

Layers with learnable parameters are not supported yet.

In theory, this means you ought to be able to write layers in Theano (making use of Theano's symbolic differentiation), embed them in caffe nets, and solve using caffe. I haven't tried that yet, but it might be worth adding some Python-level helper code for that later on.

This hasn't really been tested at all (but does build and run).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,9 @@ OBJ_BUILD_DIR := $(BUILD_DIR)/src/$(PROJECT)
 LAYER_BUILD_DIR := $(OBJ_BUILD_DIR)/layers
 UTIL_BUILD_DIR := $(OBJ_BUILD_DIR)/util
 OBJS := $(PROTO_OBJS) $(CXX_OBJS) $(CU_OBJS)
+ifeq ($(USE_PYTHON_LAYER), 1)
+	OBJS += python/$(PROJECT)/_$(PROJECT).o
+endif
 # tool, example, and test objects
 TOOL_OBJS := $(addprefix $(BUILD_DIR)/, ${TOOL_SRCS:.cpp=.o})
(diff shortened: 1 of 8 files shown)


longjon:
Upgraded and replaced by #1703.

sumitto:
Can the custom layer defined in python perform calculation in GPU? How do I define GPU operations?

PR #1074, opened 2014-09-12
Enable the users to disable optional dependencies

kloudkl:
OpenCV has [a lot of options](https://github.com/Itseez/opencv/blob/master/CMakeLists.txt#L113) for the optional libraries. Some of the newly added dependencies of Caffe are not needed by all the users and should be made optional during the building process. 

This PR adds make and CMake flags and guarding macros for those components. They are especially useful for a lean deployment. But they are still enabled by default to keep the building behavior consistent.


Nerei:
Issues https://github.com/BVLC/caffe/issues/1738
PR https://github.com/BVLC/caffe/pull/1667

`WITH_` doesn't mean `HAVE_`. Former is wish, latter is reality. Sometimes the user's desire, can't be fulfilled due to various reasons (ex. dependency not found, incompatible configurations or versions, etc). So having `WITH_` and `HAVE_` is a good practice.

It doesn't matter whether to use `WITH_` or `USE_`. Use any of them for your preference.

shelhamer:
Closing since the dev branch is deprecated. Please send PRs to master. See also #2523.

PR #1131, opened 2014-09-22
switch data layer default source to lmdb from leveldb

shelhamer:
`DATA` layers now default to lmdb storage instead of leveldb.  See #1128 for reasoning.

**Update your model definitions if the models rely on the leveldb default for loading your input data by setting `backend: LEVELDB` in `data_layer_param`.

- [ ] try both formats
- [ ] warn if mismatched


shelhamer:

PR #1135, opened 2014-09-22
Fix CMake files on centos 6

yangcha:
On Centos 6, the CMake files is broken. I have a fix and have tested on Centos 6.5.


shelhamer:
The CMake build is renovated in #1667. Any further centos tips could be added to the installation section of the wiki.

PR #1137, opened 2014-09-22
Top k layer

ducha-aiki:
This layer implements k-sparse non-linearity from paper "k-Sparse Autoencoders" http://arxiv.org/abs/1312.5663 . 
Top k values goes to the top layer, rest are set to zeros.
For now it is implemented only in CPU version.


shelhamer:

PR #1148, opened 2014-09-23
Parallel / distributed training

shelhamer:
Parallel and distributed training of Caffe models by streaming gradients among solvers. Parallelizes training without redefining models. This is the integration PR for @cypof's original contribution in #1140.

**TODO**

- [ ] switch all threads to boost::thread with [thread local storage](http://www.boost.org/doc/libs/1_35_0/doc/html/thread/thread_local_storage.html)
- [ ] fix distributed hyper-parameter schedule
- [ ] validate parallelism on multi-GPU Caffenet training
- [ ] lint

Follow-up
- [ ] distributed synchronous SGD for comparison with asynchronous / HogWild approach
- [ ] synchronization by device peer-to-peer communication instead of host reduce

Please review and collaborate here. The original details of #1140 by @cypof are:

-----------------------------------------

A set of classes to synchronize SGD between multiple solvers. Based on the Hogwild paper, and our work at Flickr to extend the model to GPUs and distributed configurations by streaming gradients between solvers.

## Features

* Models can be trained in parallel without modification. Caffe’s training code is also mostly untouched.
* Modular design. The code is broken down in simple components that synchronize one segment, e.g. CPU/GPU, CPU/LAN. They can be combined to create an architecture, either in-process or between processes by memory-mapping the weights to /dev/shm.
* Works on commodity hardware. Apparently even on 1G Ethernet, at least for mnist. Synchronization and SGD run asynchronously, to keep both compute and networking resources fully utilized. Bandwidth and latency across machines are optimized using raw sockets and user-space networking.
* No additional memory used on the GPUs.

## Limitations

* Only supports data-parallelism. Limited forms of model-parallelism should be possible using the same components but no work has been done.
* Training is less stable than on a single GPU. In particular, disabling momentum at least at the beginning of training seems to help.
* No deployment / monitoring tools. We are looking at integrating with IPython.parallel.

## Tests

Early results on MNIST seem to show linear scaling. We tested on up to 6 machines with 4 solvers each for CPU, and 2 machines with 2 GPUs each. GPUs do not perform well on this small network but still seem to scale linearly.

mnist

In the weeks to come we plan to start testing on larger networks and clusters. Currently our GPU machines are connected through 1G Ethernet, please contact us if you are interested to help benchmarking on better hardware.

## Architecture

We made the Caffe singleton thread-local, to allow multiple solvers to run in parallel on their own thread. Synchronization works by sharing the weight buffers between solvers in the same address space, and by asynchronously measuring and exchanging gradients between address spaces.

## Bugs / Todos

* If the bandwidth between the GPU and host is set too high, machines seem to hang.
* Incorrect hyper-params schedule in the distributed case. The total count of iterations needs to be tracked, maybe through the monitoring tool.
* Thread-local Caffe singletons are not destroyed, we need to design a proper shutdown strategy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROJECT := caffe
 
-CONFIG_FILE := Makefile.config
+CONFIG_FILE ?= Makefile.config
 include $(CONFIG_FILE)
 
 BUILD_DIR_LINK := $(BUILD_DIR)
@@ -269,7 +269,7 @@ endif
 
 # Debugging
 ifeq ($(DEBUG), 1)
-	COMMON_FLAGS += -DDEBUG -g -O0
+	COMMON_FLAGS += -DDEBUG -g -O0 -DBOOST_NOINLINE='__attribute__ ((noinline))'
 	NVCCFLAGS += -G
 else
 	COMMON_FLAGS += -DNDEBUG -O2
@@ -289,6 +289,11 @@ ifeq ($(CPU_ONLY), 1)
 	ALL_WARNS := $(ALL_CXX_WARNS)
 	TEST_FILTER := --gtest_filter="-*GPU*"
 	COMMON_FLAGS += -DCPU_ONLY
+endif
+
+ifeq ($(RDMA), 1)
+	COMMON_FLAGS += -DRDMA
+	LIBRARIES += ibverbs ibumad
 endif
 
 # BLAS configuration (default = ATLAS)
@@ -545,6 +550,13 @@ $(UTIL_BUILD_DIR)/%.o: src/$(PROJECT)/util/%.cpp $(HXX_SRCS) | $(UTIL_BUILD_DIR)
 
 $(GTEST_OBJ): $(GTEST_SRC) | $(GTEST_BUILD_DIR)
 	$(CXX) $< $(CXXFLAGS) -c -o $@ 2> $@.$(WARNS_EXT) \
+		|| (cat $@.$(WARNS_EXT); exit 1)
+	@ cat $@.$(WARNS_EXT)
+	@ echo
+
+$(OBJ_BUILD_DIR)/%.cuo: src/$(PROJECT)/%.cu $(HXX_SRCS) \
+		| $(LAYER_BUILD_DIR)
+	$(CUDA_DIR)/bin/nvcc $(NVCCFLAGS) $(CUDA_ARCH) -c $< -o $@ 2> $@.$(WARNS_EXT) \
 		|| (cat $@.$(WARNS_EXT); exit 1)
 	@ cat $@.$(WARNS_EXT)
 	@ echo
--- a/Makefile.config.example
+++ b/Makefile.config.example
@@ -7,6 +7,9 @@
 # CPU-only switch (uncomment to build without GPU support).
 # CPU_ONLY := 1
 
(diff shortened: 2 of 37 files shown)


bhack:
@cypof and others. Anybody already checked [ParamterServer caffe]( https://github.com/petuum/public/wiki/CNN:-Caffe-on-Petuum) effort at CMU

bhack:
See also the last doc link of [this ticket]( https://issues.apache.org/jira/plugins/servlet/mobile#issue/SPARK-4590)

immars:
compilation failure?
nvcc version 6.5, V6.5.12, g++ 4.8.2

```
src/caffe/parallel.cu(14): error: name followed by "::" must be a class or namespace name
          detected during instantiation of "void caffe::sync_master_kernel(Dtype *, Dtype **, size_t *, int, int, const cudaStream_t &, size_t) [with Dtype=float]"
```

immars:
another implementation:

https://github.com/sailorsb/caffe-parallel

data parallelism over MPI

raingo:
@cypof How this graph generated? Is it on training data or testing data?

image

cypof:
It was on testing data, the validation set.

wangsheng1001:
Hi,

I got segmentation fault when running the cifar example. 
I do not run as root, but with 'setcap cap_net_raw+ep'.
Details are as follows:

Environment:
Ubuntu 13.04  kernel 3.8.0

Command:
build/examples/parallel/raw.bin examples/parallel/cifar_solver.prototxt 1 0025908a56d8:0025908a4dfc

Error:
E0609 21:21:57.113490 15274 raw.cpp:99] # of gpus 0# of cpus 1
*** Aborted at 1433856117 (unix time) try "date -d @1433856117" if you are using GNU date ***
PC: @           0x46b802 caffe::RawSync<>::run()
*** SIGSEGV (@0x5d1aadf80) received by PID 15274 (TID 0x7d60af7fe700) from PID 18446744072932220800; stack trace: ***
    @     0x7f60bede3bd0 (unknown)
    @           0x46b802 caffe::RawSync<>::run()
    @     0x7f60c06ffaf2 thread_proxy
    @     0x7f60beddbf8e start_thread
    @     0x7f60beb05a0d (unknown)
Segmentation fault (core dumped)

More Details from gdb:
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7efff0c28700 (LWP 15611)]
0x000000000046b802 in caffe::RawSync<float>::run() ()
(gdb) bt
#0  0x000000000046b802 in caffe::RawSync<float>::run() ()
#1  0x00007ffff5999af2 in thread_proxy () from /users/dinhtta/local/lib/libboost_thread.so.1.56.0
#2  0x00007ffff4075f8e in start_thread (arg=0x7efff0c28700) at pthread_create.c:311
#3  0x00007ffff3d9fa0d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:113

shelhamer:
Instead of attacking all axes of parallelism at once, see the multi-GPU data parallelism of #2114 for a start. Closing this since sync SGD with data parallelism covers a lot of cases. However, it clearly does not cover all so this PR can serve as the historical record for the good pieces we might pick up later.

bhack:
To add something post-modern to the historical record I reference https://github.com/Yangqing/caffe2/issues/11

PR #1194, opened 2014-09-30
Add Layer::Reset method to reinitialize internal state

jeffdonahue:
The main purpose is to make testing run over the same set of data in each test pass reliably, without having to worry about setting `test_iter` and `batch_size` to sweep through the dataset exactly once.  Instead, we "Reset" to the beginning of the dataset every time.  Enable by adding `test_net_reset: true` to any SolverParameter.  With the setting enabled (and assuming all the layers in the test net support it), even if `T = test_iter * batch_size` is set to something other than the total dataset size (`D`), the same dataset will be tested in each test pass; you'll just never see some of the dataset (if `T < D`) or will see some of the dataset more than once (if `T > D`), but results will be consistent in the sense that you'll always miss/repeat the same subset of the dataset.

Current limitations:
-No unit tests (I've tried it out for `HDF5DataLayer` and it works based on the output of `debug_info`, but completely untested for other `DataLayer`s)
-Not implemented for anything using the RNG as we don't have a way to save and restore the RNG state (I've looked into this -- it seems to be completely impossible to do for curand...)


jeffdonahue:
replaced by #2101

PR #1205, opened 2014-10-02
Shuffle hdf5 input data

PatWie:
After loading a hdf5data file the order of the data can be changed
by setting "shuffle:true" in prototxt for the specific layer. Per-
mutation is done after switching to a new hdf5 file or after each
of the data in the hdf5 file was presented to the neural network.

(compilation was tested)


etienne87:
is is possible to do this in train_val directly, rather than in database creation? this would be easier to have unique solution for any database.

PatWie:
the pull request is deprecated. See https://github.com/BVLC/caffe/pull/2118

PR #1214, opened 2014-10-03
Added global_pooling to set the kernel size equal to the bottom size

sguada:
It allows to do global pooling over the height x width of the bottom blob. So we omit the kernel_size which gets equal to bottom.

Given that the #594 allows on-the-fly resizing of the blobs, this would resize the kernel_size accordingly to the size of the bottom.


Coderx7:
whats the use of global pooling ? I couldnt find any meaningful comments in the source codes!

PR #1220, opened 2014-10-05
cmake compile under Mac OS X

pluskid:
1. Use `Accelerate / vecLib` as atlas by default if under OS X. Also the path for framework header files seem to changed a bit in OS X 10.10.
2. I moved the the library finding code from the `src/caffe` directory to top-level cmake script, too. Because I have problems finding header files when compiling for the `tools` directory.
3. A small fix to `gtest` is included by adding `-DGTEST_USE_OWN_TR1_TUPLE=1` to avoid compiling conflict (`fatal error: 'tr1/tuple' file not found`) with clang.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ set(CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE}) # set release flags
 
 #    Link Flags
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")  # clang++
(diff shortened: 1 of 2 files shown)


shelhamer:
#1667 is taking care of the `Accelerate / vecLib` issue but 5cb2c41 might still be needed for Yosemite.

shelhamer:
> 3. A small fix to gtest is included by adding -DGTEST_USE_OWN_TR1_TUPLE=1 to avoid compiling conflict (fatal error: 'tr1/tuple' file not found) with clang

This is also addressed by #1667. Thanks for noting these fixes @pluskid! Since #1667 covers all kinds of CMake improvements, I'm closing this in favor of that PR.

PR #1228, opened 2014-10-07
Refactor Solver to allow interactive stepping

longjon:
In this PR, the main loop of `Solver::Solve` is factored out into `Solver::Step`, which is exposed to Python. This allows interactive stepping of the solver.

Warning: I've only actually tested a parallel version of this on another branch.

### DIFF
--- a/src/caffe/solver.cpp
+++ b/src/caffe/solver.cpp
@@ -32,6 +32,7 @@ void Solver<Dtype>::Init(const SolverParameter& param) {
   LOG(INFO) << "Initializing solver from parameters: " << std::endl
             << param.DebugString();
   param_ = param;
+  CHECK_GE(param_.average_loss(), 1) << "average_loss should be non-negative.";
   if (param_.random_seed() >= 0) {
     Caffe::set_random_seed(param_.random_seed());
   }
@@ -39,6 +40,8 @@ void Solver<Dtype>::Init(const SolverParameter& param) {
   InitTrainNet();
   InitTestNets();
   LOG(INFO) << "Solver scaffolding done.";
+  iter_ = 0;
+  current_step_ = 0;
 }
 
 template <typename Dtype>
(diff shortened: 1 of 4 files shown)


longjon:
Updated: the snapshot logic has been changed to properly snapshot after every *k*th iteration, instead of snapshotting before the iteration following every *k*th. (In other words, snapshotting has been moved to the end of the solver loop.)

This simplifies the logic (except the "snapshot after train" logic), and ensures that `Step(k)` will end up with a snapshot on disk, instead of waiting for the next `Step` call, which can be important if something goes wrong between those two calls.

src/caffe/solver.cpp
```diff
@@ -227,13 +210,44 @@ void Solver<Dtype>::Solve(const char* resume_file) {
...
+
+    // Save a snapshot if needed.
+    if (param_.snapshot() && (iter_ + 1) % param_.snapshot() == 0) {
+      Snapshot();
+    }
+  }
+}
+
+template <typename Dtype>
+void Solver<Dtype>::PreSolve() {
```
jeffdonahue: IMO since this method is common setup specific to the `Solver` class itself, it should just be given a new name like `Initialize()` (and then called separately before  `PreSolve()` @ line 236), so that anyone subclassing Solver doesn't need to remember to call it.

longjon: I agree that forcing subclassers to call `PreSolve` is awkward. Is there any reason we don't just call `PreSolve` from `Init`? That would simplify the logic and get rid of `initialized_`, as there would be no such thing as an uninitialized solver, and then it makes sense to do the base class initialization right in `Init`. I think I'll try it and update the PR.

jeffdonahue: Sounds good to me!

src/caffe/solver.cpp
```diff
@@ -155,39 +159,15 @@ void Solver<Dtype>::InitTestNets() {
...
-  iter_ = 0;
-  current_step_ = 0;
-  if (resume_file) {
-    LOG(INFO) << "Restoring previous solver status from " << resume_file;
-    Restore(resume_file);
-  }
-  // Remember the initial iter_ value; will be non-zero if we loaded from a
-  // resume_file above.
+void Solver<Dtype>::Step(int iters) {
+  vector<Blob<Dtype>*> bottom_vec;
```
jeffdonahue: can be removed since you changed the call to `ForwardPrefilled`?

src/caffe/solver.cpp
longjon: Actually, we can't do that, because virtual functions can't be called from constructors. Here's a slightly more radical idea, which I've implemented now: get rid of `Solver::PreSolve` and just use the constructors, since constructors now provide the same functionality. `SGDSolver::PreSolve`, which was the only actual `PreSolve`, remains as it was and gets called from the `SGDSolver` constructor.

src/caffe/solver.cpp
longjon: The changed call is in `Solve`, this is in `Step`, where the call is to `ForwardBackward`. An alternative is to remove `bottom_vec` but call `ForwardPrefilled` and `Backward` separately; that's what I've now implemented.

longjon: Actually, I think I'll revert that, because it makes #1663 awkward. We should probably just clean up the `Net` interface to avoid these dummy vectors at some later point.

longjon:
@jeffdonahue I think I'm satisfied with this now; merge when ready.

jeffdonahue:
Cool, thanks Jon

PR #1236, opened 2014-10-07
Legacy nvcc support

mlapin:
Hey,

unfortunately, commit 0ba046bc3ec7362b49d239d9d0947c2598a75ef0 (merge PR #1070) breaks compilation in our legacy environment (we currently use cuda toolkit 5.0 and nvcc with gcc 4.4). The compilation errors we get are of the following type:
```
emmintrin.h(1312): error: identifier "__builtin_ia32_vec_ext_v8hi" is undefined
```
This could be related to a similar issue with intrinsics when compiling under OSX (mentioned in #1070).

The proposed fix is to avoid inclusion of the opencv header when compiling with nvcc and the host gcc is < 4.7.

I kept the original workaround with the OSX flag for now, but my guess is the whole check could be simplified as:
```
#if __NVCC__
namespace cv {class Mat;}
#else
#include <opencv2/core/core.hpp>
#endif
```
that is, the cuda code does not actually require opencv, except for the class declaration `cv::Mat`.

The downside of this approach is that `ReadImageToCVMat` is not inlined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -250,7 +250,6 @@ ifeq ($(OSX), 1)
 	endif
 	# boost::thread is called boost_thread-mt to mark multithreading on OS X
 	LIBRARIES += boost_thread-mt
-        NVCCFLAGS += -DOSX
 endif
 
 # Custom compiler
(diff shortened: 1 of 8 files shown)


shelhamer:
This ended up fine on OS X. While I'm not so enthused to bring an OpenCV declaration into `common.hpp` if it helps compatibility it's fine for now.

Nerei:
@sguada 
> This solution doesn't work for vectorcv::Mat, any idea why?

From my knowledge,I guess if you use vector::resize() which takes matrix by value, class definition is required.

PR #1238, opened 2014-10-07
leveldb/lmdb refactoring

kmatzen:
This refactoring places leveldb and lmdb access behind a common interface.  The interface supports:
- open
- close
- put
- sequential scan

It would be nice to use this interface to implement an even simpler database where binary blobs are written sequentially to file and then mmapped in during training.  It's not clear why leveldb and lmdb are used when all that's being done is a sequential scan and I'd like to have timing measurements to learn what is best.

~~There are also a few bug fixes scattered throughout this pull request.  io.hpp allocates too little memory for a strcpy, there's a delete/delete[] mismatch, opencv 3.0 alpha doesn't work as-is, and more.~~ See #1261.

Existing tests pass, but none of the tools are tested as far as I can tell.  This patch might also break any special leveldb or lmdb configurations.  There wasn't an explanation as to why some flags were chosen, so I copied and pasted flags blindly.

### DIFF
--- a/include/caffe/dataset.hpp
+++ b/include/caffe/dataset.hpp
@@ -1,6 +1,8 @@
 #ifndef CAFFE_DATASET_H_
 #define CAFFE_DATASET_H_
 
+#include <google/protobuf/message.h>
+
 #include <algorithm>
 #include <iterator>
 #include <string>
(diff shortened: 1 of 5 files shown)


KangolHsu:
@shelhamer I got the similar error，but it occurs all of sudden。i swear i did change anything。


db_lmdb.hpp:16] Check failed: mdb_status == 0 (-30790 vs. 0) MDB_READERS_FULL: Environment maxreaders limit reached

Issue #1158, opened 2014-09-25
C++ extracted features are a lot worse than python and matlab wrapper


wangboyunze:
Hi, 

I am using caffe to extract CNN features from pretrained model by imagenet. Everything goes smooth. But I find a problem with C++ code that, if I use C++ code to do the feature extraction as instructed in http://caffe.berkeleyvision.org/gathered/examples/feature_extraction.html, the features are a lot different than features extracted by python and matlab wrappers( these two wrappers give same results!) and the C++ extracted features have a lot lower accuracy on my test samples... 

Do you have any comments or suggestions? many thanks!

anurikadisha:
Hi @escorciav 
I tried to run your script to extract the features from lmdb format. However I think I am making some mistake because I dont go into the for loop of your script. What should I give as an input instead of dirname in db = leveldb.LevelDB(dirname). I have two mdb files 'data.mdb' and 'lock.mdb' in 'features' folder. Currently I ran the following code:

    import leveldb
    import caffe
    from caffe.proto import caffe_pb2
    db = leveldb.LevelDB('features')
    for key, val in db.RangeIter():
        datum = caffe.io.caffe_pb2.Datum()
        datum.ParseFromString(val)
        arr = caffe.io.datum_to_array(datum) 
 
 Can someone please help?

HanuManuOm:
features_path = os.path.join(root_path, 'vgg_feats.mat')
  features_struct = scipy.io.loadmat(features_path)
Can you please give me equivalent code for this in leveldb? I do not have vgg features, I have leveldb features. i.e. MANIFEST-000002 and I want to use it in my python code. Please help me as soon as possible.

shelhamer:
Please ask usage questions on the mailing list. Reproducing results across C++, pycaffe, and matcaffe can take some care as the different interfaces have different local conventions as expected by the respective communities.

From https://github.com/BVLC/caffe/blob/master/CONTRIBUTING.md:

> _Please do not post usage, installation, or modeling questions, or other requests for help to Issues._
Use the [caffe-users list](https://groups.google.com/forum/#!forum/caffe-users) instead. This helps developers maintain a clear, uncluttered, and efficient view of the state of Caffe.

Aminullah6264:
Hello, i have issue in python feature extraction.
i extract features with python and train my model.
now i am testing it in python when i extract features it is different from feature extracted using matlab
Need your help

PR #1253, opened 2014-10-10
Fix the order in the extract_features

ducha-aiki:
Fixing #1158 by add trailing zeros to beginning of the key to extract features.


shelhamer:
Sorry this was missed! It was certainly a fix, although the corresponding fix for master was done in #2514.

PR #1263, opened 2014-10-11
Updated Reshape layer

ssafar:
Here is a layer similar to sguada's #108. Main differences:
 * like FlattenLayer, it doesn't copy blobs
 * has numpy-like "-1 for automatic dimension inference" and uses the default 0 to copy dimensions from below (just like it was planned in #108)
 * is using the latest API.

As for unifying FlattenLayer and this one, also discussed in #108: technically it could be doable but they have really different defaults. For ReshapeLayer, they are 0 0 0 0 (that is, don't change any dimensions unless explicitly requested), while for a FlattenLayer-like default behavior, we'd need 0 -1 1 1 ("keep num, infer channels, flatten everything else). Also, it might be confusing to have something called RESHAPE automatically flatten things or FLATTEN with parameters doing something entirely different from flattening. Or at least that was my reasoning behind creating another layer; it might still be wrong :)

Also, sorry if it's already work in progress by someone; I was looking at the dates on #108, concluded that I'd like to use such a thing most likely sooner than its expected arrival time, and that it might be useful if I posted it if I was going to implement it for myself anyway. Let me know if this wasn't really the right way of doing things :)

### DIFF
--- a/docs/tutorial/layers.md
+++ b/docs/tutorial/layers.md
@@ -419,6 +419,48 @@ The `SPLIT` layer is a utility layer that splits an input blob to multiple outpu
 
 The `FLATTEN` layer is a utility layer that flattens an input of shape `n * c * h * w` to a simple vector output of shape `n * (c*h*w) * 1 * 1`.
 
+#### Reshape
+
+* LayerType: `RESHAPE`
+* CPU implementation: `./src/caffe/layers/reshape_layer.cpp`
+* CUDA GPU implementation: `./src/caffe/layers/reshape_layer.cu`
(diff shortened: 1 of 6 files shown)


shelhamer:
This looks good to me but needs a `master` edition. @ssafar if you have a chance that's ideal, or else we could merge to master from this PR at some point.

jeffdonahue:
rebased in #2217

ssafar:
@jeffdonahue: thanks for the rebase & fixing it up! (& sorry for not doing it myself & disappearing for so long, I had neither a proper GPU-equipped box nor time lately. But then it's all good in the end :))

PR #1268, opened 2014-10-13
Link with atlas on centos 7.

mdenil:
On centos 7 atlas and cblas are compiled together into a single library called tatlas (there is also satlas which is a single threaded version).  This commit adds a new BLAS choice that links against this differently named library.


shelhamer:
I'd rather have cross platform details that deviate from linux "standards" be documented in the installation part of the wiki. Thank for noting how to do BLAS on centos.

PR #1291, opened 2014-10-16
Share convolution buffers to reduce memory usage

shelhamer:
Share the columnation buffer for im2col / col2im transformations across all Caffe convolution layers. The memory usage is now equal to the maximum buffer size instead of the sum over all layers. In particular this is useful for many-layered architectures like the VGG ILSVRC14 19 layer model.

Advice and Cautions:

- This is worth it for fully-convolutional models where Caffe convolution is faster than cuDNN.
- No parallelism. Only a single net can do forward / backward at a time. As the buffer is shared by all layers within and across nets, no convolution can be done in parallel. (A fix for parallel nets is to make the buffer a member of net. DAG parallelism is still out in that case, but isn't currently parallelized anyway.)
- This has no effect on cuDNN convolution, but that consumes less memory anyway.

All credit to @longjon who reshaped our world in #594 and suggested this patch in https://github.com/BVLC/caffe/pull/520#issuecomment-59000788.

**Do not merge.**


shelhamer:
Replaced by #2016.

PR #1313, opened 2014-10-17
Reshape single input batches for inputs of varying dimension

shelhamer:
To feed inputs of varying dimension, the DATA layer reshapes its prefetch and top blobs when the batch size is 1. This is useful for models of variable input size, such as fully-convolutional models.

By the grace of #594 this is a simple change.

- [x] simplify to reshape always (there's no cost)
- [x] make compatible with crop thanks to @philkr 
- [x] test data layer reshaping

### DIFF
--- a/src/caffe/layers/base_data_layer.cpp
+++ b/src/caffe/layers/base_data_layer.cpp
@@ -61,6 +61,9 @@ void BasePrefetchingDataLayer<Dtype>::Forward_cpu(
   // First, join the thread
   JoinPrefetchThread();
   DLOG(INFO) << "Thread joined";
+  // Reshape to loaded data.
+  top[0]->Reshape(this->prefetch_data_.num(), this->prefetch_data_.channels(),
+      this->prefetch_data_.height(), this->prefetch_data_.width());
   // Copy the data
   caffe_copy(prefetch_data_.count(), prefetch_data_.cpu_data(),
              top[0]->mutable_cpu_data());
--- a/src/caffe/layers/data_layer.cpp
+++ b/src/caffe/layers/data_layer.cpp
@@ -49,7 +49,7 @@ void DataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>& bottom,
   int crop_size = this->layer_param_.transform_param().crop_size();
   if (crop_size > 0) {
     top[0]->Reshape(this->layer_param_.data_param().batch_size(),
-                       datum.channels(), crop_size, crop_size);
+        datum.channels(), crop_size, crop_size);
     this->prefetch_data_.Reshape(this->layer_param_.data_param().batch_size(),
         datum.channels(), crop_size, crop_size);
     this->transformed_data_.Reshape(1, datum.channels(), crop_size, crop_size);
@@ -83,13 +83,25 @@ void DataLayer<Dtype>::InternalThreadEntry() {
   CPUTimer timer;
   CHECK(this->prefetch_data_.count());
   CHECK(this->transformed_data_.count());
+
+  // Reshape on single input batches for inputs of varying dimension.
+  const int batch_size = this->layer_param_.data_param().batch_size();
+  const int crop_size = this->layer_param_.transform_param().crop_size();
+  if (batch_size == 1 && crop_size == 0) {
+    Datum datum;
+    datum.ParseFromString(cursor_->value());
+    this->prefetch_data_.Reshape(1, datum.channels(),
+        datum.height(), datum.width());
+    this->transformed_data_.Reshape(1, datum.channels(),
+        datum.height(), datum.width());
+  }
+
   Dtype* top_data = this->prefetch_data_.mutable_cpu_data();
   Dtype* top_label = NULL;  // suppress warnings about uninitialized variables
 
   if (this->output_labels_) {
     top_label = this->prefetch_label_.mutable_cpu_data();
   }
-  const int batch_size = this->layer_param_.data_param().batch_size();
   for (int item_id = 0; item_id < batch_size; ++item_id) {
     timer.Start();
     // get a blob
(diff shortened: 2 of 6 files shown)


shelhamer:
@longjon I'll take care of this once #1568 is reviewed and finished.

bhack:
@longjon Can you check some of the last @sguada comment on our https://github.com/BVLC/caffe/pull/1416? I think that if we want to let top reshaping a forward business we need to use it in the same way in this two PR.

sguada:
@bhack yes the reshape should happen in the same way here and in #1416. So now re-reading this PR, maybe it is ok in `forward`

sguada:
@shelhamer I think to clean the reshape, what we should do is use `Data_Transformation` to process a `datum` and get the dimensions from that.

bhack:
@sguada @longjon Ok see @mtamburrano comments on https://github.com/BVLC/caffe/pull/1416 and let us know if our interpretation is aligned with this one. So we can plan to reserve time to fix last things needed to allow it to be merged.

philkr:
It seems this is incompatible with crop. Cropped images don't need to be reshaped. Here is a small diff to fix it.
```
diff --git a/src/caffe/layers/data_layer.cpp b/src/caffe/layers/data_layer.cpp
index 8e8e835..f1ed9aa 100644
--- a/src/caffe/layers/data_layer.cpp
+++ b/src/caffe/layers/data_layer.cpp
@@ -85,7 +85,8 @@ template <typename Dtype>
 void DataLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
   // Reshape on single input batches for inputs of varying dimension.
-  if (this->prefetch_data_.num() == 1) {
+  int crop_size = this->layer_param_.transform_param().crop_size();
+  if (crop_size == 0 && this->prefetch_data_.num() == 1) {
     top[0]->Reshape(1, this->prefetch_data_.channels(),
         this->prefetch_data_.height(), this->prefetch_data_.width());
   }
@@ -104,7 +105,8 @@ void DataLayer<Dtype>::InternalThreadEntry() {
 
   const int batch_size = this->layer_param_.data_param().batch_size();
   // Reshape on single input batches for inputs of varying dimension.
-  if (batch_size == 1) {
+  int crop_size = this->layer_param_.transform_param().crop_size();
+  if (crop_size == 0 && batch_size == 1) {
     Datum datum = iter_->value;
     this->prefetch_data_.Reshape(1, datum.channels(),
         datum.height(), datum.width());
```

longjon:
A note: there currently seems to be a race condition between `DataLayer::Reshape` and the prefetch thread. It should be fixed in @shelhamer's next update to this PR (the reshaping code just needs to go in `BasePrefetchingDataLayer::Forward_*` after the thread is joined).

shelhamer:
@longjon re: https://github.com/BVLC/caffe/pull/1313#issuecomment-68018501 the `DataLayerSetup` reshaping is needed if the net is to have any dimensions before a call to `Forward`, as needed for diagnostic output during initialization and for the data layer tests.

It does seem like this could be simplified so that data layers unconditionally reshape, but we need to decide what to do about `Net::Init()`'s reporting.

shelhamer:
By the way this is still broken w.r.t. crop -- I'll fix and add a test.

shelhamer:
@longjon DATA + IMAGE_DATA now reshape with tests. Could you review + merge?

src/caffe/layers/base_data_layer.cpp
```diff
@@ -61,6 +61,9 @@ void BasePrefetchingDataLayer<Dtype>::Forward_cpu(
   // First, join the thread
   JoinPrefetchThread();
   DLOG(INFO) << "Thread joined";
+  // Reshape to loaded data.
+  top[0]->Reshape(1, this->prefetch_data_.channels(),
```
longjon: Should this be `Reshape(this->prefetch_data_.num()`? This is unconditional, not just batch size one, right?

src/caffe/layers/data_layer.cpp
```diff
@@ -83,13 +83,25 @@ void DataLayer<Dtype>::InternalThreadEntry() {
...
+
+  // Reshape on single input batches for inputs of varying dimension.
+  const int batch_size = this->layer_param_.data_param().batch_size();
+  const int crop_size = this->layer_param_.transform_param().crop_size();
+  if (batch_size == 1 && crop_size == 0) {
+    Datum datum;
+    datum.ParseFromString(cursor_->value());
+    this->prefetch_data_.Reshape(1, datum.channels(),
+        datum.height(), datum.width());
+    this->transformed_data_.Reshape(1, datum.channels(),
```
longjon: This is not a comment about this PR, but the code that follows is pretty confusing wrt `transformed_data_`: `prefetch_data_` becomes `top_data`, which then becomes `transformed_data_`... somehow they're all the same!

We should probably at least rename `top_data` and maybe eliminate `transformed_data_`, (but not in this PR).

src/caffe/layers/data_layer.cpp
```diff
@@ -83,13 +83,25 @@ void DataLayer<Dtype>::InternalThreadEntry() {
   CPUTimer timer;
   CHECK(this->prefetch_data_.count());
   CHECK(this->transformed_data_.count());
+
+  // Reshape on single input batches for inputs of varying dimension.
+  const int batch_size = this->layer_param_.data_param().batch_size();
+  const int crop_size = this->layer_param_.transform_param().crop_size();
+  if (batch_size == 1 && crop_size == 0) {
```
longjon: Why not just do this unconditionally?

shelhamer: This keeps the varying input reshape logic needed for each input separate from the crop reshape logic that's needed once in setup (as discussed offline).

dxj19831029:
em... I think the python's wrapper need to adjust as well, otherwise the preprocess will always resize the input image.

PR #1324, opened 2014-10-19
extract_features: fix crash when exporting multiple features

jyegerlehner:
The problem is when we obtain the db_type argument to pass to database factory: arg_pos keeps getting incremented as we loop through num_features. It works when num_features=1, but on successive iterations of the loop the arg_pos increments past the location of the db_type argument:
`DatasetFactory<std::string, Datum>(argv[++arg_pos]);`

I also changed the number of images whose features are cached in memory before committing to disk. Prior to this change, it was hard-wired to 1000. For large nets with lots of conv channels that gets to be a lot of memory. There's no reason to keep so much in memory before writing out to disk. So I replaced the literal 1000 with an optional command-line-argument-selectable size that defaults to 100. Makes the difference between an export that crashes upon memory starvation on my 8GB machine, and one that succeeds.

### DIFF
--- a/tools/extract_features.cpp
+++ b/tools/extract_features.cpp
@@ -37,10 +37,11 @@ int feature_extraction_pipeline(int argc, char** argv) {
     LOG(ERROR)<<
     "This program takes in a trained network and an input data layer, and then"
     " extract features of the input data produced by the net.\n"
-    "Usage: extract_features  pretrained_net_param"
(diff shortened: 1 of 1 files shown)


shelhamer:
Sorry this was missed! It was certainly a fix, although the corresponding fix for master was done in #2514.

PR #1311, opened 2014-10-16
Persistent transformations

sguada:
This improvement to DataTransform allows us to have a persistent state, which allows to apply the same transformation (cropping, mirror) to multiple images. There is a `ResetState` to reset the state and compute a new transformation for the next image.

@longjon Could you take a look and let me know your opinion?

PR #1328, opened 2014-10-20
Simplify DataTransform

sguada:
This PR tries to unify the code for the different types of data (Datum, cv::Mat, Blob) to be transformed.

It adds the option to process vectors of data, persistency and state to the Transformer. Persistency as proposed in #1311 allows to apply the same transformation to different data. #1311 has been integrated here.

@longjon could you take a look? I cannot compile it due to some template error.


shelhamer:
@sguada should this be closed or is it still part of the data plan?

sguada:
@shelhamer I think we should close this PR and when needed open a new one.

PR #1347, opened 2014-10-22
shuffle data from hdf5 datasets

PatWie:
The order of read HDF5 files and the order of the entries of the HDF5 files can be shuffled when setting the flag `shuffle` in the hdf5data layer

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -499,6 +499,8 @@ message HDF5DataParameter {
   optional string source = 1;
   // Specify the batch size.
   optional uint32 batch_size = 2;
+  // Specify shuffling the order of training data
+  optional bool shuffle = 3 [default = false];
 }
 
 // Message that stores parameters used by HDF5OutputLayer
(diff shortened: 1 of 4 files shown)


shelhamer:
@jeffdonahue can you review and merge if this looks good to you?

src/caffe/proto/caffe.proto
```diff
@@ -499,6 +499,8 @@ message HDF5DataParameter {
   optional string source = 1;
   // Specify the batch size.
   optional uint32 batch_size = 2;
+  // Specify shuffling the order of training data
```
jeffdonahue: Please clarify this comment to explain that the HDF5 files themselves are shuffled but the order within any given file is fixed.

wieschoo: Everything will be shuffled: hdf5 files and entries in these hdf5 files.

jeffdonahue: Thanks, I see that now, my bad.  I think it should still be clarified though -- it's not actually a full shuffle of the dataset (i.e., some orderings of the dataset are impossible to obtain) unless you only have a single HDF5 file (or each HDF5 file only has a single entry).

jeffdonahue:
Replaced by #2118.

PR #1386, opened 2014-10-31
Data augmentation via random rotations

drdan14:
Some datasets are indifferent to which direction in the image is "up" and therefore can be augmented with random rotations. This PR implements a `rotate` option in the `TransformationParameter` message which operates analogously to `mirror`. It's `false` by default; setting it to `true` causes images to be randomly rotated by an integer multiple of 90 degrees. If `mirror` is also enabled, there are 8 possibilities for the randomly transformed output image.

An error is thrown if the input data is not square (i.e., if `datum.height() != datum.width()`); the logic for dealing with square rotations is easier to start with, and I believe that `transformed_data` is expected to have the same shape as the input data.

I added unit tests to `test_data_transformer.cpp` by copying and pasting some pre-existing mirroring unit tests; the tests show that this transformation runs, but they don't prove that it does what I expect it to do (i.e., rotate an image properly). I'm not a pro at Caffe unit tests, so if anyone can suggest how to write a unit test that confirms that the rotation is working properly, I'd be happy to implement it. I can (and probably will) copy the logic to a little test script in Matlab to convince myself that the logic is working, but that's not really good enough as a unit test.

~~I have not yet tried to train a network using the `rotate` parameter but that's next on my list. In the mean time, eyeballs on my rotation (and generalized mirroring) logic are welcome.~~

*Update:* I've trained a network using the random rotations and it worked just fine (if not necessarily any better than when random rotations were disabled).

### DIFF
--- a/src/caffe/data_transformer.cpp
+++ b/src/caffe/data_transformer.cpp
@@ -47,6 +47,7 @@ void DataTransformer<Dtype>::Transform(const Datum& datum,
   const int crop_size = param_.crop_size();
   const Dtype scale = param_.scale();
   const bool do_mirror = param_.mirror() && Rand(2);
+  const int rotation_count = (param_.rotate()) ? Rand(4) : 0;
   const bool has_mean_file = param_.has_mean_file();
   const bool has_uint8 = data.size() > 0;
   const bool has_mean_values = mean_values_.size() > 0;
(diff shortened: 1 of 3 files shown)


drdan14:
I don't have the bandwidth to rebase this onto master, so I am closing it.

bhack:
@drdan14 Nevermind. I'm not satisfied with this but often caffe PR are subject to reverend rebase over the months.

PR #1388, opened 2014-11-01
CIFAR10 example documentation bug

rohitgirdhar:
The way the scripts `examples/cifar10/create_cifar10.sh` and `examples/cifar10/train_quick.sh` are written, they expect to be run from the `$CAFFE_ROOT` (paths are with respect to root directory)


shelhamer:
Thanks for catching the typo.

PR #1397, opened 2014-11-04
Update classify.py

ariandyy:
Added some more verbose information, useful when loading a lot of files


shelhamer:
Boo, I just saw this was to dev. No matter -- I can roll this into the Python fixes I'm collecting.

PR #1411, opened 2014-11-06
Updates on API documentation

cNikolaou:
I am working on updating API's documentation, as much as I can. For a first step I have fixed some minor errors that I found (I changed the character `&` to `%` for correct presentation of the plural form). In addition I have made some updates to the `Blob`'s API.


shelhamer:
Thanks -- I cherry-picked the plural fixes to master in 256366b but the blob.hpp documentation has since changed due to #1970.

PR #523, opened 2014-06-20
Multi label Data and MultiLabel Accuracy

sguada:
This PR allows to use multi-label data {-1,0,1} with 0 means ignore, -1 negative and 1 positive.

@sergeyk review

PR #1414, opened 2014-11-07
Indirection layer

netheril96:
This is an new layer implemented, as mentioned in #523. The `IndirectionLayer` maps each single integer label into an arbitrary blob. The mapping is stored in the file specified by the `source` parameter. 

# Motivation

In a lot of cases, more than a single integer label for each image is needed. It may be a group of independent labels, or linear array/matrix, or a combination of both. The current workarounds is to create multiple `DataLayer` and `ImageDataLayer`, each outputting a data array. The problem with this approach is

1. It is not natural. The additional labels must be converted to image format that is only a dummy container of numbers, combined with dummy labels to create a `DataLayer` or `ImageDataLayer`. This approach is basically a hack that abuses the existing layers.
2. It is not flexible. The order of different set of data must be manually synchronized, and it disables the use of shuffling in a lot of cases.
3. It is error prone. The images and associated properties are tightly coupled in concept, but they are separated in implementation. This makes it very easy to change one part and then forget to change the other parts.

The `Datum` format can be reworked to allow arbitrary data in place of a single label. That approach, however, is likely to complicate the majority of use cases where a single label is already enough, and/or      break backwards compatibility.

# Network structure with indirection layer

In the first example, the single label output of `DataLayer` is transformed to three independent labels, suitable for three different loss layers.

<p align="center"></p>

In the second example, the label output is transformed into a large blob instead. Hence images (or other data) can have arbitrary data naturally associated.

<p align="center"></p>

More complex network can be constructed by adding several indirection layers so that each data can be associated with arbitrary number of arbitrary properties with arbitrary shapes.

# Basic usage

The `indirection_param` is declared as 

```protobuf
// Message that store parameters for IndirectionLayer
message IndirectionParameter {
  enum IndirectionSourceType {
    // Each line of the source file consists of numbers separated by whitespace
    SIMPLE_TEXT = 0;
    // Each line of the source file is a path to an unstructured binary file
    INDEXED_BINARY = 1;
    // Each line of the source file is a path to a binary BlobProto file
    INDEXED_BLOB = 2;
  }
  optional IndirectionSourceType type = 1;
  // The source file for index mapping
  repeated string source = 2;

  // The product of channels, height and width must equal to array length 
  // in the source
  optional uint32 channels = 3 [default = 1];
  optional uint32 height = 4 [default = 1];
  optional uint32 width = 5 [default = 1];
  enum IndirectionCacheType {
    NONE = 0;  // No cache is used
    WHOLE = 1;  // The data is cached as a whole in memory

    // Part of the data is cached; replacement is based on clock algorithm
    CLOCK = 2;
  }
  optional IndirectionCacheType cache_type = 6;

  // The following parameters are used only with partial cache
  optional uint32 cache_block_size = 7 [default = 256];
  optional uint32 cache_block_num = 8 [default = 8];
}
```

The key attributes are `type` and `source`. The `source` attribute contains the file name of a text file containing the mapping and can be repeated to produce multiple tops. In each mapping file, the zeroth line indicates the zeroth blob, the 33rd line the 33rd blob, etc. 

The `type` parameter determines how each line is interpreted. In the default case, each line composes of space separated floating numbers. They may also be interpreted as file names that contains the actual data. More `type` (such as indexed images) can be added by subclassing `IndexedDataReader`.

The `channels`, `width` and `height` attributes control the shape of top blob, and must be specified (or left with default value 1) because many type of mapping does not contain the relevant information. Changing these values can allow different interpretation of underlying data contained in the mapping.

When the `type` is not `SIMPLE_TEXT`, each lookup requires disk access and possibly format conversion, which is costly. To reduce the overhead, one can enable caching. Currently only the simplest caching mechanism are implemented: complete in memory cache and block cache with CLOCK replacement algorithm. More cache strategy can be implemented by subclassing `IndexedCachedDataReader`.


netheril96:
It has come to my attention that `HDF5DataLayer` can now produce multiple blob outputs as well, which makes this layer less needed. Strange that no one mentions it.

shelhamer:
@netheril96 `HDF5DataLayer` learned to make multiple tops in #1183.

netheril96:
@shelhamer It is very odd that no one, including you, had even hinted at it.

shelhamer:
@bhack @netheril96 `HDF5DataLayer` will not necessarily be alone in this. #1183 offered a simple +24 line patch for a helpful generalization of the data layer that's useful for problems with rich ground truth like multi-label classification, multivariate regression, multiple losses, and so on so it was merged.

From recent experience it seems that forming a new, general data pipeline without a lot of complications is a tricky task so light, decisively useful changes like #1183 are important steps.

netheril96:
@shelhamer 

So you mean that this `IndirectionLayer` is too heavy a step towards general data pipeline?

shelhamer:
@netheril96 I don't know yet -- more code takes more time to review and really work out the consequences for the framework.

It's important to have different proposals to understand the options and pick the right way, so thanks for this interpretation of how to load rich data from an integer index.

bhack:
Can anybody tag this PR abandoned by reviewers?

PR #1457, opened 2014-11-20
Preserve shape of extracted features

jyegerlehner:
Is there a reason why extract_features changes the shape of a feature datum from channels x height x width to 1 x channels*height*width x 1? This caused a problem where I extracted features to use as the training data for the next stacked encoder/decoder pair of layers. I was unable to train with them because their shape was lost: the convolutional layer needs to know the correct width and height of its input, and it had been lost. I couldn't find a way to tell the net to reshape the training data back to the correct feature dimensions. The simplest solution appears to me to just not mangle the dimensions in the first place.

### DIFF
--- a/tools/extract_features.cpp
+++ b/tools/extract_features.cpp
@@ -146,9 +146,9 @@ int feature_extraction_pipeline(int argc, char** argv) {
       int dim_features = feature_blob->count() / batch_size;
       const Dtype* feature_blob_data;
       for (int n = 0; n < batch_size; ++n) {
-        datum.set_height(dim_features);
-        datum.set_width(1);
-        datum.set_channels(1);
+        datum.set_height(feature_blob->height());
+        datum.set_width(feature_blob->width());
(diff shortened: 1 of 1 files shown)


shelhamer:
I merged this to master in c942dc1 since other users have encountered this. Although on the whole the `extract_features` tool seems like it should either go away or be rewritten especially after #1970.

Thanks for the change @jyegerlehner.

jyegerlehner:
> extract_features tool seems like it should either go away or be rewritten especially after #1970.

@shelhamer I started looking at the rewrite. Datum is still used heavily in the code, and it is hard wired to have channels/height/width explicitly. So is the way forward on that to have a deprecated V1Datum (like we have with V0LayerParameter, V1LayerParameter), and the new Datum would have repeated dim instead of channels/height/width?  If so, the scope of that change goes beyond just the extract features util. I can attempt that, but am a bit daunted by my lack of understanding of how the all these classes fit together.

PR #1460, opened 2014-11-20
fix matcaffe prepare_batch indexing

cvondrick:
The data_transformer.cpp code crops centers according to the formula

     floor((image_dim - cropped_dim) / 2) 

However, the matlab code crops centers according to the formula:

     floor((image_dim - cropped_dim + 1) / 2) + 1

where the additional +1's are for matlab's indexing at 1. However, you only need one +1. Right now, this causes matcaffe to crop centers one pixel off from the data_transformer.cpp, which makes the extracted features different. I think you should remove one of the +1s so that matcaffe matches data_transformer.cpp exactly.


shelhamer:
Follow-up if this wasn't addressed by the new matcaffe in #2505.

PR #1461, opened 2014-11-21
Fix compile error in convert_mnist_siamese_data.cpp

ixartz:
We can not print label_file which is a std::ifstream. So, we need replace label_file by label_filename.


shelhamer:
Sorry, another dev PR -- but I see this was already fixed in master@b136da59. Thanks for sending a fix @ixartz.

PR #1473, opened 2014-11-25
Python testing

longjon:
After #1472.

This PR adds Python testing of very basic functionality.
* Testing is through the built-in `unittest` module. We might want to consider other testing libraries (like `nose`) in the future, but this is fine for now.
* Hooks are added for the Makefile (`make pytest`) and Travis.
* Tests are meant to exercise tricky things that come up in the Python wrapper (like memory management), and not be redundant with existing C++ tests.
* Only very basic testing (of some functions from `_caffe`) is included so far. I'm going to go ahead and work on pycaffe on top of this, but BVLC folk are welcome to push more tests to this branch, and decide whether this should be merged right away (to at least have some testing) or a bit later (so that testing is not misleading by its paucity).
* Includes a bugfix revealed by the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,8 @@ install:
   - sudo -E $SCRIPTS/travis_install.sh
 
 before_script:
-  - export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib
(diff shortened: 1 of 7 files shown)


longjon:
Rebased after #1228 and passing again.

Issue #1349, opened 2014-10-22
Handle divergence/NaNs gracefully


longjon:
Bad data scaling and/or bad learning rates often cause training to diverge, producing NaN values and losses that quickly propagate throughout the net.

While occasional divergence is a fact of life, Caffe could do a better job of handling this situation, both in terms of user experience (there is no point in continuing to perform gradient descent on a NaN loss), and in terms of code correctness (pooling layer and probably others do not handle NaN comparisons correctly, which actually causes memory to be written out-of-bounds due to a default argmax index of -1).

See the issue with pooling layer at #1333 (and thanks @tleyden for pointing out the crash).

PR #1479, opened 2014-11-26
Added CHECK if loss is NaN

sguada:
As mentioned in #1349 If Loss is Nan it doesn't make sense to keep training.

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -65,7 +65,7 @@ message NetParameter {
 // NOTE
 // Update the next available ID when you add a new SolverParameter field.
 //
-// SolverParameter next available ID: 35 (last added: stepvalue)
+// SolverParameter next available ID: 36 (last added: stop_on_nan)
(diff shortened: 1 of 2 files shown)


seanbell:
More generally, perhaps this should die on any loss that isn't finite (i.e. NaN, +infinity, or -infinity).

PiranjaF:
Thanks a lot - this is highly needed! I've spent way too long trying to understand why pycaffe was crashing when the network was diverging. Quite annoying when trying to do a parameter search over a large parameter space.

PiranjaF:
How would I go about stopping training more silently? With this implementation any script in pycaffe would still crash immediately. It should be possible to catch the error or perhaps set as an option that training stops without stopping everything else.

seanbell:
@PiranjaF you could change it to break from the loop instead of crash.

PiranjaF:
@seanbell I did not think it could be done that easily. Thanks.

shelhamer:

Issue #1471, opened 2014-11-24
Dot product layer


mcheshkov:
I want to use dot product in network: given two bottom blobs NxKxWxH will produce one top blob Nx1x1x1, each of N will be dot product of appropriate KxWxH bottom pieces.
Is there a good way to do this with implemented layers? If not, is such layer appropriate?

shelhamer:
Elementwise product by `Eltwise` layer then sum with `Reduction` layer is a substitute for a dedicated layer.

PR #1480, opened 2014-11-26
Dot product layer

mcheshkov:
#1471
Layer take 2 bottom blobs, treat them as N vectors of CxWxH floats and produce one top blob of N floats - dot product for each pair of vectors.


shelhamer:

PR #1481, opened 2014-11-26
Fix CMake lint for out of tree build

mcheshkov:
The `../` is relative to build dir, so it only works for building from `${root}/build` cases, and breaks when building really away from source tree. Now `../` is relative to `lint.cmake`, and so it's tied to place.


shelhamer:
This was fixed in the general overhaul of the CMake build in #1667. Thanks for pointing this out at the time all the same.

PR #1482, opened 2014-11-26
Filter layer

mtamburrano:
Filter_layer developed as discussed in #1448, implemented following the suggestions by @sguada and @longjon (paragraph 2(i)).

This layer accepts 2+ bottom blobs, the first blob acts as a selector, so it should be a vector s={0,1,2, .. n} in [0,1], and the remaining blobs are blobs to be filtered. Each value of 1 in the selector vector means that the items at the corresponding indices int the bottom_blobs[1+] will be kept, on the contrary a values of 0 means the items with corresponding symbols will be filtered out.
So only non-filtered items will be forwarded and afterwards will receive backpropagation. 

As param the filter_layer needs a *need_back_prop vector*. This is needed because some blob could contains labels, and therefore will not need backpropagation. The need_back_prop vector as the same size as the top vector, and accepts 1s and 0s depending on whether it should be backpropagated or not.

This PR needs #1483 and #1484

### DIFF
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -190,7 +190,7 @@ class EltwiseLayer : public Layer<Dtype> {
 };
 
 /**
- * @brief Takes two+ Blobs, interprets first Blob as a selector and
+ * @brief Takes two+ Blobs, interprets last Blob as a selector and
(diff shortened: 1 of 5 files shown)


bhack:
@shelhamer Can you pass here?

mtamburrano:
closed, new PR is #2054

PR #1483, opened 2014-11-26
Removed propagate labels check in loss layers

mtamburrano:
Removed checks in the loss layers as discussed in #1448 with @longjon (point 7).
Now the loss layers don't check anymore if a backpropagation on labels is requested, this could now happen because #1482


mtamburrano:
closed, new PR is #2052

PR #1484, opened 2014-11-26
net.cpp now allows zero-sized batches

mtamburrano:
Implemented a way to allow zero-sized batches as discussed in #1448 with @sguada and @longjon (point 5).
The net now checks, before forwarding or backwarding anything, if some blob has *num == 0*, if so forward and backward are denyied and all subsequent layers are reshaped to avoid forward and bacward to those, too.
Actually the net doesn't allow a batch to have 0 size, but this could happen because #1482

### DIFF
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -58,6 +58,13 @@ class Net {
   string Forward(const string& input_blob_protos, Dtype* loss = NULL);
 
   /**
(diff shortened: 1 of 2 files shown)


bhack:
@shelhamer @longjon can you give us a follow-up?

mtamburrano:
closed, new PR is #2053

PR #1486, opened 2014-11-26
Blobs are N-D arrays (for N not necessarily equals 4)

jeffdonahue:
This PR gives Blobs a `vector<int>` of dimensions, rather than the old `num, channels, height, width`.  The first commit is all the changes needed to get Caffe to compile and everything to run as before with the new vector of tensor dimensions.  The remaining commits generalize some existing classes to use the new tensor dimensions (but they are not necessary to make it run, as it's still fine to just use all 4-D blobs with extra singleton dimensions where needed).

~~Currently I think the only problem is that in the `InnerProductLayer` the weight blobs won't be compatible with existing saved nets (since the weights are now 2D tensors and biases are 1D tensors), so I need to add something to handle that case for backwards compatibility.~~


jeffdonahue:
@longjon @shelhamer I just rebased this -- let me know if you get a chance to take a look.  It's been revised a bit since I think you guys last looked at it.

@souzou I'm not sure what state this was in when you tried to use it, but it was indeed in several not totally working states as I was developing.  It should be working now, so feel free to give it another try.  And yes, currently the only way to get data in dimensions other than 4 is to use the HDF5DataLayer by providing HDF5 files with datasets of the desired shape, as the other *DataLayers all have logic that assumes they are dealing with NxCxHxW (image) data.  (Someone could generalize them in various ways later on, but that's not something I want to do as part of this PR.)  There is an [example](http://nbviewer.ipython.org/github/BVLC/caffe/blob/master/examples/hdf5_classification.ipynb) of using the HDF5DataLayer on the Caffe website.

longjon:
I took a pass (carefully through the first commit, and quickly through the rest), this looks pretty awesome, comments as noted. I think you were planning to rebase this on #1615? Other than those things, I'm ready to merge.

jeffdonahue:
Thanks for the careful review @longjon! Will comment again when I've rebased on #1615 and made the fixes noted above.

futurely:
Today I came up with Jeff Dean‘s [recent talk on training large neural network](http://stanford.edu/~rezab/nips2014workshop/slides/jeff.pdf) on page 39 of which he mentioned about Eigen optimization. It was not the first time that [he was connected with Eigen](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=613). The Eigen home page even lists him in the credits as "Jeff Dean	Fix in vectorized square root for small arguments".

I know that Eigen has been involved in [a series of issues about a year ago](https://github.com/BVLC/caffe/search?q=eigen&type=Issues&utf8=%E2%9C%93). But you may want to re-evaluate Eigen because of Jeff Dean's endorsement.

jeffdonahue:
Not sure what you're talking about @futurely; this does not use Eigen.  Maybe you're thinking of Eigen's `Tensor` data structure? When I originally had 'tensor' in the title I was referring to the mathematical structure, not Eigen's implementation. (I changed it to 'ND array' because I discovered I was embarrassingly unaware of what tensors actually are as I started to read the Wikipedia article on them...)

Also, does anyone know how to see the 'outdated' GitHub inline comments in this PR's interface? There's usually a thing you can click to expand them, but I don't see any of those on this PR. (I managed to get to @longjon's comments on this PR via email, but that's not really ideal.)

futurely:
I thought Eigen had multi-dimensional arrays but it doesn't. Sorry about the previous comment. 

Speaking of real N-D arrays, there are at least three open source implementations [boost::multi_array](http://www.boost.org/doc/libs/1_57_0/libs/multi_array/doc/index.html), [ndarray](https://github.com/ndarray/ndarray), and [lite array](http://www.cs.cornell.edu/~saeed/lite/html/group__array.html). Maybe none of them is able to replace caffe::Blob because of too many features Caffe doesn't need, but their concise arithmetic API can be inspiration of future work.

If everyone can use C++11 in the future, this feature can be implemented with variadic templates]() following the example of "Chapter 29 A Matrix Design" of "The C++ programming language, Fourth Edition". But I know that many organizations are locked in older C++ and can't afford the price of upgrading to C++11.

In conclusion, this PR is very good balance between ideal and reality. Thanks!

longjon:
@jeffdonahue, if you agree, I think this is mature. If you rebase, I'll take a final read and merge.

Re: my earlier comments disappearing, I think that's because I commented on a commit instead of the PR diff, and github doesn't like that for some reason. I'll try to keep my comments on the diffs from now on (although that's sometimes less convenient).

@futurely, indeed. I think the close integration of CPU and GPU would be one of the harder things to replicate within an existing package.

dzhwinter:
@jeffdonahue @longjon how to make convolution based on N-D arra in .more than 4 axes?

PR #1496, opened 2014-11-28
Switch layer

sguada:
This layers allows to have multiple input blobs and a selector (last blob) which will select from which input the data will be copied into the top.
All the inputs (bottoms 0 to n-2) should have the same shape, and the last bottom (n-1) contain n indices between 0 and n-2 to fill the top. 
Currently if the index is out of range it will fill the corresponding part of top with zeros, this allows to do block dropout, but just passing one input bottom and a selector with 0s and 1s (the blocks corresponding with 1s will be replaced with zeros)
@longjon do you think the logic about indices out of range is confusing?

### DIFF
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -488,9 +488,9 @@ class SliceLayer : public Layer<Dtype> {
 };
 
 /**
- * @brief Takes some inputs Blob%s and a selector and outputs the data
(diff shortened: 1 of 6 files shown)


shelhamer:

bhack:
@shelhamer Is this was already merged at https://github.com/BVLC/caffe/pull/2054?

PR #1518, opened 2014-12-02
Added Net Templates to easy definition of Nets

sguada:
This PR is inspired by #1290 and the discussion there. This PR also reuse parts of the code done by @cypof.

Instead of modifying `Net.Init()` this PR provides a new tool `./build/tools/expand_net` that is able to expand templates and do the need string substitutions. 
Usage: 
```
expand_net net_proto_file_in net_proto_file_out
```
For some template examples look at `examples/templates/'
```
./build/tools/expand_net examples/templates/example_net.template examples/templates/example_net.prototxt
```

### DIFF
--- a/examples/templates/aux_classifier.template
+++ b/examples/templates/aux_classifier.template
@@ -1,0 +1,123 @@
+# Auxiliary Classifer Module
+# Requires: input
+layers {
+  bottom: "${input}"
+  top: "ave_pool"
+  name: "ave_pool"
+  type: POOLING
+  pooling_param {
+    pool: AVE
+    kernel_size: 5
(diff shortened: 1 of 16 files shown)


shelhamer:
@sguada @jeffdonahue @longjon 

I think Python net specification #1733 is a more general and concise means to generate networks. I'm not entirely comfortable with building in string processing for network definitions. I worry about the proliferation of templates too when the Python net specification can make use of whatever logic and modularity one might want. This alternative does require Python however. 

Is it alright to close this in favor of #1733?

shelhamer:
Closing since Python net specification was adopted instead.

PR #1547, opened 2014-12-09
Python log parser improvements

drdan14:
Improvements to python log parser introduced in https://github.com/BVLC/caffe/pull/1384

Highlights:
* Interface change: column order is now determined by using a list of `OrderedDict` objects instead of `dict` objects, which obviates the need to pass around a tuple with the column orders.
* The outputs are now named according to their names in the network protobuffer; e.g., if your top is named `loss`, then the corresponding column header will also be `loss`; we no longer rename it to, e.g., `TrainingLoss` or `TestLoss`.
* Fixed the bug/feature of the first version where the initial learning rate was always NaN.
* Add optional parameter to specify output table delimiter. It's still a comma by default.

You can use Matlab code from [this gist](https://gist.github.com/drdan14/d8b45999c4a1cbf7ad85) to verify that your results are the same before and after the changes introduced in this pull request. That code assumes that your `top` names are `accuracy` and `loss`, but you can modify the code if that's not true.

~~One caveat: we expect that the output layer names consist only of alphanumeric characters and the underscore, and are therefore captured with the `\w` token in the [Python regular expression syntax](https://docs.python.org/2/library/re.html#regular-expression-syntax). If other characters are allowed in layer names, we'll have to change the `\w` token to something else.~~

*Update*: any top name without white space is OK.


drdan14:
@sguada if you're still interested in these improvements, I can submit a PR to master

shelhamer:
Yeah, please submit a PR to master for your improved parsing. Thanks!

drdan14:
Replaced by https://github.com/BVLC/caffe/pull/2350

PR #1555, opened 2014-12-10
Improvements to network drawing via draw_net.py

drdan14:
A few improvements to the aesthetics and information content of graphs created using `draw_net.py`.

Highlights:
* Use `ArgumentParser` in `draw_net.py` to parse input arguments
* Plot network horizontally (left-right) by default; yes, this confuses the standard "top" and "bottom" names, but since everyone's monitor is wider than it is tall, the advantage is that you'll be able to see over 50% more of the network on your screen at once. Then you don't have to tilt the network graph on its side and make it completely unreadable during presentations. The top-bottom functionality is available via the `--rankdir BT` command line argument to `draw_net.py`
* Display number of outputs on edge for data, convolution and inner product layers
* In node names, add newline between layer name and layer type, to save horizontal space (only if graph rank direction is left-right or right-left)
* Include convolution and pooling parameters within node names
* Distinct colors for convolution, pooling and inner product nodes

If any of these changes are particularly offensive, I'm amenable to restoring the default behavior and exposing the changes via optional command line parameters.

# Change this drab, ho-hum-looking network graph:
old

# To this amazing one!
new_bt

# Plotted horizontally by default:
new_lr

### DIFF
--- a/python/caffe/draw.py
+++ b/python/caffe/draw.py
@@ -11,11 +11,11 @@ from google.protobuf import text_format
 import pydot
 
 # Internal layer and blob styles.
-LAYER_STYLE = {'shape': 'record', 'fillcolor': '#6495ED',
+LAYER_STYLE_DEFAULT = {'shape': 'record', 'fillcolor': '#6495ED',
          'style': 'filled'}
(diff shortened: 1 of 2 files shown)


zdx3578:
very good.

shelhamer:
Thanks for the improved plots and script grooming!

PR #1562, opened 2014-12-12
Adding configuration option to enable USE_CUDNN declaration for cmake bu...

sio2boss:
Changes to support building on Jetson TK1 with cmake and with cuDNN library support.  Added additional option 'CUDNN' to enable support.  Used the following procedure to build

mkdir build
cd build
cmake -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ..
make VERBOSE=1


shelhamer:
I believe this has been overtaken by #1667 which tries to make a number of improvements. Closing, but please follow-up if #1667 is missing anything.

PR #1568, opened 2014-12-12
Datum db

sguada:
This PR is intended to replace #1238 and fix some of the problems introduced by it. Now there is only one internal cursor, this avoid errors like MDB_MAX_READERS and higher memory consumption.

Introduces a new `DatumDBParameter` to be used by the Data_Layers i.e.
```
layers {
  name: "data"
  type: DATA
  top: "data"
  top: "label"
  data_param {
    batch_size: 256
  }
  datum_db_param {
    source: "examples/imagenet/ilsvrc12_train_lmdb"
    backend: LMDB
  }
  transform_param {
    crop_size: 227
    mean_file: "data/ilsvrc12/imagenet_mean.binaryproto"
    mirror: true
  }
  include: { phase: TRAIN }
}
```

For lower memory consumption use `LEVELDB` instead of `LMDB`.

All the code related to `dataset` needs to be removed, but before doing that I want to get feedback about this PR.


sguada:
@cypof I have redo the DatumDB, now it Datum_DB_factory keeps track of the opened DBs and allows only one per source.
Each DatumDb can have multiple DatumDB::Generators, each one has its own cursor (position).
I think this should simplify having multiple instances reading from the same source.

@shelhamer I thinks this should fix all the problems related to leveldb and lmdb.

shelhamer:
Ok, thanks Sergio. I'll take a look tomorrow. Fixing up the data pipeline
is key!

sguada:
@shelhamer once it is reviewed I will remove the previous dataset code and files.

longjon:
I'll also plan to take a pass by tomorrow.

include/caffe/datum_DB.hpp
```diff
@@ -0,0 +1,128 @@
...
+  explicit DatumDB(const DatumDBParameter& param)
+    : param_(param),
+      is_opened_(new bool(false)) {}
+  virtual ~DatumDB() {
+    if (is_opened_.unique()) {
+      DatumDBRegistry::RemoveSource(param_.source());
+    }
+  }
+
+  class Generator {
```
longjon: What is `Generator`? My guess is it's supposed to be an iterator-like abstraction (a "Generator" nested class in a "DB" class sounds like it ought to produce DBs, but that doesn't seem to be what's going on here). But, it just passes through everything to its `datumdb_`, holding no internal state, and it doesn't seem to be subclassed, so, what's its raison d'être?

include/caffe/datum_DB.hpp
```diff
@@ -0,0 +1,128 @@
...
+      is_opened_(other.is_opened_) {}
+  virtual void Open() = 0;
+  virtual void Close() = 0;
+  virtual bool Valid() = 0;
+  virtual bool Reset() = 0;
+  virtual bool Next() = 0;
+  virtual bool Current(string* key, Datum* datum) = 0;
+
+  DatumDBParameter param_;
+  shared_ptr<bool> is_opened_;
```
longjon: I'm baffled by this... why `shared_ptr<bool>`?

include/caffe/datum_DB.hpp
sguada: Yeah the `Generator` is iterator-like abstraction but simpler, it could be out `DatumDB` and being called `DatumGenerator` for clarity.
I explored different options, and the iterator-like need to have an state, which at the end pretty much a `DatumDB` so, instead of creating a new state, a `Generator` just receives a copy of `DatumDB` which holds the needed state.

include/caffe/datum_DB.hpp
sguada: The reason is shared is because I just want to open the `DatumDB` once per source and close it when the last `Generator` or reference is destroyed.

include/caffe/datum_DB.hpp
```diff
@@ -0,0 +1,128 @@
...
+    DISABLE_COPY_AND_ASSIGN(Generator);
+  };
+
+  virtual shared_ptr<Generator> NewGenerator() = 0;
+  virtual bool Get(const string& key, Datum* datum) = 0;
+  virtual void Put(const string& key, const Datum& datum) = 0;
+  virtual void Commit() = 0;
+
+ protected:
+  explicit DatumDB(const DatumDB& other)
```
sguada: When copying the `DatumDB` they will share `param` and `opened` state.

include/caffe/datum_DB_factory.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include <string>
+
+#include "caffe/common.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+namespace caffe {
+
+class DatumDB;
+
+class DatumDBRegistry {
```
longjon: If we're going to keep with the registry pattern, I wonder if we could template it away? That might get overly abstract though...

src/caffe/datum_imagesdb.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+}
+
+shared_ptr<DatumDB::Generator> DatumImagesDB::NewGenerator() {
+  CHECK_EQ(param_.mode(), DatumDBParameter_Mode_READ)
+    << "Only DatumDB in Mode_READ can use NewGenerator";
+  CHECK(*is_opened_);
+  LOG(INFO) << "Creating Generator for " << param_.source();
+  DatumImagesDB* datumdb = new DatumImagesDB(*this);
```
sguada: When creating a new `Generator` one make a copy of the `DatumDB`

include/caffe/datum_DB_factory.hpp
sguada: I guess we could, but I made it a bit more specific to store only one `DatumDB` per source, this avoids opening twice the same DB

include/caffe/datum_DB_factory.hpp
```diff
@@ -0,0 +1,98 @@
...
+  typedef DatumDB* (*Creator)(const DatumDBParameter&);
+  typedef std::map<string, Creator> CreatorRegistry;
+  typedef std::map<string, DatumDB* > SourceRegistry;
+
+  static CreatorRegistry& Registry() {
+    static CreatorRegistry* g_registry_ = new CreatorRegistry();
+    return *g_registry_;
+  }
+
+  static SourceRegistry& Sources() {
```
longjon: I think this is here to prevent multiple opens of the same DB? If so, it seems like an odd place for that... what I expect in a `*Registry` class is code dealing with the static-variable registration trick, but database paths come at run time. I think if we need this it would make more sense in the DB class itself (or its subclasses, if we need it for only some types of DBs).

sguada: Yeah it is there to prevent multiple opens of the same DB. So yeah, it maybe possible to move to the `Open` method of the `datumDB`, but then there will need to be a static `SourceRegistry`
The good side will be that the `*Registry` class will become more similar to the `LayerRegistry` and could be abstracted away if want to.

include/caffe/datum_DB.hpp
```diff
@@ -0,0 +1,128 @@
...
+    : DatumDB(other),
+    mdb_env_(other.mdb_env_),
+    mdb_txn_(other.mdb_txn_),
+    mdb_dbi_(other.mdb_dbi_) {}
+  virtual void Open();
+  virtual void Close();
+  virtual bool Valid();
+  virtual bool Next();
+  virtual bool Reset();
+  virtual bool Current(string* key, Datum* datum);
```
longjon: Not sure how I feel about having all these methods return `bool`... it looks like an error code return, so why put the burden of checking on the caller? But then, from the code it seems like they just return `Valid()`; why conflate those things? To me, a simpler interface (with simpler code that fails faster) would be: return `void`, fail if the cursor is not valid.

sguada: Yeah I was debating myself about this, if `Valid` should be only returning bool and the rest being `void` with failure otherwise.

src/caffe/datum_imagesdb.cpp
```diff
@@ -0,0 +1,175 @@
...
+    datum_database_.reset(new map<string, Datum>());
+    switch (param_.mode()) {
+    case DatumDBParameter_Mode_NEW: {
+      file_.open(source.c_str(), ios::out | ios::trunc);
+      CHECK(file_.good()) << "Could not create file: " << source;
+      keys_.clear();
+      batch_.clear();
+      break;
+    }
+    case DatumDBParameter_Mode_WRITE: {
```
longjon: We don't actually use `WRITE` mode for anything, do we? I'm not saying it should go away, just something to think about.

sguada: Yeah, we never `WRITE` always use `NEW`

src/caffe/proto/caffe.proto
```diff
@@ -445,6 +450,40 @@ message DataParameter {
...
 
+// Message that stores parameters used by DataLayer
+message DatumDBParameter {
+  enum Mode {
+    NEW = 0;
+    WRITE = 1;
+    READ = 2;
+  }
+  // Specify the datumDB Backend.
+  optional string backend = 1 [default = "leveldb"];
```
longjon: I'm not convinced one way or the other, but: why switch from enum to string?

src/caffe/proto/caffe.proto
```diff
@@ -445,6 +450,40 @@ message DataParameter {
...
+  // Specify the datumDB mode.
+  optional Mode mode = 2 [default = READ];
+  // Specify the datumDB source.
+  optional string source = 3;
+  // If true it will loop around the datumDB when needed
+  optional bool loop = 4 [default = true];
```
longjon: This is sorely needed. When training, this should normally be true, and when testing, this should normally be false; I wonder if we should provide that default, or leave it up to the user's explicit specification? Anyway having the option is better than the status quo.

shelhamer:
With respect to naming, how about `input_param` or `source_param` instead of `datum_db`? I'm thinking ahead to once we've deprecated datum and it might be safer to pick a general name.

src/caffe/proto/caffe.proto
shelhamer: We could favor string in the proto definition to reduce bookkeeping for future changes. The experience of writing new layers and parameters and coordinating the enum values is a hassle.

include/caffe/datum_DB.hpp
longjon: That feels confusing and underhanded; why not refer instead of copying?

Does this support multiple simultaneous cursors? If so, the interface should reflect that by implementing the cursor methods (`Next`, `Reset`, &c) in the cursor or iterator subclass (or with a cursor or iterator as arguments, which amounts to the same thing). If not, there's no need for a nested class.

include/caffe/datum_DB_factory.hpp
longjon: Yeah, there's nothing wrong with using `static` members in non-factory classes where appropriate (which is admittedly rare).

include/caffe/datum_DB.hpp
longjon: 1. For any of our existing sources, can `Valid`ity be lost between calling `Valid` and `Next`?
2. If yes, can something be done about it other than throwing an error?

I don't see a useful way to deal with such a race condition, if one exists.

src/caffe/proto/caffe.proto
longjon: And I guess that's consistent with #1685. I don't see any great reason to favor enum.

longjon:
Agree we should think about our path away from `Datum`.

I took a coarse pass, mainly at the interface; it seems serviceable but I'd like to see the `Generator` thing cleaned up with something more conceptually transparent, then I could take a finer pass (the texture seems alright though).

src/caffe/proto/caffe.proto
sguada: This is only part of the solution to avoid looping during test, the other part will be handle the case of the last batch by the prefetch thread and data_layer.

src/caffe/proto/caffe.proto
sguada: Initially I used `enum` but then changed thinking on #1685 and thinking in external extensions than don't require changing `caffe.proto`

include/caffe/datum_DB.hpp
sguada: The generator is a simplified iterator more like a stream of Datum, but without use of it.begin() and it.end() which I find annoying. I was thinking in calling cursor, which seems it will make more sense.
Currently it does support multiple cursors. The reason to use a nested class was to simplify the iteration between the cursor and the DB. 
The reason for copying instead of refer was to create new cursor related vars, each cursor need to have its own cursor, but I didn't want to have to define a different cursor for each datumDB.

sguada:
@longjon thanks for the pass over the interface and organization. I think there are three options to simplify the interface:
 1. Define Cursor as nested class of DatumDB and pass shallow copy the DatumDB.
 2. Define Cursor as nested class of DatumDB and pass a reference to the DatumDB, but this will need to create some specific variables depending of the type of DatumDB.
 3. Define Cursor as friend class of DatumDB and pass a reference to the DatumDB, and force that each type of DatumDB has to define its own type of Cursor.

Which one do you think will be clearer and easier to maintain?

@shelhamer I'm totally ok with changing the name to something more general as source_param.

I was thinking that on BlobDB could be built by combining a DatumDB and a DataTransformation, but other kinds of BlobDB could be defined. That reasoning behind of using Datum::Generator, Datum::Generator + DatumTransformation = Blob::Generator

include/caffe/datum_DB.hpp
sguada: I think the problem will likely appear if we use the same DatumDB from different threads, but since this code is not thread safe probably no one would do that.

bhack:
@sguada What is this? Is this a kind of ORM?

cypof:
Is there a way we could completely replace the Generator by queues, as part of merging with #1629? It would let multiple solvers pick from the same db, and simplify prefetching by simply filling the queue. Also it would allow running the db on it's own thread.

longjon:
@cypof maybe, although I'd worry about
* combining too much functionality in the same abstraction, as a cursor seems like a sensible thing on its own, and shouldn't be hard to wrap in a queue to obtain blocking/threading functionality; and
* I'd rather see this merged sooner than delayed by being drawn into the future,

so I think it might be better to stick to the original vision of a thin wrapper around data sources, and update or wrap the cursor functionality later.

@bhack no, it's meant to be a thin wrapper around data sources to avoid the former grossness of hand-coded switches to interface with different DBs.

@sguada here's what makes sense to me:
* (abstract) base classes define the interfaces for DB and Cursor; it's probably simplest to do both at top level; DB has a method for obtaining cursors, but no friend relationship is needed
* each implementation defines its own kind of DB and Cursor; it might very well make sense to have specific Cursors be nested classes of their specific DBs
* Cursors hold references to their owning DBs if needed (is there a reason that, e.g., for LMDB, you can't just wrap LMDB's own cursor without needing a ref to the DB?)

Indeed, I think each DB should have its own type of cursor, not that this should result in more DB-specific code, it's just that the current implementations of the cursor-like methods (`Next` &c) should go in a Cursor class rather than a DB class. Think of `std::vector` and its iterators; iterators don't hold copies of vectors, but rather refer to them as needed, and the logic of traversing the vector is the business of the iterator rather than the vector (although note that iterators are not a class per se, unlike here).

Also, how is this supposed to interact with existing data layers? It seems there is an images DB, which I guess is supposed to replace the `ImageDataLayer`? What about the other data layers; of that functionality, what is meant to be covered here, and what not?

sguada:
@longjon I have refactor the code and interface to reflect what we discussed. Please let me know your opinion before changing the rest.

include/caffe/datum_DB.hpp
```diff
@@ -0,0 +1,142 @@
...
+#include "caffe/common.hpp"
+#include "caffe/datum_DB_factory.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+namespace caffe {
+
+class DatumDBCursor {
+ public:
+  explicit DatumDBCursor(const DatumDBParameter& param)
+    : param_(param) {}
```
longjon: I don't feel too strongly about this, but I think it could be argued that param passing should be left up to subclasses, and subclasses should be explicit about what information needs to be given to cursors. E.g., it seems that LMDB and LevelDB only need the single boolean `param_.loop()`, but doing it this way obscures that fact and makes one wonder if params might affect cursors in arbitrary ways.

include/caffe/datum_DB.hpp
```diff
@@ -0,0 +1,142 @@
...
+  virtual void Put(const string& key, const Datum& value) = 0;
+  virtual void Commit() = 0;
+  virtual DatumDBCursor* NewCursor() = 0;
+
+ protected:
+  virtual void Open() = 0;
+  virtual void Close() = 0;
+
+  DatumDBParameter param_;
+  shared_ptr<bool> is_opened_;
```
longjon: Surely this trickery is no longer necessary, esp. given that copy is explicitly disabled on the next line?

longjon:
@sguada okay, I've taken a coarse pass just over the interface and {Level,LM}DB implementations. The new interface seems way more sensible to me... now it's much easier to tell at a glance what basic usage should look like. Two comments as noted. And re: namespaces, yes, I think it's a fine idea to put all this stuff in a namespace; why not just `caffe::db`?

bhack:
You can also take some idea [from uberDB](https://github.com/Pita/ueberDB/blob/master/README.md)

cypof:
@sguada I extracted all the data_layer from the parallel work into a new PR #1773. It should make it easier to merge this one. There is a lock to make sure that only one dataset is created by data file name, which might be redundant with the check you have here.

shelhamer:
@cypof actually #1748 has replaced this PR for simplicity -- sorry for this miscommunication. The highest priority was to simplify the data interface so that's what #1748 does thanks to @longjon.

Closing, although other features of this PR might follow in the future.

PR #1571, opened 2014-12-13
DataLayer to provide data from LIBSVM format file

tnarihi:
I've just implemented a new layer which provides data from LIBSVM format file.

    <label> <index1>:<feature1> ... <index M>:<feature M>

Here is an example of using this layer: <http://nbviewer.ipython.org/github/tnarihi/caffe/blob/libsvmdatalayer/examples/libsvm_format_data_into_caffe.ipynb>

### DIFF
--- a/examples/libsvm_format_data_into_caffe.ipynb
+++ b/examples/libsvm_format_data_into_caffe.ipynb
@@ -1,0 +1,717 @@
+{
+ "metadata": {
+  "name": "",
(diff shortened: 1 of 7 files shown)


shelhamer:
@tnarihi this could be done by a Python data layer #1703 or through de-centralized / personal layers as discussed in #1896 with a proof-of-concept in #2036. Closing to avoid adding more data layers since the data pipeline needs further work and special formats can be handled by Python. Thanks for sharing your code!

PR #1584, opened 2014-12-17
Check for overflow in Blob::Reshape

longjon:
~~The check on `count_` is more succinct than checking the individual dimensions, and~~ more importantly includes the previously ignored case where bad values for the dimensions cause an overflow for `count_`, resulting in a silent failure to allocate. Printing the dimension values ensures that no useful debugging information is lost ~~in the change.~~


longjon:
Replaced by #2426.

PR #610, opened 2014-07-04
Device Abstraction

shelhamer:
CPU / GPU device separation and abstraction

1. simplify the code
2. make a CPU-only build possible with less mess -- compare #561 
3. prepare layers and blobs for multi-device parallelism
4. pave the way for different GPU backends such as OpenCL

and so improves Caffe all around. That is, provided there is little to no overhead in both performance and coding. Since this requires a non-trivial set of coordinated changes, it has been promoted to the BVLC:device-abstraction feature branch. To contribute to this development, PR to BVLC:device-abstraction. When you rebase your feature branch on dev, comment with your fork to notify the maintainers to update.

See #415 for history.

This PR tracks the progress and integration of this branch for its final merge to dev.

PR #1615, opened 2014-12-22
Refactor convolution layer and add deconvolution layer

longjon:
This PR adds a `DeconvolutionLayer` that flips the forward and backward passes of `ConvolutionLayer`. (The resulting operation is still convolution, but the sense of all the parameters is reversed, so that, in particular, strided deconvolution results in upsampling whereas strided convolution results in downsampling.)

Rather than duplicate all the `ConvolutionLayer` code, common sections are factored out into a parent class, `BaseConvolutionLayer`. The tricky GEMM parameters and the column buffer are hidden from the forward and backward implementations, which I hope you agree are much more readable.

Positives
* Readable implementations of convolution and deconvolution.
* Deconvolution supports all functions of convolution, including padding (note that padding is removed from the output rather than added to the input!), groups, rectangular kernels, biases.
* Whereas `ConvolutionLayer` needs to do an `im2col` and a `col2im` in the backward pass, both operations in `DeconvolutionLayer`'s backward pass require `im2col`, so a special flag is added to avoid doing this twice.

Reservations
* The diff is pretty heavy; it was not straightforward to make the changes in a gradual way, so the history is constructed post-hoc. However, the code should be pretty understandable to the few of you intimately familiar with `ConvolutionLayer`. Also note that the diff size is really half what it appears to be, because of the point below.
* Almost all the code is duplicated via s/cpu/gpu/. This PR is not the place to address this, but I hope we can get (some part of?) #610 merged soon, because this is getting silly.
* Is "deconvolution" really the right name? This layer does not have to be used in the context of undoing a convolution (although it could be). It seems likely that this name will stick one way or another.

Design choices
* The original idea was to add convolution helper functions under `util/`. However, these end up requiring a large number of arguments (each needs most of the convolution parameters). So, one could wrap all the common arguments in a `struct`, like cuDNN... but then that `struct` may as well be the layer class itself, so we have the current design.
* Rather than use a special flag to skip the extra `im2col` in deconv backward, we could leave the column buffer out of the wrapper functions. However, this leads to more duplicated code, and cuts down significantly on readability.
* The forward helper and weight diff helper are both `im2col`s followed by gemms, so in theory we could unify them. However, then we need additional arguments and logic to figure out when to transpose the column buffer. I think this could still be done, but it's a bit of a wash.

Forthcoming
* Doc comments for `DeconvolutionLayer`.
* Tests specific to `DeconvolutionLayer`. Note that most of the functionality is exercised by the `ConvolutionLayer` tests. I've only briefly checked the forward/backward passes for upsampling/downsampling, so there could be a bug or two left.
* cuDNN deconvolution should be straightforward to implement as well, since cuDNN already provides the same kind of abstraction as this refactoring. However, I won't be able to do that right away; PRs are welcome!

### DIFF
--- a/include/caffe/vision_layers.hpp
+++ b/include/caffe/vision_layers.hpp
@@ -16,6 +16,93 @@
 
 namespace caffe {
 
+template <typename Dtype>
+class BaseConvolutionLayer : public Layer<Dtype> {
+ public:
+  explicit BaseConvolutionLayer(const LayerParameter& param)
+      : Layer<Dtype>(param) {}
+  virtual void LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+  virtual void Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+
+  virtual inline int MinBottomBlobs() const { return 1; }
+  virtual inline int MinTopBlobs() const { return 1; }
+  virtual inline bool EqualNumBottomTopBlobs() const { return true; }
+
+ protected:
+  // Helper functions that abstract away the column buffer and gemm arguments.
+  // The last argument in forward_cpu_gemm is so that we can skip the im2col if
+  // we just called weight_cpu_gemm with the same input.
+  void forward_cpu_gemm(const Dtype* input, const Dtype* weights,
+      Dtype* output, bool skip_im2col = false);
+  void forward_cpu_bias(Dtype* output, const Dtype* bias);
+  void backward_cpu_gemm(const Dtype* input, const Dtype* weights,
+      Dtype* output);
+  void weight_cpu_gemm(const Dtype* input, const Dtype* output, Dtype*
+      weights);
+  void backward_cpu_bias(Dtype* bias, const Dtype* input);
+
+  void forward_gpu_gemm(const Dtype* col_input, const Dtype* weights,
+      Dtype* output, bool skip_im2col = false);
+  void forward_gpu_bias(Dtype* output, const Dtype* bias);
+  void backward_gpu_gemm(const Dtype* input, const Dtype* weights,
+      Dtype* col_output);
+  void weight_gpu_gemm(const Dtype* col_input, const Dtype* output, Dtype*
+      weights);
+  void backward_gpu_bias(Dtype* bias, const Dtype* input);
+
+  // reverse_dimensions should return true iff we are implementing deconv, so
+  // that conv helpers know which dimensions are which.
+  virtual bool reverse_dimensions() = 0;
+  // Compute height_out_ and width_out_ from other parameters.
+  virtual void compute_output_shape() = 0;
+
+  int kernel_h_, kernel_w_;
+  int stride_h_, stride_w_;
+  int num_;
+  int channels_;
+  int pad_h_, pad_w_;
+  int height_, width_;
+  int group_;
+  int num_output_;
+  int height_out_, width_out_;
+  bool bias_term_;
+  bool is_1x1_;
+
+ private:
+  // wrap im2col/col2im so we don't have to remember the (long) argument lists
+  inline void conv_im2col_cpu(const Dtype* data, Dtype* col_buff) {
+    im2col_cpu(data, conv_in_channels_, conv_in_height_, conv_in_width_,
+        kernel_h_, kernel_w_, pad_h_, pad_w_, stride_h_, stride_w_, col_buff);
+  }
+  inline void conv_col2im_cpu(const Dtype* col_buff, Dtype* data) {
+    col2im_cpu(col_buff, conv_in_channels_, conv_in_height_, conv_in_width_,
+        kernel_h_, kernel_w_, pad_h_, pad_w_, stride_h_, stride_w_, data);
+  }
+  inline void conv_im2col_gpu(const Dtype* data, Dtype* col_buff) {
+    im2col_gpu(data, conv_in_channels_, conv_in_height_, conv_in_width_,
+        kernel_h_, kernel_w_, pad_h_, pad_w_, stride_h_, stride_w_, col_buff);
+  }
+  inline void conv_col2im_gpu(const Dtype* col_buff, Dtype* data) {
+    col2im_gpu(col_buff, conv_in_channels_, conv_in_height_, conv_in_width_,
+        kernel_h_, kernel_w_, pad_h_, pad_w_, stride_h_, stride_w_, data);
+  }
+
+  int conv_out_channels_;
+  int conv_in_channels_;
+  int conv_out_spatial_dim_;
+  int conv_in_height_;
+  int conv_in_width_;
+  int kernel_dim_;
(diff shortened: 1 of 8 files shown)


jyegerlehner:
FWIW I've been running these changes and all seems to be working well.

longjon:
Thanks for the datapoint @jyegerlehner. I've added basic doc comments and tests, so this should be ready to go pending any further comments.

shelhamer:
Thanks for adding deconvolution while making convolution look the most sane it ever has Jon!

JiaxuZhu:
Can this layer used for projecting one activation in a given feature map down to the image pixel?

ducha-aiki:
@JiaxuZhu this is exactly purpose of this layer

JiaxuZhu:
@ducha-aiki Thanks!
But I cannot find any tutorial about how to use this layer.
For example, I don't know how to set all feature maps to zeros except top 9 activations.

yyyek:
Is there any tutorial for this layer? I don't know how to use it

Tgaaly:
When doing convolution followed by a deconvolution to reconstruct the input, if the num_output parameter (i.e. the number of filters) is equal, this causes an error (below). If the num_output of only the deconvolution layer is set to 1 - it works. Is the num_output parameter different for convolution and deconvolution? Shouldn't the number of kernels be equal in both these layers? Here my input data is 5x5 in batches of 100. Below is my prototxt script.

```
Check failed: bottom[0]->count() == bottom[1]->count() (40000 vs. 2500) SIGMOID_CROSS_ENTROPY_LOSS layer inputs must have the same count.

```

```
# CONVOLUTION PART

layer {
  name: "conv1"
  type: "Convolution"
  bottom: "data"
  top: "conv1"
  convolution_param {
    num_output: 16
    kernel_size: 4
    kernel_size: 4
    stride: 1
  }
}
layer {
  name: "sigm"
  type: "Sigmoid"
  bottom: "conv1"
  top: "conv1"
}

# DECONVOLUTION PART

layer {
  name: "deconv1"
  type: "Deconvolution"
  bottom: "conv1"
  top: "deconv1"
  convolution_param {
    num_output: 16 # if set to 1 - it works
    kernel_size: 4
    kernel_size: 4
    stride: 1
  }
}
layer {
  name: "sigm"
  type: "Sigmoid"
  bottom: "deconv1"
  top: "deconv1"
}
...
layer {
  name: "loss"
  type: "SigmoidCrossEntropyLoss"
  bottom: "deconv1"
  bottom: "flatdata"
  top: "cross_entropy_loss"
  loss_weight: 1
}

```

jyegerlehner:
The two bottom blobs of the loss layer have to have the same shape. Otherwise it can't compute the differences between their elements. Apparently the blob "data", the bottom blob of the "conv1" layer, has one channel. Therefore the top blob of "deconv1" also must have 1 channel, which means num_output needs to be 1.

shelhamer:
@jyegerlehner right. This not a limit on conv / deconv layers in any way but a requirement of the `SigmoidCrossEntropyLoss` layer. Thanks for commenting.

Tgaaly:
Thanks @jyegerlehner and @shelhamer for the response! Is there a design workaround for this? to be able to do deconvolution with the same number of kernels as the convolution?

jyegerlehner:
> Is there a design workaround for this?

I'm not clear what needs to be worked around. 

If it makes you feel any better, the convolution and deconvolution layers have the same number of weights, when convolution layer's num_outputs = 16 and deconvolution layer's num_outputs=1. The convolution layer has 16 kernels, each of size 1 channel x 4 height x 4 width = 256 weights. The deconvolution layer has 1 kernel, of size 16 channels x 4 height x 4 width = 256 weights.

OK, but if you are saying you insist on having 16 deconvolution output channels, you could do that. Then follow it with a 1x1 convolution to project down to 1 channel, and run that in to your loss layer. But then you'd have  a whole lot more degrees of freedom on the decoder part of the net than the encoder (of what appears to be an autoencoder of sorts). It would be quite lopsided. But maybe that's what you want.

Tgaaly:
Every time I train this autoencoder-like network (shown above) with the deconvolutional layer, the training loss decreases once and then stays the same value (6412.21 shown below) for as long as I run epochs. Have you seen similar behavior?

```
I0626 22:42:00.732342 13649 solver.cpp:223] Learning Rate Policy: inv
I0626 22:42:00.732354 13649 solver.cpp:266] Iteration 0, Testing net (#0)
I0626 22:42:16.881786 13649 solver.cpp:315]     Test net output #0: cross_entropy_loss = 8846.19 (* 1 = 8846.19 loss)
I0626 22:42:16.881834 13649 solver.cpp:315]     Test net output #1: l2_error = 1751.28 (* 1 = 1751.28 loss)
I0626 22:42:17.051527 13649 solver.cpp:189] Iteration 0, loss = 10631.6
I0626 22:42:17.051576 13649 solver.cpp:204]     Train net output #0: cross_entropy_loss = 8873.56 (* 1 = 8873.56 loss)
I0626 22:42:17.051595 13649 solver.cpp:204]     Train net output #1: l2_error = 1758.01 (* 1 = 1758.01 loss)
I0626 22:42:17.051619 13649 solver.cpp:464] Iteration 0, lr = 0.01
I0626 22:43:20.310369 13649 solver.cpp:189] Iteration 100, loss = 7569.84
I0626 22:43:20.310458 13649 solver.cpp:204]     Train net output #0: cross_entropy_loss = 6412.21 (* 1 = 6412.21 loss)
I0626 22:43:20.310482 13649 solver.cpp:204]     Train net output #1: l2_error = 1157.63 (* 1 = 1157.63 loss)
I0626 22:43:20.310499 13649 solver.cpp:464] Iteration 100, lr = 0.00992565
I0626 22:44:23.551571 13649 solver.cpp:189] Iteration 200, loss = 7569.84
I0626 22:44:23.551831 13649 solver.cpp:204]     Train net output #0: cross_entropy_loss = 6412.21 (* 1 = 6412.21 loss)
I0626 22:44:23.551865 13649 solver.cpp:204]     Train net output #1: l2_error = 1157.63 (* 1 = 1157.63 loss)
I0626 22:44:23.551882 13649 solver.cpp:464] Iteration 200, lr = 0.00985258

```

jyegerlehner:
@Tgaaly  No, I haven't encountered a problem like that. 

One thing that looks suspicious is that l2 error is so large, when the sigmoids can only give you a response from [0,1], or [-1,1], forget which. As if the data hasn't been scaled to a range that sigmoids can produce. 

By the way in my particular case I'm using PReLUs not sigmoids. You might try the way-overcomplete approach with lots of deconv output channels, followed by perhaps PReLUs and then 1x1 convs to bring the number of channels down to what you need (1 channel ?). Those are just a few random thoughts. Hard to say more without knowing more about the problem you are solving and what you are doing.

engr3os:
@Tgaaly  You don't have to make flatdata layer since you are dealing with colored images (I suppose). You can set the group parameter of convolution parameters to 3 in both conv and deconv layers. This should take care of your problem. You can also use dropout layer if you want to achieve dropout auto encoder.

PR #1629, opened 2014-12-24
RDMA, data pipeline

cypof:
Big update, it took a while to finish and I have changes to the new data_layer so I had to branch again from the latest BVLC dev, let me know if that's fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROJECT := caffe
 
-CONFIG_FILE := Makefile.config
+CONFIG_FILE ?= Makefile.config
 include $(CONFIG_FILE)
 
 BUILD_DIR_LINK := $(BUILD_DIR)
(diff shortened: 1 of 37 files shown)


shelhamer:
I've set the parallel feature branch to this latest version. These are all good improvements but I'll note that the branch is now massive 4000+ new lines of code. It might be helpful to split `parallel.cpp` into different parts for say CPU, GPU, and distributed parallelism instead of having a monolithic implementation file.

Thanks @cypof!

cypof:
@shelhamer I started cleaning and splitting the data_layer work first and will do a separate PR that should be easy to adapt to #1568. parallel.cpp can probably be simplified a lot once everything works well. Thanks!

melgor:
@cypof  Does current version of parallel branch works with GPU? I have compiled code, but when running"gpus" I get error:
F0206 17:00:12.894520 11428 syncedmem.cpp:51] Check failed: error == cudaSuccess (77 vs. 0)  an illegal memory access was encountered

Could you said too why there are two source in examples/parallel? What is the purpose of it?

cypof:
@melgor the gpus.cpp sample might be broken. I focused on the rdma one. I am currently working on shopping off the parallel PR into simpler ones. data_layer was first, multi-gpus on one node is next, hopefully next week.

PR #1632, opened 2014-12-25
switch cifar10 example to lmdb

7hil:
I found the cifar10 example created lmdb dataset by default (see examples/cifar10/create_cifar10.sh) but the network prototxt didn't update accordingly. 
This PR switches dataset from leveldb to lmdb in prototxt in the cifar10 example.


shelhamer:
Thanks for fixing the example.

PR #1638, opened 2014-12-28
Give layers a pointer to their owning Net

longjon:
This allows layers to do essentially arbitrary things that depend on the structure of the net.

This provides a lot of power, but breaks the nice abstraction of a layer as a function-with-gradient. A future PR will depend on this functionality. If it seems sufficiently useful, perhaps we should merge this, but be wary not to abuse it. Or perhaps we should find another way to compute on the structure of nets. Or perhaps we should do the former keeping the latter as a future possibility.


include/caffe/layer.hpp
```diff
@@ -306,6 +312,9 @@ class Layer {
    *  the objective function. */
   vector<Dtype> loss_;
 
+  /** The net to which this layer belongs. */
+  Net<Dtype>* net_;
```
jeffdonahue: this should probably have a NULL initializer, otherwise LGTM

include/caffe/layer.hpp
```diff
@@ -293,6 +295,10 @@ class Layer {
     param_propagate_down_[param_id] = value;
   }
 
+  /**
+   * @brief Used by Net to give layers a pointer to their owning net.
+   */
+  void set_net(Net<Dtype>* net) { net_ = net; }
```
shelhamer: Should there really be a setter or should this be an argument at construction? Are layers going to migrate between nets?

jyegerlehner:
I suggest you consider instead giving the Layer a helper class or functor in Setup that provides a narrowly focused ability to do only whatever thing it is you need the Layer to be able to do. In so doing, you design out all the abuse (to which you allude) that the Layer could otherwise do if you just give it mutable net pointer. 

An example of this is in [1895](https://github.com/BVLC/caffe/pull/1895) where the InverseMVNLayer needs to figure out which of its bottom blobs is which by name. I provided a BlobFinder in the Setup method so it can do this.

shelhamer:
Replaced by #1974.

PR #1654, opened 2014-12-30
Add missing value support to SoftmaxLossLayer

longjon:
This PR adds an option, `missing_value`, which allows some specified label to indicate that loss should not be computed at the corresponding location.

This creates an issue for normalization. To match the existing normalization, we should normalize by the number of labels actually present. However, this means that the weight of each label depends on the number of missing values in each batch, which may be undesirable. For this reason an extra boolean, `normalize`, is added; if `normalize` is false, normalization occurs only over the batch size.

Ideally this would be done in a way that can be used uniformly across different loss layers, but this is the form it exists in now.

Not yet well tested.

### DIFF
--- a/include/caffe/loss_layers.hpp
+++ b/include/caffe/loss_layers.hpp
@@ -762,6 +762,9 @@ class SoftmaxWithLossLayer : public LossLayer<Dtype> {
   bool has_ignore_label_;
   /// The label indicating that an instance should be ignored.
   int ignore_label_;
+  /// Whether to normalize the loss by the total number of values present
+  /// (otherwise just by the batch size).
+  bool normalize_;
 };
 
 }  // namespace caffe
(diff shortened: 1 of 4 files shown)


shelhamer:
Needs tests (just check loss and gradients for a toy input and truth pair) and docs, then merge.

longjon:
I've added a doc comment describing the additional parameters, and simple tests that check the gradient with an ignore label set and with normalize set to false, so this should be ready pending any further comments.

shelhamer:
@longjon the docs and gradient tests look good, but whether the ignored labels do not actually contribute the loss or not deserves coverage. Once that test is in, let's merge.

longjon:
Indeed, the pure gradient checks leave something out. I've added a simple test that checks that labels are being ignored.

shelhamer:
Thanks Jon!

PR #1660, opened 2014-12-30
extract_features.cpp bug fix. the orignal version cause the tool to writ...

piiswrong:
Fixed a bug in extract_features. The original version cause the same data to be written to every DB when extracting from multiple blobs.


jyegerlehner:
PRs should be submitted against the dev branch, not master. This bug is fixed in dev.

shelhamer:
Fixed in dev and merged to master in the latest release, but thanks for trying to send a fix.

PR #1662, opened 2014-12-31
Fix caffe.io.resize_image for the case of constant images

seanbell:
"resize_image" in python/caffe/io.py fails for the case of a constant image.  It returns NaN from a divide-by-zero.

This patch checks whether the image is a constant and then returns a new constant image of the correct shape.


shelhamer:
Thanks for the fix @seanbell.

Issue #1168, opened 2014-09-27
Data Input by MemoryDataLayer example?


slaweks17:
Hi,
Could someone please explain how I could convert say an .csv file with 20 columns, out of which first 15 are features and last 5 are outputs into something caffe would accept?
I am trying to use the Windows version, where, it appears, hdf5 data layer has not been ported.
Using MemoryDataLayer? But how do I specify number of inputs, outputs?

Thanks,
Slawek

shelhamer:
Closing as solved by the Python layer interface.

With Python layers almost all of these special data format issues are moot since if you have a line of Python that can load your data you can make a Python data layer to load it too.

PR #1665, opened 2014-12-31
Bypassed memory_data_layer AddDatumVector to python

vpestret:
Hi,
There were a question https://github.com/BVLC/caffe/issues/1168 about invocation of memory data layer from Python which I've tried to answer but I also need this code to help. I'll also update the question with instructions and python scripts.
Happy New Year!
Vladimir

### DIFF
--- a/include/caffe/data_layers.hpp
+++ b/include/caffe/data_layers.hpp
@@ -269,12 +269,15 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
   int channels() { return channels_; }
   int height() { return height_; }
   int width() { return width_; }
+  int in_height() { return in_height_; }
+  int in_width() { return in_width_; }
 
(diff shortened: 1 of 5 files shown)


vpestret:
Closing, because evolution went another way. Preprocess is remained to be in Python. Seems, that natural way to overcome code duplication in Caffe is code triplication. Good luck!

shelhamer:
While there is pycaffe and matcaffe pre-processing code, the hope is to dispel with both `Datum` and the interface code in favor of transformations on the C++ side. However there are still complications due to disagreements in data, like keeping images as {0, ..., 255} uint8 vs [0,1] float, that need to be accounted for in user code depending on how the model was trained.

Pre-processing code may proliferate in the meantime, but such is the way of this tired, creaking world.

PR #1671, opened 2015-01-03
Categoricalhdf5data

ruipimenta:
Creates a data layer that contains categorical data. In the input file, for each category there are n possible values (numbers 0,..n-1) and the 'value' -1 for 'not set' or 'no data'. In the case of a binary category the values are 0 or -1. The number of possible values does not have to be the same for all categories.
The variable 'data' in the input file should be a matrix of integers with so many columns as the number o f categories. Besides the variables 'data' and 'label', the input hdf5 should contain another variable 'description': a vector (v) of integers, containing the number of categories (v[0]) and for each category the number of possible values, excluding -1.
This layer does the 'one hot encoding' of the categorical data. The number of channels of the top[0] blob is the sum of values of vector description (excluding v[0]).
The 'one hot encoding' work is done in the CPU, even in the GPU version: the experiments with kernels resulted in 50% more time for that work.
What happens if somebody has float and categorical data? Could it be used simoultaneously this categorical data and the hdf5datalayer?


ruipimenta:
Hi
 How could I reproduce the environment that produced the error?
 I tried setting CPU_ONLY in the Makefile.config  but it didn't reproduce the error. I would like to run it in gdb. Is the script 'travis_build_and_test.sh' available somewhere?

Rui

shelhamer:

PR #1686, opened 2015-01-07
Improve const-ness of Net

longjon:
`Net` currently has a somewhat lackluster use of `const`. Returned references should be `const`; it's not okay to modify the referents. Accessors and some other functions can be `const` methods; it is okay to call them on `const Net`s.

A few changes need to be made to fix existing code. In particular `std::map::operator[]` is not a `const` method, so the somewhat more obscure `find(...)->second` needs to be used instead in a couple places.

(I need a few of these for a future PR, so I went ahead and took a pass over the whole `Net` interface.)

If you find any of this `const`ness too intrusive, discuss.

### DIFF
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -89,7 +89,7 @@ class Net {
    * @brief For an already initialized net, implicitly copies (i.e., using no
    *        additional memory) the pre-trained layers from another Net.
    */
-  void ShareTrainedLayersWith(Net* other);
+  void ShareTrainedLayersWith(const Net* other);
   // For an already initialized net, CopyTrainedLayersFrom() copies the already
   // trained layers from another net parameter instance.
   /**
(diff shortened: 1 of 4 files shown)


include/caffe/net.hpp
```diff
@@ -89,7 +89,7 @@ class Net {
    * @brief For an already initialized net, implicitly copies (i.e., using no
    *        additional memory) the pre-trained layers from another Net.
    */
-  void ShareTrainedLayersWith(Net* other);
+  void ShareTrainedLayersWith(Net* other) const;
```
netheril96: This `const` feels wrong from a logical perspective. It clearly modifies the current network.

longjon: Oops, I was thinking this went the other way, thanks for pointing that out. I guess there is some subtle reason why the compiler does not generate any warnings in this case. I'll push a corrected version shortly...

jeffdonahue:
LGTM

Issue #1685, opened 2015-01-07
Switch LayerType to string for extensibility


shelhamer:
**Originally proposed by @jeffdonahue**

The bookkeeping for the `LayerType` enum in `caffe.proto` is a hassle. Having a coordination-free method to add layers would make life easier and open up Caffe for extension in forks or a potential "sandbox" section of the project.

The layer type could just be the C++ class name of the layer, minus the "Layer" at the end. So, it would look like:

```
layers {
  name: "accuracy"
  class: "Accuracy"  # for class AccuracyLayer
  ...
}
```

Introducing the `class` field instead of `type` gives a signature for automatically upgrading net definitions like we've done in the past.

`LayerParameter` is another obstacle. Since these are actual fields they still require unique IDs. Any thoughts on a workaround?

@jeffdonahue @Yangqing @longjon

longjon:
I wonder if it would be a good idea to coordinate this with #1474 and a new layer interface?

As an aside, I like the name `type` better than `class`, but so be it. Also consider `function` or `fun` or `fn`, although that might be confusing because our functions are implemented as classes.

shelhamer:
Agreed on coordinating with #1474 and bringing up old musings on the One True Layer Interface.

> As an aside, I like the name type better than class

Yeah, me too. We can always come up with another signature to trigger the automagic upgrade.

Yangqing:
I agree with Jon that "type" is a better name - I am a bit afraid of using a c++ reserved keyword.

RE parameters for specific layers, protobuf's extension is a good solution:

https://developers.google.com/protocol-buffers/docs/proto#extensions

We can add to LayerParameter something like "extensions 1000 to max" and individual users can choose their own extension number, which will hopefully conflict less. Of course there may be some popular numbers (1701?), but that would probably not be too bad.

shelhamer:
Done by #1694.

PR #1694, opened 2015-01-08
Layer type is a string

jeffdonahue:
addresses #1685 (thanks for the reminder @shelhamer)

Mostly works, but I still need to implement automagic upgrading of net prototxts before this is merged.  Which is actually kind of hard, because `type` is currently an `enum` so the parse of an old prototxt will fail due to the lack of quotes around the enum value (which was why I'd suggested using a different name like `class`, but I didn't particularly like the name `class` either).  One possible path that I think would probably work is to change the name of the `layers` field in `NetParameter`, but keep that field to mark the use of a "V1" prototxt (with this new one being "V2").  I've wanted to do this because it kind of annoys me that it's plural -- `layers` -- so if others agree this could be a good excuse to change the name to `layer`.

To do:

- [x] Change `layers` field name to `layer` as discussed above
- [x] Restore `type()` overrides
- [x] Update upgrade_net_proto_blah tool to also handle V1->V2 updates
- [x] Unit test(s)
- [x] Upgrade prototxts using `upgrade_net_proto_text` tool

~~Make the type case insensitive~~


shelhamer:
> change the name of the layers field in NetParameter [...]
>  I've wanted to do this because it kind of annoys me that it's plural -- layers -- so if others agree this could be a good excuse to change the name to layer

Fine by me. `type` and `layer` are both good names.

sguada:
I like layer and type too

On Wednesday, January 7, 2015, Evan Shelhamer <notifications@github.com>
wrote:

> change the name of the layers field in NetParameter [...]
> I've wanted to do this because it kind of annoys me that it's plural --
> layers -- so if others agree this could be a good excuse to change the name
> to layer
>
> Fine by me. type and layer are both good names.
>
> —
> <https://github.com/BVLC/caffe/pull/1694#issuecomment-69137839>.
>


-- 
Sergio

longjon:
> Change `layers` field name to `layer` as discussed above

Hooray!

> Make the type case insensitive

Really?

jeffdonahue:
After some discussion with @longjon I've concluded that I too think case insensitive type strings would be overly user-friendly. Sorry for that brief lapse in judgment!

jeffdonahue:
Any other field names that we should seize the (kind of) one-time opportunity to change? I was thinking it might be good to standardize the 'parameter' blob names -- we currently have `param` (for the name), `blobs_lr`, and `weight_decay`.  I could change them to `param`, `param_lr`, `param_decay`? (Other suggestions?)

I also plan to clean up all the numbers so that everything is in a nice order (not that it will stay that way, but can at least make it start with `name`, `type`, `bottom`, `top`, ...).

ducha-aiki:
>param, param_lr, param_decay
To me these ones are less meaningful, than current.

longjon:
Re: naming:
* `blobs_lr` is not very meaningful... `blobs` is overly generic, mispluralized, and `lr` is a terse abbreviation. `param_lr` is more consistent, but consider also: just `lr`, or `learning_rate`, or `rate`, or (since it's not the overall learning rate being specified) `{rate/lr/learning_rate}_{coef/mult/multiplier}`.
* `weight_decay` has the advantage of being a recognizable bigram, but `weight` is inconsistent with `param`, plus it applies to parameters that are not normally considered weights. But consider also: `decay`, or `[param_]decay_{coef/mult/multiplier}`

jeffdonahue:
Thanks for the well thought out suggestions @longjon. I had a thought on another option:

```
message LayerParameter {
  ...
  repeated ParamSpec param = 5;
  ...
}

message ParamSpec {
  optional string name = 1;
  optional float lr_mult = 2 [default = 1.0];
  optional float decay_mult = 3 [default = 1.0];
  ...
}
```

I think having a separate `ParamSpec` message (suggestions for better names than `ParamSpec` are certainly welcome) would reduce bookkeeping in code (e.g., checks that the length is 0 or N for each field) and the need to specify default values for prototxt users.  Agree/disagree?

jeffdonahue:
Other than needing to run the updated automagic upgrade tool (`upgrade_net_proto_text`) on all the existing prototxts (which I don't particularly want to do yet as it will only hugely add to the already review-discouraging +/- line count...), this is roughly ready for review.  I went ahead and did the `ParamSpec` thing in my comment above (c647d9c), but I can switch it back if people don't like it (but I'll probably still try to unify the `blobs_lr`, `weight_decay`, etc. names a bit). Feedback is welcome (@longjon @shelhamer and anyone else).

longjon:
Took a pass, just the one issue noted. The upgrade code is nasty, but maybe it has to be that way, and it's not forever...

`ParamSpec` looks good to me. I'm happy to merge this soon if there is general agreement.

jeffdonahue:
Thanks for looking over this @longjon!

> Aren't these names now stored in param_? Why not use a single implementation of this function for all layers? (I suppose I can imagine a future in which layers are initialized with something other than a LayerParameter, and this kind of introspective information is useful, however...)

Yeah, I realized this was kind of weird after doing it... But, I rely on it for the `CreateLayer` (cb50d6d) and layer type upgrade (573e8c7) unit tests, which test that `LayerFactory` actually creates a layer of the type requested. (Without the explicit type implementation, would have to just check that it sets the layer's `LayerParameter` with the correct type, which is weaker than the Layer being of the correct subclass.) Maybe there's another way to do this that I didn't think of, or it's not a useful enough check to justify the overhead of having to write the per-class `type()`?

longjon:
Well, I'm not sure yet what conclusion to reach regarding the per-layer `type` implementation, but something else to consider is RTTI (e.g., look up `std::type_info::name`). But that's hacky and compiler dependent (but all our compilers will return the same thing, and we might feel less bad about doing that only in tests...).

jeffdonahue:
I upgraded the existing net prototxts to the new format (purely using `upgrade_net_proto_text` with no other post-processing).  I think this is ready for merge, unless @longjon really wants the per-layer `type` implementation removed first -- personally I think we could punt on that decision for now and remove it later (if that's what we want to do).

longjon:
It's okay with me to punt, since we already have per-layer `type`
implementations.

shelhamer:
@jeffdonahue I'm down with this on the whole, but double-check the prototxt for stripped comments. There are explanatory notes that need to be kept.

`git log -p -S '#' -- *.prototxt` will show changes in prototxt comments in the history of this branch.

Thanks for the proto reform!

shelhamer:
@jeffdonahue I like `ParamSpec`!

seanbell:
Idea: now that there are a few different versions (and presumably there will be more), it may be worth including an explicit version number field (at the top level).    Auto-detecting the version can get fragile and tricky as you make more schema changes down the road.

If the explicit version number is omitted, you know it's either V0, V1 or the current version (V2?) and must auto-detect.  Moving forward, you require a version to be specified.

This also makes the prototxt files easier to compare as more people start sharing their caffe networks (e.g. on the Model Zoo), each with different versions.

jeffdonahue:
@shelhamer thanks for the review! Will restore the prototxt comments.

@seanbell nice idea, I thought about adding a version number at some point, but decided against it as it would need to be explicitly specified in every prototxt to be useful, and ideally these breaking changes *shouldn't* actually ever happen.  Maybe it is worth requiring specifying a version number though?

jeffdonahue:
Merged after discussion with @longjon and @shelhamer.

PR #1697, opened 2015-01-08
More generic number parsing in python log parser

mlopezantequera:
This allows for losses that are negative or have exponents

### DIFF
--- a/tools/extra/parse_log.py
+++ b/tools/extra/parse_log.py
@@ -38,8 +38,10 @@ def parse_log(path_to_log):
 
     re_iteration = re.compile('Iteration (\d+)')
     re_accuracy = re.compile('output #\d+: accuracy = ([\.\d]+)')
-    re_train_loss = re.compile('Iteration \d+, loss = ([\.\d]+)')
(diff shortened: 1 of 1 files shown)


tools/extra/parse_log.py
```diff
@@ -38,8 +38,10 @@ def parse_log(path_to_log):
 
     re_iteration = re.compile('Iteration (\d+)')
     re_accuracy = re.compile('output #\d+: accuracy = ([\.\d]+)')
-    re_train_loss = re.compile('Iteration \d+, loss = ([\.\d]+)')
-    re_output_loss = re.compile('output #\d+: loss = ([\.\d]+)')
+    re_train_loss = re.compile(
+        'Iteration \d+, loss = ([-+]?[0-9]*\.?[0-9]+[eE][-+]?[0-9]+)?')
+    re_output_loss = re.compile(
+        'output #\d+: loss = ([-+]?[0-9]*\.?[0-9]+[eE][-+]?[0-9]+)?')
     re_lr = re.compile('lr = ([\.\d]+)')
```
zhfe99: Sometimes a small learning rate would also be format as scientific notations.

PR #1705, opened 2015-01-10
Makefile: specify RPATH using $ORIGIN

longjon:
Currently, when dynamically linking against libcaffe (right now, only done for tests), RPATH is specified relative to the caffe source root, which means the resulting executable has to be executed from the same directory.
This commit uses the special $ORIGIN variable of RPATH to specify a path relative to the executable itself, so that there is no dependence on the working directory. (This is broken out into a make variable, `DYNAMIC_LDFLAGS`, that will be useful in a future PR.)


shelhamer:
Sounds good -- is the future dynamic linking for the tools and pycaffe?

PR #1714, opened 2015-01-13
Enabled the choice between OpenCV 2.4.x or OpenCV 3.x

blackibiza:
Selecting OpenCV version during CMake configuration, defining variable OPENCV_3 for newer OpenCV 3.x and OPENCV_2 for older API (2.4.6+).
Added the fix for imread to solve the compiling error

### DIFF
--- a/src/caffe/CMakeLists.txt
+++ b/src/caffe/CMakeLists.txt
@@ -43,8 +43,8 @@ if(OpenCV_FOUND)
 add_definitions(-DOPENCV_3)
 message( STATUS "OpenCV: Found (3.x)")
 else()
-#find minimum the 2.4.6 version
-FIND_PACKAGE(OpenCV 2.4.6 COMPONENTS core highgui imgproc REQUIRED)
+#find minimum the 2.3 version
+FIND_PACKAGE(OpenCV 2.3.0 COMPONENTS core highgui imgproc REQUIRED)
 add_definitions( -DOPENCV_2)
 endif(OpenCV_FOUND)
 include_directories(${OpenCV_INCLUDE_DIRS})
--- a/src/caffe/CMakeLists.txt
+++ b/src/caffe/CMakeLists.txt
@@ -38,7 +38,7 @@ include_directories(${HDF5_INCLUDE_DIRS})
 
 #    OpenCV
 #OPENCV - Git version (>=3.0 has different namespaces for variables)
-FIND_PACKAGE(OpenCV 3.0.0 EXACT COMPONENTS core highgui imgproc)
+FIND_PACKAGE(OpenCV 3.0.0 EXACT COMPONENTS core highgui imgproc imgcodecs)
 if(OpenCV_FOUND)
 add_definitions(-DOPENCV_3)
 message( STATUS "OpenCV: Found (3.x)")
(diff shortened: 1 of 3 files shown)


Nerei:
@blackibiza I think this PR should be rejected. Your changes are not actually required, CV_LOAD_IMAGE_COLOR exits in 3.0

Offtopic, but fyi: https://github.com/BVLC/caffe/pull/1667

Nerei:
But API according to our plans will diverges extremely in future. Imagine, we would remove STL from OpenCV interface in 3.x :-)

Nerei:
From other side, It seems now `#inckude opencv2/imgcodes_c.h` causes conditional compilation anyway and picture in my mind is obsolete. (I pefer 2.4.x only)

IMHO, better to create caffe::imread() function, call cv::imread() with default parameter and convert to gray-scale. But this (that is better) is just my speculations.

blackibiza:
But how long will be OpenCV 2.x around? I am of the opinion that OpenCV 3.x will break somehow the API in future. But Travis still uses 2.3.1 version of OpenCV. I believe there's still a need to have both the possibilities implemented, so in this way it's possible to keep them.

sanchom:
@Nerei Where is CV_LOAD_IMAGE_COLOR in OpenCV 3.0? As far as I can tell, it isn't part of the C++ code anymore.

Nerei:
@sanchom `#inckude <opencv2/imgcodes/imgcodes_c.h>`

@blackball Don't know. Nobody knows. Hope/Believe forever. At least for bug-fixing, new platforms support, etc. Might consider this as Python 2 and Python 3. Both branches may be developed provided that some corporate interest. We and our customers use 2.x for commercial products. But those of us who now develop 3.x more than others, advocate 3.x branch. New features are to be merged to 3.x only. But 3.x will change API many times and we know this. That's why stick to 2.x for stable commercial products.

Nerei:
@blackball 

Don't know. Nobody knows. For commercial products we stick to 2.x. Some of us advocate 3.x

**Update:** Just looked more precisely, it seems Android Manager will be released for 3.0 and and that means even binary compatibility must be supported. It seems removing STL is canseled or postponed till 4.x

sanchom:
In my opinion the 2.3.x branch of OpenCV is outdated and is not necessary to support, other than for its use by Travis CI.

From 2.4.0 and onward into 3.x, the cv::IMREAD_COLOR, etc. constants were available.

Including `opencv2/imgcodecs/imgcodecs_c.h` is non-standard, and isn't automatically included by `opencv2/imgcodecs.h`. If we're using C++, we should avoid introducing global C constants and functions.

Nerei:
Need to replace `OPENCV_3`/`OPENCV_2` definitions with `CV_VERSION_EPOCH` from `<opencv2/opencv.hpp>`, and link errors are fixed in #1667.

Nerei:
Or `CV_MAJOR_VERSION` for backward compatibility with 2.4.3 in Travis

Nerei:
All required changes are done in #1667. It compiles with OpenCV 3.0. This PR to be closed.

blackibiza:
But this change enables both OpenCV 2.x and OpenCV 3.0 to be compiled. 2.3
for Travis CI, OpenCV > 2.4 for everyone else.
It's a different solution. To me, out of the box, caffe didn't compile
using OpenCV 3.0
Am 18.01.2015 20:42 schrieb "Anatoly Baksheev" <notifications@github.com>:

Nerei:
yes. both opencv branches should compile after merging #1667

Nerei:
@blackibiza I believe this can be closed. Just take the latest `dev` branch.

blackibiza:
no problem!

PR #1716, opened 2015-01-13
Repack layer

philkr:
A new layer that allows to run fully convolutional nets efficiently with a denser stride. It's best illustrated with an example: Take VGG net, which has a convolutional stride of 16 (without the last pooling). There is currently no good way of running running the net at a denser stride (let say 4).
The easiest solution would be to just reduce the stride of the pooling (eg. from 2 to 1). This unfortunately only works if there is no other layer following the pooling. In all other cases it will mess up the order of the data.
The repack layer addresses this ordering issue. It reorganizes the data (eg. multiple shifted versions of an image) and packs them along the 'num' dimension in a blob.
After all convolutions are done the data can then simply be unpacked into a complete image again.
Here is a small example on VGG: Let's say we want to decrease the stride from 16 to 8. We simply replace the bottom most pooling of stride 2 with a stride 1 pooling followed by repacking:
```
layer {
  bottom: "conv4_3"
  top: "pool4"
  name: "pool4"
  type: "Pooling"
  pooling_param {
    pool: MAX
    kernel_size: 2
    stride: 2
  }
}
```
gets replaced by
```
layer {
  bottom: "conv4_3"
  top: "pool4p"
  name: "pool4"
  type: "Pooling"
  pooling_param {
    pool: MAX
    kernel_size: 2
    stride: 1
  }
}
layer {
  bottom: "pool4p"
  top: "pool4"
  name: "pack2"
  type: "Repack"
  repack_param {
    stride: 2
    operation: PACK_IMAGE
  }
}
```
The will produce an output that contains four images stacked along the 'num' direction. To unpack the stack simply add another repack layer to the end of the network
```
layer {
  bottom: "conv5"
  top: "conv5p"
  name: "unpack2"
  type: "Repack"
  repack_param {
    stride: 2
    operation: UNPACK_IMAGE
  }
}
```

The repacking trick allows us to evaluate VGG with a stride of 4 instead of 16 (16 times the samples) in roughly 3 times the runtime.


nian-liu:
Great！This makes dense prediction tasks easier!

shelhamer:
This will come after #1639 as discussed with @philkr.

shelhamer:
@philkr once #1976 is in this will need a master edition (close this PR and send another to master).

shelhamer:

PR #1724, opened 2015-01-14
Tell users to go to the caffe-users mailing list

pannous:


shelhamer:
Thanks. The README is definitely due for an overhaul to better direct everyone where they need to go.

PR #1731, opened 2015-01-15
cuDNN R2

shelhamer:
**See #1854 for the initial cuDNN R2 compatible branch by @slayton58.**

Although cuDNN R2 is still experimental it is in the release candidate stage of development, so this PR will ready Caffe for R2. This should not be made the default until R2 is done or we add version switching for compatibility.

Note that R2 brings OS X support.

Compatibility
- [x] create general tensor, but set 4D tensor
- [x] set conv 2D descriptors
- [x] set pooling 2D descriptors
- [x] replace accumulation flags with alpha, beta scaling
- [x] let pooling have padding
- [x] pick fastest by default
- [ ] pick algorithm in `Reshape()` instead of `Forward()`
- [ ] expose choice of the forward convolutional algorithm / preference

Perhaps
- keep R1 compatibility `#if CUDNN_VERSION < 20` ?
- switch to the N-D interface in coordination with #1486 
- 3D convolution


ziyuang:
cuDNN v2 is now final (libcudnn.so.6.5.48).

Issue #1474, opened 2014-11-25
Treat bottoms and params more uniformly?


longjon:
#1471 reminded me of something that's been weighing on my mind lately.

Caffe is designed with a hard distinction between "data blobs", the intermediate results of computation, and "parameter blobs", the variables on which gradient descent is performed.

This distinction is artificially imposed on the inputs to layers, which are just functions (with derivatives). For example, inner product layer performs a matrix multiply, `C = AB`, where `B` must be a bottom, `C` must be a top, and `A` must be a parameter blob. Meanwhile, @mcheshkov wants to compute _the same function_, except with `A` and `B` both as bottoms.

A layer should really just be a function with its first derivative, leaving you free to compose layers into functions (i.e., nets) which have free variables (parameters) and bound variables (intermediate data). The distinction between bottoms and parameters is really the business of `Net` and `Solver` rather than `Layer`, as it's only after composition that one learns which variables are free.

You can find some symptoms of this imposition in the code, e.g., we have both `propagate_down` and `param_propagate_down`. It'll also come up when generating nets from Python.

Obviously changing this would be rather major. I don't want to propose any specific course of action at this time, but just to make a note and let other minds compute.

Yangqing:
+1

Issue #1169, opened 2014-09-27
Modular model definitions: network layer, layer modules, and protobuf generation


jyegerlehner:
Does anyone see any value in a new layer type that is itself a network? This would allow a recursive inclusion of networks within networks. 

The inception module used in GoogLeNet is an example of a network-in-network. I provided an implementation of GoogLeNet's Inception "module" in [this](https://github.com/BVLC/caffe/issues/1106). With Caffe as it is, if you want to build GoogLeNet, you would have to copy all those Inception module layers 9 times thereby duplicating the same thing many times, changing the names of layers and top and bottom blobs to be unique within the net. By contrast, if we have a Network layer type, then each Inception module  would appear as a single layer instance of type NETWORK (perhaps). It would refer to the network's prototxt somehow, and specify values for the the things that are distinctive about each instance of the Inception module (such as the number of output channels of each of the convolutions and the pool).  And also the bottom and top blob names. 

Am I on my own if I want this implemented? Anyone already working something like this? I wouldn't want to duplicate effort.

PR #1733, opened 2015-01-15
Python net specification

longjon:
Protobuf is powerful, but is not a general-purpose language, and can be cumbersome in its current use for writing down complicated chains of functions, i.e., nets. See previous discussion at (nonexhaustively) #1290 and #1169.

Python is a general-purpose language with clean syntax and an intimate relationship with Caffe. This PR allows net specification as (domain-specific) Python code (and consequently, effortless programmatic net construction).

Currently this is just a proof-of-concept sketch, which lightly sprinkles Python magic over the `caffe_pb2` protobuf interface.

LeNet can be written as:
```python
from caffe import layers as L
from caffe import params as P

data = L.data(batch_size=64, backend=P.data.LMDB, source='examples/mnist/mnist_train_lmdb')
conv1 = L.convolution(data, kernel_size=5, num_output=20)
pool1 = L.pooling(conv1, kernel_size=2, stride=2, pool=P.pooling.MAX)
conv2 = L.convolution(pool1, kernel_size=5, num_output=50)
pool2 = L.pooling(conv2, kernel_size=2, stride=2, pool=P.pooling.MAX)
ip1 = L.relu(L.inner_product(pool2, num_output=500))
prob = L.softmax(L.inner_product(ip1, num_output=10))
```

And can be instantiated via this hackiness, to be cleaned up later:
```python
with open('tmp.prototxt', 'w') as f:
    print >>f, prob.to_proto()
net = caffe.Net('tmp.prototxt')
```

Things to think about:
* Perhaps there should be a way to construct `Net`s directly from `Layer`s, without going through protobuf.
* The magic in use here means syntax doesn't really gets checked until the `Net` is created.
* This will need to be rebased on #1694 and more significantly #1474.
* It would be nicer if the functions constructed here simply _were_ `Net`s rather than having to make `Net`s out of them.
* There is no way to control blob names right now, but there are ways to make that happen...


netheril96:
> Perhaps there should be a way to construct Nets directly from Layers, without going through protobuf.

This is because the constructor of `net` unnecessarily involves I/O. It should instead take a `NetParameter` as argument, instead of (or in addition to) a filename as string.

jeffdonahue:
This looks nifty.  What happens with multiple tops? (e.g. in your example, can you do `data, label = L.data(...)`?)

> the constructor of net unnecessarily involves I/O. It should instead take a NetParameter as argument, instead of (or in addition to) a filename as string.

I agree, there should definitely be a way to construct a pycaffe Net by passing a caffe_pb2.NetParameter thing (if you have one) rather than writing it to a file -- PRs welcome (or I'll probably try to do this at some point if nobody else does).

longjon:
Re: multiple tops, I had in mind your syntax @jeffdonahue, but I don't think I implemented it here yet.

Yes, there should be a way to construct `Net`s from `NetParameter`s, although that was not what I had in mind with the comment above, but rather a way to construct `Net`s without involving protobuf at all. In this sketch, one constructs some special Python class which specifies a net and then turns it into a `Net`. Why doesn't one just construct a `Net` to begin with?

longjon:
Just pushed an update to this; still a bit hacky, but now more feature complete:
* rebased on string types
* supports layers with multiple tops (with an explicit `ntop` kwarg)
* supports in-place layers (`in_place` kwarg)
* supports output of nets with multiple disconnected tops
* supports `transform_param`, `loss_param`, and other non-layer-specific params
* the automatic names have been replaced with an explicit name dict, which, combined with an (arguably atrocious) hack allows the actual Python variable names to be used as blob/layer names

Here's one way to generate train and test for an AlexNet variant (striped down for unrelated reasons) with the current code:
```python
from caffe import layers as L, params as P, to_proto

# helper functions for common structures

def conv_relu(bottom, ks, nout, stride=1, pad=0, group=1):
    conv = L.Convolution(bottom, kernel_size=ks, stride=stride,
                                num_output=nout, pad=pad, group=group)
    return conv, L.ReLU(conv, in_place=True)

def fc_relu(bottom, nout):
    fc = L.InnerProduct(bottom, num_output=nout)
    return fc, L.ReLU(fc, in_place=True)

def max_pool(bottom, ks, stride=1):
    return L.Pooling(bottom, pool=P.Pooling.MAX, kernel_size=ks, stride=stride)

def alexnet(lmdb, batch_size=256, include_acc=False):
    data, label = L.Data(source=lmdb, backend=P.Data.LMDB, batch_size=256, ntop=2,
        transform_param=dict(crop_size=227, mean_value=[104, 117, 123], mirror=True))

    # the net itself
    conv1, relu1 = conv_relu(data, 11, 96, stride=4)
    pool1 = max_pool(relu1, 3, stride=2)
    conv2, relu2 = conv_relu(pool1, 5, 256, pad=2, group=2)
    pool2 = max_pool(relu2, 3, stride=2)
    conv3, relu3 = conv_relu(pool2, 3, 384, pad=1)
    conv4, relu4 = conv_relu(relu3, 3, 384, pad=1, group=2)
    conv5, relu5 = conv_relu(relu4, 3, 256, pad=1, group=2)
    pool5 = max_pool(relu5, 3, stride=2)
    fc6, relu6 = fc_relu(pool5, 4096)
    fc7, relu7 = fc_relu(relu6, 4096)
    fc8 = L.InnerProduct(relu7, num_output=1000)
    loss = L.SoftmaxLoss(fc8, label)

    if include_acc:
        acc = L.Accuracy(fc8, label)
        return to_proto((loss, acc), {v: k for k, v in locals().iteritems()})
    else:
        return to_proto(loss, {v: k for k, v in locals().iteritems()})

with open('train.prototxt', 'w') as f:
    print >>f, alexnet('/path/to/caffe-train-lmdb')

with open('test.prototxt', 'w') as f:
    print >>f, alexnet('/path/to/caffe-val-lmdb', batch_size=50, include_acc=True)
```

I'll be dogfooding this along presently.

python/caffe/layers.py
```diff
@@ -0,0 +1,93 @@
...
+
+def uncamel(s):
+    """Convert CamelCase to underscore_case."""
+
+    return re.sub('(?!^)([A-Z]+)', r'_\1', s).lower()
+
+def assign_proto(proto, name, val):
+    if isinstance(val, list):
+        getattr(proto, name).extend(val)
+    else:
```
sontran: To generate weight filler & bias filler, I would add the following here 
elif isinstance(val, dict):
        new_proto = getattr(proto, name)
        for k, v in val.iteritems():
            assign_proto(new_proto, k, v)

python/caffe/layers.py
```diff
@@ -0,0 +1,93 @@
+from collections import OrderedDict
+import re
+
+from .proto import caffe_pb2
+from google import protobuf
+
+def uncamel(s):
+    """Convert CamelCase to underscore_case."""
+
+    return re.sub('(?!^)([A-Z]+)', r'_\1', s).lower()
```
sontran: Doesn't yet work for "LRN"

longjon:
Closing to reopen against master.

kevin-li:
Is there a way to modify existing network from within python? Or is there a way to edit prototxt by coding instead of manully changing them? 

My experiment requires freezing/training different layers, how can I do this through python code? 

Thank you.

PR #1739, opened 2015-01-15
Use new cuDNN R2 API

slayton58:
Modify cuDNN layers to use the R2 API for cuDNN

### DIFF
--- a/src/caffe/layers/cudnn_conv_layer.cu
+++ b/src/caffe/layers/cudnn_conv_layer.cu
@@ -21,21 +21,51 @@ void CuDNNConvolutionLayer<Dtype>::Forward_gpu(
 
     // Forward through cuDNN in parallel over groups.
     for (int g = 0; g < this->group_; g++) {
+      const Dtype alpha = 1.0;
+      const Dtype beta = 0.0;
+
+      cudnnConvolutionFwdAlgo_t algo;
+
+      // get the desired convolution algorithm
+      CUDNN_CHECK(cudnnGetConvolutionForwardAlgorithm(handle_[g],
+                                                      bottom_descs_[i],
+                                                      filter_desc_,
+                                                      conv_descs_[i],
+                                                      top_descs_[i],
+                                                      CUDNN_CONVOLUTION_FWD_NO_WORKSPACE,
+                                                      0, // memoryLimitInBytes,
+                                                      &algo));
+
+      // get minimum size of the workspace needed for the desired algorithm
+      size_t workspaceSizeInBytes;
+
+      CUDNN_CHECK(cudnnGetConvolutionForwardWorkspaceSize(handle_[g],
+                                                          bottom_descs_[i],
+                                                          filter_desc_,
+                                                          conv_descs_[i],
+                                                          top_descs_[i],
+                                                          algo,
+                                                          &workspaceSizeInBytes));
+
+      void *workspace = NULL;
+
       // Filters.
-      CUDNN_CHECK(cudnnConvolutionForward(handle_[g],
-          bottom_descs_[i], bottom_data + bottom_offset_ * g,
-          filter_desc_, weight + weight_offset_ * g,
-          conv_descs_[i],
-          top_descs_[i], top_data + top_offset_ * g,
-          CUDNN_RESULT_NO_ACCUMULATE));
+      CUDNN_CHECK(cudnnConvolutionForward(handle_[g], (void *)(&alpha),
+                                          bottom_descs_[i], bottom_data + bottom_offset_ * g,
+                                          filter_desc_, weight + weight_offset_ * g,
+                                          conv_descs_[i],
+                                          algo, workspace, workspaceSizeInBytes, // algo, workspace, workspacebytes,
+                                          (void *)(&beta),
+                                          top_descs_[i], top_data + top_offset_ * g));
 
       // Bias.
       if (this->bias_term_) {
         const Dtype* bias_data = this->blobs_[1]->gpu_data();
-        Dtype alpha = 1.;
-        CUDNN_CHECK(cudnnAddTensor4d(handle_[g], CUDNN_ADD_SAME_C, &alpha,
-            bias_desc_, bias_data + bias_offset_ * g,
-            top_descs_[i], top_data + top_offset_ * g));
+        Dtype alpha = 1.0;
+        Dtype beta = 1.0;
+        CUDNN_CHECK(cudnnAddTensor(handle_[g], CUDNN_ADD_SAME_C, (void *)(&alpha),
+                                     bias_desc_, bias_data + bias_offset_ * g, (void *)(&beta),
+                                     top_descs_[i], top_data + top_offset_ * g));
       }
     }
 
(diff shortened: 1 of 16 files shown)


bhack:
See https://github.com/BVLC/caffe/pull/1731

shelhamer:
@slayton58 thanks for the API integration! I'll cherry-pick your commit for these changes into #1731.

slayton58:
Let me know if there's more changes you need, or API questions that need clearing up

shelhamer:
Please lint your branch by `make lint` and make any corrections -- thanks.

slayton58:
Ok, done

immars:
Hi Layton,
I got a error compiling:

```
./include/caffe/util/cudnn.hpp:113:13: error: ‘CUDNN_POOLING_AVERAGE’ was not declared in this scope
     *mode = CUDNN_POOLING_AVERAGE;
```

I got cudnn-6.5-linux-x64-v2-rc2 installed ( libcudnn.so.6.5.41), with

```
grep CUDNN_POOLING_AVERAGE -R /usr/local/cuda/include/
/usr/local/cuda/include/cudnn.h:    CUDNN_POOLING_AVERAGE_COUNT_INCLUDE_PADDING = 1, // count for average includes padded values
/usr/local/cuda/include/cudnn.h:    CUDNN_POOLING_AVERAGE_COUNT_EXCLUDE_PADDING = 2 // count for average does not include padded values
```

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -21,21 +21,54 @@ void CuDNNConvolutionLayer<Dtype>::Forward_gpu(
...
+      void *workspace = NULL;
+
       // Filters.
       CUDNN_CHECK(cudnnConvolutionForward(handle_[g],
-          bottom_descs_[i], bottom_data + bottom_offset_ * g,
-          filter_desc_, weight + weight_offset_ * g,
-          conv_descs_[i],
-          top_descs_[i], top_data + top_offset_ * g,
-          CUDNN_RESULT_NO_ACCUMULATE));
+            reinterpret_cast<void *>(&alpha),
```
longjon: Why all the `reinterpret_cast`s?
* If we must cast, `static_cast` should work here, no?
* Shouldn't the cast be unneeded here, just as it's not needed for the data arguments?

slayton58: We always explicitly cast to (void *), and reinterpret_cast<void *> captures the inherent type-unsafe nature of that operation. Original commit used (void *) casting, (void *) -> reinterpret_cast<void *> was requested by your lint tool.

slayton58:
The compile issue is because the new cuDNN drop last night changes the pooling enums slightly - this pull request was originally written for the older RC

src/caffe/layers/cudnn_conv_layer.cu
longjon: Right, C-style casting is right out; whether one should use `reinterpret_cast` or `static_cast` here is a stylistic issue that can be argued either way. What bothers me is that the `alpha` and `beta` arguments are being treated differently than the data arguments, even though (according to the header I am looking at), all of those arguments have type `void*` and are subject to the same danger. Is there a reason for that?

slayton58: I don't see a reason that the casts have to be there, it was a stylistic decision by the author of some of our code that I followed when I wrote this patch -- I personally prefer the explicit cast, but it is extraneous

longjon: Right, I'm not asking why the casts are there, but why the coefficients are being treated differently from the data. The code suggests that the coefficients are being passed untyped, while the data is typed, so that the data is safer than the coefficients. But from the headers it seems that everything is untyped.

Ideally one could a gain little safety with (templated, perhaps) wrappers, or perhaps y'all will introduce a typed interface at some point. But, to me it seems misleading to mark some arguments as unsafe when all are unsafe. Of course, marking every argument with `reinterpret_cast` would be noisy, so maybe it is better to leave them out and make a note of the unsafe interface...

kerkilchoi:
While building @slayton58's commits I got this error. Can someone help me how to resolve this error? Am I missing any dependencies?
```
/usr/local/cuda/include/thrust/iterator/detail/iterator_traits.inl:60:53: error: no type named 'iterator_category' in 'thrust::iterator_traits<thrust::device_ptr<void> >'
        typename thrust::iterator_traits<Iterator>::iterator_category
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
/usr/local/cuda/include/thrust/detail/device_malloc.inl:36:19: note: in instantiation of template class 'thrust::iterator_system<thrust::device_ptr<void> >' requested here
  typedef thrust::iterator_system< thrust::device_ptr<void> >::type system;
                  ^
/usr/local/cuda/include/thrust/detail/device_malloc.inl:41:35: error: no matching function for call to 'malloc'
  return thrust::device_ptr<void>(thrust::malloc(s, n).get());
                                  ^~~~~~~~~~~~~~
/usr/local/cuda/include/thrust/memory.h:304:29: note: candidate template ignored: could not match 'execution_policy_base<type-parameter-0-0>' against 'int'
pointer<void,DerivedPolicy> malloc(const thrust::detail::execution_policy_base<DerivedPolicy> &system, std::size_t n);
                            ^
/usr/local/cuda/include/thrust/memory.h:341:26: note: candidate template ignored: could not match 'execution_policy_base<type-parameter-0-1>' against 'int'
pointer<T,DerivedPolicy> malloc(const thrust::detail::execution_policy_base<DerivedPolicy> &system, std::size_t n);
                         ^
In file included from src/caffe/layers/cudnn_softmax_layer.cpp:6:
In file included from /usr/local/cuda/include/thrust/device_vector.h:25:
In file included from /usr/local/cuda/include/thrust/device_malloc_allocator.h:27:
In file included from /usr/local/cuda/include/thrust/device_malloc.h:102:
/usr/local/cuda/include/thrust/detail/device_malloc.inl:50:64: error: no type named 'type' in 'thrust::iterator_system<thrust::device_ptr<void> >'
  typedef thrust::iterator_system< thrust::device_ptr<void> >::type system;
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```

shelhamer:
Hey @slayton58 I only just now realized this is based on `master`. This branch should be rebased on `dev` and then submitted as a replacement PR since that is where new development comes in and `master` is far out of date. Comment if the rebase causes any trouble -- I'm happy to help.

kerkilchoi:
Got a new error while building @slayton58's commit:
Can anyone help?
```
src/caffe/layers/cudnn_conv_layer.cu(113): error: no operator "*" matches these operands
            operand types are: * const std::__1::vector<caffe::Blob<float> *, std::__1::allocator<caffe::Blob<float> *>>
          detected during instantiation of "void caffe::CuDNNConvolutionLayer<Dtype>::Backward_gpu(const std::__1::vector<caffe::Blob<Dtype> *, std::__1::allocator<caffe::Blob<Dtype> *>> &, const std::__1::vector<__nv_bool, std::__1::allocator<__nv_bool>> &, const std::__1::vector<caffe::Blob<Dtype> *, std::__1::allocator<caffe::Blob<Dtype> *>> &) [with Dtype=float]" 
(142): here

src/caffe/layers/cudnn_conv_layer.cu(126): error: no operator "*" matches these operands
            operand types are: * const std::__1::vector<caffe::Blob<float> *, std::__1::allocator<caffe::Blob<float> *>>
          detected during instantiation of "void caffe::CuDNNConvolutionLayer<Dtype>::Backward_gpu(const std::__1::vector<caffe::Blob<Dtype> *, std::__1::allocator<caffe::Blob<Dtype> *>> &, const std::__1::vector<__nv_bool, std::__1::allocator<__nv_bool>> &, const std::__1::vector<caffe::Blob<Dtype> *, std::__1::allocator<caffe::Blob<Dtype> *>> &) [with Dtype=float]" 
(142): here

src/caffe/layers/cudnn_conv_layer.cu(113): error: no operator "*" matches these operands
            operand types are: * const std::__1::vector<caffe::Blob<double> *, std::__1::allocator<caffe::Blob<double> *>>
          detected during instantiation of "void caffe::CuDNNConvolutionLayer<Dtype>::Backward_gpu(const std::__1::vector<caffe::Blob<Dtype> *, std::__1::allocator<caffe::Blob<Dtype> *>> &, const std::__1::vector<__nv_bool, std::__1::allocator<__nv_bool>> &, const std::__1::vector<caffe::Blob<Dtype> *, std::__1::allocator<caffe::Blob<Dtype> *>> &) [with Dtype=double]" 
(142): here

src/caffe/layers/cudnn_conv_layer.cu(126): error: no operator "*" matches these operands
            operand types are: * const std::__1::vector<caffe::Blob<double> *, std::__1::allocator<caffe::Blob<double> *>>
          detected during instantiation of "void caffe::CuDNNConvolutionLayer<Dtype>::Backward_gpu(const std::__1::vector<caffe::Blob<Dtype> *, std::__1::allocator<caffe::Blob<Dtype> *>> &, const std::__1::vector<__nv_bool, std::__1::allocator<__nv_bool>> &, const std::__1::vector<caffe::Blob<Dtype> *, std::__1::allocator<caffe::Blob<Dtype> *>> &) [with Dtype=double]" 
(142): here
```

slayton58:
Have you pulled the latest version of the code? I’m using CUDA 6.5, gcc 4.6.3 and cuDNN v2 RC2 and it compiles without issue on my machine

kerkilchoi:
@slayton58 I am using
1) CUDA 7.0
2) cuDNN v2 RC2
3) Clang: 
Apple LLVM version 6.0 (clang-600.0.56) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin14.0.0
Thread model: posix
on Mac OSX Yosemite.

shelhamer:
Replaced by #1854 to dev.

Issue #1743, opened 2015-01-16
New Development Workflow: The One True Branch


shelhamer:
Caffe will switch to a one branch development workflow to reduce friction and confusion. All development will branch from and merge to master. Tags will mark releases. Official feature branches will host radical or involved changes to not block routine development and releases.

- no overhead for core developers handling back-merges and release merges
- no confusion about where to contribute PRs
- release tagging still gives stable checkpoints for downstream

Historically Caffe development has followed a master (stable) and dev (unstable) branch model. This lets research live at the bleeding edge in dev while applications stem from master. Experimental ideas can be tried out in dev, hammered out, then bundled into release merges to master. This isn't sustainable.

Of course master and dev need to be reconciled to found the one branch workflow. This should be accomplished with the next release, which will be soon. Open PRs can still be checked out and merged. We're still deciding how and when exactly to coordinate all this.

shelhamer:
It turns out our comrades in open source on pylearn-dev are likewise [discussing their development workflow](https://groups.google.com/forum/#!msg/pylearn-dev/g3KdkxMcYkQ/4x9isjsCDhIJ) but they are considering the switch to the two branch workflow.

There's no cure-all to scientific open source development.

longjon:
From skimming the pylearn-dev thread, the main source of tension there seems to be between interface stability and ability to rearchitect. In this regard, Caffe may be better off, as its external interface is smaller.

Although it's not entirely clear what all constitutes Caffe's external interface, the core part is certainly the net definition through protobuf, a rigidly structured language. So far we've been able to provide an (automatic) upgrade path for all changes to that interface. We've also not been shy about changing arguably internal interfaces such as C++ method types. So, I think our situation is a bit nicer than theirs for a single-branch model, and it's probably a good idea to consider keeping our interface small in the future.

shelhamer:
Carried out in #1849 and #1887.

PR #1748, opened 2015-01-19
Simple database wrappers

longjon:
After discussing with @shelhamer, we'd like to merge a simplified version of #1568 very soon, before addressing #1743, and in line with the discussion at #1238. To accelerate this process, I created this pared-down version of that PR. (History has been heavily rewritten, but keeps @sguada's authorship.)

This PR is meant to instantiate the original conception of #1238, i.e., a simple DB wrapper class that abstracts LevelDB vs LMDB without switch statements, and nothing more. The remaining functionality of #1568 may be rebased for continuing consideration.

Notes:
* The DB interface lives in `caffe::db` as suggested in https://github.com/BVLC/caffe/pull/1568#issuecomment-69127445.
* The interface is a further simplified version of that from #1568. One reads (and only reads) from `Cursor`s (and only `Cursor`s), and writes (and only writes) from `Transaction`s and (only `Transaction`s). The wrappers are uncoupled from protobuf.
* I used `boost::scoped_ptr` in some instances where `unique_ptr` would be appropriate, but is disallowed by our avoidance of C++11.
* Tools are not well-tested. The changes are pretty light, so hopefully nothing is broken.


sguada:
@longjon I took a quick pass, and although the interface is cleaner, I'm concerned that the management of cursors and transactions for LMDB could create problems. 

I'm not sure if the proposed way of creating a new lmdb_txn for each cursor will cause problems, specially since they are never closed. 

Also creating a new transaction for each batch implies creating a new lmdb_txn that it not clearly closed. 

As it is now one could try to write while READ, or try to read while NEW.

Have you done some stress test? I mean trying to train a ImageNet like model? Have you tried that you can train and test on the same lmdb?

include/caffe/util/db.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+class Cursor {
+ public:
+  Cursor() { }
+  virtual ~Cursor() { }
+  virtual void SeekToFirst() = 0;
+  virtual void Next() = 0;
+  virtual string key() = 0;
+  virtual string value() = 0;
+  virtual bool valid() = 0;
```
sguada: I think this should `Valid()` to align better with leveldb

include/caffe/util/db.hpp
```diff
@@ -0,0 +1,188 @@
...
+  leveldb::Iterator* iter_;
+};
+
+class LevelDBTransaction : public Transaction {
+ public:
+  explicit LevelDBTransaction(leveldb::DB* db) : db_(db) { CHECK_NOTNULL(db_); }
+  virtual void Put(const string& key, const string& value) {
+    batch_.Put(key, value);
+  }
+  virtual void Commit() {
```
sguada: What would happen if the Transaction is delete before `Commit`?

include/caffe/util/db.hpp
```diff
@@ -0,0 +1,188 @@
...
+  leveldb::DB* db_;
+};
+
+inline void MDB_CHECK(int mdb_status) {
+  CHECK_EQ(mdb_status, MDB_SUCCESS) << mdb_strerror(mdb_status);
+}
+
+class LMDBCursor : public Cursor {
+ public:
+  explicit LMDBCursor(MDB_txn* mdb_txn, MDB_cursor* mdb_cursor)
```
sguada: Why do your pass `mdb_txn` if it is never used?

include/caffe/util/db.hpp
```diff
@@ -0,0 +1,188 @@
...
+    : mdb_txn_(mdb_txn), mdb_cursor_(mdb_cursor), valid_(true) {
```
sguada: I think `valid_` should be initially `false`

src/caffe/util/db.cpp
```diff
@@ -0,0 +1,84 @@
...
+    flags = MDB_RDONLY | MDB_NOTLS;
+  }
+  MDB_CHECK(mdb_env_open(mdb_env_, source.c_str(), flags, 0664));
+  LOG(INFO) << "Opened lmdb " << source;
+}
+
+LMDBCursor* LMDB::NewCursor() {
+  MDB_txn* mdb_txn;
+  MDB_cursor* mdb_cursor;
+  MDB_CHECK(mdb_txn_begin(mdb_env_, NULL, MDB_RDONLY, &mdb_txn));
```
sguada: If for every cursor one creates a new mdb_txn, it can run out of readers pretty quickly.

src/caffe/util/db.cpp
```diff
@@ -0,0 +1,84 @@
...
+  MDB_CHECK(mdb_dbi_open(mdb_txn, NULL, 0, &mdb_dbi_));
+  MDB_CHECK(mdb_cursor_open(mdb_txn, mdb_dbi_, &mdb_cursor));
+  return new LMDBCursor(mdb_txn, mdb_cursor);
+}
+
+LMDBTransaction* LMDB::NewTransaction() {
+  MDB_txn* mdb_txn;
+  MDB_CHECK(mdb_txn_begin(mdb_env_, NULL, 0, &mdb_txn));
+  MDB_CHECK(mdb_dbi_open(mdb_txn, NULL, 0, &mdb_dbi_));
+  return new LMDBTransaction(&mdb_dbi_, mdb_txn);
```
sguada: Why do you pass `mdb_dbi_`?

tools/extract_features.cpp
```diff
@@ -121,13 +121,15 @@ int feature_extraction_pipeline(int argc, char** argv) {
...
   for (size_t i = 0; i < num_features; ++i) {
     LOG(INFO)<< "Opening dataset " << dataset_names[i];
-    shared_ptr<Dataset<std::string, Datum> > dataset =
-        DatasetFactory<std::string, Datum>(argv[++arg_pos]);
-    CHECK(dataset->open(dataset_names.at(i), Dataset<std::string, Datum>::New));
-    feature_dbs.push_back(dataset);
+    shared_ptr<db::DB> db(db::GetDB(argv[++arg_pos]));
+    db->Open(dataset_names.at(i), db::NEW);
+    feature_dbs.push_back(db);
+    shared_ptr<db::Transaction> txn(db->NewTransaction());
```
sguada: Creating a new Transaction each time could create problems for LMDB.

tools/extract_features.cpp
```diff
@@ -158,10 +160,13 @@ int feature_extraction_pipeline(int argc, char** argv) {
...
             image_indices[i]);
-        CHECK(feature_dbs.at(i)->put(std::string(key_str, length), datum));
+        string out;
+        CHECK(datum.SerializeToString(&out));
+        txns.at(i)->Put(std::string(key_str, length), out);
         ++image_indices[i];
         if (image_indices[i] % 1000 == 0) {
-          CHECK(feature_dbs.at(i)->commit());
+          txns.at(i)->Commit();
+          txns.at(i).reset(feature_dbs.at(i)->NewTransaction());
```
sguada: Creating a new Transaction each time could create problems for LMDB.

include/caffe/util/db.hpp
longjon: I went with `valid` since it's an accessor and we generally lowercase those, but presumably LevelDB is also following the Google C++ style guide, so I could see the argument either way.

include/caffe/util/db.hpp
longjon: Bad things... fixing.

include/caffe/util/db.hpp
longjon: Why, so the transaction can be freed when finished with the cursor... which isn't currently done... fixing.

include/caffe/util/db.hpp
longjon: I was sloppy about this, assuming a nonempty database... will do it properly now.

src/caffe/util/db.cpp
longjon: It's needed as an argument to `mdb_put`.

tools/extract_features.cpp
longjon: See http://symas.com/mdb/doc/group__mdb.html#ga846fbd6f46105617ac9f4d76476f6597; transactions are freed after commit by LMDB, there is no other way (and they should be freed properly if not committed after the next fix).

longjon:
Thanks for your apt comments @sguada, I've fixed the sloppiness as noted.

Re: writing after READ, LMDB is opened read-only, I don't immediately see a corresponding option for LevelDB. We could enforce that in the wrapper or not; in this patch I've tried to provide nothing more than the functionality in master, and master of course doesn't enforce that. For LevelDB, READ still has the semantics of "don't create if missing".

Re: reading after WRITE or NEW, that seems fine to me, just as files are usually opened read-only or read/write.

Re: stress testing, I haven't done it yet, but can do.

include/caffe/util/db.hpp
sguada: In this case for consistency with leveldb I would vote for `Valid`

src/caffe/util/db.cpp
sguada: ok

include/caffe/util/db.hpp
shelhamer: I'm fine with lowercase valid by the accessor rule. No need to match LevelDB exactly since this is an interface over DBs.

shelhamer:
I plan to merge this shortly, so let me know if there are any objections.

@longjon @sguada I stress tested this with 8x K40s doing simultaneous CaffeNet train and test on the same LMDB at a rate of 20 iter / s each. I'll leave a CaffeNet solver running overnight to check for leaks or other issues.

PR #1753, opened 2015-01-19
Enhance debug info

jeffdonahue:
- a882248 adds debug info about inputs and parameters in forward, and for test nets
- 9f94917 adds `sumsq_{data,diff}` methods to `Blob` compute a sum of squares of all data
- 2fcab20 uses the `sumsq_*` methods to compute net-wide parameter stats (L1 and L2 norm)
- 4280ac8 adds `debug_info` to `NetParameter` so it can be enabled outside of training


jeffdonahue:
I think this should be pretty uncontroversial, so I'll merge this if it passes Travis after a rebase.

PR #1755, opened 2015-01-19
SoftmaxLayer GPU optimization

jeffdonahue:
Two of the four kernel calls in `SoftmaxLayer` have unnecessary inner loops.  This turns those inner loops into separate kernel calls, which improves performance a bit (but still not nearly to the performance of `CuDNNSoftmaxLayer`).


shelhamer:
LGTM -- what can be unrolled must be unrolled...

PR #1756, opened 2015-01-19
Max out the protobuf file read size limit

jeffdonahue:
This sets the max possible size of protobuf reads to 2 GB minus 1 byte (= 2^31 - 1), the largest `int` value.  (If we want to start supporting nets with total param size >= 2 GB this we'll have to come up with a new solution.  For example, we could store each parameter in a separate binary proto file, or switch to HDF5 for parameter serialization.)

### DIFF
--- a/src/caffe/util/io.cpp
+++ b/src/caffe/util/io.cpp
@@ -17,6 +17,8 @@
 #include "caffe/proto/caffe.pb.h"
 #include "caffe/util/io.hpp"
 
+const int kProtoReadBytesLimit = INT_MAX;  // Max size of 2 GB minus 1 byte.
+
 namespace caffe {
 
 using google::protobuf::io::FileInputStream;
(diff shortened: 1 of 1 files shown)


longjon:
Yes, I've run into this before. Perhaps we should make this less mysterious by using `INT_MAX`?

jeffdonahue:
yup, good idea, fixed

src/caffe/util/io.cpp
```diff
@@ -17,6 +17,8 @@
 #include "caffe/proto/caffe.pb.h"
 #include "caffe/util/io.hpp"
 
+#define kProtoReadBytesLimit INT_MAX  // Max size of 2 GB minus 1 byte.
```
longjon: Since this is C++, why not `const int`?

jeffdonahue: done

longjon:
Look good, merged. It would be nice to also break out the other mysterious constant at some point...

PR #1757, opened 2015-01-19
Gradient clipping

jeffdonahue:
This lets you add a field `clip_gradients` to a `SolverParameter` to set the maximum L2 norm of the parameter gradients.  If the norm of the gradients exceeds the value set, the gradients are scaled down to have exactly that L2 norm.  The L2 norm is computed prior to applying learning rate, etc. -- it applies to the actual loss gradient.

### DIFF
--- a/src/caffe/blob.cpp
+++ b/src/caffe/blob.cpp
@@ -279,6 +279,72 @@ Dtype Blob<Dtype>::sumsq_diff() const {
   return sumsq;
 }
 
+template <> void Blob<unsigned int>::scale_data(unsigned int scale_factor) {
+  NOT_IMPLEMENTED;
+}
+
+template <> void Blob<int>::scale_data(int scale_factor) {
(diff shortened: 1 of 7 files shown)


src/caffe/blob.cpp
```diff
@@ -205,6 +205,142 @@ Dtype Blob<Dtype>::asum_diff() const {
...
+  return 0;
+}
+
+template <> int Blob<int>::sumsq_data() const {
+  NOT_IMPLEMENTED;
+  return 0;
+}
+
+template <typename Dtype>
+Dtype Blob<Dtype>::sumsq_data() const {
```
jeffdonahue: I don't like writing or reading these duplicate methods for `data` and `diff` -- should these be in `SyncedMemory`, or perhaps a new abstraction layer between `SyncedMemory` and `Blob`?

jeffdonahue:
Merging after discussion with @shelhamer.

PR #1780, opened 2015-01-22
Simplify the Solver class

netheril96:
The current `caffe::Solver` class is a giant behemoth that handles too many things at once. Not only does it compute and update the network, it also does I/O on its own, automatically determines snapshot filenames and time point, evaluates test networks at regular intervals, automatically determines learning rate and stop criteria, and formats the output of all the networks it handles. 

This PR attempts to reduce the `Solver` class to its core, that is, only for computation. A separate `SolvingDriver` class, which can be seen as an adaptor, conforms to the past interface and allows minimal changes to existing code. Users of `caffe.bin` won't find anything different, but those who link to `libcaffe` are now free to do the following things, which are impossible before:

* Control the pace of learning. For example, better stop criteria than the arbitrarily predetermined maximum number of iterations can be devised. One may stop the training process once the loss is below certain threshold, when the test accuracy no longer improves, or continue indefinitely until interrupted.

* Retrieve the intermediate outputs of networks by programming API. Before this PR the only way to get such information is by parsing the unstructured log file, which is error-prone and unnecessarily involves I/O for something as pure as computation.

* Serialize and deserialize at will. One can save the intermediate result wherever and whenever he/she likes. No more waste of cycles when you want to interrupt the training process but it has not proceeded to the next checkpoint.

* Implement your own learning rate policy. Instead of decaying the learning rate through predetermined function, it can now be dynamically adjusted in any ways the library users want.

### DIFF
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -133,7 +133,7 @@ PySGDSolver::PySGDSolver(const string& param_file) {
   // as in PyNet, (as a convenience, not a guarantee), create a Python
   // exception if param_file can't be opened
   CheckFile(param_file);
-  solver_.reset(new SGDSolver<float>(param_file));
+  solver_ = boost::make_shared<SolvingDriver<float> >(param_file);
   // we need to explicitly store the net wrapper, rather than constructing
   // it on the fly, so that it can hold references to Python objects
   net_.reset(new PyNet(solver_->net()));
--- a/include/caffe/caffe.hpp
+++ b/include/caffe/caffe.hpp
@@ -12,6 +12,7 @@
 #include "caffe/net.hpp"
 #include "caffe/proto/caffe.pb.h"
 #include "caffe/solver.hpp"
+#include "caffe/solving_driver.hpp"
 #include "caffe/util/benchmark.hpp"
 #include "caffe/util/io.hpp"
 #include "caffe/vision_layers.hpp"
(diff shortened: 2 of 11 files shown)


python/caffe/_caffe.cpp
```diff
@@ -133,7 +133,7 @@ PySGDSolver::PySGDSolver(const string& param_file) {
   // as in PyNet, (as a convenience, not a guarantee), create a Python
   // exception if param_file can't be opened
   CheckFile(param_file);
-  solver_.reset(new SGDSolver<float>(param_file));
+  solver_ = boost::make_shared<SolvingDriver<float> >(param_file);
```
jeffdonahue: I've never seen this `boost::make_shared`...why use it instead of `reset`?

netheril96: @jeffdonahue 

See https://stackoverflow.com/questions/20895648/difference-in-make-shared-and-normal-shared-ptr-in-c

bhack:
@netheril96 Why you have closed this?

netheril96:
@bhack 

I abandoned the idea because the simplified `Solver` class is still heavily coupled. I resort to write my own `Solver` class without the cruft of the myriad parameters instead.

shelhamer:
Thanks for posting the code of this attempt at decoupling. It's likely this
will need to be done at some point to better expose the solver to tools
like monitoring.

PR #1787, opened 2015-01-23
[fix] align pytest for solver with #1728

shelhamer:
The test sets the mode by `caffe.set_mode_cpu()` as defined the interface change
in #1728.


jeffdonahue:
sweet, thanks Evan

PR #1790, opened 2015-01-23
Give phase to Net

### LINKED ISSUE #1430: data_transformers applies jitter in TEST phase
```phase_``` is set when ```data_transformer``` is constructed using the global ```Caffe::phase()``` state.
https://github.com/BVLC/caffe/blob/master/include/caffe/data_transformer.hpp#L18

This is done when the test net is constructed in ```InitTestNets```.
https://github.com/BVLC/caffe/blob/master/src/caffe/solver.cpp#L153

```Caffe::phase()``` is still ```Caffe::TRAIN``` at this point.

### PR DESCRIPTION
**Pull request by shelhamer**
After #1794 since `dump_network` is dropped for incompatibility here.

Make Net master of phase as is its proper due and let layers inherit its phase. This fixes #1430 and satisfies the phase part of #1500. Here's to taking a small step toward abolishing the singleton.

- phase is set at instantiation and should never change. 
- phase defaults to TEST by the [caffe.proto default](https://github.com/BVLC/caffe/blob/dev/src/caffe/proto/caffe.proto#L183). This fits the wishes in #192, but should there be a default phase at all?

Misgivings:
-  the `explicit Net(const string& param_file, Phase phase)` constructor works, but what of `stage` and `level` and further extensions to net state? Should the parameter file constructor be done away with and all these set through `NetParameter`? The `caffe` tool and wrappers will need to learn about `NetState`.
- exposing the `Phase` enum seemed like a hassle, so 001f060 just exposes the proto values as `caffe.{TRAIN,TEST}` for instantiating `caffe.Net`. I suppose this could have a pytest.

### DIFF
--- a/include/caffe/data_transformer.hpp
+++ b/include/caffe/data_transformer.hpp
@@ -16,7 +16,7 @@ namespace caffe {
 template <typename Dtype>
 class DataTransformer {
  public:
-  explicit DataTransformer(const TransformationParameter& param);
+  explicit DataTransformer(const TransformationParameter& param, Phase phase);
   virtual ~DataTransformer() {}
 
   /**
@@ -104,7 +104,7 @@ class DataTransformer {
 
 
   shared_ptr<Caffe::RNG> rng_;
-  Caffe::Phase phase_;
+  Phase phase_;
   Blob<Dtype> data_mean_;
   vector<Dtype> mean_values_;
 };
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -104,8 +104,6 @@ class Caffe {
     return *singleton_;
   }
   enum Brew { CPU, GPU };
-  enum Phase { TRAIN, TEST };
-
 
   // This random number generator facade hides boost and CUDA rng
   // implementation from one another (for cross-platform compatibility).
@@ -137,16 +135,12 @@ class Caffe {
 
   // Returns the mode: running on CPU or GPU.
   inline static Brew mode() { return Get().mode_; }
-  // Returns the phase: TRAIN or TEST.
-  inline static Phase phase() { return Get().phase_; }
   // The setters for the variables
   // Sets the mode. It is recommended that you don't change the mode halfway
   // into the program since that may cause allocation of pinned memory being
   // freed in a non-pinned way, which may cause problems - I haven't verified
   // it personally but better to note it here in the header file.
   inline static void set_mode(Brew mode) { Get().mode_ = mode; }
-  // Sets the phase.
-  inline static void set_phase(Phase phase) { Get().phase_ = phase; }
   // Sets the random seed of both boost and curand
   static void set_random_seed(const unsigned int seed);
   // Sets the device. Since we have cublas and curand stuff, set device also
@@ -163,7 +157,6 @@ class Caffe {
   shared_ptr<RNG> random_generator_;
 
   Brew mode_;
-  Phase phase_;
   static shared_ptr<Caffe> singleton_;
 
  private:
--- a/include/caffe/data_layers.hpp
+++ b/include/caffe/data_layers.hpp
@@ -48,8 +48,7 @@ class BaseDataLayer : public Layer<Dtype> {
 
  protected:
   TransformationParameter transform_param_;
-  DataTransformer<Dtype> data_transformer_;
-  Caffe::Phase phase_;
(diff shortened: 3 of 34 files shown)

### CODE REVIEW

File: include/caffe/data_transformer.hpp [hunk 1]
```diff
@@ -16,7 +16,7 @@ namespace caffe {
 template <typename Dtype>
 class DataTransformer {
  public:
-  explicit DataTransformer(const TransformationParameter& param);
+  explicit DataTransformer(const TransformationParameter& param, Phase phase);
```
jeffdonahue: Maybe it would make sense to add `phase` to `TransformationParameter`, setting the phase from the owning net by default? (Keeps the single-argument constructor, and allows one to override defaults, e.g., if one wanted to randomly crop/mirror at test time for some reason.)

shelhamer:
@jeffdonahue @longjon what if `phase` is brought down into the layers that need it? These layers would have `phase` added to their parameter message and it would default to the `Net`'s during `Net::Init()`. This marks which layers actually care about phase and does away with the need for the `Net` pointer to sidestep needing to make a dummy `Net` for all the tests, which is pretty annoying as-is.

jeffdonahue:
Sounds fine, but seems like you might as well just add it to `Layer` itself and have the Net set it for all Layers (rather than worrying about whether they need it).

shelhamer:
@longjon @jeffdonahue I think this could work -- let me know what you think.

longjon:
My thoughts on this:

Phase is just a net-level parameter. It would be nice if phase could be ignored for nets that don't have layers that depend on phase, but it's also important that phase be specified explicitly for nets that do depend on it, since test/train confusion is very bad.

So I'm okay with the way this is done now as an incremental step, because per-layer phase-dependence isn't already explicitly noted, and doing so is a pain. But I think I would actually prefer if phase-dependent layers took phase explicitly. That way the user who just wants to compute doesn't have to think about this irrelevant phase thing, while at the same time the user who runs a train/val net must explicitly and immutably choose a phase, at the threat of construction-time failure (I am both users, at times).

jeffdonahue:
This LGTM.  I don't have much of an opinion on the issues @longjon raised.

shelhamer:
Note to self: all this needs is new pyreformation constructors to cast phase.

shelhamer:
This is reconciled with pyreformation #1703 and now `caffe.Net`s are instantiated with phase as `caffe.{TRAIN,TEST}`.

- I'm content with exposing the enum as done here -- while there is a boost::python enum sharing protobuf's enum in this way will lead to gross identifiers and type woes.
- The phase arg should be converted to a keyword argument in the `Net` constructors, and so with `level`, `stage`, and eventually mode and device too.

A follow-up PR will further change the pycaffe interface to yield to `Net` all that is `Net`'s.

lukeyeager:
@shelhamer you ended up creating a Python interface with this PR which isn't actually a "valid" Python interface. Boost.Python lets you do some cool stuff with pattern matching like allowing these two function signatures:
```python
net1 = Net('deploy.prototxt', caffe.TEST)
net2 = Net('deploy.prototxt', 'weights.caffemodel', caffe.TEST)
```
But you can't actually do that with pure Python:
```python
>>> class Net():
...     def __init__(self, network, weights, phase):
...         pass
...     def __init__(self, network, phase):
...         pass
... 
>>> net1 = Net('deploy.prototxt', 1)
>>> net2 = Net('deploy.prototxt', 'weights.caffemodel', 1)
  File "<stdin>", line 1, in <module>
TypeError: __init__() takes exactly 3 arguments (4 given)
```
The second definition of `__init__` overrides the first.

I wish this would have been done with named keyword arguments instead, like:
```python
class Net():
    def __init__(self, network, phase=0, weights=None):
        pass

net1 = Net('deploy.prototxt', phase=1)
net2 = Net('deploy.prototxt', phase=1, weights='weights.caffemodel')
```
Actually, if you would have just put "phase" as the second positional argument for both signatures, we could have salvaged it because you can provide keyword arguments as positional arguments if they line up with the order in which you specified the keyword arguments. As it is, we can't change it now without breaking code. :disappointed: 

NOTE - I'm pointing this out because my upcoming PR adding `stage` and `level` as optional arguments will either not be backwards compatible, or will include duplicated code.

shelhamer:
@lukeyeager right, thanks for the clear exposition. I would have liked keyword args too and I agree it makes for a clearer interface. I'm traveling now but will look at #3863.

PR #1794, opened 2015-01-25
drop dump_network tool

shelhamer:
This tool is old and brittle.
Nets are better serialized however desired through the Python and MATLAB interfaces.

Speak up if you think it should stay and deserves to be updated.


jeffdonahue:
LGTM

PR #1796, opened 2015-01-26
Support Net Surgery in Matlab

xianjiec:
add functions set_weights, save in matcaffe.cpp to enable transplanting parameters / net surgery from matlab.

add net_surgery_demo.m to show how to do net surgery in matlab.


shelhamer:
Thanks for the pull request and the demo, but matcaffe has been updated altogether in #2505. The new interface includes net surgery and more. If you see any improvements to make in the new interface, please do send a PR!

PR #1798, opened 2015-01-27
add print_result code block from https://github.com/jetpacapp/caffe/blob...

stray-leone:
.../master/python/classify.py


shelhamer:

PR #1838, opened 2015-02-06
Bug in MemoryData that does not allow using arrays with n_ * size_ > 2^31

DmitryUlyanov:
So the problem is in here: 
```
top[0]->set_cpu_data(data_ + pos_ * size_);
```
`pos_` and `size_` are `int`, so I had segmentation fault due to overflow in there, I think, when using huge array.  To overcome this we need to change some variables types, so please comment how you think to do it better since I did the first thing that came to my mind that would fix the crashes I had.

### DIFF
--- a/include/caffe/data_layers.hpp
+++ b/include/caffe/data_layers.hpp
@@ -281,7 +281,7 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
   Dtype* data_;
   Dtype* labels_;
   int n_;
-  int pos_;
+  size_t pos_;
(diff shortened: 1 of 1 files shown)


longjon:
Thanks for the fix; only issue is this should probably be `size_t`, and please squash your commits into one (`git rebase -i` and `git push -f` to this branch). Using `int` for `size_` is still a limitation, but it's a known issue, see discussion at #1584.

DmitryUlyanov:
Done

longjon:
Thanks!

PR #1841, opened 2015-02-07
HDF5_DATA + MEMORY_DATA refuse loudly to transform

### LINKED ISSUE #1718: HDF5DataLayer and MemoryDataLayer silently doesn't do data transformations
The `HDF5_DATA` layer does not do transformations, since it is intended to be completely general w.r.t. the input data. However, setting transformation parameters is completely valid in the protobuf definition since `TransformationParameter` is nested in the layer message so it can be used across data layers.

`MEMORY_DATA` is more complicated as transformations are done for inputs added as `Datum` or `cv::Mat` but not for raw data taken by pointer through `reset()`.

They should complain or even die loudly so that users know this.

shelhamer:
These layers are meant for generic inputs so they do not do transformations. As `transform_param` was pulled out of data layers' proto definitions for reuse, these layers silently ignored transformation parameters until now instead of signalling their refusal.

Transforming could be added in a future PR. Better to make sure users know the current state first.

See #1718.

### DIFF
--- a/src/caffe/layers/hdf5_data_layer.cpp
+++ b/src/caffe/layers/hdf5_data_layer.cpp
@@ -58,6 +58,9 @@ void HDF5DataLayer<Dtype>::LoadHDF5FileData(const char* filename) {
 template <typename Dtype>
 void HDF5DataLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
+  // Refuse transformation parameters since HDF5 is totally generic.
+  CHECK(!this->layer_param_.has_transform_param()) <<
+      this->type() << " does not transform data.";
   // Read the source to parse the filenames.
   const string& source = this->layer_param_.hdf5_data_param().source();
   LOG(INFO) << "Loading list of HDF5 filenames from: " << source;
--- a/src/caffe/layers/memory_data_layer.cpp
+++ b/src/caffe/layers/memory_data_layer.cpp
@@ -81,6 +81,9 @@ void MemoryDataLayer<Dtype>::Reset(Dtype* data, Dtype* labels, int n) {
   CHECK(data);
   CHECK(labels);
   CHECK_EQ(n % batch_size_, 0) << "n must be a multiple of batch size";
+  // Refuse transformation parameters since a memory array is totally generic.
+  CHECK(!this->layer_param_.has_transform_param()) <<
+      this->type() << " does not transform data.";
   data_ = data;
   labels_ = labels;
   n_ = n;


DmitryUlyanov:
I can create PR with transforms for memory_data, do you need that ?

shelhamer:
@DmitryUlyanov if extending `MemoryDataLayer::Reset()` to transform is useful to you then you could send a PR -- however we've always thought of this data layer as taking input arrays exactly as they are. Note that it does transform when adding vectors of `Datum` or `cv::Mat` as-is.

@longjon thoughts?

DmitryUlyanov:
@shelhamer well, I don't actually think It's a good way to transform things when doing reset (or Add...Vector) because during training you obviously want to have some random transformations every time. So I did it in `Forward_cpu` and didn't feel net started to work any slower on my 40x40 images with `batch_size = 300`, I did it like that (this version is outdated a bit): 

```c++
void MemoryDataLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
  CHECK(data_) << "MemoryDataLayer needs to be initalized by calling Reset";

  if (this->layer_param_.has_transform_param()) {
    
    // Copy to blob
    Blob<Dtype> b;
    b.Reshape(batch_size_, channels_, height_, width_);
    b.set_cpu_data(data_ + pos_ * size_);

    // transform
    this->data_transformer_.Transform(&b, top[0]);

    // Copy labels
    top[1]->set_cpu_data(labels_ + pos_);

    } else {
    top[0]->set_cpu_data(data_ + pos_ * size_);
    top[1]->set_cpu_data(labels_ + pos_);
  }
  pos_ = (pos_ + batch_size_) % n_;
  has_new_data_ = false;
}
```
And anyway user can omit transform_params in config so memory_data will take input arrays exactly as they are with no overhead at all.
So if this way is satisfactory I will send a PR.

rohitgirdhar:
Even for `MemoryDataLayer`, I don't see how the transformation will work in the `AddMatVector` or `AddDatumVector` functions.. Since we don't specify the transformation parameters in the prototxt, I don't see any way to specify them for a `MemoryDataLayer` object through code (or is there? Please let me know). The `Mat`s I pass to `AddMatVector` are used as is, and it even complains if I don't resize them to 227x227 (or whatever other dimensions specified in prototxt) before passing..

rohitgirdhar:
Even with MemoryDataLayer, I use the prototxt to specify transform parameters to use (in `AddMatVector` function). However the check added in this PR fails, and the program aborts. Is there another way to add `DataTransfomer` parameters through the code? (whatever examples I've seen take it from the prototxt file).

shelhamer:
This check was accidentally too stringent for `MemoryDataLayer`. @DmitryUlyanov your approach is better, so please send a PR for transforming memory data as outlined.

(While transforming during forward isn't ideal it's the best solution at present.)

PR #1842, opened 2015-02-07
Dynamic Linking

shelhamer:
Link the tools, tests, examples and pycaffe dynamically, drop the static tests, and set `rpath` so everything works within the `caffe` dir. This should reduce the build size and prepares for the Python layer in #1703. 

Thoughts:

- resolving `LINKFLAGS`, `LDFLAGS`, and `DYNAMIC_LDFLAGS` could be a good idea.
- is the origin `rpath` worth it? I suppose it is to make everything work out of the box.
- matcaffe is still static -- I don't know anything about mex linking and there isn't a MATLAB layer to motivate it (yet)

@longjon should surely check this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,10 +123,8 @@ TEST_CU_BINS := $(addsuffix .testbin,$(addprefix $(TEST_BIN_DIR)/, \
 TEST_CXX_BINS := $(addsuffix .testbin,$(addprefix $(TEST_BIN_DIR)/, \
 		$(foreach obj,$(TEST_CXX_OBJS),$(basename $(notdir $(obj))))))
 TEST_BINS := $(TEST_CXX_BINS) $(TEST_CU_BINS)
-# TEST_ALL_BIN is the test binary that links caffe statically.
+# TEST_ALL_BIN is the test binary that links caffe dynamically.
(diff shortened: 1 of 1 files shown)


longjon:
@shelhamer I've updated this:
* `DYNAMIC_LDFLAGS` serves no purpose when rpath has to be set explicitly anyway, so get rid of it.
* Set rpath everywhere it's needed, so Travis passes now.
* Make `DYNAMIC_NAME` an order dependency.
* Add the appropriate incantations for OS X. This builds and "runs" for me, but `make pytest` crashes on my machine. I think this is a personal configuration problem, so you may wish to try with your own OS X setup.

shelhamer:
Sweet -- this checks out for me on OS X and a linux machine. Thanks Jon!

shelhamer:
Fixed rpath for examples in 5c81243.

jeffdonahue:
Just wanted to say this is awesome, thanks for doing this @shelhamer & @longjon.  Changing core code and doing an incremental build with `make -j everything` is now practically instant since the tests aren't rebuilt, and the build dir with both debug and release builds is <500 MB now (rather than ~2 GB).

PR #1740, opened 2015-01-16
Support OS X Yosemite / 10.10

shelhamer:
Compiling Caffe on OS X Yosemite almost just works, but this takes care of the details:

- pick libstdc++ for OS X (regardless of version)
- make gtest rely on its own tuple to not conflict with clang (I'm not sure this is necessary, but it seems to be needed sometimes and does no harm)
- 10.10 has Accelerate while 10.9 has vecLib for BLAS

Thanks to @leonardt @drdan14 @pluskid scottsievert and others for reporting these issues with fixes.

Note: this PR is to dev because a release will be made soon.

Issue #1700, opened 2015-01-09
Python can't set the device / phase for net initialization


shelhamer:
cuDNN handles and such are acquired at net initialization when the cuDNN layers that require these resources are constructed. Since the Python interface only exposes `set_device()` as a method of `Net` instead of a module function, it is too late to actually set the device for cuDNN computation once the net is made. All cuDNN computation from Python is run on GPU 0 for this reason, and attempts to set other devices will fail with

    status == CUDNN_STATUS_SUCCESS (8 vs. 0)  CUDNN_STATUS_EXECUTION_FAILED

due to the disagreement between initialization and execution.

@longjon has the workaround for now: use the environment variable CUDA_VISIBLE_DEVICES instead of using set_device.

The simplest fix is to expose `set_device()` and `set_phase` from the `caffe` module itself as functions. This is only a bandaid and changes the interface.

Of course The Right Idea is to make Net responsible for device and phase, set them at initialization, and never switch #1500... but that involves a few details.

alfredox10:
How do I make this modification for DIGITS? I am getting that same error there after the first epoch.

lukeyeager:
@alfredox10, can you open a new issue against DIGITS? I bet your issue is unrelated to this one.
https://github.com/NVIDIA/DIGITS/issues/new

PR #1849, opened 2015-02-09
Next: release candidater

shelhamer:
A lot has happened since the last release! This packages up 400+ commits by 27 authors. Thanks all! A follow-up release will come in ~3 weeks to cover documentation, cuDNN v2, and as always more new features.

- BVLC models for unrestricted use #1650: the BVLC models including their weights are licensed for unrestricted use (already announced but included for completeness).
- New Development Workflow #1743 #1887: switch to a "one true branch" model in which all changes stem from `master` and releases are tagged. This is simpler and frees up maintainer time for real work.
- Layer Registry #1167 #1270: the layer factory is now a registry to decentralize the inclusion of layers in the framework. Now `REGISTER_LAYER_CLASS` follows `INSTANTIATE_CLASS`.
- Pythonified Caffe + Python layer #1703: expose `Layer` and all `Solver` to Python and implement `Layers` in Python via the `PythonLayer` type. **This changes the pycaffe interface.**
- Give Phase to Net #1790: phase is an instantiation argument to `Net` and immutable. This is the first step to dismantling the Caffe singleton and prevents Phase missteps like #1430. **This changes the Net constructor and pycaffe and matcaffe interfaces.**
- Dynamic Linking #1842: link the tools, tests, examples, and pycaffe dynamically.
- CMake build #1667: the CMake build, while unofficial, has been vastly improved and should work automatically in many cases. Feedback and follow-up improvements are welcome!
- Layer type is a string #1694: eliminates the use of an enum for the `LayerParameter` `type`; adds a `message ParamSpec` unifying parameter-related proto fields for a single parameter (`name`, `blobs_lr`, `weight_decay`, ...) into a single structure. **You should update your model definitions.**
- ExpLayer #1204 adds a layer to compute the exp of the input
- Reshaping data #1313 `Data` and `ImageData` layers now reshape when the batch size == 1 to handle inputs of varying dimensions
- Encoded image inputs #1239 #1878: save storage by storing + loading encoded bytes instead of decoded uint8 or float data.
- OS X 10.10 / Yosemite #1740: Makefile configured for Yosemite + CUDA 7 + libc++. No more libstdc++ torture and editing homebrew formulae -- if you have already installed Caffe by the old method then clear out the dependencies and do it the easy way.
- 1x1 conv. optimization #1118: speed-up 1x1 convolution with the `CAFFE` engine by skipping `im2col` when it's the identity.
- make LMDB default #1128: concurrent access wins.

**Fixes**

- Test phase jitter #1430: Random cropping was accidentally done at test time *during training* for models with cropping configured. Results of `caffe test` were unaffected. Fixed by #1790.
- Pycaffe device / mode / phase #1700: Python couldn't set the device, mode, or phase for proper init. since these methods were attached to `caffe.Net`. Fixed by exposing these methods at the module level in #1790. Call `caffe.set_{mode,device}` before making a `Net`. This also solves issues with pycaffe and cuDNN.
- Image loading failures #1837: `ImageDataLayer` would silently skip images it failed to load, but now it dies loudly.

**Documentation**: The follow-up release will concentrate on pycaffe + matcaffe examples and interface documentation.

- new installation guides are split up by platform
- the examples reflect the protobuf definition and interface changes in #1694 and elsewhere

**Dependencies**: unchanged, except

- CUDA 7 is strongly suggested -- especially on OS X
- cuDNN compatibility is still at v1 but see #1731 for v2 release candidate compatibility -- the switch to v2 will come when cuDNN v2 is released

**Interface Change / Deprecation**

- the new protobuf model definition format is detailed in #1694. Old models are currently automagically upgraded but **you should upgrade with the included tools** `upgrade_net_proto_{text,binary}`.
- layers are registered for layer factory #1167 #1270 -- see the [layer development guide](https://github.com/BVLC/caffe/wiki/Development)
- Phase is now set by `NetParameter::set_type` or the `Net(string file, Phase phase)` constructor #1790.
- Pycaffe phase, mode, and device handling are done by module methods or set at construction. Call `caffe.set_{mode,device}` before making a `Net` and include the phase as an argument as `net = caffe.Net('def.prototxt', caffe.TRAIN` for example.

:coffee:


bhack:
We are interested in https://github.com/BVLC/caffe/pull/1482 and https://github.com/BVLC/caffe/pull/1484 for the next release . This PR was opened in November and already partially reviewed. We are available to reserve working time to finalize that if you have other feedbacks. /cc @mtamburrano

shelhamer:
@bhack re: #1482 and #1484 I need to give conditional computation more thought before we pick any one approach. However we are working on a means to decentralize layers to have a "layer zoo" for quicker integration of layers while canonical BVLC/caffe editions are developed.

We'll think about conditional computation and pass over #1482 and #1484 for the follow-up release coming sooner than not.

bhack:
@shelhamer Ok. In this PRs we have have followed BVLC members design feedbacks for conditional computation. We are available to contribute working time to improve  or generalize more the design with other feedbacks. Instead generally we are not interested to contribute in a decentralized "layer zoo" in working hours cause probably there  will not be any Quality Assurance policy on this kind of solution.
But I still think that could be useful for maintain spare contributions.

ducha-aiki:
@shelhamer
Actually, we have done some bets in our office if any layer from @bhack  and Co would be ever merged (while some others branches are merged very quickly).  Now stacks are 70 to 30 that they would be never merged :) No offence.

mtamburrano:
@ducha-aiki I too would like to bet 10 bucks against us, too ;)

PR #1851, opened 2015-02-10
Fixes for CuDNN layers: only destroy handles if setup

### LINKED ISSUE #1850: Failed in test_layer_factory.cpp
Hi, Folks

Have anyone experienced this problem: failed the test in LayerFactoryTest/3.TestCreateLayer?

It seems that it is due to the update 4 days ago, or do I have some configuration errors?

Thanks

jeffdonahue:
Fixes #1850 -- layer factory test I added (which constructs and destroys each layer type) with #1694 breaks for CuDNN layers.  This fixes that.  (I had to do a similar thing for a couple of other layers in #1694 (794cd7f), but wasn't compiling with CuDNN enabled while developing #1694 so didn't see the CuDNN failures.)


shelhamer:
Thanks Jeff.

PR #1865, opened 2015-02-14
Implement Krizhevsky-style relighting augmentation

yanii:
Hi all,

I'm trying to add the various augmentation tricks to the Caffe code which are sorely missed at the moment, and this is my first patch towards that. Later much more involved patch for full-image cropping/scale augmentation to come.

This attempts to implement the relighting augmentation as described in the Krizhevsky supervision paper. Towards this, there is a utility to calculate both the pixelwise mean (of floating point precision) and pixelwise principal components of a given data set (e.g. LMDB).

There are some design decisions to think about - most notably how to store/load the eigenvectors/eigenvalues. I took the approach of loading these directly from the prototxt parameters, just as pixelwise mean is done, since they are not very large and this is a bit more transparent. I could also see the view that they should be stored as a separate proto file however, as the image-wise mean is done. 

I really don't like the code duplication that exists in data_transformer at the moment (i.e. for the mean/croppping), etc, but I followed the present format and repeated the relighting code for each of the various data formats, but I personally think there should be some work into refactoring this in future. I recognize that due to differences in the formats this isn't that simple though.

Yani

### DIFF
--- a/include/caffe/data_transformer.hpp
+++ b/include/caffe/data_transformer.hpp
@@ -5,6 +5,7 @@
 
 #include "caffe/blob.hpp"
 #include "caffe/common.hpp"
+#include "caffe/filler.hpp"
 #include "caffe/proto/caffe.pb.h"
 
 namespace caffe {
@@ -107,6 +108,10 @@ class DataTransformer {
   Phase phase_;
   Blob<Dtype> data_mean_;
   vector<Dtype> mean_values_;
+  vector<Dtype> eigen_values_;
+  Blob<Dtype> eigen_vectors_;
+  shared_ptr<Filler<Dtype> > relight_filler_;
+  Blob<Dtype> relight_;
 };
 
 }  // namespace caffe
--- a/src/caffe/data_transformer.cpp
+++ b/src/caffe/data_transformer.cpp
@@ -32,6 +32,35 @@ DataTransformer<Dtype>::DataTransformer(const TransformationParameter& param,
       mean_values_.push_back(param_.mean_value(c));
     }
   }
+
+  int eigen_val_size = param_.eigen_value_size();
+  if (eigen_val_size > 0) {
+    CHECK_GT(param_.eigen_vector_component_size(), 0) <<
+      "Must specify both eigen vectors and eigen values" <<
+      " to do re-lighting augmentation";
+    CHECK_EQ(param_.eigen_vector_component_size(),
+      eigen_val_size*eigen_val_size) <<
+      "Specify an eigen vector per channel, each vector of channel length";
+
+    for (int c = 0; c < eigen_val_size ; ++c) {
+      eigen_values_.push_back(param_.eigen_value(c));
+    }
+
+    CHECK_EQ(eigen_val_size*eigen_val_size,
+      param_.eigen_vector_component_size()) <<
+      "Eigen vectors should be "<< eigen_val_size
+      << "^2 matrix for eigen vectors of length " << eigen_val_size;
+
+    eigen_vectors_.Reshape(1, 1, eigen_val_size, eigen_val_size);
+    Dtype *eigen_vec_data = eigen_vectors_.mutable_cpu_data();
+    for (int i = 0; i < param_.eigen_vector_component_size(); ++i) {
+        eigen_vec_data[i] = param_.eigen_vector_component(i);
+    }
+
+    relight_filler_ =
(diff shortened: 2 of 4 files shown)


yanii:
Travis CI build for "WITH_CUDA=false WITH_CMAKE=false" failed on BlobMathTest, which as far as I can see has no relation to my patch. Running same test on my machine with CPU only and not using CMake this test passes. Looks like numerical error:

```
[----------] 3 tests from BlobMathTest/0, where TypeParam = caffe::FloatCPU
[ RUN      ] BlobMathTest/0.TestSumOfSquares
src/caffe/test/test_blob.cpp:117: Failure
Value of: this->blob_->sumsq_diff()
  Actual: 16266.95
Expected: expected_sumsq * kDiffScaleFactor * kDiffScaleFactor
Which is: 16266.945
[  FAILED  ] BlobMathTest/0.TestSumOfSquares, where TypeParam = caffe::FloatCPU (1 ms)
```

yanii:
Just discovered there is an overflow issue running the compute_image_pca util on very large datasets (i.e. imagenet training set) since it is summing over all pixels, had only used it on smaller datasets so far. Will re-implement using a numerically stable algorithm ASAP.

shelhamer:
Thanks for the augmentation!

> how to store/load the eigenvectors/eigenvalues. I took the approach of loading these directly from the prototxt parameters, just as pixelwise mean is done

Good choice. It's best to bundle these in the definition for clarity.

> I really don't like the code duplication that exists in data_transformer at the moment
> there should be some work into refactoring this in future

No argument. The data transformer was at least a step to reduce duplication across data layers, but the data pipeline deserves attention to make it simpler and more general.

shelhamer:
Re: https://github.com/BVLC/caffe/pull/1865#issuecomment-74391707 @jeffdonahue could you relax the `EXPECT_FLOAT_EQ` usage in the blob math tests? They're apparently not totally reliable on cheap / free hardware.

yanii:
Wow, doing the covariance calculations on Imagenet is a numerical nightmare! I've implemented an online method for calculating mean/co-variance that works well though - the side-benefit is it only needs one pass of the dataset now, and it seems to work nicely for Imagenet. I get the following eigenvectors/values, which are around what I got when I ran the naive implementation on a 1% subset of Imagenet:

```
Processed 1281167 files.
Mean channel values: [103.8597331040042, 116.7835176771538, 123.6662784959614]
Channel Covariance: [5244.208369419691, 4492.485855678318, 3891.093127656954;
  4492.485855678318, 4770.908271347861, 4356.913884984975;
  3891.093127656954, 4356.913884984975, 5044.990455324214]
mean_value: 103.86
mean_value: 116.784
mean_value: 123.666
eigen_value: 13515.7
eigen_value: 1252.4
eigen_value: 292.057
eigen_vector_component: 0.582858
eigen_vector_component: 0.581974
eigen_vector_component: 0.567083
eigen_vector_component: -0.691932
eigen_vector_component: -0.0104332
eigen_vector_component: 0.721888
eigen_vector_component: -0.426036
eigen_vector_component: 0.813141
eigen_vector_component: -0.396605
```

There doesn't seem to be any published numbers to compare to unfortunately. 

Future work might be to vectorize this code by calculating the covariance for a block of samples at a time, and combining with the online covariance according to the rule (see end of covariance section of http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Covariance) . Tried doing this already but it's tricky to get it numerically correct for an image-sized block, and didn't speed things up as much as I'd hoped. Might make more sense to just run on a smaller subset of the data however if speed is critical.

daodaofr:
Do you have a demo about how to use the compute_image_pca.cpp ?

yanii:
Simply run it like you would the compute_image_mean util, i.e. give it the lmdb/leveldb as a parameter - that's all.

It prints out usage when you call it with no params, although it looks like it's slightly wrong at the moment, in that there is no output required (or created). The mean and PCA are simply printed to the screen.

vimalthilak:
@yanii What type of accuracy improvements did you see after adding this augmentation step?

yanii:
@vimalthilak sorry for the late response! Was working like crazy towards the ICCV deadline.

I haven't run with with a standard network I'm afraid - I will try to do so when I have a spare GPU and report it though. Please feel free to try it out and let me know though.

shelhamer:

PR #1870, opened 2015-02-15
RMSprop Solver implementation

erogol:
I implemented and tested RMSprop as suggested by G. Hinton's Coursera Lecture 6. This is my fitsr pull reuest to the community and in the github as well please be nice :)

My implementation is basically an alternation of AdaGradSolver. I just copied the code below and change the computation lines based on RMSprop formulation. In the future, I also like to add Momentum step as well in to RMSprop. Basically the idea is to compute an update step  based on Momentum then another step by RSMprop then aggregate both steps into a single update. Regarding my experiences this combination is better than AdaGrad ans plain SGD with faster converge rate and lower overfitting.


KeyKy:
In the future, I also like to add Momentum step as well in to RMSprop. However, Check failed: 0 == this->param_.momentum() (0 vs. 0.9) Momentum cannot be used with RMSProp.

Noiredd:
@KeyKy Because [RMSprop does not have a momentum term](http://ruder.io/optimizing-gradient-descent/index.html#rmsprop) at all - it's how it works mathematically. If you're looking for an adaptive solver with momentum, try Adam.

PR #1872, opened 2015-02-15
EmbedLayer

jeffdonahue:
Based on #1486 (N-D blobs) and #1663 (parameter gradient accumulation).  This adds `EmbedLayer` (should probably change the name to `EmbeddingLayer` for consistency with `PoolingLayer` etc.), which essentially learns a lookup table for integer inputs, useful for language modeling and such.   Its computation is equivalent to an `InnerProductLayer` where the inputs are "one-hot" vectors, but instead of explicitly representing the one-hot vectors (which wastes lots of memory), this assumes the input itself is the indices of the "hot" index of those one-hot vectors (like the label inputs for the categorical losses). This should probably be replaced with SparseInnerProduct (#937) once that's merged, assuming that's faster -- this is a more lightweight change (or at least it will be once #1486 is merged) that continues the unfortunate trend of casting floats to ints as labels.


jzhang533:
EmbedLayer has a blob stores the vocabulary_size X embedding_size embeddings, during Forwards/Backward, only invloved words' embeddings are used for computation, but all the embeddings (the whole blob) are updated during solving (Solver::ComputeUpdateValue, Blob::Update).  

Is my understanding correct?

jeffdonahue:
@jzhang533 yes that's correct; it has the same behavior as other parameter layers (`InnerProductLayer`, `ConvolutionLayer`).

jzhang533:
@jeffdonahue thanks for clarify, am trying to learn embeddings for a large vocabulary, will try to figure out a way to avoid needless computation during solving.

PR #1878, opened 2015-02-16
Encoded cleanup

philkr:
I decided to clean the encoded flag up a bit and add some more options.
convert_imageset can now store images encoded with resize and gray options on. Internally io.cpp will use opencv to re-encode a resized image if needed. It is also possible to change the encoding, using the encode_type flag.

The DecodeDatumToCVMat and DecodeDatum functions were renamed to DecodeDatumToCVMatNative and DecodeDatumNative . The ...Native functions will load images with their native color channels (grey or BGR) instead of always converting encoded images to BGR. This allows the encoding of 1d label masks (such as image class labels).

### DIFF
--- a/src/caffe/test/test_io.cpp
+++ b/src/caffe/test/test_io.cpp
@@ -289,8 +289,8 @@ TEST_F(IOTest, TestDecodeDatum) {
   string filename = EXAMPLES_SOURCE_DIR "images/cat.jpg";
   Datum datum;
   EXPECT_TRUE(ReadFileToDatum(filename, &datum));
-  EXPECT_TRUE(DecodeDatum(&datum));
-  EXPECT_FALSE(DecodeDatum(&datum));
+  EXPECT_TRUE(DecodeDatumNative(&datum));
+  EXPECT_FALSE(DecodeDatumNative(&datum));
   Datum datum_ref;
   ReadImageToDatumReference(filename, 0, 0, 0, true, &datum_ref);
   EXPECT_EQ(datum.channels(), datum_ref.channels());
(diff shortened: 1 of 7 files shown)


tools/convert_imageset.cpp
```diff
@@ -78,11 +81,14 @@ int main(int argc, char** argv) {
...
   LOG(INFO) << "A total of " << lines.size() << " images.";
 
-  if (encoded) {
-    CHECK_EQ(FLAGS_resize_height, 0) << "With encoded don't resize images";
-    CHECK_EQ(FLAGS_resize_width, 0) << "With encoded don't resize images";
-    CHECK(!check_size) << "With encoded cannot check_size";
-  }
+  if (encode_type.size() && !encoded)
+    LOG(INFO) << "encode_type specified, assuming encoded=true.";
+//   if (encoded) {
```
shelhamer: Why keep these around as comments? Just drop.

src/caffe/test/test_io.cpp
```diff
@@ -309,38 +309,17 @@ TEST_F(IOTest, TestDecodeDatumToCVMat) {
...
   Datum datum;
   EXPECT_TRUE(ReadFileToDatum(filename, &datum));
-  cv::Mat cv_img = DecodeDatumToCVMat(datum);
+  cv::Mat cv_img = DecodeDatumToCVMatNative(datum);
   EXPECT_EQ(cv_img.channels(), 3);
   EXPECT_EQ(cv_img.rows, 360);
   EXPECT_EQ(cv_img.cols, 480);
 }
 
-TEST_F(IOTest, TestDecodeDatumToCVMatResized) {
```
shelhamer: Tests that check the dimensions for resizing should be kept.

shelhamer:
@philkr thanks for improving the encoding. Can you squash this into a single commit and take care of the inline comments? Thanks.

src/caffe/test/test_io.cpp
philkr: Since the decode-resize functions, and decode-gray functions were never used, I decided to drop them. The decoding should really only produce a datum of the encoded image directly and not do any (hidden) conversion. The tests I removed correspond to the removed functions.

shelhamer: You're right -- I was confused and thought you had removed `TestReadImageToCVMatResized`. This is fine.

shelhamer:
Please combine your commits for the sake of the history http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html then this can be merged.

philkr:
ahh, the wonders of git :)

PR #1879, opened 2015-02-16
Correct 'epochs' to 'iterations'

bamos:
See https://github.com/BVLC/caffe/blob/master/models/bvlc_reference_caffenet/solver.prototxt


sguada:
Thanks @bamos

PR #1880, opened 2015-02-16
Add PReLULayer

tnarihi:
I implemented parameterized ReLU unit, which is described in http://arxiv.org/abs/1502.01852 . In my caffe fork, I tried RReLU for CIFAR10 quick examples and the performance was improved about 3%. I am happy If it is worth merging. I don't care even if it is not because I implemented it just for my purpose.

### DIFF
--- a/include/caffe/neuron_layers.hpp
+++ b/include/caffe/neuron_layers.hpp
@@ -654,7 +654,11 @@ class ThresholdLayer : public NeuronLayer<Dtype> {
 };
 
 /**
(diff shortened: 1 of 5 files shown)


pannous:
Not sure, but doesn't relu_layer already have negative_slope, which could be the same as the alpha in the paper?
** https://github.com/BVLC/caffe/blob/dev/src/caffe/layers/relu_layer.cpp

tnarihi:
@pannous The difference between them is that 

1. slopes can be learned.
2. slopes can vary across channels.

pannous:
I see, thanks

tnarihi:
Modified to allow in-place computation.

tnarihi:
Replaced by #1940

PR #1884, opened 2015-02-17
Fix `draw` to support new protobuf format

tnarihi:
This PR fixes the problem of drawing layers which has additional info to be shown. It was due to string type of layer. This PR complements 1344d1b which have been merged by cherry-pick.


shelhamer:
Thanks for the fix Takuya!

PR #1886, opened 2015-02-17
new conv algorithm

songjmcn:
new conv algorithm(in src/caffe/conv2),faster,small memory use, I want to replace old caffe convolution algotirhm


longjon:
Do you have benchmarks for this on a standard model? Pardon my skepticism, but it's a little hard to believe that this will beat the highly optimized gemm-based approach (or cuDNN, for that matter).

shelhamer:
Closing as this is unproven, undocumented, and unlikely to improve on either GEMM or cuDNN.

PR #1894, opened 2015-02-17
Making python3 work with cmake and the new python wrapper

philkr:
A small change than allows the cmake based build to use python3.
Simply specify the "python_version" in the cmake configuration (either 2 or 3).

Also replaced "import_array" with "import_array1" which works for both python 2.7 and 3.? .

Protobuf and pycaffe should now work with python3. Updated the docs to reflect the fact that protobuf 3.0 alpha is required for python3 (not for python2).


philkr:
Rebased the PR to master #1923.

PR #1895, opened 2015-02-18
InverseMVNLayer

jyegerlehner:
This PR extends the MVNLayer to allow the mean and variance blobs to be exported as top blobs. It adds a new layer type InverseMVNLayer which takes the mean and variance as bottom blobs, and performs the inverse operation (adding the mean back, and denormalizing for variance). A use case for this is an autoencoder that feeds input into the MVNLayer, and generates the output from the InverseMVNLayer, with autoencoding layers in-between.

There was also a problem with the MVNLayer that was fixed: if it is given input that has exactly zero variance (e.g. a solid color RGB image, with across_channels=false), it computes the variance as E(X^2) - (EX)^2, but the result isn't usually exactly zero, but has small negative values due to floating point resolution. The subsequent square root operation then produces NaN. This PR also fixes this issue.


jyegerlehner:
Oops this is broken for across_channels=true. I will fix and update as am able. Or close.

jyegerlehner:
Replaced with [1979](https://github.com/BVLC/caffe/pull/1979).

PR #1897, opened 2015-02-18
Fixed RPATH for python and tools cmake scripts.

BlGene:
These changes were necessary to get the new cmake ( #1667)  scripts to work on my system.
@Nerei @shelhamer

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -3,16 +3,23 @@ if(NOT HAVE_PYTHON)
   return()
 endif()
 
+# when building, don't use the install RPATH already
+# (but later on when installing)
+SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
+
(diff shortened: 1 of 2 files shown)


Nerei:
But this PR breaks examples with build-tree. Give me some time, I'll take look more precisely a bit later and suggest how to solve this.

Nerei:
@BlGene  (CC: @shelhamer)

I took couple hours before I realized where is the bug :-) Explanation is here: https://github.com/BVLC/caffe/pull/1939. Please take a look if this patch fixes your problems. One more proof that to write something in source folder (symlink in this case) is a very bad practice.

Nerei:
@BlGene Please don't forget to clean install-tree before testing.

BlGene:
This is fixed by @Nerei  in a more elegant way in PR #1939.

shelhamer:
Thanks for the initial fix @BlGene and follow-up @Nerei.

PR #1900, opened 2015-02-18
Full-sized image cropping and Scale Augmentation

yanii:
At the moment Caffe is missing a range of training time augmentation tricks, which is one of the main reasons that networks such as AlexNet, VGG and GoogleNet in the model zoo do not realize their published accuracy. One such augmentation is image relighting, for which I have a separate pull request. This patch on the other hand adds full-size image based cropping and scale augmentation support to data_transformer which is a bit more involved.

State-of-the-art networks are trained on Imagenet using random crops from original images (not necessarily square or of minimum image size) resized up or down such that one side is of a minimum size (i.e. 256) - rather than from pre-cropped/an-istropically scaled square images (i.e. 256x256) as Caffe currently uses. Furthermore scale augmentation is usually used, in the form of varying the size of the minimum side an image is resized to (i.e. between 256 - 512 in VGG, or similar usage in GoogLeNet), or using a set of sizes. This patch allows for the former which is apparently more effective.

Luckily it turns out the current DB backend supports storing full-sized images already, with the exception that for some rare, very large images in Imagenet, the Google protobuf message `ParseFromString` will refuse to parse such a long string. A common work-around for using protobuf messages with large messages is to use a `ZeroCopyInputStream`/`CodedInputStream`  and `ParseFromCodedStream` for which maximum message sizes can be set. Google however says using protobuf messages for large objects should be avoided - so perhaps a transition away from this in future makes sense.

Two new parameters are added to `TranformationParameter`:

`min_size` - the minimum image size to scale the smallest image side to (i.e. 256 in the existing framework)
`min_size` - the maximum image size to scale the smallest image side to

Setting min_size will enable full-sized image cropping, while setting both min_size and max_size will enable scale-augmentation.

One observation is that to do image resizing on Blobs (rather than CvMat encoded images), I first copy the format into a CvMat, call cv::resize, and then copy back. It doesn't make sense to me to re-implement such functionality from scratch, and from a maintainability standpoint I lean towards using the OpenCV resize. 

In fact it might make sense to remove the current code duplication in data_transformer by merging the code for both CvMat encoded images and Blobs to both use CvMats, this allows the appropriate flexibility to easily add transformations in future (which seems likely) while keeping the code more maintainable. I recognize this might incur a slight performance penalty for the non-encoded DBs though.

Yani

### DIFF
--- a/include/caffe/util/io.hpp
+++ b/include/caffe/util/io.hpp
@@ -148,6 +148,12 @@ cv::Mat DecodeDatumToCVMat(const Datum& datum);
 
 void CVMatToDatum(const cv::Mat& cv_img, Datum* datum);
 
+template <typename Dtype, int channels>
+cv::Mat BlobToCVMat(Blob<Dtype> *blob, int num = 0);
+
+template <typename Dtype, int channels>
+void CVMatToBlob(cv::Mat img, Blob<Dtype> *blob);
+
 template <typename Dtype>
 void hdf5_load_nd_dataset_helper(
     hid_t file_id, const char* dataset_name_, int min_dim, int max_dim,
--- a/src/caffe/data_transformer.cpp
+++ b/src/caffe/data_transformer.cpp
@@ -1,4 +1,5 @@
 #include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
 
 #include <string>
 #include <vector>
@@ -206,8 +207,6 @@ void DataTransformer<Dtype>::Transform(const cv::Mat& cv_img,
   const int num = transformed_blob->num();
 
   CHECK_EQ(channels, img_channels);
-  CHECK_LE(height, img_height);
-  CHECK_LE(width, img_width);
   CHECK_GE(num, 1);
 
   CHECK(cv_img.depth() == CV_8U) << "Image data type must be unsigned byte";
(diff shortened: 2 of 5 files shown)


yanii:
Note: This code works well on the HEAD I branched off of the dev branch from last week (5e61f551220c0babda4ac9674e4604d0ba61df9e), however new changes to DecodeDatum in 52465873d08cfd58a4c84ee8b239908814de4f39 have broken it at runtime:
```
data_transformer.cpp:209] Check failed: channels == img_channels (3 vs. 1) 
```
I'm looking into it now, I think this is because some of the original imagenet images are actually greyscale, but I can't see how this would have changed for a LMDB that worked fine before. If anyone has any tips I'd love to hear!

shelhamer:
@philkr `DecodeDatum` might need a mode to force color for making input LMDBs and such when a few inputs might be greyscale even though it is a color model.

avdmitry:
@yanii,
there is problem if somebody wants to use scale augmentation with full image mean stored in file. In this case mean should be subtracted before scaling.

p.s. I also think it'll be usefull to remove code duplication and use OpenCV for image transformation. I found this issue trying to use this code for "Data" layer.

yanii:
@avdmitry Yes, this doesn't support using an image mean file, since the full images (of arbitrary dimension) are being used and randomly resized before crop - it doesn't really make sense to be using a mean image (how would you produce it for example?).

Question is if I've put an assert there to inform the user of this though (i.e. only mean pixel supported with scale augmentation) - which may be a good point.

avdmitry:
@yanii,
I made several changes: https://gist.github.com/avdmitry/bbddbc0be6ad564114fa
This code isn't absolutely perfect, but seems it works for me, maybe you'll find something useful there.
I added assert for full image mean, but I think it'll be better to handle this case proper, because nothing really doesn't interfere to it.

shelhamer:
Closing this PR since it is against the deprecated dev branch. Further augmentations could be PR'd to master.

PR #1907, opened 2015-02-19
relax MemoryData transform check to warning

shelhamer:
warning is still useful to keep from accidentally running data without
pre-processing.


mtngld:
@shelhamer, I agree with @yudanaz. Is this warning required?

danst18:
@shelhamer I just came across this and I don't understand it either. What's the purpose of this warning if the data seems to be always transformed as @yudanaz says. Could you please clarify it?

mtngld:
@shelhamer I agree with all of the above...

Issue #1901, opened 2015-02-18
Training with encoded LMDB broken by #1878


yanii:
Some change that's happened in the last week seems to have broken training with the dev branch and encoded LMDB for me, and I think it may be the changes from 52465873d08cfd58a4c84ee8b239908814de4f39 to the functions for reading encoded data. It seems that it will now load data in native format, including using the same number of channels as the original image (imdecode has the flag set to -1), and this is not handled properly datalayer or data_transformer.

Instead, at least for an existing encoded LMDB, greyscale (single-channel) images from Imagenet now make it all the way through to data_transfomer and throw an assert: 

```
F0218 23:24:28.115943  2648 data_transformer.cpp:208] Check failed: channels == img_channels (3 vs. 1) 
*** Check failure stack trace: ***
    @     0x7f102732ac6c  (unknown)
    @     0x7f102732abb8  (unknown)
    @     0x7f102732a5ba  (unknown)
    @     0x7f102732d551  (unknown)
    @     0x7f1027894094  caffe::DataTransformer<>::Transform()
    @     0x7f1027888a45  caffe::DataLayer<>::InternalThreadEntry()
    @     0x7f10278234cf  boost::_mfi::mf0<>::operator()()
    @     0x7f1027823432  boost::_bi::list1<>::operator()<>()
    @     0x7f102782317c  boost::_bi::bind_t<>::operator()()
    @     0x7f1027822e12  boost::detail::thread_data<>::run()
    @     0x7f101f03ad3a  (unknown)
    @     0x7f101d15d0a5  start_thread
    @     0x7f102690988d  (unknown)
    @              (nil)  (unknown)
Aborted
Done.

```

shelhamer:
It seems like the encoding / decoding need to have a flag for forcing decoding to color or grayscale. This could be exposed through caffe.proto in `transform_param`.

yanii:
Yeah, that makes sense. We should make sure it's backwards compatible with existing networks though - so perhaps it should default to forcing colour? Doesn't seem like a great solution, but I can't see an alternative.

philkr:
One issue with force_color (or is_color) for the DataLayer is that encoded and non-encoded images will have different semantics. We load a non-encoded datum as is (however many dimensions it has), while we convert encoded datums.
We could just add a encoded_is_color flag and keep having different semantics between non-encoded and encoded images, but at least make it explicit.
Maybe a nicer solution would be to store a native_encoding flag in the database, and add it to newly created db's.

[...]

PR #1910, opened 2015-02-19
force_encoded_color flag and warning against mixed color-gray encoded DBs

philkr:
This PR fixes issue #1901, where older encoded DBs are not compatible with PR #1878. It adds a force_encoded_color flag to data_param, to force mixed encoded DBs to be read properly (as all color DBs).
It also warns users against the use of mixed color-gray DBs.


src/caffe/layers/data_layer.cpp
```diff
@@ -98,7 +101,15 @@ void DataLayer<Dtype>::InternalThreadEntry() {
 
     cv::Mat cv_img;
     if (datum.encoded()) {
-       cv_img = DecodeDatumToCVMatNative(datum);
+      if (force_color)
+        cv_img = DecodeDatumToCVMat(datum, true);
+      else
+        cv_img = DecodeDatumToCVMatNative(datum);
+      if (cv_img.channels() != this->transformed_data_.channels())
+        LOG(WARNING) << "You dataset contains encoded images with mixed "
```
jeffdonahue: s/you/your/

otherwise, this looks good to me

src/caffe/layers/window_data_layer.cpp
```diff
@@ -281,7 +281,7 @@ void WindowDataLayer<Dtype>::InternalThreadEntry() {
       if (this->cache_images_) {
         pair<std::string, Datum> image_cached =
           image_database_cache_[window[WindowDataLayer<Dtype>::IMAGE_INDEX]];
-        cv_img = DecodeDatumToCVMatNative(image_cached.second);
+        cv_img = DecodeDatumToCVMat(image_cached.second, true);
```
shelhamer: This doesn't seem right -- `ImageData` has its own color / not color flag that should be respected https://github.com/BVLC/caffe/blob/dev/src/caffe/proto/caffe.proto#L522... or maybe it should do the native decoding unless `is_color` is set to true?

shelhamer:
@philkr thanks for working on the fix. Please remember to squash and rebase for a clean merge.

shelhamer:
@philkr this still wasn't a clean merge, so please rebase to the latest dev next time (remember your clone might be out of date). For the sake of this being done with I did the merge manually in 9114424. Thanks.

yanii:
Just started trying to use encoded dbs again, and after re-creating a new encoded database with the latest caffe-dev (which I see is now deprecated) the problem persists - so this doesn't seem to be a problem restricted to older databases.. even newly created DBs need the flag to force it.

Would advise testing this on a dataset with mixed image encodings - I'm using the original imagenet images (without them being re-encoded) and finding this.

philkr:
You are right. I was trying to be a bit too smart about it in ```src/caffe/util/io.cpp:120```. I'll fix it over the weekend.

PR #1911, opened 2015-02-19
Transforms for memory data layer

DmitryUlyanov:
Confused a lot with the code inside, to be true,  let's make it clear and better.

```c++
size_t num = datum_vector.size();
```
There's no reason to make it size_t, anyway it goes right to the functions where it is casted to int.

And because you had no refs to top blob you reshaped it every time in Forward, not the best thing to do probably, are you ok with storing ref to `vector<Blob<Dtype>*>& top` in setup in order to reshape the blob immediately in `set_batch_size` ?

This comment is very confusing, at least because there is no variable num_images, I really could not get what it is about.
```c++
 // num_images == batch_size_
```

and I need to remove this lines somehow: 
```c++
// Apply data transformations (mirror, scale, crop...)
  this->data_transformer_->Transform(mat_vector, &added_data_);
```
```c++
  // Apply data transformations (mirror, scale, crop...)
  this->data_transformer_->Transform(datum_vector, &added_data_);
```
Please post a piece of code how to convert these `mat_vector` and  `datum_vector` to `added_data_` without transformations, didn't find a good looking way to do it. 

But it looks like it is transforming something :) Added setup test, don't know how to organize Forward test with all this random crops, how would you do it ?

### DIFF
--- a/include/caffe/data_layers.hpp
+++ b/include/caffe/data_layers.hpp
@@ -277,14 +277,19 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
   virtual void Forward_cpu(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top);
 
-  int batch_size_, channels_, height_, width_, size_;
+  int batch_size_, channels_, size_;
+  // This is input data shape
+  int height_, width_;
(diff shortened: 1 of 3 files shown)


shelhamer:

Issue #1919, opened 2015-02-20
Dev branch is broken, error loading libcaffe.so


mees:
Currently the dev branch at 079e4168f55112f6cf62f1d9ee7f4927f15978c5 seems broken.
$ build/tools/caffe test
build/tools/caffe: error while loading shared libraries: libcaffe.so: cannot open shared object file: No such file or directory
I reproduced this on two different computers, anybody knows what is causing this?

shelhamer:
Fixed in #1921 -- sorry for the linking issue.

shelhamer:
By the way, the fix is be made to `master` since the `dev` branch will be discontinued in favor of single branch development.

PR #1920, opened 2015-02-20
f29b8e5 Revert "[build] fix rpath for examples" as it breaks libcaffe.so...

mees:
fixes issue #1919

This reverts commit 5c81243ad72234d8de6c347ec704653afc963bca.


shelhamer:
Yeah, so 5c81243 fixes the examples but messed up the tools -- my mistake -- but this will now break the examples again. I'll push an fix to both to `master` shortly -- note that `dev` is discontinued in favor of a one branch development model #1743 and http://caffe.berkeleyvision.org/development.html#versioning.

shelhamer:
Fixed in #1921. Thanks for the report @mees!

mees:
thanks for the fix and the info @shelhamer !

PR #1922, opened 2015-02-20
Cross-channel LRN bounds checking for GPU implementation

### LINKED ISSUE #1864: normalizing using LRN layer
I wish to normalize across channels (I have 30 channels). To do so (unless I misunderstand) I can use the LRN layer with norm_region=ACROSS_CHANNEL and local_size = 61. The local_size = 2*num_of_channels+1 with zero padding should ensure the normalisation factor is constant across channels. 

If I do this in CPU, it works fine. If I run it in GPU when norm_region is larger then the number of channels I get NAN's. 

Any ideas?

Thanks

erictzeng:
Fixes #1864.

When the `local_size` of a `LRNLayer` in `ACROSS_CHANNELS` mode is larger than the number of channels in a blob, we get some nasty indexing errors in the GPU implementation (the CPU implementation is fine). This pull request consists of two contributions:

1. Two new test cases, `TestForwardAcrossChannelsLargeRegion` and `TestGradientAcrossChannelsLargeRegion` that use a `local_size` of 15 on a blob with 7 channels. These tests break the current master implementation.
2. Bound checks in the GPU implementation to ensure that we are no longer indexing out of bounds. This fixes the additional test cases introduced in point 1.

As far as I can tell, this PR is complete, though if anyone finds anything wrong with it of course I'd be happy to fix.

### DIFF
--- a/src/caffe/layers/lrn_layer.cu
+++ b/src/caffe/layers/lrn_layer.cu
@@ -26,26 +26,24 @@ __global__ void LRNFillScale(const int nthreads, const Dtype* in,
     Dtype accum_scale = 0;
     // fill the scale at [n, :, h, w]
     // accumulate values
-    while (head < post_pad) {
+    while (head < post_pad && head < channels) {
       accum_scale += in[head * step] * in[head * step];
-      ++head;
-    }
-    // until we reach size, nothing needs to be subtracted
-    while (head < size) {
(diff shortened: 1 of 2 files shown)


jeffdonahue:
Thanks for the fix @erictzeng!

PR #1925, opened 2015-02-21
provide a solver option to limit runtime (max_seconds)

jsupancic:
This patch lets you say "run for 12 hours = 43200 seconds". 
It's useful when you need exact limits with shared computing resources. 
The runtime limit is enabled by setting  max_seconds to a value > 0.
Maybe it helps? 

Best,
James

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -6,6 +6,8 @@
 
 #include "caffe/net.hpp"
 
+#include <boost/date_time/posix_time/posix_time.hpp>
+
 namespace caffe {
 
 /**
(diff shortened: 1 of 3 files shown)


jsupancic:
Closing because of failed lint tests.

shelhamer:
Actually you should re-open and force push your changes to revise the PR. It seems like this could be useful for the shared cluster situation you describe.

jsupancic:
Hi Evan,

I didn't know I could revise a push request. I'm just getting up to speed on multi-user git. Thanks for the feedback. 

I'll correct the patch (to pass lint) and then re-open. 

Best,
James

jsupancic:
I've updated the branch (with a squashed commit) to pass the lint tests. 

Apparently it is not possible to re-open a pull request after a force push? I should have opened first then forced the push? Anyway, I've submitted a new PR here #1956

PR #1926, opened 2015-02-21
Check caffe tool runs

shelhamer:
Simply run the caffe tool in `runtest` to catch obvious issues like the linking
problem fixed in #1921.


longjon:
One thing to keep in mind here: this means `caffe` needs to return exit status zero when called with no arguments, rather than considering that an error condition. Given this patch we'll definitely notice if we change that, so... merging, thanks.

Issue #1928, opened 2015-02-21
Cropping the mean file


lukeyeager:
Since the huge update, I'm having to fix lots of API changes with my python interfaces to caffe. How do I wade through the commit stream to find out what all needs to change?

This is my most recent problem. I try this:

```python
classifier = caffe.Classifier(deploy_file, snapshot_file,
             image_dims      = (height, width),
             mean            = image_mean,
             raw_scale       = 255,
             channel_swap    = channel_swap,
             )
classifier.predict([image], oversample=False)
```

And I get this error:

    ValueError: operands could not be broadcast together with shapes (3,227,227) (3,256,256) (3,227,227)

Apparently, the mean_file has to be cropped before passing it to the Classifier constructor now. How was I supposed to know that? I don't see it in any of [these commits](https://github.com/BVLC/caffe/commits/master/python/caffe/classifier.py). And how am I supposed to do it? Should I be cropping it myself manually, or should I try to use Classifier.transformer.preprocess()?

shelhamer:
It's true a lot has changed in the last release, which is why we're shooting for more gradual releases from here on. While the release notes are still coming in at #1849 you can look there for a list of changes with interface differences bolded.

With regards to the new `Transformer` in particular you can look at the [classification example history](https://github.com/BVLC/caffe/commits/master/examples/classification.ipynb) instead of `classify.py` itself. In particular 26c16bf414dc0496628f7e34bd03e8f26f2c309d notes

> This is a quick translation of the examples to the caffe.io.Transformer
interface. The results are not strictly identical to the earlier
implementation! The models now use a mean pixel instead of a mean image
for simplicity. The output classifications and detections are preserved
but scores may differ.
> 
> Note: the examples will be rewritten to make use of `caffe.Net` alone
since it is the true interface, but not yet.

In general this step tried to be less magic while keeping configurability. The coming switch to `caffe.Net` examples should clarify the standard practices with the interface, whereas it seems `caffe.{Classifier,Detector}` have confused many.

Re: what to do now about the mean, you could crop it yourself or I suggest taking the mean pixel as in the example. The mean pixel approach simplifies working with images of different size / aspect / resolution and has been shown to work for classification, detection, and segmentation. This is how the BVLC GoogLeNet was learned: https://github.com/BVLC/caffe/blob/master/models/bvlc_googlenet/train_val.prototxt#L13-L15.

[...]

PR #1931, opened 2015-02-21
Add error checking for image mean

lukeyeager:
When setting the mean, assert that it is either one pixel or an array with shape equal to the input data size.

Helps others who run into #1928 by giving them a meaningful error message.

### DIFF
--- a/python/caffe/io.py
+++ b/python/caffe/io.py
@@ -238,11 +238,16 @@ class Transformer:
         """
         self.__check_input(in_)
         if mean.ndim == 1:
+            # broadcast pixel
             mean = mean[:, np.newaxis, np.newaxis]
-        mk, mh, mw = mean.shape
-        in_k, in_h, in_w = self.inputs[in_][1:]
-        #if mk != in_k or (mh, mw) != (in_h, in_w) and (mh, mw) != (1, 1):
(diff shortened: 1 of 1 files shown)


python/caffe/io.py
```diff
@@ -231,11 +231,14 @@ def set_mean(self, in_, mean):
         """
         self.__check_input(in_)
         if mean.ndim == 1:
+            # broadcast pixel
             mean = mean[:, np.newaxis, np.newaxis]
-        mk, mh, mw = mean.shape
-        in_k, in_h, in_w = self.inputs[in_][1:]
-        #if mk != in_k or (mh, mw) != (in_h, in_w) and (mh, mw) != (1, 1):
```
shelhamer: @lukeyeager ha this was actually my check for just this but I accidentally commented it out. Your checks cover more cases so when I merge I'll combine them.

shelhamer:
Could you update this PR to `raise ValueError` instead of `assert` then I can just merge? Thanks.

lukeyeager:
Sure thing.

lukeyeager:
Closing due to inactivity. I'll rebase and move this to a different branch for the sake of sanity in my repo, then repost.

PR #1939, opened 2015-02-22
Cmake Bugfix: install rpath for pycaffe (replacement of #1897)

Nerei:
This is proper replacement of #1897 and contains bugfix with RPATH for pycaffe in install-tree.

The problem was:

1. at build time cmake created symlink 
`<caffe_source>/python/caffe/_caffe.so -> <cmake_build_dir>/lib/_caffe.so`

2. at install time cmake installed `_caffe.so` and updated RPATH properly

3. next at install time cmake installed content of `<caffe_source>/python/caffe` directory, and overwrote installed earlier `_caffe.so` with the symlink ;-)))


BlGene:
@Nerei Hey, thanks for figuring this out. This works for me.

@shelhamer  I'm for merging this.

shelhamer:
Thanks all.

PR #1941, opened 2015-02-22
fixes: cpp_lint.py fails silently with Python3 

jsupancic:
cpp_lint.py fails silently with Python3 (which is the default on some systems).
This commit specifies Python2 with which cpp_lint.py works :-)

Best,
James


longjon:
This should work in general given PEP 394 (https://www.python.org/dev/peps/pep-0394/). Merging, thanks.

PR #1946, opened 2015-02-23
MSRA weight filler

nickcarlevaris:
This PR adds MSRAFiller, which implements an Xavier-like filler designed for use with ReLUs instead of tanh, based on the paper: He et al, "Delving Deep into Rectifiers: Surpassing Human-Level Performance on ImageNet Classification," 2015. 

It also adds a VarianceNorm option to FillerParameters which allows one to normalize by fan_in, fan_out or their average. VarianceNorm applies to the MSRAFiller and the XavierFiller (default behavior unchanged). It also adds tests for MSRAFiller and XavierFiller.

Replaces #1883 (updates based on that discussion and rebased against master). 

Like the XavierFiller, the fan_in and fan_out dimensions are not correct for inner product layers (as pointed out by @seanbell in #1883). However, I did update the documentation to note this.

### DIFF
--- a/include/caffe/filler.hpp
+++ b/include/caffe/filler.hpp
@@ -127,17 +127,18 @@ class PositiveUnitballFiller : public Filler<Dtype> {
 };
 
 /**
- * @brief Fills a Blob with values @f$ x \sim U(-a, +a) @f$ where @f$ a @f$
- *        is set inversely proportional to the number of incoming nodes.
+ * @brief Fills a Blob with values @f$ x \sim U(-a, +a) @f$ where @f$ a @f$ is
+ *        set inversely proportional to number of incoming nodes, outgoing
+ *        nodes, or their average.
  *
  * A Filler based on the paper [Bengio and Glorot 2010]: Understanding
- * the difficulty of training deep feedforward neuralnetworks, but does not
- * use the fan_out value.
+ * the difficulty of training deep feedforward neuralnetworks.
  *
- * It fills the incoming matrix by randomly sampling uniform data from
- * [-scale, scale] where scale = sqrt(3 / fan_in) where fan_in is the number
- * of input nodes. You should make sure the input blob has shape (num, a, b, c)
- * where a * b * c = fan_in.
+ * It fills the incoming matrix by randomly sampling uniform data from [-scale,
+ * scale] where scale = sqrt(3 / n) where n is the fan_in, fan_out, or their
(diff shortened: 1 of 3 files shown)


seanbell:
Note that #1970 should fix the `fan_in` and `fan_out` calculations for `InnerProductLayer` since the weights will now be 2D with shape `output x input`.

include/caffe/filler.hpp
```diff
@@ -127,17 +127,18 @@ class PositiveUnitballFiller : public Filler<Dtype> {
...
  *
- * It fills the incoming matrix by randomly sampling uniform data from
- * [-scale, scale] where scale = sqrt(3 / fan_in) where fan_in is the number
- * of input nodes. You should make sure the input blob has shape (num, a, b, c)
- * where a * b * c = fan_in.
+ * It fills the incoming matrix by randomly sampling uniform data from [-scale,
+ * scale] where scale = sqrt(3 / n) where n is the fan_in, fan_out, or their
+ * average, depending on the variance_norm option. You should make sure the
+ * input blob has shape (num, a, b, c) where a * b * c = fan_in and num * b * c
+ * = fan_out. Note that this is currently not the case for inner product layers.
```
shelhamer: #1970 is in so this filler is now right for InnerProduct layers too.

shelhamer:
The only issue I see is naming and attribution. The same `sqrt(2)` scheme has been suggested by Andrew Saxe et al. through derivations in http://arxiv.org/abs/1312.6120v3. Although I suggested "MSRA" earlier, I think a citation to both and a functional name is perhaps best.

@nickcarlevaris you suggested "ReLU" since this is intended for use with the so-named nonlinearity. It could be this is the right choice.

@longjon ?

futurely:
#1940 has been merged for a month. Can these two work together to reproduce the paper's results?

omgteam:
This issue has been open for a long time. Hope it merged quickly.

omgteam:
Why hasn't this been merged into master? anything wrong?

shelhamer:
Merged to master in c255709997b68ddbf56353f8fb27123e88cc51f3. Thanks @nickcarlevaris!

I did a manual merge to re-format the commit message and add my own commit to note potentially related work. Closing since my edit threw off the github merge.

happynear:
Why there is no parameter to specify the \alpha defined in Equation 15?
Since PReLU layer has been added to Caffe, I think we should also introduce this parameter into the filler.

PR #1948, opened 2015-02-24
APPLE was misspelled in Line 27

spmallick:


longjon:
@Nerei is this the logic you want here? Not sure why `-Wall` would be set conditionally, but maybe there is something I don't know. I'm going to go ahead and merge this since it's a clear typo fix.

@spmallick, thanks for the fix.

Nerei:
@longjon These are gcc/clang flags. You can't use them on Windows. But eventually I'll add windows support. That's why the condition.

Thank you all.

PR #1955, opened 2015-02-25
Decoding the datum before feeding it into the reshaping data layer

philkr:
Fixes a small bug with encoded data #1878 in the reshaping data layer #1313 . The transformed_data_ was not initialized with the decoded datum dimensions and instead used (1,0,0,0).
This PR decodes the data before reshaping.


shelhamer:
Thanks Philipp!

PR #1956, opened 2015-02-25
provide a solver option to limit runtime (max_seconds)

jsupancic:
This is an updated version of PR #1925 which should pass the lint checks . For some reason I cannot reopen a PR after force pushing, but can force push to an open PR? Strange behavior to keep in mind. Ref. https://github.com/isaacs/github/issues/361

This patch lets you say "run for 12 hours = 43200 seconds".
It's useful when you need exact limits with shared computing resources.
The runtime limit is enabled by setting max_seconds to a value > 0.
Maybe it helps?

Best,
James

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -1,6 +1,8 @@
 #ifndef CAFFE_OPTIMIZATION_SOLVER_HPP_
 #define CAFFE_OPTIMIZATION_SOLVER_HPP_
 
+#include <boost/date_time/posix_time/posix_time.hpp>
+
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


jsupancic:
Heads up, I accidentally targeted the master branch when I meant the dev branch. Sorry, I'm still getting the hand of github. :-/

Best,
James

shelhamer:
No no! master is right as of the last release. All development is now off
master. I'll make a further post to the mailing list to try and clarify
this.

Thanks for the cleaned up contribution.

longjon:
This seems sort of reasonable and convenient, but I'm not sure I see why this should be a feature of Caffe instead of something handled externally. Do the other devs think the clutter is worth the convenience?

(Also, what happens if you set `max_seconds` and then change your mind and want to keep running? If some external process is responsible for halting Caffe, you may still be able to do that. With this patch, too bad.)

(A related question: should Caffe attempt to snapshot on `SIGINT`?)

shelhamer:
On further thought I agree with @longjon. I too can see the convenience but this seems like a responsibility better left to the caller / cluster. Thanks for sending the code all the same @jsupancic so users who want it can merge it. Let's close this but comment if there is an overwhelming reason this should be part of the framework itself.

See #2012 for the snapshot on signal idea.

PR #1957, opened 2015-02-25
Fix matcaffe printout to specify correct # of arguments

forresti:
In a recent revision to Matcaffe, phase (train or test) was added to the of caffe('init', ...) arguments. Fixed the error printout to reflect this.


shelhamer:
Thanks Forrest! I cherry-picked this to master in 4a3887a since we've switched to a single branch workflow #1743.

PR #1960, opened 2015-02-25
Replaced illegal tab in Makefile with spaces.

gustavla:
Commands, such as `$(error ...)`, are not allowed to be indented with tabs
outside of targets, throwing an error instead of outputting the actual
error. The solution is to use innocuous spaces instead. Ideally, spaces
should be used everywhere outside targets, but since make does not mind
it if variable assignments are tab-indented outside targets, a complete
overhaul is not necessary. However, if more errors are added, it might
make more sense to be consistent.

You can easily see what happens by creating a simple Makefile:

    ifeq (,)
        $(error test)
    endif

Try tab-indenting the error, and you will see the error (at least for GNU make 3.81):

    Makefile:2: *** commands commence before first target.  Stop.

The problem is resolved if you switch to spaces.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,8 @@ ifneq (,$(findstring clang++,$(CXX)))
 else ifneq (,$(findstring g++,$(CXX)))
 	STATIC_LINK_COMMAND := -Wl,--whole-archive $(STATIC_NAME) -Wl,--no-whole-archive
 else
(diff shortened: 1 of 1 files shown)


longjon:
Oh gee, what a mess. You do seem to be correct, thanks for the fix.

PR #1961, opened 2015-02-25
Makefile fix for OS X 10.10

sergeyk:
Fixes the following make error on OS X:

```
In file included from ./include/caffe/util/math_functions.hpp:11:
./include/caffe/util/mkl_alternate.hpp:11:10: fatal error: 'cblas.h' file not found
#include <cblas.h>
```


shelhamer:
This works for installs with the standalone command line tools since Xcode isn't really needed. Thanks for the fix Sergey!

Issue #1943, opened 2015-02-23
master vs dev?


jyegerlehner:
I must have missed something. Docs used to instruct us to rebase PRs against dev, and now I see the docs say master. Is that right? What is the status of the dev branch? Deprecated? So any PR that's open would need to be updated before it can be merged, and is that simply a matter of rebasing off master? Thanks in advance if anyone can unconfuse me.

bhack:
If you have time you can rebase your PR branch on master and open a new PR. Dev branch is "de facto" deprecated but core members can still pickup your PR code and merge to master. There was a discussion on https://gitter.im/BVLC/caffe

shelhamer:
Please see the release notes for rc2 and the links to issues explaining the
workflow change. @bhack has summarized it: dev has been deprecated for
simplicity.

jyegerlehner:
OK thanks everyone.

PR #1965, opened 2015-02-25
Batch normalization layer with test and examples

ducha-aiki:
Implemented batch normalization layer (see http://arxiv.org/abs/1502.03167) based on @ChenglongChen and @Russell91 code with fixes and improvements. 
Also added shuffling pool by @jjkjkj of the input data to the data_layer to not to have same files together in same batch. Tests passes and rebased on master.
For illustration of the effectiveness two examples of CIFAR-10 classifier with sigmoid non-linearity with and without batch normalization.
bn_cifar


weiliu89:
Nice! I guess cifar_baseline is using ReLU instead of Sigmoid? Do you have any training examples of using ReLU + BN? Does it converge much faster than not using BN as stated in the paper?

ChenglongChen:
@ducha-aiki,
Nice work!
I myself have the problem to figure out how to implement per-epoch (or mini-batch) shuffling.

As for the fixed mean and variance for inference, I think we can hack that by using two extra vars to keep track of the (exponential) moving averaging mean or variance, and use those instead of the current batch_mean and batch_variance for normalization in TEST phase.

Besides, the current implementation keeps two "copies" of the blob (buffer_blob and x_norm) which might be a bit memory consuming when using big deep net. It might worth considering switching to for-loop rather than the current BLAS vectorization as @Russell91 did in his init commit.

weiliu89:
@ChenglongChen 

I used different "stage" at the beginning of TEST "phase" to compute mean and variance from a few training mini-batches. Moving average might work during training as mentioned in Section 3.1 in the paper.

ChenglongChen:
@weiliu89,
I haven't totally understood what Caffe's params referred to. I have to check what's the difference between stage and phase. But, it sounds like you are doing what Algorithm 2 in the paper describes (the inference part), right?

weiliu89:
@ChenglongChen 

Yes. I refer to how Caffe handle phase, and add several stages that bn_layer does different thing in different stage. This is the easiest way I can think of to implement Algorithm 2. It seems working well, but I haven't debug it though

sunbaigui:
@ducha-aiki  does master branch supports googlenet network now?
why not merge this into dev branch, and experiment on
image
Figure 4 in the paper.

ducha-aiki:
@weiliu89, we will add some graphs with ReLU-CIFAR later. For now, BN-model converges very fast, but a bit less accurate than non-BN. If we make a bit deeper model, than cifar-baseline, than BN converges faster and to more accurate one.

@sunbaigui, sure, if you sponsor us with some GPU. When I have trained my modification of the GoogLeNet, it took 3 weeks. Even if it woulb be 7 times faster, it is too much GPU time to spent for us :)
About master and dev - there is no dev anymore  (see #1943  )and master support GoogLeNet.

jjkjkj:
cifar_baseline is example/cifar/train_full.sh
@weiliu89 I could not archive cifar_baseline accuracy with batch normalization, without net modification( simply removing lrn, adding bn before or after every relu in examples/cifar10/cifar10_full_train_test.prototxt and only playing with learning rate)
I think this due to over-simplicity of this net and the fact than lrn for this net is enough normalization already.

I also trained variation of vgg16 on cifar with and without batch normalization
The only difference in architecture, compared to original vgg16, is shrieked down ip layers and adding dropout(not sure this is necessary for bn network)
Also I forgot to turn off PReLU and MSR weight filler in this networks, but I think that this makes little to no difference.

https://gist.github.com/jjkjkj/39e87099e9381e6886a5

For no_bn net  base_lr=0.001 causes net to diverge. For bn net lr is first guess, so maybe with bigger lr it will converge faster and better.

@ChenglongChen  @weiliu89  @ducha-aiki  About test phase. I tested cifar_vgg16 whith range of batch sizes (2-250) in test phase and found very small changes in accuracy (with batch size 2 accuracy is only 1% less than with 250)

ducha-aiki:
@shelhamer @longjon @jeffdonahue Could you please review this PR?

weiliu89:
I think that current PR doesn't compute mean and variance from training images (or moving mean and variance) during testing phase, but it compute mean and variance from test mini-batch, which I think is not exactly the same as described in the paper. I am not sure how much it affects the test accuracy.

yangyi02:
@jjkjkj For the cifar experiment, do you try the comparison adding bn before or after every relu? Does that matter?

jjkjkj:
@yangyi02 Yes i tried and found no difference(with examples/cifar10/cifar10_full_train_test.prototxt). As i said i think that this net is bad example for batch normalization.
As for cifar_vgg16, now i'm training net without dropout.

nakosung:
To feed evaluation network mean/var, setting mean = beta, var = (1/gamma)^2 will be OK? (Learned beta, gamma is similar to true mean/var?)

justfortest1:
@ducha-aiki  when batchsize=1 in test phase, can it also work?

ducha-aiki:
@justfortest1 no.

yangyi02:
@jjkjkj Any update on cifar experiments? We tried it and batch norm does not really show improvement. We are planning moving to larger dataset now.

lsy1993311:
@weiliu89  Could you please share you implementation? I think different stages should be considered.

weiliu89:
@lsy1993311 My Caffe version is old and I am not familiar with how to upload the code and haven't tested it. The high level idea is to include set_stage(string) and stage() in include/caffe/common.hpp (Refer to set_phase() and phase() in the same file). Then in src/caffe/solver.cpp, I add a function at the beginning of TestAll() which tries to compute mean & std. In the function, I set phase to TRAIN, and include two stages by using set_stage() as described before. The first stage is called "aggregation" which does several iteration of Forward pass to aggregate mean & std from a few mini-batches; the second stage is called "finalize" which compute the final mean & std by dividing the number of mini-batches you have passed. Finally, in batch_norm_layer, I can call Caffe::stage() and implement some additional thing in order to handle different stage (e.g. "aggregate" and "finalize"). I won't go in details into how to do it as it should be trivial.

However I don't have time to really debug this thoroughly. One thing to notice is that what I described above needs to compute mean & std every time I call TestAll() which might not necessary because it costs extra computation during training. On the other hand, you can only call the function in Snapshot() and use moving mean & std during training as described in the paper (can set a different stage for doing this during training).

melgor:
I have been testing that version with my data set using VGG16 model. And it works, speed up converge.
But it does not produce better result than normal version. Even more, batch size on test influence on accuracy.  
@weiliu89  propose resolving that problem. I will try to implement it. But this is only mismatch between Caffe code and Paper?

ducha-aiki:
@melgor, as far as I see - yes. It would be great, if you help to implement @weiliu89 solution.

melgor:
I have just found implementation of @ChenglongChen  which implement BN with right code in Test Phase. It save the mean and variance in BN layer. It looks like better implementation because it does not need to change Solver code. But it does not calculate mean and  variance within all Train data, only update the value of statistic using ex: S_{t+1} = decay * Y_{t+1} + (1 - decay) * S_{t}, where decay is parameter.

What do you think about such implementation? @ChenglongChen, does it work better than mini-batch statistic?

More information here:
https://github.com/ChenglongChen/caffe-windows/blob/master/src/caffe/layers/bn_layer.cpp

ChenglongChen:
Sorry guys. I have been caught up with work this moment, so I don't have time to test it out thoroughly. The use of exponentially weighted moving average (EWMA) is simply due to the fact that BN tends to keep the distribution of activation stable (?).

The algo2 in the paper is a bit complicate:
1. before the TEST phase, we forward a few mini-batch to compute the mean & var for the 1st BN layer, then we save this mean & var for other round inference (& forward)
2. we then forward those mini-batch to compute the mean & var for the 2nd BN layer, notice that the normalization part of the 1st BN layer is carried out using mean & var computed in step1 not the mini-batch statistics.
3. similarly, we perform the above for the rest BN layers.
4. after computing all the mean & var, we then have the inference BN network.

lsy1993311:
@jjkjkj  How is your result without using dropout?

jjkjkj:
@lsy1993311 As exepcted: slightly faster initial training but strong overfitting(it's natural whet parameters >> dataset). So, BN does not always remove need of dropout.

nakosung:
What if removing BN layers at testing phase? I mean that no normalization/reconstruction will be used during testing.

src/caffe/layers/bn_layer.cpp
```diff
@@ -0,0 +1,349 @@
...
+        spatial_mean_.cpu_data(),
+        spatial_sum_multiplier_.cpu_data(), Dtype(1), bottom_diff);
+
+    caffe_cpu_axpby(buffer_blob_.count(), Dtype(1),
+        buffer_blob_.cpu_data(), Dtype(-1. / (N_ * H_ * W_)),
+        bottom_diff);
+
+    // put the squares of bottom into buffer_blob_
+    caffe_powx(buffer_blob_.count(), bottom_data, Dtype(2),
+        buffer_blob_.mutable_cpu_data());
```
weiliu89: line 326 - 327 is redundant and can be deleted. So is the same for the GPU code.

borisgin:
Hi Dmitro, nice work! 
When I tryied to train cifar_bn, I got endless warnings "force_color: ...". Looks like a minor bug,  So I just commented these lines in data_layer. cpp.
Another observation: when I put bn layer just before convolutional & Ip layer, I obtained faster convergence then when bn_layer is located after convolutional layer:
cifar_fast_bn

ducha-aiki:
@weiliu89, thanks for catch!

Hi Boris, thanks for for observation. It is interesting, that is it very architecture dependent: we have tried on other architectures and there was no difference, as stated in original paper. Still lot of place for exploring :) 
The force_color bug is fixed now, thanks.

wafflette:
weiliu89>>I think that current PR doesn't compute mean and variance from training images (or moving mean and variance) during testing phase, but it compute mean and variance from test mini-batch, which I think is not exactly the same as described in the paper. I am not sure how much it affects the test accuracy.

FWIW I agree, that's different from the paper description. Also what if the test data contains only one sample (single image inference)?

anxiaoxi45:
bug ??   occurs "nan"

anxiaoxi45:
bug??  occurs   "nan" .........

AIROBOTAI:
Hi @ducha-aiki, and others, thanks for your excellent work!  Here I have a question about your code. Could you please explain to me what is the purpose for the codes starting from the line No. 153 to line NO. 185 in the function "void DataLayer<Dtype>::InternalThreadEntry()" in data_layer.cpp? I just cannot figure out why these codes should be there when the "datum.encoded()" is true. Thx a lot in advance!

ducha-aiki:
@AIROBOTAI it needs for data shuffling. With batch normalization, it is important that network don`t see same images together in batch, so this lines implement shuffling.
As for the encoded datum - it is only mode we use, so we implemented shuffling only for it.

AIROBOTAI:
@ducha-aiki Thanks for your prompt reply! But what should I do if my datum is NOT encoded? I have checked the return value of datum.encoded() to find it to be false. So in this case, those lines of code for shuffling will be jumped over.

ducha-aiki:
@AIROBOTAI then you can regenerate LMDB with encoded key, or add same lines to unencoded branch of if :)

waldol1:
One alternative to true shuffling is to do random skips.  The DataLayer has
a parameter called rand_skip that causes the DB cursor to start in a random
position in the DB.  It is easy to extend that concept so that the DB skips
a random number of instances every time it advances its cursor.  You can do
so by modifying the last part of the InternalThreadEntry() to run the
cursor->next() in a loop for a random number of iterations.  This way the
mini-batch is never composed of the same instances and the probability of
two instances being in the same batch is inversely proportional to the size
of the skip.  I find in practice that this doesn't slow down DB reads very
much because it is still kind of sequential.

Hope that helps.

AIROBOTAI:
@ducha-aiki I have modified some seemingly confusing codes in your code to make it more straight to me. Now it works, thanks again!

AIROBOTAI:
@waldol1 thanks for your suggestion! Your method seems more easy to use than the shuffling pool proposed in this pull request. I'd also like to know whether you have tested the test accuracy using your method and how is the performance. @ducha-aiki what's your comments on this new shuffling method? Thanks for your all!

waldol1:
I haven't tested this shuffling method with regards to BatchNorm, but it
seemed to help a little on an mnist autoencoder that I was testing when the
training set was small.

talda:
@ducha-aiki @shelhamer - is there still plan to pull this in?
If so two suggestions:
1. separate the data shuffling into another PR, also find a solution that works for all types of data layers.
2. keep a running average of batch statistics so this PR could be used for classifying a single image.

bhack:
@talda This is still to young to be reviewed. It is only six months old. :)

ducha-aiki:
@bhack actually, this PR is not needed at all, if believe to Google. 
Here are slides http://lsun.cs.princeton.edu/slides/Christian.pdf 
Pre-last one says:
"Releasing Pretrained Inception and MultiBox
Academic criticism: Results are hard to reproduce
We will be releasing pretrained Caffe models for:
● GoogLeNet (Inception 5)
● BN-Inception (Inception 6)
● MultiBox-Inception proposal generator (based on
Inception 6)
Contact @Yangqing "

talda:
@bhack too bad Github does not have like/upvote button for comments. I would defiantly upvote your previous comment.

erogol:
Can I classify a single image by using that PR's modifications and batch normalization ?

ducha-aiki:
@erogol No.

erogol:
@ducha-aiki I applied moving average and it now works.

ducha-aiki:
Closed, because of #3229 merge.The scale + bias could be taken from #2996

PR #1966, opened 2015-02-25
cmake and python3 bugfixes for #1939 and #1923

philkr:
#1939 replaced PROJECT_SOURCE_DIR with CMAKE_SOURCE_DIR, which breaks building caffe as a subproject (see #1899). This PR reverts to PROJECT_SOURCE_DIR.

Python3 requires ```list``` around ```self.blobs.keys()```.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -4,7 +4,7 @@ if(NOT HAVE_PYTHON)
 endif()
 
 include_directories(${PYTHON_INCLUDE_DIRS} ${NUMPY_INCLUDE_DIR} ${Boost_INCLUDE_DIRS})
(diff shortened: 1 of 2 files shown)


longjon:
Looks good to me. @Nerei, this reverts a change made by you; do you agree?

Nerei:
Ok change. Never knew that someone includes caffe in this way.

andresromero:
I have tried to use Caffe with Python 3 and I am still finding some bugs related to Python 3 `protobuf` package, what version are you using?

longjon:
Okay, merging. @andresromero, I believe you need protobuf 3.0.0-alpha.

PR #1969, opened 2015-02-25
Fix incorrectly storing empty param_name to param_names_index_

tnarihi:
Shouldn't it be `param_name.size()` if you intend to check if `param.name()` is specified for this parameter blob? It does not work if `ParamSpec` is given only for one of parameter blobs in a layer while the layer has multiple parameter blobs, e.g. Convolution, InnerProduct etc..


tnarihi:
I find that it does not affect actual behavior of weight sharing. Even if empty param_name is added to `param_names_index_`, it is never referenced.

jeffdonahue:
Thanks for the fix @tnarihi!

PR #1913, opened 2015-02-20
Matcaffe2

shelhamer:
This PR resurrects #501. The good work started by @sguada will be finished by @bharath272 and @s-gupta who have a lot of MATLAB and Caffe experience.

PR #1970, opened 2015-02-25
Blobs are N-D arrays (for N not necessarily equals 4)

jeffdonahue:
Replaces #1486 -- this PR is to master instead of dev.  This is rebased and is ready for review @longjon.

> This PR gives Blobs a `vector<int> shape_` of dimensions, rather than the old num, channels, height, width. The first commit is all the changes needed to get Caffe to compile and everything to run as before with the new vector of tensor dimensions. The remaining commits generalize some existing classes to use the new tensor dimensions (but they are not necessary to make it run, as it's still fine to just use all 4-D blobs with extra singleton dimensions where needed).


jeffdonahue:
I had added a commit intended to add matcaffe support for N-D array blobs, but decided to defer to #1913 to add the support, as I don't have much experience with matcaffe and it might require some special cases (e.g., I remembered that MATLAB doesn't support arrays with <2 axes so would need to figure out what to do with those cases -- maybe the MATLAB default of adding extra axes with dimension 1 would work fine, but not sure).   But if that commit might be helpful in the development of #1913, feel free to cherry-pick or just refer to it [here](https://github.com/jeffdonahue/caffe/commit/af72f0ccc6da7cb459534eb83d43c87361c17581).

So for now, matcaffe will continue to work fine for blobs with <= 4 dimensions, but will die on attempts to access blobs with >4 dimensions (per their use of `num`/`channels`/`height`/`width`, which call `LegacyShape`).

src/caffe/proto/caffe.proto
```diff
@@ -2,13 +2,21 @@ syntax = "proto2";
 
 package caffe;
 
+// Specifies the shape (dimensions) of a Blob.
+message BlobShape {
+  repeated int64 dim = 1 [packed = true];
```
Yangqing: My 2 cents - uint32 should be enough? A 2G blob (8G in floats) looks unlikely.

jeffdonahue: My feeling is that it's probably worth future-proofing given the relatively tiny amount of extra storage cost (unless you're really using blobs with `INT_MAX` axes, as are now allowed...).  I could make it `uint32` for now and then later change it to `int64` if/when we really find practical uses for blobs that large, but that might cause backward compatibility problems with (de)serialization that I don't even want to think about...  (Why not `uint64`? I think it's worth keeping -1 and other <0 values reserved for various purposes -- e.g. setting `axis [default = -1]` to signify "all axes", or that an axis must be manually specified and should fail otherwise, etc. And it *does* seem pretty unlikely that in the reasonably foreseeable future we'll need 2^64 byte blobs, so 2^63-1 bytes seems like an okay max...)

I'm open to discussion on any of this though.

Yangqing: Future-proof sounds better :) Thanks for checking with me!

shelhamer: But Jeff, there is no reason for any individual to have a > 2GB blob in their home. Haha no, I'm all for `int64` dimensions.

longjon:
Oops, just realized that I've been commenting on commits again. Agree with `int64` blob dimensions. I'm ready to merge this today, pending comments made, the new type for Python `Blob.reshape` (which I'll do), and a nod from @shelhamer.

longjon:
Added my Python update... this should be ready pending your approval of that, Travis, and @shelhamer.

jeffdonahue:
Thanks Jon! Your updated `Blob_Reshape` looks good to me.

shelhamer:
The `axis` field in the proto is inconsistently `(u)int32` -- should it be `int32` everywhere to allow positive and negative indexing in Slice, Concat, and InnerProduct dimension picking?

src/caffe/layers/cudnn_softmax_layer.cpp
```diff
@@ -26,10 +26,10 @@ template <typename Dtype>
 void CuDNNSoftmaxLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
   SoftmaxLayer<Dtype>::Reshape(bottom, top);
-  int N = bottom[0]->num();
```
shelhamer: I don't think this should be changed -- in our usage the cuDNN softmax is run in "channel mode" so that the softmax is done independently for each (N, H, W). This will break models expecting this behavior for softmax over spatial dimensions.

We should perhaps expose the choice of "instance mode" too.

src/caffe/layers/inner_product_layer.cpp
```diff
@@ -15,7 +15,11 @@ void InnerProductLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
   const int num_output = this->layer_param_.inner_product_param().num_output();
   bias_term_ = this->layer_param_.inner_product_param().bias_term();
   N_ = num_output;
-  K_ = bottom[0]->count() / bottom[0]->num();
+  const int dim = this->layer_param_.inner_product_param().axis();
+  // Dimensions starting from "axis" are "flattened" into a single
+  // length K_ vector. For example, if bottom[0]'s shape is (N, C, H, W),
```
shelhamer: For example, if bottom[0]'s shape is (N, C, H, W) **and axis = 1**

shelhamer:
@jeffdonahue this is sweet. My only comments were minor, so do what you want with them and merge!

src/caffe/layers/softmax_loss_layer.cpp
```diff
@@ -35,6 +35,9 @@ void SoftmaxWithLossLayer<Dtype>::Reshape(
     const vector<Blob<Dtype>*>& bottom, const vector<Blob<Dtype>*>& top) {
   LossLayer<Dtype>::Reshape(bottom, top);
   softmax_layer_->Reshape(softmax_bottom_vec_, softmax_top_vec_);
+  softmax_axis_ = this->layer_param_.softmax_param().axis();
+  outer_num_ = bottom[0]->count(0, softmax_axis_);
+  inner_num_ = bottom[0]->count(softmax_axis_ + 1);
```
shelhamer: This could be a good time to add a check that the prediction and target dimensions (except channel) agree?

jeffdonahue: done in 8b6e912

jeffdonahue:
Hey Evan, thanks for the review!  I made the additional commits above in response to your comments -- currently marked with `fixup!` to be autosquashed later.  I thought you might want to take a look since, re negative indexing, I applied it in SliceLayer, ConcatLayer, SoftmaxLayer, and InnerProductLayer, and ended up adding an additional `Blob` method `CanonicalAxisIndex` (as I started to realize I was writing the same checks several times).  I added a single test that actually uses it to `TestConcatLayer`.

jeffdonahue:
This should be ready to go -- I'll merge as soon as @shelhamer or @longjon quickly OKs the small fixups I made last night in response to @shelhamer's suggestions (can see just those changes at https://github.com/jeffdonahue/caffe/compare/eca584d...jeffdonahue:tensor-blob-presquash).

bhack:
This will probably cross impact every new layer in the PRs queue. How many new layers we have in the "waiting list"? The merging order is not neutral because will impact the adaptation charge to developer of one PR instead of another.

shelhamer:
Thanks for the fixups @jeffdonahue -- this all looks good to me. This PR can be a guide to converting in-progress and future layers to N != 4. I'm happy to see brewing transcend to higher dimensions.

Next up: N-D convolution and pooling!

ddetone:
@shelhamer I am currently working on my own 3-D convolution and pooling implementation. Do you have a branch that this is currently being developed on that I could take a look at?

shelhamer:
@jeffdonahue can you align https://github.com/BVLC/caffe/blob/master/docs/tutorial/net_layer_blob.md with the transcendence to N-D blob sometime?

PR #1971, opened 2015-02-25
Use regular expressions to parse image data text files.

### LINKED ISSUE #1951: bug in convert_imageset for file names with spaces
When creating a leveldb using a list of files, convert_imageset file names with a spaces in it. One solution is to treat everything before the first .jpg or .JPEG extension to be regarded as part of the file name.
Another option is to enable wrapping the filename in "".

erictzeng:
Fixes #1951.

This pull request consists of two changes:

1. Rather than the brittle `ifstream` method of parsing image data files, this pull request uses regular expressions for more robust matching.
2. Previously, the parsing code was duplicated across two files, `tools/convert_imageset.cpp` and `src/caffe/layers/image_data_layer.cpp`. This pull request pulls that common code out into a new function in `src/caffe/util/io.cpp` for ease of maintenance.

More details follow.

Each line of the input text file is matched against the following regular expression:

````
\h*("?)(.+?)\1\h+(\d+)\h*
````

Feel free to play around with [an interactive version](https://regex101.com/r/xK3hN7/3) so you can test it out and see what it matches. This regular expression handles a lot of cases that would've been difficult to handle using the previous naive approach. It captures whitespace within a filename, and enables quoting of filenames in case for some insane reason you have a space at the beginning of a file name.

Some concrete examples of really degenerate cases that will parse correctly:

````
file name with spaces.jpg 1
" file_name_with_leading_space.jpg" 2
file_name_with_"_symbol.jpg 3
" really disgusting " file  ""name  .jpg" 4
```

One drawback is that this introduces `boost_regex` as an additional dependency. However, since we already require Boost, this seems like an acceptable tradeoff.

Implementation-wise, this pull request should be complete, though it's lacking tests, which I will get around to writing at some point in the near future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,7 +162,7 @@ ifneq ($(CPU_ONLY), 1)
 	LIBRARIES := cudart cublas curand
 endif
 LIBRARIES += glog gflags protobuf leveldb snappy \
-	lmdb boost_system hdf5_hl hdf5 m \
+	lmdb boost_regex boost_system hdf5_hl hdf5 m \
 	opencv_core opencv_highgui opencv_imgproc
 PYTHON_LIBRARIES := boost_python python2.7
 WARNINGS := -Wall -Wno-sign-compare
--- a/include/caffe/util/io.hpp
+++ b/include/caffe/util/io.hpp
@@ -3,6 +3,8 @@
 
 #include <unistd.h>
 #include <string>
+#include <utility>
+#include <vector>
 
 #include "google/protobuf/message.h"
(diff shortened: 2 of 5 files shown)


shelhamer:
@erictzeng this looks right -- thanks for fixing the brittle format -- but I think you need to update the travis script to install boost regex: https://github.com/BVLC/caffe/blob/master/scripts/travis/travis_install.sh.

bchu:
Any updates on this?

PR #1974, opened 2015-02-25
Give layers a pointer to their owning Net

shelhamer:
`master` edition of @longjon's #1638:

> This allows layers to do essentially arbitrary things that depend on the structure of the net.

> This provides a lot of power, but breaks the nice abstraction of a layer as a function-with-gradient. A future PR will depend on this functionality. If it seems sufficiently useful, perhaps we should merge this, but be wary not to abuse it. Or perhaps we should find another way to compute on the structure of nets. Or perhaps we should do the former keeping the latter as a future possibility.

### DIFF
--- a/include/caffe/layer.hpp
+++ b/include/caffe/layer.hpp
@@ -13,6 +13,8 @@
 
 namespace caffe {
 
+template <typename Dtype> class Net;
+
 /**
  * @brief An interface for the units of computation which can be composed into a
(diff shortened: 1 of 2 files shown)


shelhamer:
Unnecessary given the separation of responsibilities by the new crop layer #3570 and python net spec coordinates and cropping #3613.

Issue #1938, opened 2015-02-22
I think MVNLayer with across_channels = true is broken


jyegerlehner:
I think MVNLayer is broken when `across_channels: true` is specified in MVNParameter, even though the unit test is passing.

In `MVNLayerTest::TestForwardAcrossChannels`, we create the layer parameter with:

`layer_param.ParseFromString("mvn_param{across_channels: true}");` ParseFromString is returning false and `layer_param.mvn_param().across_channels()` is false, after the call, even though we specified `true` in the prototxt. So the test is running with default value of across_channels, which is `false`, the  default value. 

If instead in `MVNLayerTest::TestForwardAcrossChannels` we call:
`CHECK(google::protobuf::TextFormat::ParseFromString("mvn_param{across_channels: true}", &layer_param ));`

it succeeds, and `layer_param.mvn_param().across_channels()` is true. And then you will see that `MVNLayerTest::TestForwardAcrossChannels` fails, as the Forward_* produces NaN values.

I have no idea why `layer_param.ParseFromString` fails whereas `google::protobuf::TextFormat::ParseFromString' suceeds. I don't have protocol source installed and would rather debug Caffe than Protobuffers. If anyone knows please educate me.

It at first seems surprising that  test `TestForwardAcrossChannels` succeeds even though across_channels = false when you'd think it would need to be true in order to pass the test. I think it's because the values in bottom blob are drawn from a normal distribution using GaussianFiller<Dtype>, and the only difference between across_channels=true vs. false, is that there are 20 values from which one is computing the mean and variance instead of 60. So you still get a zero mean and variance 1.0 in the top blob which is what the test checks for, regardless of whether you compute them for a single channel or across channels.

I'm working on this in connection with [1895](https://github.com/BVLC/caffe/pull/1895). I will say trying to understand the vectorization approach using BLAS calls used to perform these ops is making my head hurt : /.

jyegerlehner:
I think [1895](https://github.com/BVLC/caffe/pull/1895) now fixes this.

seanbell:
Just to explain why `layer_param.ParseFromString` is failing -- `layer_param.ParseFromString` parses serialized binary data stored as a string, and `google::protobuf::TextFormat::ParseFromString` parses text-format data (also stored as a string).  It's the wrong function to use.

The same `ParseFromString` issue was raised in #1355, but it seems that the problem is still not fixed.

jyegerlehner:
[...]

PR #1979, opened 2015-02-26
InverseMVNLayer

jyegerlehner:
Replaces [1895](https://github.com/BVLC/caffe/pull/1895).

This PR extends the MVNLayer to allow the mean and variance blobs to be exported as top blobs. It adds a new layer type InverseMVNLayer which takes the mean and variance as bottom blobs, and performs the inverse operation (adding the mean back, and denormalizing for variance). A use case for this is an autoencoder that feeds input into the MVNLayer, and generates the output from the InverseMVNLayer, with autoencoding layers in-between.

There was also a problem with the MVNLayer that was fixed: if it is given input that has exactly zero variance (e.g. a solid color RGB image, with across_channels=false), it computes the variance as E(X^2) - (EX)^2, but the result isn't usually exactly zero, but has small negative values due to floating point resolution. The subsequent square root operation then produces NaN. This PR also fixes this issue.

### DIFF
--- a/docs/tutorial/layers.md
+++ b/docs/tutorial/layers.md
@@ -485,4 +485,78 @@ The `SLICE` layer is a utility layer that slices an input layer to multiple outp
 
 #### Mean-Variance Normalization
 
-`MVN`
+* LayerType: `MVN`
+* CPU implementation: `./src/caffe/layers/mvn_layer.cpp`
+* CUDA GPU implementation: `./src/caffe/layers/mvn_layer.cu`
+* Parameters (`MvnParameter mvn_param`)
+    - Optional
+        - `normalize_variance` [default true]: false to only subtract the mean. true will make the variance of each output instance 1.
+        - `across_channels` [default false]: true will perform normalization by computing a single mean and variance across all the channels. Setting to false will compute a separate mean and variance for each channel.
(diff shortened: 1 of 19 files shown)


seanbell:
See https://github.com/BVLC/caffe/issues/1938#issuecomment-76260921 regarding the case of mean subtraction

jyegerlehner:
Could our Brewers give some indication if they would prefer the MVNLayer fixes to be separated out of this PR, as cdoersch [suggests here that they should](https://github.com/BVLC/caffe/issues/2017)? It makes abundant sense that reviewing bug fixes for an existing feature is a higher priority than reviewing code for a proposed new feature. Especially since the MVNLayer fixes are a smallish subset of the the changes in this PR.  I can do that, but it's a fair bit of work and don't want to expend it if it is no more likely to get review attention. Thanks.

seanbell:
@shelhamer Are there any plans to merge this?  If not, at least the fix to the MVNLayer gradient should be cherry-picked in.  The gradient is currently incorrect (and it's not detected by the gradient checker because the test is also incorrect).

jyegerlehner:
MVNLayer fixes put in [2964](https://github.com/BVLC/caffe/pull/2964), InverseMVNLayer doesn't seem interesting to others.

Issue #1834, opened 2015-02-05
different crop_size for width and height?


graphitash:
When defining an image transformation in a protobuf file. is there a way to set a rectangular crop size, i.e. a separate size for width and height? I can only see how to set a single crop size for both width and height as in AlexNet.

lukeyeager:
Sadly, no. Caffe only supports square crops in the [protobuf definition](https://github.com/BVLC/caffe/blob/e3c895baac64c7c2da4f30ab7ae35a1c84e63733/src/caffe/proto/caffe.proto#L351) at the moment.

ChenglongChen:
To achieve that, you have to modify the data layer and replace crop_size with crop_height and crop_width as discussed in: https://github.com/BVLC/caffe/issues/284
But I have no luck with it so far :(

nmerrill67:
I have a working crop_width and crop_height in [my fork](https://github.com/nmerrill67/caffe). However, if you have to specify crop_width and crop_height even if they are equal in my patch.

magomezs:
I have a python layer to make rectangular cropping in caffe
Please see: https://github.com/magomezs/data_augmentation_layers

PR #1980, opened 2015-02-26
Non-square cropping

okuta:
Add `crop_h` and `crop_w` to `TransformationParameter`.
This PR is implemented similar to h/w kernel size.
Related issue: #1834

### DIFF
--- a/src/caffe/data_transformer.cpp
+++ b/src/caffe/data_transformer.cpp
@@ -14,6 +14,11 @@ template<typename Dtype>
 DataTransformer<Dtype>::DataTransformer(const TransformationParameter& param,
     Phase phase)
     : param_(param), phase_(phase) {
+  CHECK(param_.crop_size() == 0 ||
+        param_.crop_h() == 0 && param_.crop_w() == 0)
+        << "Crop size is crop_size OR crop_h and crop_w; not both";
+  CHECK((param_.crop_h() != 0) == (param_.crop_w() != 0))
+        << "For non-square crops both crop_h and crop_w are required.";
   // check if we want to use mean_file
   if (param_.has_mean_file()) {
     CHECK_EQ(param_.mean_value_size(), 0) <<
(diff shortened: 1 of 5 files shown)


seanbell:
This is a great addition.

Small suggestion: I think it's a good idea to add in some checks with helpful messages like:
```
CHECK(!param_.has_crop_size() !=
      !(param_.has_crop_h() && param_.has_crop_w()))
      << "Crop size is crop_size OR crop_h and crop_w; not both";
CHECK(param_.has_crop_size() ||
      (param_.has_crop_h() && param_.has_crop_w()))
      << "For non-square crops both crop_h and crop_w are required.";
```
(this was adapted from checks done in the convolution layer https://github.com/BVLC/caffe/blob/master/src/caffe/layers/base_conv_layer.cpp#L16 )

okuta:
@seanbell Tahnk you for your suggestion.
`crop_size`, `crop_h` and `crop_w` have default value.
https://github.com/okuta/caffe/blob/4d445f420e7e6fe6426ffbc62e6c212f5b9fc6d8/src/caffe/proto/caffe.proto#L333-335
Therefore I wrote a check code that does not use methods such as `has_crop_size()`.

haoyangz:
This is very helpful for Deep Learning work on non-CV fields . Any chance that it could be merged soon?

liyongsea:
I am working with 1d time series data and a 1d crop could be very useful. I am wondering if this change would be merged?

src/caffe/data_transformer.cpp
```diff
@@ -243,18 +262,18 @@ void DataTransformer<Dtype>::Transform(const cv::Mat& cv_img,
...
+      h_off = Rand(img_height - crop_h + 1);
+      w_off = Rand(img_width - crop_w + 1);
     } else {
-      h_off = (img_height - crop_size) / 2;
-      w_off = (img_width - crop_size) / 2;
+      h_off = (img_height - crop_h) / 2;
+      w_off = (img_width - crop_w) / 2;
     }
-    cv::Rect roi(w_off, h_off, crop_size, crop_size);
+    cv::Rect roi(w_off, h_off, crop_h, crop_w);
```
bobbens: I think the crop_h and crop_w are swapped.

lukeyeager:
:+1:

elezar:
Can this non-square crop be implemented using the ND crop layer as in #3570?

shelhamer:
Closing as better handled by a Python data layer. Thanks for extending crop, but I think including all these varieties of augmentation adds too much code which is more clearly expressed in Python anyway.

PR #1984, opened 2015-02-26
Adapt python net drawing to new proto format

axel-angel:
The current version of draw_net.py that's using draw.py plots an empty figure on the LeNet prototxt. I've heard the format of this file changed/is changing and thus I made the adaptation that works.

Can you review this?

### DIFF
--- a/python/caffe/draw.py
+++ b/python/caffe/draw.py
@@ -92,11 +92,11 @@ def choose_color_by_layertype(layertype):
     """Define colors for nodes based on the layer type
     """
     color = '#6495ED'  # Default
-    if layertype == 'Convolution':
+    if layertype == 'CONVOLUTION':
         color = '#FF5050'
-    elif layertype == 'Pooling':
+    elif layertype == 'POOLING':
         color = '#FF9900'
-    elif layertype == 'InnerProduct':
+    elif layertype == 'INNER_PRODUCT':
         color = '#CC33FF'
     return color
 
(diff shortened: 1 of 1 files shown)


shelhamer:
@axel-angel thanks for checking this but this change is actually in reverse. The current code is right as layer type has changed to a string #1694. The examples should all be in this new format. @jeffdonahue is there old proto somewhere?

jeffdonahue:
No, there isn't anything left in the old format that I know of, but let me know if I'm mistaken.  These changes seem to reverse @tnarihi's fixes in #1884, so I'll close this.

axel-angel:
Okay then, that would explain why I have a deprecation warning*. I need to adapt my prototxt to the new format then, I'll look into it. I had the case where "layers" were not recognized and thus I switched to "layer" and such, and that's probably due to my repository not updated to last version (seems the modification went the 5th February). Thank you for the clarification!

* E0226 11:58:30.210986 10845 upgrade_proto.cpp:618] Attempting to upgrade input file specified using deprecated V1LayerParameter: snapshots/lenet_mnist_v2_iter_10000.caffemodel
I0226 11:58:30.211860 10845 upgrade_proto.cpp:626] Successfully upgraded file specified using deprecated V1LayerParameter

tnarihi:
The problem is the protobuf itself does not support automatically updating old format. Should we add something to prohibit to draw old format protofile? I think no message is not good. Otherwise, merge layrers and layer after reading net proto?

PR #1985, opened 2015-02-26
Less memory: Blob sharing across nets  by name

tnarihi:
This PR implements blob sharing feature which can reduce memory consumption in solver. In the previous implementation, blobs storing intermediate results are independently used among different nets, which usually causes redundant memory consumption for blob data. In this PR, Nets in solver, i.e. `net`, `test_nets`, share the blobs by name to save memory. However, in the case that the shapes of blobs over nets have different shapes, e.g. different batch size, it will break. Layers except data layers is okay since there are reshaped before every `Forward` call. I made some changes that data layers always reshape their top blobs in `Forward`, and it will changes the default behavior of `DummyDataLayer`. Another issue is that it will break if the nets are used simultaneously, e.g. multi-threads, but I believe nobody does like that so far.
I have only thought about standard classification net and segmentation net. I am not sure if it will work for every possible network. And I am also not sure whether this is best solution (another idea is to use global `Blob` pool or memory pool) but definitely helpful for anyone who concerns memory usage. In my experiment, It reduces the memory usage 1GB-ish in VGG16 training (batch_size=16). Again, I summarize this PR here:

* Share blobs with different nets (by setting share_blobs=true in `SolverParameter`. default=false)
* Data layers always reshapes top blobs
* Change the default behavior of `DummyDataLayer` (refill_constant=false will keep the behavior but might breaks share_blobs=true)


src/caffe/net.cpp
```diff
@@ -650,6 +650,61 @@ void Net<Dtype>::ShareTrainedLayersWith(const Net* other) {
...
+        net_input_blob_indices_.begin(),
+        net_input_blob_indices_.end(), target_blob_id) !=
+      net_input_blob_indices_.end()) {
+      LOG(INFO) << "Ignoring source blob " << source_blob_name;
+      continue;
+    }
+    LOG(INFO) << "Sharing source blob " << source_blob_name;
+    // We do not care blob shapes because every call of `Forward` method
+    // reshapes output blobs into right shapes, except input blobs.
+    blobs_[target_blob_id] = other->blobs()[i];
```
jyegerlehner: Is this better than using Blob<Dtype>::ShareData() (and possibly ShareDiff)? E.g. 
```c++
blobs_[target_blob_id]->ShareData( *other->blobs()[i]));
```
Then you don't need UpdateBlobPointers because each net just keeps its own. I think this is more robust; ShareData() doesn't require that blob pointers be reassigned as in  UpdateBlobPointers. If you assign blob pointers instead of using ShareData, UpdateBlobPointers has to know all the places that blob pointers reside and keep them in sync. That introduces tight coupling to the rest of the implementation which is not robust but rather prone to breaking. 
Actually, if you do ShareData(),  I think you have to reshape before sharing because ShareData checks that the shape is the same.

src/caffe/net.cpp
```diff
@@ -650,6 +650,61 @@ void Net<Dtype>::ShareTrainedLayersWith(const Net* other) {
 }
 
 template <typename Dtype>
+void Net<Dtype>::ShareBlobsWith(const Net* other) {
+  int num_source_blobs = other->blobs().size();
```
jyegerlehner: OK here's a naive question, since  I'm not sure I understand everything about this change. But the Net::blobs() method just returns the blobs that are outputs of layers. What about the blobs that are the layer's learnt parameters i.e. weights? Do we not want to share those too, so as not to redundantly have to store them as well?
In fact, now that I ask this question, I realize I don't know why there needs to be a test net that's different from the trained net. Why can't we just generate test scores using the current state of the trained net.

src/caffe/net.cpp
tnarihi: Yeah, first I tried to do like that. However, I noticed that `ShareData` actually do sharing `SyncedMemory` instance (shared_ptr), and if you reshape the blob after sharing it and the allocated capacity is not large enough as you need for reshaping, it will create a new synced memory instance. Then, the coupling will be gone. This will happen in the special case that an input layer produces different shape blobs for each data. That is why I decide to share blobs directly. But, yes, your point is right. Using `UpdateBlobPointers` makes it a little complicated. If you have any better idea to implement this, please help me. Thanks!

src/caffe/net.cpp
tnarihi: Parameter blobs are shared by `ShareTrainedLayerWith` method. In the implementation of Caffe, `Solver` class calls it before testing the trained net in `Test()` method.

As far as I am understanding, the test nets could be partially different or even totally different from train net. At least, in almost cases, `DataLayer`s are different from train net.

mtamburrano:
any news about this pr?

PR #1987, opened 2015-02-26
Fix siamese ipynb example

tnarihi:
Just fix. It used old interface. http://nbviewer.ipython.org/github/BVLC/caffe/blob/101055c840ef3a713341735c97ca49b2a8de253e/examples/siamese/mnist_siamese.ipynb


shelhamer:
Thanks Takuya!

PR #1999, opened 2015-02-27
Small fix (visualization) on SLICE layer's documentation

boechat107:
The sample was missing some additional spaces to be correctly rendered on the HTML. The mistake was mine.


longjon:
Seems consistent with the other samples now. Thanks!

PR #2001, opened 2015-02-28
[Python] PythonLayer takes parameters by string and better exception handling

tnarihi:
This PR changes two things:

* Exceptions are handled correctly. Please see commit messages for details.
* `PythonLayer` can take parameters by string format through protobuf. You can see a example of how it can be used at `python/caffe/test/test_python_layer_with_param_str.py` that can be run `make pytest`.

I only tested this on my Ubuntu 14.04. I think it should work on different platform if you install boost and python correctly.

\# Please make sure to uncomment `WITH_PYTHON_LAYER=1` in Makefile.config before trying `PythonLayer`.


src/caffe/layer_factory.cpp
```diff
@@ -160,14 +160,18 @@ REGISTER_LAYER_CREATOR(TanH, GetTanHLayer);
...
 shared_ptr<Layer<Dtype> > GetPythonLayer(const LayerParameter& param) {
+  string module_name = param.python_param().module();
+  string layer_name = param.python_param().layer();
   Py_Initialize();
   try {
-    bp::object module = bp::import(param.python_param().module().c_str());
-    bp::object layer = module.attr(param.python_param().layer().c_str())(param);
+    bp::object globals = bp::import("__main__").attr("__dict__");
+    bp::exec(("import " + module_name).c_str(), globals, globals);
+    bp::object layer_class = bp::eval((module_name + "." + layer_name).c_str(),
```
seanbell: I could be mistaken (and/or paranoid), but isn't it dangerous to call `eval` on a string that hasn't been checked?  That is, we should check that `module_name` and `layer_name` both match the regex `^[a-zA-Z_][a-zA-Z0-9_]*$.

jeffdonahue: Agreed with @seanbell that if we're going to `eval`/`exec`it's worth doing a check like the one he suggests.  Is it necessary to switch away from `bp::import` though?  I'm not sure why it's needed, is it to load `__dict__` first? Could `param_str` just be a string, and then Python layer writers could `eval ` it as a dict on their own if they want it to take a dict?

seanbell: @tnarihi explains the use of `eval`/`exec` in one of the commit messages ( https://github.com/tnarihi/caffe/commit/64a2b049d8aac624188159e8f273c99107dce99d )

jeffdonahue: Ah, thanks, missed that.

tnarihi: Thanks for commenting. I also agree with @seanbell . To avoid injections would be nice. I will modify it soon using boost regex. Thanks for giving me the regex pattern, @seanbell ! I will use this.

tnarihi:
I just pushed the change to avoid injecting. I didn't add any test case but I checked it works manually.

By the way, I am wondering whether we should abort the process or should throw exception when invalid characters are found. I know Caffe doen't throw exceptions and many C++ people don't like to throw exceptions (but I don't know why). From Python perspective, it would be nicer if we could catch all Caffe errors. I know Google logging library always aborts if check failed. Besides that, is there any specific reason that Caffe does not throw? Is there any discussion regarding it?

seanbell:
See #1683 for a discussion of exceptions vs fatal errors.

tnarihi:
Thanks for the pointer, @seanbell.

tnarihi:
Just committed so that regex allows to import nested modules.

cmake/Dependencies.cmake
```diff
@@ -128,6 +128,7 @@ if(BUILD_python)
   if(PYTHONLIBS_FOUND AND NUMPY_FOUND AND Boost_PYTHON_FOUND)
     set(HAVE_PYTHON TRUE)
     if(BUILD_python_layer)
+      find_package(Boost 1.46 COMPONENTS regex)
```
stas-sl: I changed this line in order to compile successfully with cmake:
`find_package(Boost 1.46 COMPONENTS regex python)`

Anyway thanks for the PR!

tnarihi:
Thanks for the fix @stas-sl! Actually I didn't test that in my local since Travis test passed. I will modify it soon.

cmake/Dependencies.cmake
tnarihi: Thanks for the fix @stas-sl! Actually I didn't test that in my local since Travis test passed. I will modify it soon.

shelhamer:
Closing now that #2456 #2462 and #2871 have cherry-picked or contributed alternatives. Thanks @tnarihi

PR #2004, opened 2015-02-28
Python classifier: multi-label prediction support

boechat107:
- The preprocessing procedures of the inputs were moved to a separated function. Its code is just the same.
- The new prediction function, `predict_multi`, does almost the same thing `predict` does, but it returns a dictionary of probabilities where the keys are the names of the top blobs of the last network's layer.
- The function `predict` has the same behavior of its original code; now it is just a special case of `predict_multi`, taking only the first output label's probabilities.

## Related References

> Goodfellow, I., Bulatov, Y., & Ibarz, J. (2013). Multi-digit Number Recognition from Street View Imagery using Deep Convolutional Neural Networks. Retrieved from http://arxiv.org/abs/1312.6082

> Jaderberg, M., Simonyan, K., Vedaldi, A., & Zisserman, A. (2014). Synthetic Data and Artificial Neural Networks for Natural Scene Text Recognition, 1–10. Retrieved from http://arxiv.org/abs/1406.2227v4

### DIFF
--- a/python/caffe/classifier.py
+++ b/python/caffe/classifier.py
@@ -89,6 +89,7 @@ class Classifier(caffe.Net):
         multipreds = {}
         for bname in self.outputs:
             multipreds[bname] = out[bname]
+            # For oversampling, average predictions across crops.
             if oversample:
                 tmparray = multipreds[bname]
                 tmparray = tmparray.reshape((len(tmparray) / 10, 10, -1))
--- a/python/caffe/classifier.py
+++ b/python/caffe/classifier.py
@@ -90,8 +90,8 @@ class Classifier(caffe.Net):
         for bname in self.outputs:
             multipreds[bname] = out[bname]
             if oversample:
-                tmparray = multipreds[bname].reshape(
-                    (len(predictions) / 10, 10, -1))
+                tmparray = multipreds[bname]
+                tmparray = tmparray.reshape((len(tmparray) / 10, 10, -1))
                 multipreds[bname] = tmparray.mean(1)
         return multipreds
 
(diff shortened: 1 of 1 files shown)


boechat107:
Hasn't anybody built a network with multiple top blobs?

wuhang:
@boechat107  
90 of your code has an undefined variable "predictions", I want to know what it is

boechat107:
Very good question @wuhang! That was a simple mistake and now it's fixed (I think...). If I have written tests, that wouldn't probably have happened. 

In addition the bug fix, I updated the code accordingly to last version of the file in the master branch.

If this PR is found to be useful, I can write some test for the modifications.

boechat107:
Ah, just to justify this PR... Without these modifications, I was not able to use the python interface for something like described in these two papers:

> Goodfellow, I., Bulatov, Y., & Ibarz, J. (2013). *Multi-digit Number Recognition from Street View Imagery using Deep Convolutional Neural Networks.* Retrieved from http://arxiv.org/abs/1312.6082

> Jaderberg, M., Simonyan, K., Vedaldi, A., & Zisserman, A. (2014). *Synthetic Data and Artificial Neural Networks for Natural Scene Text Recognition*, 1–10. Retrieved from http://arxiv.org/abs/1406.2227v4

I was successful with my forked version.  =)

shelhamer:
Sorry the lack of attention—the `caffe.Net` class is used almost universally while the `Classifier` is not so common. I'm planning to deprecate the `Classifier` class in favor of the notebook examples that show how to do classification and preprocessing. Thanks for extending all the same.

Issue #1997, opened 2015-02-27
Pillow needed for filter_visualization.ipynb


danielhamngren:
Hi, I was having some problems when following the ipython notebook examples. I got the error:
```
AttributeError: 'builtin_function_or_method' object has no attribute 'iterkeys'
```
when running the following line
```
scores = net.predict([caffe.io.load_image(caffe_root + 'examples/images/cat.jpg')])
```
The problem was resolved by installing the package Pillow with pip.
Since I'm new to this project I don't know the process on contributing, can I just add the line 
```
pillow>=2.70
```
to https://github.com/BVLC/caffe/blob/master/python/requirements.txt and do a pull request?

shelhamer:
> can I just add the line
> pillow>=2.70
> to https://github.com/BVLC/caffe/blob/master/python/requirements.txt and do a pull request?

Please do, then close this issue with a reference to your pull request. Thanks.

danielhamngren:
All right! It is done :-)

The pull request is over here https://github.com/BVLC/caffe/pull/2010

PR #2010, opened 2015-03-01
Added Pillow to requirements.txt

danielhamngren:
This change will prevent the caffe python package break due to lack of supported image formats. I don't know which image formats would be supported without Pillow (or PIL) but now at least JPEG and PNG are supported.

See https://github.com/BVLC/caffe/issues/1997

Most people in this field probably already have Pillow or PIL and wouldn't notice about this, but it'll make installing pycaffe a little bit easier on a fresh OS install.

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -14,3 +14,4 @@ python-dateutil>=1.4,<2
 protobuf>=2.5.0
 python-gflags>=2.0
 pyyaml>=3.10
+Pillow>=2.7.0


shelhamer:
Thanks!

lukeyeager:
Is 2.7.0 really required? 2.3.0 comes with Ubuntu 14.04 (`apt-get install python-pil`), and seems to do the job for me.

@danielhamngren, was there a reason you chose 2.7.0? Or was that simply the default version that came when you did a `pip install Pillow`? It looks to be an overly stringent requirement.

danielhamngren:
Hi, I used 2.7.0, as you guessed,  because that was the default one at the time when doing a 'pip install pillow'. 

If everything works with 2.3.0,  then I agree 2.7.0 is to stringent.

PR #2028, opened 2015-03-04
fixed missing climits header in build

beniz:
Missing climits header was killing my builds with brand new commits.


jeffdonahue:
Fixed in #2035

PR #2031, opened 2015-03-04
Add error checking for image mean

lukeyeager:
When setting the mean, assert that it is either one pixel or an array with
shape equal to the input data size.

(Previously #1931)


shelhamer:
Thanks for the descriptive check -- and the rebase.

PR #2033, opened 2015-03-04
Unrolled recurrent layers (RNN, LSTM)

jeffdonahue:
(Replaces #1873)

Based on #2032 (adds EmbedLayer -- not needed for, but often used with RNNs in practice, and is needed for my examples), which in turn is based on #1977.

This adds an abstract class `RecurrentLayer` intended to support recurrent architectures (RNNs, LSTMs, etc.) using an internal network unrolled in time.  `RecurrentLayer` implementations (here, just `RNNLayer` and `LSTMLayer`) specify the recurrent architecture by filling in a NetParameter with appropriate layers.

`RecurrentLayer` requires 2 input (bottom) Blobs.  The first -- the input data itself -- has shape `T x N x ...` and the second -- the "sequence continuation indicators" `delta` -- has shape `T x N`, each holding `T` timesteps of `N` independent "streams".  `delta_{t,n}` should be a binary indicator (i.e., value in {0, 1}), where a value of 0 means that timestep t of stream n is the beginning of a *new* sequence, and a value of 1 means that timestep t of stream n is *continuing* the sequence from timestep t-1 of stream n.  Under the hood, the previous timestep's hidden state is multiplied by these delta values.  The fact that these indicators are specified on a per-timestep and per-stream basis allows for streams of arbitrary different lengths without any padding or truncation.  At the beginning of the forward pass, the final hidden state from the previous forward pass (`h_T`) is copied into the initial hidden state for the new forward pass (`h_0`), allowing for exact inference across arbitrarily long sequences, even if `T == 1`.  However, if any sequences cross batch boundaries, backpropagation through time is approximate -- it is truncated along the batch boundaries.

Note that the `T x N` arrangement in memory, used for computational efficiency, is somewhat counterintuitive, as it requires one to "interleave" the data streams.

There is also an *optional* third input whose dimensions are simply `N x ...` (i.e. the first axis must have dimension `N` and the others can be anything) which is a "static" input to the LSTM.  It's equivalent to (but more efficient than) copying the input across the `T` timesteps and concatenating it with the "dynamic" first input (I was using my `TileLayer` -- #2083 -- for this purpose at some point before adding the static input).  It's used in my captioning experiments to input the image features as they don't change over time.  For most problems there will be no such "static" input and you should simply ignore it and just specify the first two input blobs.

I've added scripts to download COCO2014 (and splits), and prototxts for training a language model and LRCN captioning model on the data. From the Caffe root directory, you should be able to download and parse the data by doing:

```
cd data/coco
./get_coco_aux.sh # download train/val/test splits
./download_tools.sh # download official COCO tool
cd tools
python setup.py install # follow instructions to install tools and download COCO data if needed
cd ../../.. # back to caffe root
./examples/coco_caption/coco_to_hdf5_data.py
```

Then, you can train a language model using `./examples/coco_caption/train_language_model.sh`, or train LRCN for captioning using `./examples/coco_caption/train_lrcn.sh` (assuming you have downloaded `models/bvlc_reference_caffenet/bvlc_reference_caffenet.caffemodel`).

Still on the TODO list: upload a pretrained model to the zoo; add a tool to preview generated image captions and compute retrieval & generation scores.


cvondrick:
Firstly, thanks for the fantastic code. I had been playing with my own LSTM, and found this PR, and it is above and beyond any of my own attempts. Really nice job.

There seems to be a bug in the ReshapeLayer of this PR. In some cases, the ReshapeLayer will produce all zeros instead of actually copying the data. I've created a minimal test case that shows this failure for this PR:

    # Load a random dataset 
    layer {
      name: "ToyData_1"
      type: "DummyData"
      top: "ToyData_1"
      dummy_data_param {
        shape {
          dim: 101 
          dim: 7
          dim: 3
        }
        data_filler {
          type: "gaussian"
          std: 1
        }
      }
    }
    
    layer {
      name: "ZeroData"
      type: "DummyData"
      top: "ZeroData"
      dummy_data_param {
        shape {
          dim: 101 
          dim: 7
          dim: 3
        }
        data_filler {
          type: "constant"
          value: 0
        }
      }
    }
    
    
    # Reshape ToyData_1 to be the same size
    layer {
      name: "Reshape"
      type: "Reshape"
      bottom: "ToyData_1"
      top: "ToyData_2"
      reshape_param {
        shape {
          dim: 101
          dim: 7
          dim: 3
        }
      }
    }
    
    
    
    
    
    # Since ToyData_1 should be ToyData_2, we expect that this loss be zero. But, it is not.
    layer {
      name: "EuclideanLoss"
      bottom: "ToyData_1"
      bottom: "ToyData_2"
      top: "ToyData_1_vs_2_Difference"
      type: "EuclideanLoss"
    }
    
    # We expect this loss to be non-zero, and it is non-zero. 
    layer {
      name: "EuclideanLoss"
      bottom: "ToyData_1"
      bottom: "ZeroData"
      top: "ToyData_1_vs_Zero_Difference"
      type: "EuclideanLoss"
    }
    
    # Since ToyData_2 should be ToyData_1 and ToyData_1 is non-zero, we expect this loss to be non-zero. But, it is zero.
    layer {
      name: "EuclideanLoss"
      bottom: "ToyData_2"
      bottom: "ZeroData"
      top: "ToyData_2_vs_Zero_Difference"
      type: "EuclideanLoss"
    }

Above, it loads a random dataset, ToyData_1. It then reshapes it to the exact same size (identity) to create ToyData_2. We would expect that || ToyData_1 - ToyData_2 ||_2 == 0

However, if you train with the above model on this branch, you will see that the Euclidean loss between ToyData_1 and ToyData_2 is non-zero. Moreover, the loss between ToyData_2 and a blob of all zeros is zero. Note that, as expected, the loss between ToyData_1 and all zeros is zero. 

It seems there is a bug with reshape. I've fixed it here by copying an older version of Reshape into this branch: https://github.com/cvondrick/caffe/commit/3e1a0ff73fef23b8cb8adc8223e0bb2c8900e56b

Unfortunately, I didn't have time to write a real unit test for this. But, hope this bug reports helps.

Carl

jeffdonahue:
Well that's disturbing... I don't have time to look into it now but thanks
a lot for reporting Carl! Will follow up when I've figured something out.

jeffdonahue:
Oops, failed to read to the end and see that you already had a fix.  Thanks
for posting the fix! (I *think* the current version of my reshapelayer PR
may do what your fix does, in which case I'll just rebase this onto that PR
as I should anyway.)

cvondrick:
Thanks Jeff -- yeah, we fixed it by copying a ReshapeLayer from somewhere else. Unfortunately, we have lost track of exactly where that layer came from, but I'm sure somebody (maybe even you) wrote it at some point.

hf:
When is this feature going to be ready? Is there something to be done?

thuyen:
For the captioning model, can anyone show me how to generate captions after the training is done? Current LSTM layers process the whole input sequence (20 words in the coco example) across time, but we need to generate one by one at each time step (current time step is the input to the next).

vadimkantorov:
I've just tried to run train_lcrn.sh (after running coco_to_hdf5_data.py and other scripts) and I get a "dimensions don't match" error:
```
F0324 16:37:24.435840 20612 eltwise_layer.cpp:51] Check failed: bottom[i]->shape() == bottom[0]->shape()
```
The stack-trace and log are here: http://pastebin.com/fWUxsSmv

I've uncommented line 471 in net.cpp to find the faulty layer (the only modification). It seems it happens in lstm2 which blends input from the language model and from the image CNN.

train_language_model.sh runs fine without errors.

Ideas?

ih4cku:
By the way, does Caffe's recurrent layer support bi-directional RNN?

vadimkantorov:
Both factored and unfactored setups are concerned. Seems there are some dimensions problems while blending CNN input with embedded text input.

ritsu1228:
I have the same question as @thuyen. My understanding is that the current unrolled architecture slices an input sentence and feed the resulting words to each time step at once. So, for both train and test nets, the ground truth sentences are fed to the unrolled net. However, for captioning an image, there is no sentence to give to the net. But I don't think it is correct to give the start symbol to each layer. Did I miss anything?

ritsu1228:
The dimension check fails for the static input (the image feature) with size 100*4000 vs 1*100*4000. It seems to be caused by Reshape layer; @cvondrick 's fix seems to solve this.

jeffdonahue:
Yes,  as noted by @cvondrick, this works with the older version of the ReshapeLayer which puts everything in `Reshape` (as opposed to the newer one that uses `LayerSetUp` -- see discussion with @longjon in #2088).  I don't yet have any idea why the `Reshape` version would work but the `LayerSetUp` version wouldn't, but I've just force pushed a new version of this branch that uses the previous `ReshapeLayer` version, and confirmed that both example scripts (`train_lrcn.sh` & `train_language_model.sh`) run.  Sorry for breaking the LRCN one.

jeffdonahue:
> By the way, does Caffe's recurrent layer support bi-directional RNN?

You can create a bi-directional RNN using 2 RNN layers and feeding one the input in forward order and the other the input in backward order, and fusing their per-timestep outputs however you like (e.g. eltwise sum or concatenation).

vadimkantorov:
@jeffdonahue @ritsu1228 Thanks for figuring it out! A bit newbie quesiton: how do I get the fix?
The recurrent branch of @jeffdonahue doesn't have new commits.

vadimkantorov:
Thanks @jeffdonahue , training lrcn now works! Same question as @thuyen @ritsu1228. Does anyone have an idea how to hook up to when the first word after the start symbol gets produced and put the next symbol on the input_sentence tensor in memory before the next round of unrolled net will get to run?

read-mind:
As @jeffdonahue mentioned, bidirectinal RNN can be built with two RNNs, it's easy to prepare reversed input sequence, but how to reverse the output of one RNN when fusing two RNN outputs in Caffe? It seems no layer does the reverse.

jeffdonahue:
> As @jeffdonahue mentioned, bidirectinal RNN can be built with two RNNs, it's easy to prepare reversed input sequence, but how to reverse the output of one RNN when fusing two RNN outputs in Caffe? It seems no layer does the reverse.

True; one would need to implement an additional layer to do the reversal.  You'd also need to be careful to ensure that your instances do not cross batch boundaries (as is allowed by my implementation as it works fine for unidirectional architectures) since inference at each timestep is dependent on all other timesteps in a bidirectional RNN.

> Same question as @thuyen @ritsu1228. Does anyone have an idea how to hook up to when the first word after the start symbol gets produced and put the next symbol on the input_sentence tensor in memory before the next round of unrolled net will get to run?

In the not-too-distant future I'll add code for evaluation, including using the model's own predictions as input in future timesteps as you mention.

jeffdonahue:
I've also gotten a number of questions on the optional third input to RecurrentLayer -- I've added some clarification in the original post:

> There is also an *optional* third input whose dimensions are simply `N x ...` (i.e. the first axis must have dimension `N` and the others can be anything) which is a "static" input to the LSTM.  It's equivalent to (but more efficient than) copying the input across the `T` timesteps and concatenating it with the "dynamic" first input (I was using my `TileLayer` -- #2083 -- for this purpose at some point before adding the static input).  It's used in my captioning experiments to input the image features as they don't change over time.  For most problems there will be no such "static" input and you should simply ignore it and just specify the first two input blobs.

liqing-ustc:
Thanks for the fantastic code. But the code of the Reshape function in Recurrent layer makes me confused. when passing data from "output_blobs_" to "top blobs", why it is
```C++
    output_blobs_[i]->ShareData(*top[i]);
    output_blobs_[i]->ShareDiff(*top[i]);
```
rather than
```C++
    top[i]->ShareData(*output_blobs_[i]);
    top[i]->ShareDiff(*output_blobs_[i]);
```
it seems that the top blobs is just reshaped and empty.

the original code is here:

```C++
template <typename Dtype>
void RecurrentLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
  CHECK_EQ(top.size(), output_blobs_.size());
  for (int i = 0; i < top.size(); ++i) {
    top[i]->ReshapeLike(*output_blobs_[i]);
    output_blobs_[i]->ShareData(*top[i]);
    output_blobs_[i]->ShareDiff(*top[i]);
  }
  x_input_blob_->ShareData(*bottom[0]);
  x_input_blob_->ShareDiff(*bottom[0]);
  cont_input_blob_->ShareData(*bottom[1]);
  if (static_input_) {
    x_static_input_blob_->ShareData(*bottom[2]);
    x_static_input_blob_->ShareDiff(*bottom[2]);
  }
}
```

sunbaigui:
I've gotten a lrcn model according your example, I'm curious about how to predict the sentence of a given image. It seems there's no such a tool or deploy.prototxt yet, are you working on it?

niuchuang:
Could someone give me some guidance about how to construct a RNN with jeffdonahue's PR? I have downloaded the lrcn.prototxt , unfortunately I cannot understand most of its contents , such as  include { stage: "freeze-convnet" }, include { stage: "unfactored" } and so on. In fact,I have some time sequence image data , each of which has a label. I have trained reference model in caffe with these data,  and now I try to use RNN to classify them.  What document I should read so that I can understand lrcn.prototxt  and something like this,and then train a RNN model with my data. Much thanks !

shls:
@jeffdonahue Hi. Could you please give me an example about using my own images based on LRCN? Thanks in advance.

Edward12138:
Hi, I'm wondering if anyone can explain that all the elewise product operations are implemented as the sum operations..Thanks!

Natanada:
@jeffdonahue Hi, thank you for this code. I'm able to train the LRCN on mscoco and my own data set, but I am unable to load the network to view the generated sentences. My understanding is that the layers are unrolled at all times, making sentence generation given an image rather unintuitive. Could you provide some clarification on using the LRCN for generation?
Thanks again!

raingo:
@jeffdonahue Hi, I had a look at the code, and have one question.

Between line 141-149 of rnn_layer.cpp, x_cont is combined with previous hidden state. Why they are combined using element sum operation, rather than element production?

Thanks!

Edward12138:
@raingo  I've got the same question. Not sure if they use some trick in the sum layer.

pakjce:
In RecurrentLayer class, all recurrent networks are unfolded in the LayerSetUp() method. However, in case of a real-time application such as robotic control (batch size is one), unfolding is not necessary. Is there any plan to considering this case?

jeffdonahue:
@raingo EltwiseLayer with the addition operation optionally takes scalar coefficients for each of the addends.  I modified the layer to accept an additional bottom *Blob* of coefficients when the `coeff_blob` option is set.  It is a little hacky and should probably be in a separate layer.

@pakjce as far as I can tell there is no reason to have a special case for `T==1`; just provide `1xN` blobs to indicate to the layer you're using data with a single timestep -- this is what I do for inference.

I pushed a [very unpolished branch](https://github.com/jeffdonahue/caffe/blob/recurrent-rebase-cleanup) to my public fork containing inference code (including deploy prototxts as many people have requested).  I used [`retrieval_experiment.py`](https://github.com/jeffdonahue/caffe/blob/recurrent-rebase-cleanup/examples/coco_caption/retrieval_experiment.py) for all image captioning results in the paper.  (You'll probably have to manually modify the variable declarations at the beginning of the `main` function as appropriate for your desired experimental config.) The file `captioner.py` in that same directory contains most of the relevant inference code.

lostoy:
Can anyone share some ideas about : with this lstm, how to do the temporal average pooling, for video classification?

mgarbade:
In @jeffdonahue's post from Mar 4 (first comment on top of this thread) he explains how to set up his LRCN caffe branch for COCO.
However I cannot make it run. 
First of all I had to manually download COCO since "get_coco_aux.sh" only downloaded the filename lists "coco2014_filename.train.txt" etc.
When I try to execute "./examples/coco_caption/coco_to_hdf5_data.py" I get the error:
IOError: [Errno 2] No such file or directory: './data/coco/coco/annotations/captions_train2014.json'

I would adapt the python script "coco_to_hdf5_data.py" myself if I could but it is quite lengthy and I don't know how to repair it. Also I don't have the file "captions_train2014.json" required by the script. It wasn't downloaded.

Any help would be appreciated.

seed93:
@mgarbade You need to install pycocotools using `download_tools.sh`

mgarbade:
@seed93  I tried to do this already. It's not downloading the dataset however. It only produces the following error:

    data/coco/download_tools.sh 
    ~/libs/caffe/caffe-recurrent-rebase-cleanup ~/libs/caffe/caffe-recurrent-rebase-cleanup
    fatal: destination path 'coco' already exists and is not an empty directory.
    ~/libs/caffe/caffe-recurrent-rebase-cleanup
    Cloned COCO tools to: /home/garbade/libs/caffe/caffe-recurrent-rebase-cleanup/data/coco/coco
    To setup COCO tools (and optionally download data), run:
    cd /home/garbade/libs/caffe/caffe-recurrent-rebase-cleanup/data/coco/coco
    python setup.py install
    and follow the prompts.

In the meantime I found the files "captions_train2014.json" which are also part of the coco dataset which I downloaded manually. However the proplem remains that since the dataset is not downloaded automatically the paths to the the manually downloaded dataset are probably wrong...
Did you manage to get the example running on your computer?

liuxinfengabc:
an example of training an LRCN model for image captioning in the COCO dataset   has been provided , I want to know when the activity recognition example is  released?

ryaneleary:
Any rough timeline for when this work will be merged into master?

seed93:
@mgarbade branch 'recurrent-rebase-cleanup' uses a newer version of cocotools, which doesn't download dataset automatically. You may download by yourself. PS: another problem you may encounter is in file `coco_to_hdf5_data.py`, you should modify line 19 as
```
from pycocotools.coco import *
```

odellus:
+1

jinserk:
Anyone can explain how to run the retrieval_experiment.py in the branch 'recurrent-rebase-cleanup'? There are lots of prototxt here, and the python file internally uses 'lrcn_finetune_' prefixed caffemodel, and uses 'lrcn_word_to_preds.deply.prototxt', but I have no idea how to train the lrcn_finetune_... or lrcn_word_to_preds....prototxts at all.
Moreover, when I changed the lrcn_finetune_... to lrcn_iter_... then ran it, an error occurred as follows:
[libprotobuf ERROR google/protobuf/text_format.cc:245] Error parsing text-format caffe.NetParameter: 17:15: Message type "caffe.LayerParameter" has no field named "embed_param".
F0818 11:44:12.371534  8094 upgrade_proto.cpp:928] Check failed: ReadProtoFromTextFile(param_file, param) Failed to parse NetParameter file: ./examples/coco_caption/lrcn_word_to_preds.deploy.prototxt
Aborted (core dumped)

And could you help me how to fix it?

FabHan:
Is there a plan to merge this to the main branch?

jeffdonahue:
I've updated this with a new version that uses `ScalarLayer` (#3021) instead of `EltwiseLayer` with `coeff_blob` (this confused a lot of people previously), and includes the eval tools and language model notebook that were only in `recurrent-rebase-cleanup` before.  The eval tools and notebook still need to be groomed heavily.

For reference, the previous version is still available in my fork under a new [`recurrent-v3ish`](https://github.com/jeffdonahue/caffe/tree/recurrent-v3ish) branch.

(Also relies on #3022.)

PiranjaF:
> The first -- the input data itself -- has shape T x N x ... and the second -- the "sequence continuation indicators" delta -- has shape T x N, each holding T timesteps of N independent "streams". delta_{t,n} should be a binary indicator (i.e., value in {0, 1}), where a value of 0 means that timestep t of stream n is the beginning of a new sequence, and a value of 1 means that timestep t of stream n is continuing the sequence from timestep t-1 of stream n.

Despite the nice description I'm trying to understand exactly how to input the data to the RNNLayer/LSTMLayer. I want to do a many-to-one classification similar to a sentiment analysis (http://karpathy.github.io/2015/05/21/rnn-effectiveness/), but in another area than NLP. Say that I have `M` observations and each of them have `T` sequences (of varying length) with `V` features/dimensions related to each element in the sequence. For instance, this could be `M=5` sentences with `T=[10, 7, 7, 7, 12]` words in each of those and `V=100` dimensions from the word2vec representation. How would I store that? For my use case I cannot use an EmbedLayer/LookupLayer as none of the elements in the sequences are identical (unlike words in NLP). Therefore I need to store this data directly in the HDF5 file.

@jeffdonahue, I don't understand exactly what you mean by a "stream" - how is it related to the variables `M`, `T` and `V` I've identified above? What should the dimensions be for my dataset in the HDF5 file?

Others should also feel free very to drop a hint. Any help would be very appreciated!

szcom:
gradient scaling question:
It seems that I need to specify batch size in # of timesteps. That is if my average unaligned sequence length is 50 timesteps then in order to get mini-batch of about 10 sequences I should put 500 as my batch size:
layer {
  name: "data"
  type: "HDF5Data"
  top: "cont_phonemes"
  top: "input_phonemes"
  top: "target_duration"
  include { phase: TRAIN }
  hdf5_data_param {
    source: "./n.save/train_phonemes.txt"
    batch_size: 500
  }
}
Result in the gradient is scaled by 500 - number of timesteps instead of number of sequences?
As for ex [here](https://github.com/BVLC/caffe/blob/master/src/caffe/layers/euclidean_loss_layer.cpp#L29)

prophecies:
@jeffdonahue The updates to caffe-recurrent branch a week back were really welcome additions to an already great code. Are you also planning to provide an examples for activity recognition anytime soon?
In the meanwhile, can you please point me in the right directions as to how I can average or get output only from final time-step from LSTM (something akin to slide 10 from your CVPR presentation). I am trying to do one-of-k classification on some sequences. 
So, few hints about how to extract output from only specific time-steps could get me started. Of course example about a activity recognition, sentiment analysis, or any other many-to-one classification on sequential data would be very helpful too.

karthikthiagu:
@PiranjaF In the description, the input data is given to have dimensions T x N x ... If the dimension of each input time step is D, then the input data would have dimension T x N x D, that is the reason the description has those dots following N. It would be written to a .h5 file as a numpy array with shape TxNxD.

In a TxN array, each column represents an independent stream. Treat a stream like a list of sequences one below the othe. The terms of the sequence become the columns of the array. In the case of sentences, each word corresponds to a time step and the dimensionality of the embedded-word vector would be V. If there are 10 sentences each of length 20 (after padding and truncating them to a maximum length of 20) and 5 parallel streams with the dimensionality of 100 for the word-vectors, the variables would look like this:
T = 40 (2 sentences in each stream, 20 words in a sentence)
N = 5 (5 parallel streams)
V = 100 (dimension of word-vector

SunnyWay:
Could someone tell me why *input_sentence* and *target_sentence* use -1 as paddings ?  and How does the *EmbedLayer* deal with it ? Much Thanks !

PiranjaF:
@karthikthiagu Thanks! So just to clarify. Based on your answer and the initial post by @jeffdonahue it seems that the reason for having parallel streams is for optimization purposes. I'm trying to understand why. Are all of these streams loaded onto the GPU and then processed one-by-one to update the weights? Is it a good number of parallel streams then the highest number possible to hold in memory by the GPU? 

In your example you have 20 words in each sentence. How would specify that I want to do a many-to-one classification of these sentences so that each value in the label `y` is matched to a sentence? This is linked to what @prophecies is asking about. If possible, could you both specify how I should output the two values for `y` in this case and where I should specify the many-to-one relationship? Thanks for your help!

ijhong:
Did anyone have troubles to run the retrieval_experiment.py ? 
I did clone the original Caffe, fetch pull 2033, and checkout 2033 on my computer. 
For running retrieval_experiment.py, I ran train_lrcn.sh and finetune_lrcn.sh. ahead. 

When running retrieval_experiment.py, it seemed that everything goes well -- loading annotations, initializing net from parameters, computing descriptors for images, and so on. 
And then scoring captions for image begun and the program was killed around 146/40504. 
The error message shown was following: 

Scoring captions for image 146/40504Traceback (most recent call last):
  File "./examples/coco_caption/retrieval_experiment.py", line 362, in <module>
    main()
  File "./examples/coco_caption/retrieval_experiment.py", line 359, in main
    experimenter.retrieval_experiment()
  File "./examples/coco_caption/retrieval_experiment.py", line 156, in retrieval_experiment
    self.score_captions(image_index)
  File "./examples/coco_caption/retrieval_experiment.py", line 63, in score_captions
    outputs = pickle.load(caption_scores_file)
MemoryError
Killed

I checked the memory state using 'top' command later again and found that the percentage of memory usages increases rapidly as it scores numbers of captions. (like memory leaks) 

Could anyone tell how to test it? Or it's okay to tell me some similar cases and its solutions. 
Thanks, 
Injeong.

examples/coco_caption/captioner.py
```diff
@@ -0,0 +1,402 @@
...
+  def set_caption_batch_size(self, batch_size):
+    self.lstm_net.blobs['cont_sentence'].reshape(1, batch_size)
+    self.lstm_net.blobs['input_sentence'].reshape(1, batch_size)
+    self.lstm_net.blobs['image_features'].reshape(batch_size,
+        *self.lstm_net.blobs['image_features'].data.shape[1:])
+    self.lstm_net.reshape()
+
+  def preprocess_image(self, image, verbose=False):
+    if type(image) in (str, unicode):
+      image = plt.imread(image)
```
ronghanghu: I just tried to use it today to predict captions on some PNG images, and it doesn't work, because for png images, `plt.imread` returns (0, 1) np.float32 rather than uint8.

ronghanghu: Although this might not be a issue with COCO, it may still be better to add
```
    if image.dtype == np.float32:
        image *= 255
        image = image.astype(np.uint8)
```

miliadis:
Hi,

I am wondering if matlab's code net.forward() supports RNN Layers. I tried to use it but I got an error:

Error parsing text-format caffe.NetParameter: 18:19: Message type "caffe.LayerParameter" has no field named "recurrent_param".

davheld:
Is there a way to have separate RNN's for encoding and decoding using this framework, and to backprop through both?

bhack:
What do you think of this [benchmarks](https://github.com/junhyukoh/caffe-lstm/blob/master/README.md)? Are still valid?

cuihenggang:
Hi,

I'm trying the "coco_caption" example of the "jeffdonahue:recurrent" branch. I find I need to do the following changes in order t make it work:

''''
diff --git a/examples/coco_caption/coco_to_hdf5_data.py b/examples/coco_caption/coco_to_hdf5_data.py
index 300e474..0bcf71f 100755
--- a/examples/coco_caption/coco_to_hdf5_data.py
+++ b/examples/coco_caption/coco_to_hdf5_data.py
@@ -10,13 +10,13 @@ import sys
 sys.path.append('./examples/coco_caption/')
 
 COCO_PATH = './data/coco/coco'
-COCO_TOOL_PATH = '%s/PythonAPI/build/lib/pycocotools' % COCO_PATH
+COCO_TOOL_PATH = '%s/PythonAPI/build/lib.linux-x86_64-2.7/pycocotools' % COCO_PATH
 COCO_IMAGE_ROOT = '%s/images' % COCO_PATH
 
 MAX_HASH = 100000
 
 sys.path.append(COCO_TOOL_PATH)
-from coco import COCO
+from pycocotools.coco import COCO
 
 from hdf5_sequence_generator import SequenceGenerator, HDF5SequenceWriter
'''

It might have been fixed in other branches, but I guess it might be useful for people who are trying this to see it before they give up.

Thanks,
Cui

cuihenggang:
When I'm using the LRCN example, it reports out of memory error. Can I just use a smaller batch size to make it work? I'm asking this because I find in the generated HDF5 data directory, it explicitly says "buffer_100". The directory is called "examples/coco_caption/h5_data/buffer_100", so can I use batch sizes other than 100?

PS. My GPU has 4.8 GB memory. I'm wondering how much memory does it use?

Thanks,
Cui

The error message:

...
I1120 13:29:15.401499 41584 caffe.cpp:211] Starting Optimization
I1120 13:29:15.401648 41584 solver.cpp:293] Solving lrcn_caffenet_to_lstm
I1120 13:29:15.401669 41584 solver.cpp:294] Learning Rate Policy: step
I1120 13:29:15.406371 41584 solver.cpp:346] Iteration 0, Testing net (#0)
I1120 13:29:16.967797 41584 blocking_queue.cpp:50] Data layer prefetch queue empty
I1120 13:30:01.502318 41584 solver.cpp:414]     Test net output #0: accuracy = 6.72326e-05
I1120 13:30:01.502837 41584 solver.cpp:414]     Test net output #1: cross_entropy_loss = 9.31544 (* 20 = 186.309 loss)
I1120 13:30:01.502871 41584 solver.cpp:346] Iteration 0, Testing net (#1)
I1120 13:30:22.846604 41584 solver.cpp:414]     Test net output #0: accuracy = 0.000105017
I1120 13:30:22.846714 41584 solver.cpp:414]     Test net output #1: cross_entropy_loss = 9.30936 (* 20 = 186.187 loss)
F1120 13:30:23.124014 41584 syncedmem.cpp:58] Check failed: error == cudaSuccess (2 vs. 0)  out of memory
    @     0x7f14c5770daa  (unknown)
    @     0x7f14c5770ce4  (unknown)
    @     0x7f14c57706e6  (unknown)
    @     0x7f14c5773687  (unknown)
    @     0x7f14c5e4ac21  caffe::SyncedMemory::to_gpu()
    @     0x7f14c5e49fa9  caffe::SyncedMemory::mutable_gpu_data()
    @     0x7f14c5d447d3  caffe::Blob<>::mutable_gpu_diff()
    @     0x7f14c5eaee97  caffe::InnerProductLayer<>::Backward_gpu()
    @     0x7f14c5e7af57  caffe::Net<>::BackwardFromTo()
    @     0x7f14c5e7b0c1  caffe::Net<>::Backward()
    @     0x7f14c5e9c791  caffe::Solver<>::Step()
    @     0x7f14c5e9cfd5  caffe::Solver<>::Solve()
    @           0x4092b3  train()
    @           0x406c01  main
    @     0x7f14c44c1ec5  (unknown)
    @           0x40739b  (unknown)

cuihenggang:
Hi, I'm a little confused about how I should understand the testing accuracy of this example application.

When I train the language model (without using any image data), I can get an accuracy of about 50% in about a day. I think the accuracy layer calculates the accuracy by just comparing each word of the generated sentence with the ground truth sentence (word by word). So I don't understand why I can get 50% of the image caption words correct without looking at any images. Am I missing anything here?

deepsemantic:
Hi @cuihenggang , I have same question, did you find a way to reduce the batch size? Do you know how the buffer size relate to image batch size and sentence batch size? Thanks

cuihenggang:
Hi @deepsemantic, yeah, I figured it out. To use a different batch size, you will need to modify the "coco_to_hdf5_data.py" script to change the "BUFFER_SIZE" variable (it is the batch size), and then run this script to generate the HDF5 data with that batch size.

deepsemantic:
@cuihenggang Thanks for the hints!

elliotchencv:
Hi, I would like to know if I understood pycaffe and this branch correctly.

Did `net.predict()` produce all the unrolled _outputs_ while `net.forward()` produces _single_ output only?

Will the model produces exact same results using both ways?

stas-sl:
I have a question about the third 'static' input. Let's say, T=10 and I have quite short sequences of length=4 describing images based on image features fed into the LSTM layer . So these sequences will be stacked one below other in one 'stream' in one batch. But they should have different image features as 'static' input. Should I pad sequences so that there would be only one of them per stream per batch?

elliotchencv:
@stas-sl According to my understanding, the LSTM layer reads two or three inputs:
```
bottom: "TimeVariantFeature" #T x N x D1 x D2 ...
bottom: "ContinuityFlag" #T x N
bottom: "StaticFeature" #D1' x D2' ...
```

I guess the term "static" means it never changes as time **t_i** varies throughout stream **n_i** (I suppose sequence ≈ stream).

You can define the _static_ or _dynamic_ input shape whatever you want.

If you have multiple sequences with different _static_ or _dynamic_ inputs, just make multiple streams.

Please correct me if I was wrong.

shaibagon:
@elliotchencv I believe the static feature should be of dimension `N x D1 x D2` - you should have the same static feature for all time steps T, but different for each stream/batch.

stas-sl:
@elliotchencv, the reason why I'm asking is because sequence != stream, there could be a lot of sequences in one stream - one after another. You can have multiple short sequences in one stream, or there could be a long sequence spread along multiple batches. 

I want to give image features as input to LSTM and it should produce some sequence as output. In order to do this, as I understand, I need to repeat these features multiple times - for every timestep, but they will be the same. So, I was thinking that is what static input for. The input is static along a single sequence, but because there could be multiple sequences per stream per batch, it could be not static along the whole batch.

So, I see 2 options now:
1. Do not use static input, but manually repeat image features multiple times as first dynamic input.
2. Pad sequences to have length T, so that there could be only one sequence per stream per batch and use static input. Actually, in this case I don't need dynamic input, probably I can fill it with some zeros.

twinanda:
I am still having a problem setting up my input data for the network. I am solving a classification problem with T input and T output.

layer {
  name: "data"
  type: "HDF5Data"
  top: "fc"
  top: "labels"
  top: "cont_seq"
  include {
    phase: TRAIN
  }
  hdf5_data_param {
    source: "/media/free2_data1/data/matlab/predictOR/ImageList/Time-Estimation/Fine-Tune.txt"
    batch_size: 100
  }
}
layer{
  name: "lstm1"
  type: "LSTM"
  bottom: "fc"
  bottom: "cont_seq"
  top: "lstm1"
  recurrent_param{
    num_output: 1000
    weight_filler {
      type: "uniform"
      min: -0.08
      max: 0.08
    }
    bias_filler {
      type: "constant"
      value: 0
    }
  }
}
...

in this case, I have the dimensions of fc: 5828 x 40 x 7; labels: 5828 x 40; and cont_seq: 5828 x 40.
When I tried to train this architecture, I got an error
`F0215 18:23:16.143862  3488 hdf5_data_layer.cpp:54] Check failed: hdf_blobs_[i]->shape(0) == num (5828 vs. 7) `

I double- and triple-checked my input data in the hdf5 files. The dimensions are correct, but I still have this problem. Has anybody encountered this problem?

deepsemantic:
Hi all, 
I have a problem of memory usage when doing retrieval experiment with "retrieval_experiment.py" on MS coco 5000 test images. 
In computing recalls, it needs to compute the score for each (image, caption) pair  by calling  `def score_captions(self, image_index, output_name='probs') in "captioner.py" . I found the last line of this function: self.caption_scores[image_index] = outputs`. With the size of list "caption_scores" keep increasing, the memory usage reach to14GB (my machine with 16GB memory) and keep increasing.  I also found each computed ".pkl" file is 16.9M(for example, scores_image_000000.pkl), which means for 5000 test images, we totally need almost 85GB memory.  
Do you have same problem and how do you fix it? Thanks in advance.

chriss2401:
Hi,

I am trying to train a RNN (with donahue's code) by feeding feature maps extracted by a pre-trained CNN. Theoretically the LSTM layer accepts an input of TxNxW, where T is number of sequences, N is number of streams and W is number of dimensions. My W is three dimensional (256 28 28) but when I reshape my data to TxNx256x28x28 I get a notification from the recurrent layer that my bottom blob is of size TxNx256. Was thinking if the layer only receives the 256 dim of my feature map. If yes, should I maybe flatten the layer before passing it in? Thanks!

shaibagon:
@chriss2401 I worked with @jeffdonahue 's LSTM and I fed it 5-D blobs without any problem.
However, recently I came across some limitations to this LSTM implementation, so I am now using NetSpec() to manually unroll my own LSTMs: nasty prorotxt, but much more control over the architecture.

chriss2401:
@shaibagon thanks a lot for the heads up. Is NetSpec advantage mainly that you can feed two LSTM's at a given time step ?

chriss2401:
I have a few more general questions. 

1) I am attempting on training an RNN with @jeffdonahue 's LSTM, but for regression. So I removed the accuracy layer and replaced softmax loss with euclidean loss and changed the last IP layer to output 1 as suggested in #512 . Just wondering, will the TxN results that are fed in the euclidean loss output the averaged loss among the streams inputed ? So if I give a patch of 40 and resize it to 10x4 will the loss be the average of the losses among these ?

2) @jeffdonahue states that due to the clipping markers there is no need to pad or truncate sequences in order to "fit in the sequence size ( so if I give a 10x4 and my sequence is larger than 10 it is still ok since the clippers 1 will indicate that we are still working on the same sequence). But in his work "sequence to sequence - video to text" he actually does pad and truncate the videos in order to fit the 80 sequence size. I am a bit confused, does anyone have experience on this ? 

Any help would be appreciated, thanks.

shaibagon:
I have posted a simple example of LSTM implementation using `caffe.NetSpec()` on stackoverflow: http://stackoverflow.com/a/35967589/1714410

This implementation should work with regular caffe branch (master), and does not need this recurrent PR. 

Advantages of using NetSpec():
1. No need to use "non-master" branches of caffe, no need to merge PR to a private branch...
2. More flexibility: you can easily replace the `TanH` activation with other form (e.g., `PReLU`), you can feed one time step LSTM output as an input to the next time step (AFAIK, this PR does not yet support this feature).

Disadvantages:
1. LSTM is not encapsulated as a single layer in the prototxt: you end up with a very long prototxt describing the model and it is more difficult to debug architecture issues.

fl2o:
@shaibagon, Thank you for your stackoverflow post!

However, I need to process signals of different sizes using LSTM, and I am wondering if this is possible using your code? I also have a label associated to each time step (it's a regression problem), how to take that into account using the _euclideanLossLayer_? Thanks!

chriss2401:
@shaibagon thanks allot for your reference, I appreciate it. Ideally I would like to build a many-to-many regression RNN ( one output for each time step) , but it seems a bit not straightforward with this implementation (you have to use two LSTM layers and define which stage you're at I guess).

shaibagon:
@chriss2401 @fl2o The implementation I posted at SO is only a sketch: once you understand how to build the proper prototxt (`NetSpec`) for a single time step LSTM unit using basic caffe layers, the sky is the limit: you can use the `h1` output of `single_time_step_lstm(...)` as the per-time step descriptor on top of which you can have a `EuclideanLoss` layer for each time step. I suppose you can build a more complex python script that stack together several LSTM units (per time step) etc. 
For instance I currently have an architecture with a stack of 3 LSTM units per time step, unrolling it for 10 time steps results with a prototxt file with ~6.5K lines... 

BTW, I find `draw_net` utility very useful for inspecting the resulting model for structural "bugs"

jeffdonahue:
I totally agree (and I think I've commented elsewhere) that NetSpec is definitely the way to go for the most part -- and I may never have bothered to write the C++ versions here if I'd been aware of NetSpec at the time.  So thanks for posting your example @shaibagon!

The one advantage these C++ implementations have over the NetSpec route is the ability to remember the final timestep hidden state and carry it over as the zeroth timestep hidden state of the next batch for the ability to handle arbitrarily long sequences at training time (with gradients between batches truncated).  This is (AFAIK) impossible to do through NetSpec, even if you're allowed to add extra layers; it would require some extra primitives in the `Net` class itself, I think.  And in practice I don't know how much of an advantage this feature is for any application, but there is an example of it being used in `train_language_model.sh` if anyone is interested...

shaibagon:
@jeffdonahue  the feature of carrying `h` and `c` to the next temporal batch is a very important one. Do you need the entire `LSTM` layer for this feature, or is it already implemented in the `LSTMUnit` layer? 
The reason I am asking is that I which to have a net that feeds one time step's `h` as the bottom of the next time step LSTM. AFAIK, this is not possible to do using your `LSTM` layer in the `recurrent` branch. However, I managed to do that using `NetSpec()`. If I can use your `LSTMUnit`s combined with `NetSpec()` I believe I can enjoy both worlds: having `h` and `c` carried across temporal batches **AND** managing to feed `h_{t-1}` as a bottom to `LSTM_{t}`!

jeffdonahue:
Nope -- `LSTMUnit` is just a specialized non-linearity that computes c_t and h_t given the gate input values (applying sigmoid/tanh and summing), written with the hope that there's a speed/memory advantage to doing it in one GPU kernel vs. the equivalent composition of sigmoid/tanh/slice/concat/eltwise sum/prod layers.  (I'm pretty sure there is a (slight) memory advantage; not so sure of the speed advantage...)

I thought of doing this at some point, but probably won't get to it (so someone else should feel free to try if they like).  You'd want something like a "`SourceLayer`", taking no input, that can provide the 0th timestep hidden state (would be very simple, copying its memory to its top, similar to @longjon's parameter layer which is a PR that I can't find right now...), and a "`SinkLayer`", taking the final timestep hidden state as input and producing no output.  Then you'd have to add some logic to `Net` so that it copies the `SinkLayer`'s stored state to the `SourceLayer`'s internal state at the beginning of the next call to `Forward`.  (I think this primitive could allow for some very complex behavior beyond the obvious application to recurrent nets -- you might be able to implement parameter storage and SGD itself all in `Forward` with these primitives.)

shaibagon:
@jeffdonahue I was thinking of using the  "HDF output" data layer as the "sink" layer and a respective "HDFData" layer as the "source" layer. Less efficient than using specialized in-memory layers, but better than hacking SGDSolver... (IMHO).

shaibagon:
@jeffdonahue BTW, working with `caffe.NetSpec()` to construct networks starts to feel like working with Theano... In that case, I suppose Theano can offer more flexibility...

jeffdonahue:
I suppose the HDF5 combination could work (but probably requires more hacking of at least the data layer to accept changes to the file list at runtime? Sounds a bit hairy...).  And introducing more IO into the training loop is never going to be great for speed.  But may be okay for some applications, and I wouldn't doubt it's the quickest way to go.

> @jeffdonahue BTW, working with caffe.NetSpec() to construct networks starts to feel like working with Theano... In that case, I suppose Theano can offer more flexibility...

Of course -- see also Caffe2, TensorFlow, Torch, [...].  (BTW, I definitely wasn't actually advocating implementing learning using the primitive I described.)  Caffe (still) has its advantages, but it's certainly not the most flexible framework. (Don't want to go into that discussion any further here, though...)

shaibagon:
@jeffdonahue you override the same hdf file all the time, thus you don't need to change the hdf5 file list at all... Assuming hdf5data layer does not have prefetching :O

jeffdonahue:
Ah, I didn't know the HDF5OutputLayer worked that way, I see... sounds a little scary, but might work... good luck!

fl2o:
@shaibagon Thanks for the hightlight but I struggle to see how to handle signals with different lengths (ie timestep) for the training process using _NetSpec_? I can't change my unrolled net architecture during training...
Should I use a very long LSTM and stop the forward pass after I have reached the end of the signal being processed then start the backward pass?

shaibagon:
@fl2o AFAIK, if you want exact backprop for recurrent nets in caffe, there's no way around explicitly unrolling the net across ALL time steps. 
However, if you use @jeffdonahue 's recurrent branch, you will be able to achieve exact forward estimation, and backprop exact to the limit of the temporal batch size. This can alleviate the need to explicitly unroll very long temporal nets.

Regarding working with very long sequences:
1. You may define `maxT` and explicitly unroll your net to span `maxT` time steps, padding shorter sequences with some "null" data/label. 
2. Since caffe uses SGD for training, it is better to have more than one sequence participating in a forward-backward pass (i.e., mini-batch of size > 1). Otherwise gradient estimation will be very noisy.

Can you afford all these bolbs in memory at once?

shaibagon:
@jeffdonahue BTW, is there a reason why this PR is not merged into `master`?

fl2o:
@shaibagon I am gonna try padding shorter sequences with some "null" data/label (Should I use a special term or just 0 ?) in order to avoid the gradient estimation problem, but I am not sure yet about the memory issue..! (maxT will be around 400! while minT ~50)

shaibagon:
@fl2o  I'm not certain just using 0 is enough. You want no gradients to be computed from these padded time steps. You might need to have an "ignore_label" and implement your loss layer to support "ignore_label". 
**Make sure no gradients from the padded time steps are propagated into the "real" time steps**

fl2o:
That's what I was wondering ....
Wonder if it's not "easier" to use this PR directly ^^
Gonna figure it out! Thank you @shaibagon

shaibagon:
@fl2o in the future, I think it would be best to keep this github issue thread for PR related comments only. For more general inquires and questions about LSTM in Caffe, it might be better to ask a question in stackoverflow.

chriss2401:
@shaibagon Cheers for all the helpful comments.

lood339:
Hi, I used the LRCN code to generate captions form an image. I replace the alexNet with google net. The result likes this:
"A brown cat sitting top top top top ...."
The sentence repeats the word "top" a lot. Is there any reasons?
I also tried other modifications. It seams the LSTM is very sensitive to the learning parameters. Is this conclusion right in general?
Thanks.

examples/coco_caption/hdf5_sequence_generator.py
```diff
@@ -0,0 +1,132 @@
...
+          batch_indicators[t, i] = 0 if self.stream_indices[i] == 0 else 1
+          self.stream_indices[i] += 1
+          if self.stream_indices[i] == len(self.streams[i][self.substream_names[0]]):
+            num_completed_streams += 1
+        if not exhausted[i]: all_exhausted = False
+      if all_exhausted and truncate_at_exhaustion:
+        print ('Exhausted all data; cutting off batch at timestep %d ' +
+               'with %d streams completed') % (t, num_completed_streams)
+        for name in self.substream_names:
+          batch[name] = batch[name][:t, :]
```
yangfly: words at timestep t might not be deleted:
`batch[name] = batch[name][:(t+1), :]`

examples/coco_caption/hdf5_sequence_generator.py
```diff
@@ -0,0 +1,132 @@
...
+        batch_indicators = batch_indicators[:t, :]
```
yangfly: `batch_indicators = batch_indicators[:(t+1), :]`

liminchen:
Could anyone tell me what's the difference between C_diff and C_term_diff in the backward_cpu function? I'm trying to understand the code and write a GRU version. Thanks in advance!
`template <typename Dtype>
void GRUUnitLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
    const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom) {
  CHECK(!propagate_down[2]) << "Cannot backpropagate to sequence indicators.";
  if (!propagate_down[0] && !propagate_down[1]) { return; }

  const int num = bottom[0]->shape(1);
  const int x_dim = hidden_dim_ * 4;
  const Dtype* C_prev = bottom[0]->cpu_data();
  const Dtype* X = bottom[1]->cpu_data();
  const Dtype* flush = bottom[2]->cpu_data();
  const Dtype* C = top[0]->cpu_data();
  const Dtype* H = top[1]->cpu_data();
  const Dtype* C_diff = top[0]->cpu_diff();
  const Dtype* H_diff = top[1]->cpu_diff();
  Dtype* C_prev_diff = bottom[0]->mutable_cpu_diff();
  Dtype* X_diff = bottom[1]->mutable_cpu_diff();
  for (int n = 0; n < num; ++n) {
    for (int d = 0; d < hidden_dim_; ++d) {
      const Dtype i = sigmoid(X[d]);
      const Dtype f = (*flush == 0) ? 0 :
          (*flush * sigmoid(X[1 * hidden_dim_ + d]));
      const Dtype o = sigmoid(X[2 * hidden_dim_ + d]);
      const Dtype g = tanh(X[3 * hidden_dim_ + d]);
      const Dtype c_prev = C_prev[d];
      const Dtype c = C[d];
      const Dtype tanh_c = tanh(c);
      Dtype* c_prev_diff = C_prev_diff + d;
      Dtype* i_diff = X_diff + d;
      Dtype* f_diff = X_diff + 1 * hidden_dim_ + d;
      Dtype* o_diff = X_diff + 2 * hidden_dim_ + d;
      Dtype* g_diff = X_diff + 3 * hidden_dim_ + d;
      const Dtype c_term_diff =
          C_diff[d] + H_diff[d] * o * (1 - tanh_c * tanh_c);
      *c_prev_diff = c_term_diff * f;
      *i_diff = c_term_diff * g * i * (1 - i);
      *f_diff = c_term_diff * c_prev * f * (1 - f);
      *o_diff = H_diff[d] * tanh_c * o * (1 - o);
      *g_diff = c_term_diff * i * (1 - g * g);
    }
    C_prev += hidden_dim_;
    X += x_dim;
    C += hidden_dim_;
    H += hidden_dim_;
    C_diff += hidden_dim_;
    H_diff += hidden_dim_;
    X_diff += x_dim;
    C_prev_diff += hidden_dim_;
    ++flush;
  }
}`

maydaygmail:
@jeffdonahue captioner.py for generating sentence, to generate the current word, captioner.py only use the previous one word not all the previous words?

anguyen8:
Does any know if there is a pre-trained image captioning LRCN model out there? I'd greatly appreciate if this is included in the Model Zoo. 

@jeffdonahue : would you be able to release the model from your CVPR'15 paper?

anteagle:
Has this branch been landed to the master ? The layers are in the master, but it seems the examples are not there. Could anyone point to me to the right way to get this branch ?  I did git  pull #2033, but just showed Already up-to-date.

shaibagon:
@anteagle it seems like the PR only contained the LSTM RNN layers and not the examples (too much to review). You'll have to go to Jeff Donahue's "recurrent" branch.

anteagle:
@shaibagon thanks, I got from Jeff's repo, though it has not been updated for a while.

jeffdonahue:
Closing with the merge of #3948 -- though this PR still contains examples that PR lacked, and I should eventually restore and rebase those on the now merged version. In the meantime I'll keep my `recurrent` branch (and other mentioned branches) open and in their current form for reference.

jeffdonahue:
The methods you refer to are all automatically generated by protobuf.  See `caffe.proto` for the definitions of `top`, `bottom`, etc., which result in the protobuf compiler automatically generating the `add_top`, `add_bottom` methods.  (The resulting C definitions are in the protobuf-generated header file `caffe.pb.h`.)

yangzhikai:
oh , Thank you very much. I have not find this file(caffe.pb.h) because I haven't complied it before!

soulslicer:
Hi, is there any working example of the layer in caffe?

cuixing158:
The same question, is there any working example of the layer in caffe?

longjj:
@cuixing158 @soulslicer   jeffdonahue's example for coco image caption task.

https://github.com/BVLC/caffe/pull/2033#issuecomment-118142621

Go for his caffe branch and you will find the example `.prototxt` files and others which may be helpful.

Issue #1896, opened 2015-02-18
Allow self-contained development of Caffe layers


hannes-brt:
Judging from the "Development" page on the wiki, it seems that the only way to implement new Caffe layers is to fork the entire Caffe repository and implement the layer in the same manner as the standard layers. Only the actual implementation of the layer is self-contained in its own file, while the declarations go into the standard header files and the layer registration goes into the `caffe.proto` file which involves finding an unused ID for the layer.

As far as I can see there are several problems with this approach:
 - Researchers routinely develop new kinds of network layers and with the current approach they need to keep all of their custom layers in the same repository. Often a developer will work on several new layers simultaneously and it would make development of several layers much easier if they could be developed in a self-contained manner in separate repositories.
 - Developers who implement new types of layers will want to publish them on their own terms or often not publish them at all. If it is necessary to fork Caffe for every addition, then developers will have to always publish a complete fork, which might lead to a proliferation of Caffe forks.
 - When a developer of custom layers makes a fork of Caffe they will have to continuously merge new commits back into their own fork. When people publish their own forks and don't continuously merge new code, then there will be many Caffe forks that are out of sync and incompatible with each other. Most obviously, the official Caffe project might assign the protobuffer ID a developer has picked for their layer to a different layer which will create a continuous series of merge conflicts for the layer developer.
 - A user who would like to use custom layers created by different developers will have to integrate all of those layers into a single Caffe fork themselves which will routinely create more merge conflicts that the user has to resolve.

Suggestions:
I think it should be possible to design a system by which developers can create new layers completely separately from Caffe, compile them separately and then register their layer in some configuration file to make Caffe aware of its existence. Ideally, this could be done such as to not require recompilation of Caffe every time a new layer is added.
An obvious issue that will have to be solved is that `caffe.proto` must dynamically issue layer IDs based on availability. Possibly this could be solved by generating `caffe.proto` dynamically from a script and a templating engine to issue dynamic layer IDs. Using this approach a developer would create a short file with all the content that needs to go into `caffe.proto`, register this file with Caffe and then Caffe will use a script to stitch all those snippets together and issue a unique ID to every registered layer.

[...]

PR #2036, opened 2015-03-04
Use proto extensions with hashed field numbers for layer-specific parameters

jeffdonahue:
This is the proof-of-concept mentioned by @shelhamer in #1896.  **DO NOT MERGE** -- I'm not really sure if this is a good way to go, this PR is just for discussion for now.

This would facilitate self-contained independent development of layers by addressing one of the major current pain points with that -- conflicting proto field numbers.  The idea is that instead of picking a field number consecutively after the previous one, which of course causes conflicts whenever other new layers are added, as long as your field *name* (e.g., `convolution_param`) doesn't conflict with anything else in `LayerParameter`, the field number is generated as a function of the name, and (with high probability) you get a unique number.  (The collision odds for any given pair of extensions are 1 in 536850911; haven't done the birthday problem analysis to figure out what that actually means in practice, but if we were going to use this we might need to eventually add a "salt" field to escape collisions.)

See `src/caffe/proto/reshape_param.proto.varextnum` for an example of what it looks like to declare a LayerParameter extension this way.  The script `scripts/fill_proto_ext_num.py` creates the actual `reshape_param.proto` file from the `*.varextnum` version.

Build notes: there is currently something finicky about my Makefile changes such that parallel builds with `make -j` sometimes don't work, and I haven't attempted add CMake support.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,10 @@ EXAMPLE_SRCS := $(shell find examples -name "*.cpp")
 BUILD_INCLUDE_DIR := $(BUILD_DIR)/src
 # PROTO_SRCS are the protocol buffer definitions
 PROTO_SRC_DIR := src/$(PROJECT)/proto
-PROTO_SRCS := $(wildcard $(PROTO_SRC_DIR)/*.proto)
+PROTO_VAR_EXT_SRCS := $(wildcard $(PROTO_SRC_DIR)/*.proto.varextnum)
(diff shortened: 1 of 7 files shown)


jeffdonahue:
As hacky/scary of an idea as this might be, it's looking more attractive as it looks like we currently have at least 11 mutually unmergable PRs, which would probably all be safely mergable in any order if not for the fact that they all grab `LayerParameter` ID `132`. Thoughts?

bhack:
I think it is needed cause id in proto are actually a natural way to have cross PR conflicts. Is there also a commit of ReshapeLayer?

PR #2048, opened 2015-03-05
Dynamic setting of input batch size and support for multiple nets in Matlab wrapper

kencoken:
I'm proposing a pull request which introduces the following new functionality:

* The ability to override the first dimension of any input blob(s) by specifying the optional `force_input_count` parameter when initialising a network
* Extensions to the Matlab wrapper to allow multiple network instances to be maintained simultaneously (by keeping a hashmap of handles)

These are both prerequisites for an extended Matlab wrapper for mid-level feature computation using Caffe which I am planning to distribute soon, ported from the [deepeval-encoder package](http://www.robots.ox.ac.uk/~vgg/software/deep_eval/), with the intention of replacing it with a new package which builds upon caffe.

Please let me know if there is anything I can do to assist in the integration of these changes to the main Caffe branch, or if there is a more appropriate way in which to port these changes. In particular, am aware that the `force_input_count` parameter is quite specific, but nonetheless opted to issue a PR given that, since it is an optional argument, it does not change the existing API if not used. It may be that I maintain these changes in a separate forked repo, but thought I'd issue a PR just in case these changes were useful to others.

### DIFF
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -23,8 +23,10 @@ namespace caffe {
 template <typename Dtype>
 class Net {
  public:
-  explicit Net(const NetParameter& param, const size_t force_input_count = 0);
-  explicit Net(const string& param_file, Phase phase, const size_t force_input_count = 0);
+  explicit Net(const NetParameter& param,
+               const size_t force_input_count = 0);
+  explicit Net(const string& param_file, Phase phase,
+               const size_t force_input_count = 0);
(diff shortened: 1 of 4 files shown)


shelhamer:
Thanks for the pull request and apologies for the silence, but matcaffe has been updated altogether in #2505. The new interface includes multiple nets, reshaping, and more. If you see any improvements to make in the new interface, please do send a PR!

I hope the new interface is useful for your deepeval package.

PR #2052, opened 2015-03-06
Removed propagate labels check in loss layers

mtamburrano:
This is the new PR based on old #1483 which was not mergeable anymore

Old description:

> Removed checks in the loss layers as discussed in #1448 with @longjon (point 7).
> Now the loss layers don't check anymore if a backpropagation on labels is requested, this could now happen because #1482


mtamburrano:
this PR is no longer needed because #2095, so I close it

PR #2054, opened 2015-03-06
Filter layer rebased

mtamburrano:
This is the new PR based on old #1482 which was not mergeable anymore
The blobs are now treated as N dimensional entities.
To correctly deal with 0 sized batches (it could be happen when all the data is filtered) this PR needs #2053
To allow discriminating backprop this layer needs #2052

Possible use case:
[HD-CNN: Hierarchical Deep Convolutional Neural Network for Image Classification](http://arxiv.org/abs/1410.0736)

Old description:

> Filter_layer developed as discussed in #1448, implemented following the suggestions by @sguada and @longjon (paragraph 2(i)).

> This layer accepts 2+ bottom blobs, the first blob acts as a selector, so it should be a vector s={0,1,2, .. n} in [0,1], and the remaining blobs > are blobs to be filtered. Each value of 1 in the selector vector means that the items at the corresponding indices int the bottom_blobs[1+] will be kept, on the contrary a values of 0 means the items with corresponding symbols will be filtered out.
> So only non-filtered items will be forwarded and afterwards will receive backpropagation.

> As param the filter_layer needs a need_back_prop vector. This is needed because some blob could contains labels, and therefore will not > need backpropagation. The need_back_prop vector as the same size as the top vector, and accepts 1s and 0s depending on whether it should be backpropagated or not.

### DIFF
--- a/src/caffe/layers/filter_layer.cpp
+++ b/src/caffe/layers/filter_layer.cpp
@@ -1,0 +1,128 @@
+#include <algorithm>
+#include <vector>
+
+#include "caffe/layer.hpp"
+#include "caffe/util/math_functions.hpp"
+#include "caffe/vision_layers.hpp"
+
+namespace caffe {
+
+template <typename Dtype>
+void FilterLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  CHECK_EQ(top.size(), bottom.size()-1) <<
+      "Top.size() should be equal to bottom.size() - 1";
+  first_reshape_ = true;
+}
+
+template <typename Dtype>
+void FilterLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  // bottom[0...k-1] are the blobs to filter
+  // bottom[last] is the "selector_blob"
+  int selector_index = bottom.size() - 1;
+  for (int i = 1; i < bottom[selector_index]->num_axes(); ++i) {
+    CHECK_EQ(bottom[selector_index]->shape(i), 1)
+        << "Selector blob must have all shapes == 1 (except the first one)";
+  }
+  for (int i = 0; i < bottom.size()-1; i++) {
+    CHECK_EQ(bottom[selector_index]->shape(0), bottom[i]->shape(0)) <<
+        "Each bottom should have the same dimension as the selector blob";
+  }
+
+  const Dtype* bottom_data_selector = bottom[selector_index]->cpu_data();
+  indices_to_forward_.clear();
+
+  // look for non-zero elements in bottom[0]. Items of each bottom that
+  // have the same index as the items in bottom[0] with value == non-zero
+  // will be forwarded
+  for (int item_id = 0; item_id < bottom[selector_index]->shape(0); ++item_id) {
+    // we don't need an offset because item size == 1
+    const Dtype* tmp_data_selector = bottom_data_selector + item_id;
+    if (*tmp_data_selector) {
+      indices_to_forward_.push_back(item_id);
+    }
+  }
+  // only filtered items will be forwarded
+  int new_tops_num = indices_to_forward_.size();
+  // init
+  if (first_reshape_) {
+    new_tops_num = bottom[0]->shape(0);
+    first_reshape_ = false;
+  }
+  for (int t = 0; t < top.size(); t++) {
+    int num_axes = bottom[t]->num_axes();
+    vector<int> shape_top(num_axes);
+    shape_top[0] = new_tops_num;
+    for (int ts = 1; ts < num_axes; ts++)
+      shape_top[ts] = bottom[t]->shape(ts);
+    top[t]->Reshape(shape_top);
+  }
+}
+
+template <typename Dtype>
+void FilterLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  int new_tops_num = indices_to_forward_.size();
+  // forward all filtered items for all bottoms but the Selector (bottom[last])
+  for (int t = 0; t < top.size(); t++) {
+    const Dtype* bottom_data = bottom[t]->cpu_data();
+    Dtype* top_data = top[t]->mutable_cpu_data();
+    int dim = bottom[t]->count() / bottom[t]->shape(0);
+    for (int n = 0; n < new_tops_num; n++) {
+      int data_offset_top = top[t]->offset(n);
+      int data_offset_bottom =  bottom[t]->offset(indices_to_forward_[n]);
+      caffe_copy(dim, bottom_data + data_offset_bottom,
+          top_data + data_offset_top);
+    }
+  }
+}
+
+template <typename Dtype>
+void FilterLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
+      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom) {
+  if (propagate_down[bottom.size() - 1]) {
+    LOG(FATAL) << this->type()
+               << "Layer cannot backpropagate to filter index inputs";
+  }
+  for (int i = 0; i < top.size(); i++) {
+    // bottom[last] is the selector and never needs backpropagation
+    // so we can iterate over top vector because top.size() == bottom.size() -1
+    if (propagate_down[i]) {
+      const int dim = top[i]->count() / top[i]->shape(0);
+      int next_to_backward_offset = 0;
+      int batch_offset = 0;
+      int data_offset_bottom = 0;
+      int data_offset_top = 0;
+      for (int n = 0; n < bottom[i]->shape(0); n++) {
+        data_offset_bottom = bottom[i]->offset(n);
+        if (next_to_backward_offset >= indices_to_forward_.size()) {
+          // we already visited all items that were been forwarded, so
+          // just set to zero remaining ones
+          caffe_set(dim, Dtype(0),
+              bottom[i]->mutable_cpu_diff() + data_offset_bottom);
+        } else {
+          batch_offset = indices_to_forward_[next_to_backward_offset];
+          if (n != batch_offset) {  // this data was not been forwarded
+            caffe_set(dim, Dtype(0),
+                bottom[i]->mutable_cpu_diff() + data_offset_bottom);
+          } else {  // this data was been forwarded
+            data_offset_top = top[i]->offset(next_to_backward_offset);
+            next_to_backward_offset++;  // point to next forwarded item index
+            caffe_copy(dim, top[i]->mutable_cpu_diff() + data_offset_top,
+                bottom[i]->mutable_cpu_diff() + data_offset_bottom);
+          }
+        }
+      }
+    }
+  }
+}
+
+#ifdef CPU_ONLY
+STUB_GPU(FilterLayer);
+#endif
+
+INSTANTIATE_CLASS(FilterLayer);
+REGISTER_LAYER_CLASS(Filter);
+
+}  // namespace caffe
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -181,6 +181,69 @@ class EltwiseLayer : public Layer<Dtype> {
 };
 
 /**
+ * @brief Takes two+ Blobs, interprets last Blob as a selector and
+ *  filter remaining Blobs accordingly with selector data (0 means that
+ * the corresponding item has to be filtered, non-zero means that corresponding
+ * item needs to stay).
+ */
+template <typename Dtype>
+class FilterLayer : public Layer<Dtype> {
+ public:
+  explicit FilterLayer(const LayerParameter& param)
+      : Layer<Dtype>(param) {}
+  virtual void LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+  virtual void Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+
+  virtual inline const char* type() const { return "Filter"; }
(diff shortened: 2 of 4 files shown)


ducha-aiki:
>Possible use case:
HD-CNN: Hierarchical Deep Convolutional Neural Network for Image Classification

BTW, I have asked guys who wrote this paper about their plans releasing their caffe code. The answer was no, until paper got`s accepted.

bhack:
Another one with some possible relation to this PR: [Error-Driven Incremental Learning in Deep Convolutional Neural Network for Large-Scale Image Classification](http://research.microsoft.com/apps/mobile/publication.aspx?id=238746)

jeffdonahue:
> As param the filter_layer needs a need_back_prop vector. This is needed because some blob could contains labels, and therefore will not > need backpropagation. The need_back_prop vector as the same size as the top vector, and accepts 1s and 0s depending on whether it should be backpropagated or not.

This still doesn't make sense to me.  If your labels are not a function of the network weights, Caffe will already figure out that they do not need backpropagation.  This is how all of our current loss layers work -- Caffe figures out that the labels are a net "input" (i.e., unaffected by the weights) and hence we have  `d(labels)/dw = 0` and don't need to compute anything explicitly, thereby setting `propagate_down = false` for the labels.  On the other hand, if the learned weights of the net DO contribute to the labels that your layer takes as input, Caffe will *correctly* determine that we need to know how these labels change as we change the weights, hence propagate_down will be set to true for the labels in this case.

Maybe there is a use case for adding a proto field `skip_propagate_down` for certain bottoms when you don't care about backpropagation being mathematically correct, but it certainly seems to me that it shouldn't be called `need_back_prop` when the mathematical "need" for backprop is already inferred by Caffe.

ducha-aiki:
@jeffdonahue
Let me make an example, even not considering this filter layer. Imagine, I`d like to train three networks in one (they branch from, i.e. conv4 in alexnet). Their average accuracy is little bit better, that accuracy of each single network. So I use probabilities of the ensemble as additional optimization goal (with L2 loss). But I don`t want to network to move ensemmble prediction close to each individual. I want only otherwise.
Or - similar case - pseudolabel on unsupervised dataset. I use labels from argmax on ensebmle as true labels. Caffe determines that they depends on weights with layers and then tell me "Caanot propagate to label inputs" and dies. So this proto field "skip_back_prop" would solve this issues.

jeffdonahue:
Thanks for the example @ducha-alki -- it does seem useful to have a way of indicating that backprop should be truncated for particular bottoms of particular layers, or more generally a way to scale the bottom diff by a particular scale factor other than 1 (e.g., 0 if you want to truncate).  It would also be a convenient way to freeze layers of the net without having to set `lr_mult = 0` everywhere, which has been suggested before.

I am happy to review a separate PR which does the following:

* Add a new proto field to the top level of `message LayerParameter` that is either `repeated bool skip_propagate_down` or `repeated float bottom_diff_scale`, which may be specified either 0 times or N times, where N is the number of bottoms.
* If specified 0 times, `skip_propagate_down` would default to all false and `bottom_diff_scale` would default to all 1.0.
* In the case where `skip_propagate_down == true` or `bottom_diff_scale == 0` for some bottom(s), `Net::Init` would modify its `propagate_down` settings accordingly (potentially resulting in more than just the layer where `skip_propagate_down=true` was specified skipping backprop)
* Adds appropriate tests for the new behavior to `TestNet` (`src/caffe/test/test_net.cpp`)

For this PR, I can review it again when the `need_back_prop`-related stuff is removed.

mtamburrano:
thank you for your example @ducha-aiki.
To return on the specific case of `filter_layer`, the issue occurs when the labels are filtered according to the selector: because the `filter_layer` takes labels as input, `Caffe` thinks that the labels
contribute to weights computation and sets `propagate_down` to true, but they are just filtered and not influenced by successive computations, so they don't need backpropagation.
By the way, @jeffdonahue when you open the new PR with `skip_propagate_down`, I'll remove the redundant `need_back_prop` param

bhack:
I think that @jeffdonahue told that he want only review a separte PR with skip_propagate_down and not to write himself one. 
Actually we cannot remove need_back_prop param here without this new PR. So i want know from @jeffdonahue if we are stalled.

bhack:
@longjon Do you remember our discussion on this?

jeffdonahue:
> To return on the specific case of filter_layer, the issue occurs when the labels are filtered according to the selector: because the filter_layer takes labels as input, Caffe thinks that the labels contribute to weights computation and sets propagate_down to true, but they are just filtered and not influenced by successive computations, so they don't need backpropagation.

You are misunderstanding -- Caffe will *not* think your labels need backpropagation if they aren't the output (direct or indirect) of a layer with weights.  For example, if you have a data layer that directly produces the FilterLayer labels, FilterLayer will work fine without any `need_back_prop` or `skip_back_prop` hackery -- Caffe's `Net::Init` will correctly set `propagate_down = false`  for the label input.  On the other hand, if for example you take an input and multiply it by *learnable weights* then take an argmax to produce your FilterLayer labels, and FilterLayer's output influences a loss, then Caffe will correctly set `propagate_down = true` for the labels as the weights influence the loss through your FilterLayer.

It is important that if you put a layer in a network and it needs to be able to backpropagate (i.e., the loss is influenced by the weights via the layer), but can't, that Caffe will correctly complain and fail to train in this case.  However, I agree that sometimes it could be desirable to train the net in spite of that, as in @ducha-aiki's example and other cases, but such an option (1) shouldn't be called `need_back_prop` as that to me implies a lack of *mathematical* need for backprop when backprop may in fact be needed, and (2)  shouldn't have an ad-hoc implementation for one specific layer type as it creates redundancy (now have to listen to both `propagate_down` and `need_back_prop`) and it's equally useful for other layers.

So yes, I will wait to review this PR once the `need_back_prop` option is removed, regardless of whether that immediately satisfies your use case, and you or anyone else are welcome to also create the `skip_propagate_down` PR I described.

mtamburrano:
I agree with point 2, maybe a more general param could be useful in cases beyond this filter layer and to avoid redundancy the `need_back_prop` vector should be removed, but I still think that `filter_layer` needs either this vector or the new parameter you proposed.
Let's make an example:
We can have a net of this type, where the selector is obtained applying a `sigmoid_layer` followed by a `threshold_layer` to obtain the selctor for the `filter layer`:

![filter layer example]
(http://i1367.photobucket.com/albums/r795/jokulhaupssky/Untitled%20Diagram_zpsj9bxvszs.jpg)

In this case the `filter_layer` receives the output from `pooling_layer` and the labels directly from the `data_layer`. Both of these input needs to be filtered according to the third input, the selector, and the `filter_layer` outputs two filtered `top_blobs` (pooling and labels) for further computations.
`Caffe` correctly detects that `filter_layer` needs backpropagation and sets `propagate_down` to true for all of his `bottom_blobs`, but the label values don't need to be updated since are only used to calculate the loss and to be filtered and provided to further loss layers.

By the way me and @bhack are thinking about implementing this `skip_propagate_down` param, so we would like to clarify some doubt:
 - we should override the `propagate_down` values at the end of `Net::Init`?
 - can you suggest some test cases to be covered?

sguada:
Meanwhile you could have two filter layers, one that filter the
pooling_layer and another one that filter the labels. In that case one of
the filter layers will propagate_down and the other will not.

Sergio

2015-03-09 12:12 GMT-07:00 Manuele <notifications@github.com>:

bhack:
@sguada Is your comment the same of https://github.com/BVLC/caffe/pull/1482#issuecomment-67251914 and that @mtamburrano replied with https://github.com/BVLC/caffe/pull/1482#issuecomment-67348458? Is there something different?

sguada:
@bhack yeah my comment is the same, for the label part of filter one doesn't need to back-propagate while for the data part of filter it will back-propagate as needed.

I think you should try to implement the simplest filter layer possible and address problems in future PRs, otherwise this will never reach a merge-able point.

jeffdonahue:
@mtamburrano right,  I understand the use case, and your example makes sense.  Note that in your example you would add `skip_propagate_down: false skip_propagate_down: true` to the `SigmoidCrossEntropyLoss` layer definition to avoid the problem; `FilterLayer` itself should not need to be changed as Caffe should infer `propagate_down = false` for the labels if `skip_propagate_down: true` is set in the loss layer (though also adding `skip_propagate_down: true` for the labels in the `FilterLayer` would not affect anything).

> we should override the propagate_down values at the end of Net::Init?

It is probably easier/better to weave in the logic where Net is already setting `propagate_down`; you can likely use similar logic to the logic used to handle the `lr_mult: 0` cases not needing backprop.

> can you suggest some test cases to be covered?

How about the example you drew? You can turn it into a prototxt string, and add a new test case to `TestNet` (`src/caffe/test/test_net.cpp`) that checks `propagate_down == false` for FilterLayer's label input with `skip_propagate_down` set in SoftmaxWithLossLayer, and checks `propagate_down == true` without `skip_propagate_down`. (May have to add a public accessor to `Net` for `propagate_down` if there isn't one already.)

mtamburrano:
>How about the example you drew? You can turn it into a prototxt string, and add a new test case to TestNet (src/caffe/test/test_net.cpp) that checks propagate_down == false for FilterLayer's label input with skip_propagate_down set in SoftmaxWithLossLayer, and checks propagate_down == true without skip_propagate_down. (May have to add a public accessor to Net for propagate_down if there Isn't one already.)

I'd like to use my example for tests, but if I have to open a separate PR for the new parameter, I can't use this unmerged layer for running tests. Or rather, I can write the code, but the travis build will not succeed...

jeffdonahue:
The test doesn't need to use a `FilterLayer` specifically; the result should be the same regardless of layer type (so any layer that takes two inputs should work).  But if it's really an issue you can base the other PR on this one (thereby requiring that this one be merged first).

mtamburrano:
@jeffdonahue the new PR is live at #2095
tomorrow I'll remove the `need_back_prop` vector from this PR and probably #2052 is no more needed and can be closed

mtamburrano:
`need_back_prop` vector and all related stuff removed

bhack:

bhack:
@jeffdonahue Can you pass here?

bhack:
@mtamburrano Seems that It needs a rebase again.

bhack:
@jeffdonahue We want to program our activity to reserve a little bit of time for this. Do you need other changes or is it mergeable?

bhack:
@jeffdonahue Can we squash also this? Do you think is it ready?

src/caffe/proto/caffe.proto
```diff
@@ -512,6 +513,12 @@ message ExpParameter {
   optional float shift = 3 [default = 0.0];
 }
 
+// Message that stores parameters used by FilterLayer
+message FilterParameter {
+  // specify which blobs need to be backpropagated and which not
+  repeated uint32 need_back_prop = 1 [packed = true];
```
jeffdonahue: This shouldn't be needed after #2095 right?

mtamburrano:
removed `need_back_prop` param from Caffe.proto (I had already removed from the other files, but forgot that one) and merged again with Master.

@jeffdonahue I would like to remember that this PR now works, but to correctly treats with zero-sized batches, it still needs #2053 (although seems it is not only `filter_layer` to need it, but it applies to other cases, too)

bhack:
@jeffdonahue Can we try to merge  this before CVPR? We can allocate a little bit more time if you need some other little fixes.

mtamburrano:
rebased on master again.
@jeffdonahue, have you any idea when this PR could be merged? (and #2053 too)

bhack:
@jeffdonahue (@sguada amarcord) Since 18 Nov 2014 in https://github.com/BVLC/caffe/pull/1448. This PR it becoming barrel edition!

src/caffe/layers/filter_layer.cpp
```diff
@@ -0,0 +1,124 @@
...
+      int data_offset_bottom =  bottom[t]->offset(indices_to_forward_[n]);
+      caffe_copy(dim, bottom_data + data_offset_bottom,
+          top_data + data_offset_top);
+    }
+  }
+}
+
+template <typename Dtype>
+void FilterLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
+      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom) {
```
jeffdonahue: This should have something like `CHECK(!propagate_down[bottom.size() - 1]) << this->type() <<  "Layer cannot backpropagate to filter index inputs"` (like `SoftmaxWithLossLayer` for labels, etc.).

jeffdonahue:
Okay, please add a check as I commented above, squash, and I'll merge.

mtamburrano:
Check added, test updated and squashed all commits.

I would to point up again that without #2053, if a `Filter_Layer` filters all the input data for a given batch, it outputs a 0-sized batch that will cause a crash because `net.cpp` doesn't allow that.
This was discussed with @longjon [here](https://github.com/BVLC/caffe/pull/1448#issuecomment-64333009)(in particular see Point 5) and in the old PR #1484

mtamburrano:
seems like Travis fails first 2 builds because not every tests pass.
I don't know why, that tests should not regard my changes, I also tried to build the same branch on my repositories and everything works fine: https://travis-ci.org/mtamburrano/caffe/builds/64395042

jeffdonahue:
Not sure what happened but your diff is now huge.  Did you rebase on the latest master?

bhack:
@mtamburrabo The problem seems that now it is rebased on https://github.com/mtamburrano/caffe/commit/596add755cdd1178ffd3ca5598a64909b361739d but caffe master is at https://github.com/BVLC/caffe/commit/b12c17104cd3446b876346ea8342ecec8551fab4

mtamburrano:
I think I did something wrong when I squashed commits, everything should be fine now.
I'm sorry for the closed-reopen PR spam ;)

bhack:
@jeffdonahue Seems rebased correctly now. Can you pass here?

jeffdonahue:
Thanks @mtamburrano! Merged @ 21b6bf8.  I fixed a bug in the GPU backward implementation (gradient test was segfaulting) and did some minor comment/style cleanup @ 0ccf336.

PR #2056, opened 2015-03-06
CMAKE: Adding missing hdf5 include directory

philkr:
If ```hdf5.h``` is not located in ```/usr/include``` ```HDF5_INCLUDE_DIRS``` might not be set correctly, but ```HDF5_HL_INCLUDE_DIR``` still contains the correct directory.


shelhamer:
Thanks for the CMake work Philipp!

PR #2057, opened 2015-03-06
Allow python solver creation from prototxt string (not just file)

philkr:
A simple change that allows us to specify prototxt strings for solver settings from python, instead of just filenames. It makes it easier to train models using the new python interface.


shelhamer:
@philkr did you mean to close this? Seems like a reasonable `Solver` constructor.

philkr:
That was an accident, new version is in #2801.

Issue #2041, opened 2015-03-05
Python classifier.py crash when running classification.ipynb


dgmp88:
I was getting the following error running the classification.ipynb notebook on a fresh installation on OSX:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-7167c0e21af1> in <module>()
----> 1 prediction = net.predict([input_image])  # predict takes any number of images, and formats them for the Caffe net automatically
      2 print 'prediction shape:', prediction[0].shape
      3 plt.plot(prediction[0])
      4 print 'predicted class:', prediction[0].argmax()

/Users/george/Documents/caffe/python/caffe/classifier.py in predict(self, inputs, oversample)
     84             caffe_in[ix] = self.transformer.preprocess(self.inputs[0], in_)
     85         out = self.forward_all(**{self.inputs[0]: caffe_in})
---> 86         predictions = out[self.outputs[0]].squeeze(axis=(2,3))
     87 
     88         # For oversampling, average predictions across crops.

ValueError: 'axis' entry 2 is out of bounds [-2, 2)
```

which I was able to fix by deleting the arguments for squeeze so that 86 looks like this:

```python
predictions = out[self.outputs[0]].squeeze()
```

dgmp88:
Hmm, though doing this means that in the next section of classification.ipynb, when oversampling is switched off, the results don't make sense:

```python
print 'prediction shape:', prediction[0].shape
plt.plot(prediction[0])
print 'predicted class:', prediction[0].argmax()
```

prediction shape: ()
predicted class: 0

unless I remove the indexing:

```python
print 'prediction shape:', prediction.shape
plt.plot(prediction)
print 'predicted class:', prediction.argmax()
```

prediction shape: (1000,)
predicted class: 281

seanbell:
I think that the problem is that classification.ipynb hasn't yet been updated to match #1970 which was recently merged in.

Perhaps there should be a Python integration test that checks all of the steps in http://nbviewer.ipython.org/github/BVLC/caffe/blob/master/examples/classification.ipynb so that these kinds of bugs can be identified earlier in the future.  If Travis can't handle the model download, the test could just use random weights.

shelhamer:
@seanbell is right about the cause. We haven't come up with a good method for auto-checking the notebooks but these would be valuable integration checks that catch these kinds of issues. I suppose we could at least check they not crash.

[...]

Issue #1937, opened 2015-02-22
web demo and classify.py broken


EvanWeiner:
The python tools and examples are currently broken on master.

I'd like to suggest adding a test case for this before releasing future Caffe updates.

shelhamer:
> The python tools and examples are currently broken on master.

No, pycaffe and the examples actually work when installed according to the instructions as evidenced by `make pytest` checked by continuous integration and ourselves and individually checking the ipython notebook examples for release. Without any detail it's hard to know your problem. Remember to `make clean` between releases at least.

> I'd like to suggest adding a test case

Adding more Python tests -- in particular integration tests of small examples that run a network and check output -- is a good idea to not rely on manual checking. If you have a concrete idea for a test, please send a PR.

If you need installation / build help join the [caffe-users group](https://groups.google.com/forum/#!forum/caffe-users).

EvanWeiner:
@shelhamer - I kindly request this issue be re-visited because in Caffe's latest release much of the python code is indeed broken, here's the details:

1. The python-based web_demo in the examples folder is no longer working.
2. The classifiy.py script in the python folder is no longer working.
3. `sudo make pytest` fails for me after upgrading to latest code base and doing a make clean. Note that `sudo make runtest` does work fine for me after compiling the main source. Here's the error I'm getting for the failed `make pytest`:

F0223 08:14:56.364547  7935 layer_factory.hpp:77] Check failed: registry.count(type) == 1 (0 vs. 1) Unknown layer type: Python (known types: AbsVal, Accuracy, ArgMax, BNLL, Concat, ContrastiveLoss, Convolution, Data, Deconvolution, Dropout, DummyData, Eltwise, EuclideanLoss, Exp, Flatten, HDF5Data, HDF5Output, HingeLoss, Im2col, ImageData, InfogainLoss, InnerProduct, LRN, MVN, MemoryData, MultinomialLogisticLoss, Pooling, Power, ReLU, Sigmoid, SigmoidCrossEntropyLoss, Silence, Slice, Softmax, SoftmaxWithLoss, Split, TanH, Threshold, WindowData)
*** Check failure stack trace: ***
Aborted (core dumped)
make: *** [pytest] Error 134

shelhamer:
Thanks for coming back with detail. 1 & 2 can be looked into -- @sergeyk
could you check the web demo? -- and should likely be given pytests for
automatic checking.

3. there is a pytest for PythonLayer so if you install without it that test
[...]

PR #2058, opened 2015-03-07
Pycaffe fixes and example reformation

shelhamer:
Fix recent issues with pycaffe examples.

The `caffe.Net` interface and solver are the core of pycaffe. The `Classifier` and `Detector` "convenience" classes were a mistake on my part that seem to have mostly confused. To better illustrate the main interface this PR will switch to `caffe.Net` throughout (or at least take steps in that direction).

- [x] fix misc. issues #1937 #2041 
- [x] groom logreg + net surgery examples
- [x] include print output by @ariandyy 
- [x] switch to caffe.Net interface ~~everywhere~~ except for classification and detection examples

Next up

- switch to `caffe.Net` everywhere and deprecate `Classifier`, `Detector`
- pycaffe tour: demonstrate `caffe.Net`, `caffe.io`, solver interface
- Python net specification #1733 will make logreg + net surgery better
- Python layer: explain interface and show data and loss examples


python/classify.py
```diff
@@ -104,15 +104,20 @@ def main(argv):
 
     if args.gpu:
         print('GPU mode')
+    else:
+        print 'CPU mode'
```
lukeyeager: Print statements without the parentheses break the Python 3 compatibility that @philkr added in #1923.

As @longjon said, 
> Regarding future PRs, yes, let's try to maintain Python 3 compatibility as long as it's easy to do so.

shelhamer: Right -- thanks for the reminder. This was just a cherry-pick so I'm still making changes and combing over.

shelhamer:
@longjon I'm merging these fixes to the broken parts of pycaffe and examples but the real switch to `caffe.Net` everywhere is still to come.

PR #2064, opened 2015-03-07
Update readme.md for ImageNet example

drohit92:
In the resume training option.. the name of the solver state changed to 'caffenet_train_iter_10000.solverstate' from 'caffenet_train_10000.solverstate'


shelhamer:
Thanks for the proofreading.

PR #2066, opened 2015-03-07
Update leveldb include variable name to match FindLevelDB.cmake

caotto:
Currently the variable name used in cmake/Dependencies.cmake does not match the one set in FindLevelDB.cmake (LEVELDB_INCLUDE vs. LeveDB_INCLUDE), so if leveldb is not installed in a standard path cmake will fail.


jeffdonahue:
Thanks @caotto!

PR #2067, opened 2015-03-08
Thread-specific singleton (from @cypof)

longjon:
This is the first in a series of PRs pulling some work from the parallel branch/@cypof that's ready for immediate consideration, and that will be also be used by future PRs implementing orthogonal work on synchronous parallelism.

This PR switches from a single, global `Caffe` singleton to a per-thread singleton, and then performs the necessary circumlocutions to avoid boost/nvcc conflicts. This is both necessary and sufficient for most multi-device work. (Further refactoring to eliminate the singleton is possible, but this is a much simpler sufficient step.)

Using the same singleton in different threads is pretty certainly a bad idea, so I see no reason not to merge this change as soon it's verified to be in good working order.

@cypof, your other commits seem to be missing your name and have a different email from your github account. I used your github information for the commit I added here; please confirm that the author information is as you desire.

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -2,7 +2,6 @@
 #define CAFFE_COMMON_HPP_
 
 #include <boost/shared_ptr.hpp>
-#include <boost/thread.hpp>
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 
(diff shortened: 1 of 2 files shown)


jeffdonahue:
Closing -- merged in #2903

PR #2071, opened 2015-03-08
Correctly (re)encoding grayscale images as color

philkr:
ReadImageToDatum was not encoding gray scale images correctly. Thanks to @yanii for reporting this. If a dataset contains gray scale images the function now re-encodes those images as color images.


shelhamer:
@philkr is this fix still needed? I don't remember the state of encoding at the moment after your and @sguada's changes.

philkr:
I don't know. I figured if nobody needed it for 300 or so commit's maybe not...

PR #2072, opened 2015-03-09
Bugfix: final snapshot iter number is off by one

jeffdonahue:
Since #1228 the final snapshot has saved as `*_iter_N.{solverstate,caffemodel}` where `N = max_iter + 1`.  This fixes that.  @longjon


longjon:
Looks good, this logic is cleaner and more correct. Thanks @jeffdonahue!

PR #2073, opened 2015-03-09
Keep track of the layer graph

longjon:
Currently, `Net` keeps track of each layer's input and output blobs. However, since layers can be in-place, this information isn't sufficient to easily obtain the dependencies present in the computation graph.

This PR adds `layer_parents` and `layer_children` to `Net`. The parents of a layer (index) are the layers (indices) whose forwards must come before that layer's forward. The children of a layer are the layers whose backwards must come before that layer's backward.

Note that the hard distinction that we've maintained so far between backward and forward means that there are additional, untracked dependencies in the computation graph; namely, a layer's forward must come before its backward. I don't anticipate any serious consequences of this.

This graph will be used in a future PR, to allow correct multithreaded computation.

### DIFF
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -62,7 +62,8 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
   // set the input blobs
   for (int input_id = 0; input_id < param.input_size(); ++input_id) {
     const int layer_id = -1;  // inputs have fake layer ID -1
-    AppendTop(param, layer_id, input_id, &available_blobs, &blob_name_to_idx);
+    AppendTop(param, layer_id, input_id, &available_blobs, &blob_name_to_idx,
+        NULL);
   }
   DLOG(INFO) << "Memory required for data: " << memory_used_ * sizeof(Dtype);
   // For each layer, set up its input and output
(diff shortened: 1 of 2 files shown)


src/caffe/net.cpp
```diff
@@ -369,17 +383,21 @@ void Net<Dtype>::AppendTop(const NetParameter& param, const int layer_id,
 template <typename Dtype>
 int Net<Dtype>::AppendBottom(const NetParameter& param,
     const int layer_id, const int bottom_id,
-    set<string>* available_blobs, map<string, int>* blob_name_to_idx) {
+    set<string>* available_blobs, const map<string, int>& blob_name_to_idx,
+    const map<int, int>& top_blob_idx_to_layer) {
```
jeffdonahue: `available_blobs` should be last arg

jeffdonahue:
LGTM other than my style comment.  Ideally it'd have a test or two but I guess it can't break anything given that it's just adding some new member vars...

futurely:
There is a [very simple solution](https://github.com/zxie/nn/blob/master/nets/graph.py) to manage the forward and backward computation of the predecessors and successors in a [general flow graph](http://cs224d.stanford.edu/lectures/CS224d-Lecture5.pdf).

shelhamer:
Closing as #2219 was closed.

PR #2081, opened 2015-03-09
Clear all content out of dev except README.md

jeffdonahue:
Not sure if this is a good idea or if this has even been an issue, but it might be clearer that dev should no longer be used or PRed to if it has nothing in it besides the README.md saying to use master.  (Obviously, all of its content and commit history can still be accessed via git.)


shelhamer:
It seems there is sometimes confusion about dev still. Should this be merged to make the end of dev totally obvious?

bhack:
Probably it is better to delete the branch now.

lukeyeager:
I agree. Why do we care to keep the `devcommit history in this fork?

shelhamer:
@jeffdonahue @longjon should we delete dev? All the history did make it to master in the final merge so nothing is lost. The last few PRs to dev will close but we can mark them accordingly.

Thanks for commenting @bhack and @lukeyeager.

bhack:
All this have borderline history.

shelhamer:
We have decided to close the last dev PRs and delete the branch to make it clear that master is the base of all development.

Issue #966, opened 2014-08-22
N-d tiling (aka repmat)


petergehler:
I am looking for a N-d tiling operation (aka repmat in matlab) and did not find one in the current dev branch. The concat layer does provide some of its functionality already. Is there an easy way to stich it together with current layers already, otherwise I would create a new one based on the concat implementation. Any hints?

shelhamer:
Satisfied by #2083.

haihaoshen:
Hi, would you take a look TileLayer Usage #4422?

PR #2083, opened 2015-03-09
TileLayer

jeffdonahue:
This adds a layer to duplicate an input Blob `N` times along any axis.  AKA `repmat`.  (This is less general than either `np.tile` or `repmat` as those allow arbitrary tiling along all axes, but this satisfies a lot of use cases and could be generalized later, or even now, chained together with other `Tile` layers to get the more general behavior in a less convenient and probably less efficient way.)

### DIFF
--- a/src/caffe/layers/tile_layer.cpp
+++ b/src/caffe/layers/tile_layer.cpp
@@ -1,0 +1,62 @@
+#include <vector>
+
+#include "caffe/common_layers.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/util/math_functions.hpp"
+
+namespace caffe {
+
+template <typename Dtype>
+void TileLayer<Dtype>::Reshape(
+    const vector<Blob<Dtype>*>& bottom, const vector<Blob<Dtype>*>& top) {
+  const TileParameter& tile_param = this->layer_param_.tile_param();
+  axis_ = bottom[0]->CanonicalAxisIndex(tile_param.axis());
+  CHECK(tile_param.has_tiles()) << "Number of tiles must be specified";
+  tiles_ = tile_param.tiles();
+  CHECK_GT(tiles_, 0) << "Number of tiles must be positive.";
+  vector<int> top_shape = bottom[0]->shape();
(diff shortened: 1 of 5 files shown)


jeffdonahue:
Oh yeah, as alluded to in #966 this could already be done by a combination of `SplitLayer` with N tops and `ConcatLayer` with the corresponding N bottoms and `concat_param { axis: K }` where you'd set `tile_param { axis: K }` here.  I'm not actually sure if this is more efficient than the combination of those layers, but I think it's a bit easier and more natural to write...

src/caffe/layers/tile_layer.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+    bottom_diff += inner_dim_;
+  }
+}
+
+INSTANTIATE_LAYER_GPU_FUNCS(TileLayer);
+
+#endif  // #ifdef CPU_ONLY
+
+INSTANTIATE_CLASS(TileLayer);
```
myfavouritekk: `INSTANTIATE_LAYER_GPU_FUNCS(TileLayer);` and `INSTANTIATE_CLASS(TileLayer);` seem to cause duplicate explicit instantiation of `Forward_gpu` and `Backward_gpu`. I put the `else` block in a new `tile_layer.cu` and solved the problem.

jeffdonahue:
This is rebased and ready for review.  I also added CUDA kernels for the GPU implementation in the latter commit.

shelhamer:
@jeffdonahue this looks ready to me once it's rebased for the layer param ID since #2032 was merged. Feel free to update and merge.

jeffdonahue:
Rebased.  Thanks for reviewing @shelhamer!

artheru:
please add a feature, that take two blobs as input and the second blob is automatically tiled to the same shape as the first blob. e.g.,:

blobA: [16, 3, 240, 240],
blobB: [16, 3, 1, 1],

output blob: repmat(blobB, 1, 1, 240, 240)

haihaoshen:
Hi, would you please take a look at TileLayer Usage #4422

PR #2086, opened 2015-03-10
Python net specification

longjon:
`master` edition of #1733. Still rough, but should be usable by the intrepid.

* Now including an AlexNet (CaffeNet variant) generation example.
* Fixed an error in `uncamel` which broke acronym names (e.g., LRN) (thanks @sontran).
* Now supports fillers (thanks @sontran).

### DIFF
--- a/examples/pycaffe/caffenet.py
+++ b/examples/pycaffe/caffenet.py
@@ -1,0 +1,54 @@
+from caffe import layers as L, params as P, to_proto
+from caffe.proto import caffe_pb2
+
+# helper function for common structures
+
+def conv_relu(bottom, ks, nout, stride=1, pad=0, group=1):
+    conv = L.Convolution(bottom, kernel_size=ks, stride=stride,
+                                num_output=nout, pad=pad, group=group)
+    return conv, L.ReLU(conv, in_place=True)
+
+def fc_relu(bottom, nout):
+    fc = L.InnerProduct(bottom, num_output=nout)
+    return fc, L.ReLU(fc, in_place=True)
+
+def max_pool(bottom, ks, stride=1):
+    return L.Pooling(bottom, pool=P.Pooling.MAX, kernel_size=ks, stride=stride)
+
+def caffenet(lmdb, batch_size=256, include_acc=False):
+    data, label = L.Data(source=lmdb, backend=P.Data.LMDB, batch_size=batch_size, ntop=2,
+        transform_param=dict(crop_size=227, mean_value=[104, 117, 123], mirror=True))
+
+    # the net itself
+    conv1, relu1 = conv_relu(data, 11, 96, stride=4)
+    pool1 = max_pool(relu1, 3, stride=2)
+    norm1 = L.LRN(pool1, local_size=5, alpha=1e-4, beta=0.75)
+    conv2, relu2 = conv_relu(norm1, 5, 256, pad=2, group=2)
+    pool2 = max_pool(relu2, 3, stride=2)
+    norm2 = L.LRN(pool2, local_size=5, alpha=1e-4, beta=0.75)
+    conv3, relu3 = conv_relu(norm2, 3, 384, pad=1)
+    conv4, relu4 = conv_relu(relu3, 3, 384, pad=1, group=2)
+    conv5, relu5 = conv_relu(relu4, 3, 256, pad=1, group=2)
+    pool5 = max_pool(relu5, 3, stride=2)
+    fc6, relu6 = fc_relu(pool5, 4096)
+    drop6 = L.Dropout(relu6, in_place=True)
+    fc7, relu7 = fc_relu(drop6, 4096)
+    drop7 = L.Dropout(relu7, in_place=True)
+    fc8 = L.InnerProduct(drop7, num_output=1000)
+    loss = L.SoftmaxWithLoss(fc8, label)
+
+    if include_acc:
+        acc = L.Accuracy(fc8, label)
+        return to_proto(loss, acc)
+    else:
+        return to_proto(loss)
+
+def make_net():
+    with open('train.prototxt', 'w') as f:
+        print >>f, caffenet('/path/to/caffe-train-lmdb')
+
+    with open('test.prototxt', 'w') as f:
+        print >>f, caffenet('/path/to/caffe-val-lmdb', batch_size=50, include_acc=True)
+
+if __name__ == '__main__':
+    make_net()
--- a/python/caffe/__init__.py
+++ b/python/caffe/__init__.py
@@ -4,3 +4,4 @@ from .proto.caffe_pb2 import TRAIN, TEST
 from .classifier import Classifier
 from .detector import Detector
 from . import io
+from .net_spec import layers, params, NetSpec, to_proto
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -1,0 +1,203 @@
+"""Python net specification.
+
+This module provides a way to write nets directly in Python, using a natural,
+functional style. See examples/python_nets/caffenet.py for an example.
+
+Currently this works as a thin wrapper around the Python protobuf interface,
+with layers and parameters automatically generated for the "layers" and
+"params" pseudo-modules, which are actually objects using __getattr__ magic
+to generate protobuf messages.
+
+Note that when using to_proto or Top.to_proto, names of intermediate blobs will
+be automatically generated. To explicitly specify blob names, use the NetSpec
+class -- assign to its attributes directly to name layers, and call
+NetSpec.to_proto to serialize all assigned layers.
+
+This interface is expected to continue to evolve as Caffe gains new capabilities
+for specifying nets. In particular, the automatically generated layer names
+are not guaranteed to be forward-compatible.
+"""
+
+from collections import OrderedDict
+
+from .proto import caffe_pb2
+from google import protobuf
+
+
+def param_name_dict():
+    """Find out the correspondence between layer names and parameter names."""
+
+    layer = caffe_pb2.LayerParameter()
+    # get all parameter names (typically underscore case) and corresponding
+    # type names (typically camel case), which contain the layer names
+    # (note that not all parameters correspond to layers, but we'll ignore that)
+    param_names = [s for s in dir(layer) if s.endswith('_param')]
+    param_type_names = [type(getattr(layer, s)).__name__ for s in param_names]
+    # strip the final '_param' or 'Parameter'
+    param_names = [s[:-len('_param')] for s in param_names]
+    param_type_names = [s[:-len('Parameter')] for s in param_type_names]
+    return dict(zip(param_type_names, param_names))
+
+
+def to_proto(*tops):
+    """Generate a NetParameter that contains all layers needed to compute
+    all arguments."""
+
+    if not isinstance(tops, tuple):
+        tops = (tops,)
+    layers = OrderedDict()
+    autonames = {}
+    for top in tops:
+        top.fn._to_proto(layers, {}, autonames)
+    net = caffe_pb2.NetParameter()
+    net.layer.extend(layers.values())
+    return net
+
+
+def assign_proto(proto, name, val):
+    """Assign a Python object to a protobuf message, based on the Python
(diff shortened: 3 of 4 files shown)


longjon:
This should now support repeated `Message`s as `list`s of `dict`s (like `param` or `dummy_data_param`'s `shape`; @erictzeng, you asked about this earlier).

I think that means you can now specify any `NetParameter` as Python code. Once layer naming has been cleaned up a bit (for which I have an idea in mind), I think this will have reached mergeability as a thin wrapper around prototxt.

muupan:
uncamel('HDF5Data') wrongly returns 'hd_f5_data'.

muupan:
And uncamel('PReLU') wrongly returns 'p_relu'.

I think 'HDF5Data' -> 'hdf5_data' is normal uncamelling but 'PReLU' -> 'prelu' is exceptional. Maybe we have to find some other way.

seanbell:
@muupan The layers that break the rule could just be hardcoded in the "uncamel" function, e.g. with a dictionary.

Something like this:

```
_UNCAMEL_EXCEPTIONS = {
    'HDF5Data': 'hdf5_data',
    'PReLU': 'prelu',
}

def uncamel(s):
    """Convert CamelCase to underscore_case."""
    return _UNCAMEL_EXCEPTIONS.get(s, 
        re.sub('(?!^)([A-Z])(?=[^A-Z])', r'_\1', s).lower())
```

Shaunakde:
I am having an issue using this PR. Running the example: http://nbviewer.ipython.org/github/BVLC/caffe/blob/tutorial/examples/01-learning-lenet.ipynb gives me the following error:
```
 File "/home/shaunak/caffe-pr2086/examples/wine/classify.py", line 18, in lenet
    n = caffe.NetSpec()

  File "../../python/caffe/layers.py", line 84, in __init__
    super(NetSpec, self).__setattr__('tops', OrderedDict())

TypeError: must be type, not None
```

escorciav:
Hi @Shaunakde , you safe me hours of reading code so I felt that I should help you (sorry if you already noticed that). You should use the [comment](https://github.com/BVLC/caffe/pull/2086#issuecomment-94569931) of seanbell for your example. I guess that the reason is that ```HDF5Data``` is an CamelCase tricky layer.

Thank you longjon for this tool. I have spent hours debugging prototxt without noticing minor difference such as layers instead of layer.

examples/python_nets/caffenet.py
```diff
@@ -0,0 +1,54 @@
+from caffe import layers as L, params as P, to_proto
```
BlGene: I was looking over this and from what I can tell the to_proto function from PR #1733 has moved
into NetSpec [here](https://github.com/BVLC/caffe/pull/2086/files#diff-821d83c04a94134a26dae418beeb622eR92). 

In this case NetSpec should be imported and all layer variables should be defined to it. This would then also print the layers in their correct order.

The alexnet function would look as follows:
def alexnet(lmdb, batch_size=256, include_acc=False):
    n = NetSpec()

    n.data, n.label = L.Data(source=lmdb, backend=P.Data.LMDB, batch_size=batch_size, ntop=2,
        transform_param=dict(crop_size=227, mean_value=[104, 117, 123], mirror=True))

    # the net itself
    n.conv1, n.relu1 = conv_relu(n.data, 11, 96, stride=4)
    n.pool1 = max_pool(n.relu1, 3, stride=2)
    n.norm1 = L.LRN(n.pool1, local_size=5, alpha=1e-4, beta=0.75)
    n.conv2, n.relu2 = conv_relu(n.norm1, 5, 256, pad=2, group=2)
    n.pool2 = max_pool(n.relu2, 3, stride=2)
    n.norm2 = L.LRN(n.pool2, local_size=5, alpha=1e-4, beta=0.75)
    n.conv3, n.relu3 = conv_relu(n.norm2, 3, 384, pad=1)
    n.conv4, n.relu4 = conv_relu(n.relu3, 3, 384, pad=1, group=2)
    n.conv5, n.relu5 = conv_relu(n.relu4, 3, 256, pad=1, group=2)
    n.pool5 = max_pool(n.relu5, 3, stride=2)
    n.fc6, n.relu6 = fc_relu(n.pool5, 4096)
    n.drop6 = L.Dropout(n.relu6, in_place=True)
    n.fc7, n.relu7 = fc_relu(n.drop6, 4096)
    n.drop7 = L.Dropout(n.relu7, in_place=True)
    n.fc8 = L.InnerProduct(n.drop7, num_output=1000)
    n.loss = L.SoftmaxWithLoss(n.fc8, n.label)

    if include_acc:
        n.acc = L.Accuracy(n.fc8, n.label)

    return n.to_proto()

BR, Max

BlGene:
Hi,

I've been using this PR for a few days and I have been able to write all the models I wanted to using it. I am very pleased with it and Ty for writing it. :+1: @longjon 

The PR mentions that it fillers, but I don't see how these can be accessed from python, was this omitted from the PR?

I was wondering if there is an easy way to specify parameters for python layers, here a few ideas came to mind:
1. Just generating the python layer code with the variables in place
2. Extending PythonParameter in order to smuggle a dict through to the python layer:
```prototext
message PythonParameter {
  optional string module = 1;
  optional string layer = 2;
  repeated string param_keys = 3;
  repeated string param_values = 3;
}
```
So for this I was wondering if there is an easier way.

Also in case it was overlooked, its worthwhile looking at the Theano version of this, called [Mariana](http://bioinfo.iric.ca/~daoudat/Mariana/index.html).

bhack:
See also [Lasagne](http://lasagne.readthedocs.org/en/latest/) on Theano.

longjon:
Getting ready to update this; here's list of TODOs:

Cleanups and fixes before first mergeable version, which should be considered basically a "protobuf wrapper/generator":
- [ ] Try a better solution to the implicit layer <-> parameter correspondence (currently the messy `uncamel` function); the fact that the current protobuf system leaves this unspecified means there is no clean answer, but I have an idea for a hack that's a little more robust.
- [ ] Finish the transition to the `NetSpec` class (basically a namespace to avoid the previous atrocious hack of abusing `locals` (this was left as a WIP commit).
- [ ] Minimal tests.
- [ ] Minimal docs/examples (as begun).

Desiderata and future work to come after merge (PRs welcome!):
- [ ] Operator overloading for arithmetic.
- [ ] Update in accord with the solution to #1474.
- [ ] Make Python layer integration more natural (possibly also by updating the way  Python layers are specified in protobuf, in accord with #1685).

longjon:
TODOs all done for now; marking this ready for review!

@muupan @seanbell and others, `uncamel` is gone; the layer-parameter correspondence is now determined through inspection of the `caffe_pb2` module. The only assumption is that the parameter type of a layer named `X` is `XParameter`, which is true for all existing layers and should remain true.

@Shaunakde, note that we generally aren't able to provide support for PRs, especially in-progress ones. You're welcome to contribute to the development discussion, but otherwise please use caffe-users or other venues.

@BlGene, see the tests for an example of specifying fillers. Parameters for Python layers are a different (though related) issue; see, e.g., #2001.

examples/python_nets/caffenet.py
```diff
@@ -0,0 +1,54 @@
...
+    return conv, L.ReLU(conv, in_place=True)
+
+def fc_relu(bottom, nout):
+    fc = L.InnerProduct(bottom, num_output=nout)
+    return fc, L.ReLU(fc, in_place=True)
+
+def max_pool(bottom, ks, stride=1):
+    return L.Pooling(bottom, pool=P.Pooling.MAX, kernel_size=ks, stride=stride)
+
+def alexnet(lmdb, batch_size=256, include_acc=False):
```
shelhamer: Trivial, but this should be called `caffenet` since it has our usual inversion.

shelhamer:
@longjon this is great. Only comments are on naming: should we add an `examples/pycaffe` dir instead of `examples/python_nets`? I'm thinking that we should include example PythonLayers at some point too.

shelhamer:
@longjon amended names and the `(AttributeError, KeyError)` exception handling and pushed. Merging.

kjmonaghan:
Very helpful, thank you! 

How do I go about changing the decay_mult parameter?

jeffdonahue:
e.g., `L.Convolution(data, kernel_size=5, num_output=20, param=[dict(decay_mult=0.5)])`.   Any proto-generated object (in this case a `ParamSpec`) can be specified using a dict.  The brackets `[]` are needed to make a list of them, since `param` is a `repeated` field (to support multiple parameters). You could also explicitly use a `ParamSpec` object (rather than a `dict` that gets converted into one) by doing:

```
filter_spec = caffe_pb2.ParamSpec()
filter_spec.decay_mult = 0.5
L.Convolution(data, kernel_size=5, num_output=20, param=[filter_spec])
```

(untested, could be slightly wrong)

jmerkow:
Is it possible to specify dummy data with this?
I've tried about 1000 things to get it to work and can't seem to get it...

Based on caffenet.py, I would expect that it would be something like this:

```python
import caffe
from caffe import layers as L, params as P, to_proto
from caffe.proto import caffe_pb2
from __future__ import print_function
def gen_inputs():
    data = L.DummyData(name="data",ntop=1,shape=dict(dim=[1,2,3]))
    label = L.DummyData(name="label",ntop=1,shape=dict(dim=[1,2,3]))
    return data,label

def caffenet(include_acc=False):
    data,label=gen_inputs()

    loss = L.SoftmaxWithLoss(data, label)
    return to_proto(loss)

def make_net(output_dir='./',net_name='train'):
    fname = os.path.join(output_dir,net_name+'.prototxt')
    with open(fname, 'w') as f:
        print(caffenet(), file=f)
make_net()
```

this produces:

```
layer {
  name: "data"
  type: "DummyData"
  top: "DummyData1"
}
layer {
  name: "label"
  type: "DummyData"
  top: "DummyData2"
}
layer {
  name: "SoftmaxWithLoss1"
  type: "SoftmaxWithLoss"
  bottom: "DummyData1"
  bottom: "DummyData2"
  top: "SoftmaxWithLoss1"
}
```
I've tried various other things using caffe_pb2 and caffe.params.
And I tried assigning dummy_data_param explicitly and with dictionaries:

```python
def gen_inputs():
    data = L.DummyData(name="data",ntop=1,dummy_data_param=dict(shape=dict(dim=[1,2,3])))
    label = L.DummyData(name="label",ntop=1,dummy_data_param=dict(shape=dict(dim=[1,2,3])))
    return data,label
```

```python
def gen_inputs():
    data_shape = caffe_pb2.BlobShape()
    data_shape.dim = [1,2,3]
    data_param = caffe_pb2.DummyDataParameter()
    data_param.shape = data_shape
    data = L.DummyData(name="data",ntop=1,dummy_data_param=data_param)
    label = L.DummyData(name="label",ntop=1,dummy_data_param=data_param)
    return data,label
```

Any thoughts?
--Jameson

BlGene:
Maybe this?:
(Because shape is a repeated field in DummyDataParameter it should be given as a list.)

```python
from __future__ import print_function
import caffe
from caffe import layers as L, params as P, to_proto
from caffe.proto import caffe_pb2
import os

def gen_inputs():
    data  = L.DummyData(name="data", ntop=1,dummy_data_param=dict(shape=[dict(dim=[1,2,3])]))
    label = L.DummyData(name="label",ntop=1,dummy_data_param=dict(shape=[dict(dim=[1,2,3])]))
    return data,label

def caffenet(include_acc=False):
    data,label=gen_inputs()

    loss = L.SoftmaxWithLoss(data, label)
    return to_proto(loss)

def make_net(output_dir='./',net_name='train'):
    fname = os.path.join(output_dir,net_name+'.prototxt')
    with open(fname, 'w') as f:
        print(caffenet(), file=f)

make_net()
```

BR, Max

longjon:
@BlGene is right. If bad parameters are being silently ignored, however (in current `master`), that's a bug and you're welcome to open an issue.

jmerkow:
I can open it with this as a test case.  It may be the name 'shape' which has meaning in some contexts?

dfagnan:
@longjon  Does this Alexnet example get all the weight_filler and bias_fillers correct? I'm not able to easily see that the default weight_filler is somehow gaussian  with std = 0.01 here. Is this true?

PR #2082, opened 2015-03-09
FlattenLayer gets a FlattenParameter with an axis, end_axis

jeffdonahue:
This generalizes `FlattenLayer` in the spirit of how other layers (e.g. `InnerProductLayer`) were generalized in #1970.

PR #2088, opened 2015-03-10
Very simple version of ReshapeLayer

jeffdonahue:
(Cherry-picked from my RNN PR -- #2033)

This is a version of `ReshapeLayer` that just takes a `BlobShape` and reshapes its input to be that shape.  It doesn't have any of the options  from #108 or #1263 (infer special defaults when dim is set to  0 or 1 for some axes), but works with the new ND blobs (#1970), and I think the added options to `FlattenLayer` in #2082 satisfy a lot of the use cases of those options.  But I'm happy to see those options added later, or maybe this shouldn't be merged without them?  (I was originally in favor of replacing `FlattenLayer` with `ReshapeLayer` altogether, but the implementations are very light (all in  `SetUp`), and I now think it's probably more intuitive just to call a "flatten" layer than figuring out the special dimension values for reshape layer to get the equivalent output.)

### DIFF
--- a/src/caffe/layers/reshape_layer.cpp
+++ b/src/caffe/layers/reshape_layer.cpp
@@ -1,0 +1,26 @@
+#include <vector>
+
+#include "caffe/common_layers.hpp"
+#include "caffe/layer.hpp"
+
+namespace caffe {
+
+template <typename Dtype>
+void ReshapeLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  top[0]->Reshape(this->layer_param_.reshape_param().shape());
+  top[0]->ShareData(*bottom[0]);
+  top[0]->ShareDiff(*bottom[0]);
+}
+
+template <typename Dtype>
+void ReshapeLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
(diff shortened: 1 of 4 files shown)


src/caffe/layers/reshape_layer.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "caffe/common_layers.hpp"
+#include "caffe/layer.hpp"
+
+namespace caffe {
+
+template <typename Dtype>
+void ReshapeLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  top[0]->Reshape(this->layer_param_.reshape_param().shape());
```
jeffdonahue: @longjon I guess this should technically all be done in `LayerSetUp` since the top shape doesn't vary with the bottom shape?  (But then if special options were added when certain dims of the shape are set to 0 or -1, the top would then vary with the bottom shape, so it would be moved to `Reshape` then?  Maybe better to keep it in `Reshape` then?)

longjon: Right, to me it feels like this should go in `LayerSetUp`. If special options were added so that top shape depended on bottom shape, then `Reshape` would be needed. But that isn't the case in this code.

jeffdonahue: Thanks, fixed.  I kept the `CHECK_EQ` for `count`s in `Reshape` though, which still seems right.

jeffdonahue:
replaced by #2217

PR #2089, opened 2015-03-10
ReductionLayer

jeffdonahue:
Performs a "reduction" operation (currently `SUM`, `MEAN`, `ASUM` for sum of absolute values, and `SUMSQ` for sum of squares) to turn a number of "tail" axes into a single scalar value.  The `MEAN` operation, in combination with a `loss_weight`, is useful for creating custom losses that don't have an obnoxious amount of output.  For example, this `EuclideanLoss`:
```
layer {
  type: "EuclideanLoss"
  bottom: "preds"
  bottom: "targets"
  top: "mean_squared_error"
  loss_weight: 1
}
```

...is equivalent to this `Reduction`:
```
layer {
  type: "Eltwise"
  bottom: "preds"
  bottom: "targets"
  top: "error"
  eltwise_param {
    operation: SUM
    coeff: 1
    coeff: -1
  }
}
layer {
  type: "Reduction"
  bottom: "error"
  top: "sum_squared_error_per_instance"
  reduction_param { operation: SUMSQ axis: 1 }
}
layer {
  type: "Reduction"
  bottom: "sum_squared_error_per_instance"
  top: "mean_squared_error"
  reduction_param { operation: MEAN axis: 0 coeff: 0.5 }
  loss_weight: 1
}
```
 (would be more efficient to do as a single `Reduction` with `SUM_OF_SQUARES` and a certain `coeff` setting, but then you have to compute the batch size and everything is less pretty...)

Eventually, this should support reduction along inner axes (e.g. support an `end_axis`), but that makes the implementation substantially more involved than this...

### DIFF
--- a/src/caffe/layers/reduction_layer.cpp
+++ b/src/caffe/layers/reduction_layer.cpp
@@ -1,0 +1,132 @@
+#include <algorithm>
+#include <cfloat>
+#include <vector>
+
+#include "caffe/layer.hpp"
+#include "caffe/util/math_functions.hpp"
+#include "caffe/vision_layers.hpp"
+
+namespace caffe {
+
+template <typename Dtype>
+void ReductionLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  op_ = this->layer_param_.reduction_param().operation();
+}
+
+template <typename Dtype>
+void ReductionLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  axis_ = bottom[0]->CanonicalAxisIndex(
+      this->layer_param_.reduction_param().axis());
+  // In the output, we'll keep all axes up to the reduction axis, but
+  // throw away any after that.
+  // Note: currently reducing along non-tail axes is not supported; otherwise,
+  // we'd need to also copy any axes following an "end_axis".
+  vector<int> top_shape(bottom[0]->shape().begin(),
+                        bottom[0]->shape().begin() + axis_);
+  top[0]->Reshape(top_shape);
+  num_ = bottom[0]->count(0, axis_);
+  dim_ = bottom[0]->count(axis_);
+  CHECK_EQ(num_, top[0]->count());
+  if (op_ == ReductionParameter_ReductionOp_SUM ||
+      op_ == ReductionParameter_ReductionOp_MEAN) {
+    vector<int> sum_mult_shape(1, dim_);
+    sum_multiplier_.Reshape(sum_mult_shape);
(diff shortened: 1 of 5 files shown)


src/caffe/layers/reduction_layer.cpp
```diff
@@ -0,0 +1,131 @@
...
+  for (int i = 0; i < num_; ++i) {
+    switch (op_) {
+    case ReductionParameter_ReductionOp_SUM:
+    case ReductionParameter_ReductionOp_MEAN:
+      *top_data = caffe_cpu_dot(dim_, mult_data, bottom_data);
+      break;
+    case ReductionParameter_ReductionOp_ASUM:
+      *top_data = caffe_cpu_asum(dim_, bottom_data);
+      break;
+    case ReductionParameter_ReductionOp_SUM_OF_SQUARES:
```
shelhamer: `SUM_OF_SQUARES` is a little unlike the other op names. `SUMSQ` fits in with `ASUM`. Your pick.

src/caffe/test/test_reduction_layer.cpp
```diff
@@ -0,0 +1,209 @@
...
+    delete blob_bottom_;
+    delete blob_top_;
+  }
+
+  void TestForward(ReductionParameter_ReductionOp op, float coeff = 1) {
+    LayerParameter layer_param;
+    ReductionParameter* reduction_param = layer_param.mutable_reduction_param();
+    reduction_param->set_operation(op);
+    if (coeff != 1.0) {
+      reduction_param->set_coeff(2.3);
```
shelhamer: `reduction_param->set_coeff(coeff);` might be more robust for cases when `coeff` != 2.3.

shelhamer:
Once the comments are addressed this looks fine, although there could be a test for reducing over a tail that isn't all dimensions.

p.s. N-D blobs are nice. `count()` and `CanonicalAxisIndex()`make for more obvious code.

src/caffe/test/test_reduction_layer.cpp
jeffdonahue: er, yeah, thanks, not sure what I was thinking when I wrote this...

src/caffe/layers/reduction_layer.cpp
jeffdonahue: fair enough -- I'll change it to `SUMSQ`

jeffdonahue:
Thanks for all the reviews @shelhamer! Tests for non-zero axis added; will merge after Travis.

jeffdonahue:
Wow... especially thanks for the suggestion to test other axes. I had a bug in Backward for `asum` and `sumsq` (wasn't incrementing `bottom_data`).  Should be good now...

shelhamer:
Tests are more trustworthy than me. Good catch!

PR #2090, opened 2015-03-10
LogLayer

jeffdonahue:
This adds `LogLayer`, a `NeuronLayer`, which by default takes the natural log of its inputs.  It's designed analogously to `ExpLayer` and `PowerLayer`.  (In general computes `log_{\gamma}(\alpha x + \beta) ` with `log_param { base: \gamma scale: \alpha shift: \beta }`.)

### DIFF
--- a/include/caffe/util/mkl_alternate.hpp
+++ b/include/caffe/util/mkl_alternate.hpp
@@ -33,6 +33,7 @@ extern "C" {
 
 DEFINE_VSL_UNARY_FUNC(Sqr, y[i] = a[i] * a[i]);
 DEFINE_VSL_UNARY_FUNC(Exp, y[i] = exp(a[i]));
+DEFINE_VSL_UNARY_FUNC(Ln, y[i] = log(a[i]));
 DEFINE_VSL_UNARY_FUNC(Abs, y[i] = fabs(a[i]));
 
 // A simple way to define the vsl unary functions with singular parameter b.
--- a/include/caffe/neuron_layers.hpp
+++ b/include/caffe/neuron_layers.hpp
@@ -268,6 +268,72 @@ class ExpLayer : public NeuronLayer<Dtype> {
 };
 
 /**
+ * @brief Computes @f$ y = log_{\gamma}(\alpha x + \beta) @f$,
+ *        as specified by the scale @f$ \alpha @f$, shift @f$ \beta @f$,
+ *        and base @f$ \gamma @f$.
+ */
+template <typename Dtype>
+class LogLayer : public NeuronLayer<Dtype> {
+ public:
+  /**
+   * @param param provides LogParameter log_param,
+   *     with LogLayer options:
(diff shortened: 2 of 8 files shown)


pannous:
can be confused with "LoggingLayer" 
Other names are verbose as well, suggestion: LogarithmLayer

longjon:
@pannous I do think `LogLayer` is the best name here; we have `ExpLayer`, not `ExponentiationLayer`, `log` is the commonly used mathematical abbreviation, Caffe has distinct mechanisms for logging and an existing nomenclature for output layers.

include/caffe/util/mkl_alternate.hpp
```diff
@@ -33,6 +33,7 @@ extern "C" {
 
 DEFINE_VSL_UNARY_FUNC(Sqr, y[i] = a[i] * a[i]);
 DEFINE_VSL_UNARY_FUNC(Exp, y[i] = exp(a[i]));
+DEFINE_VSL_UNARY_FUNC(Ln, y[i] = log(a[i]));
```
longjon: I kinda prefer the name `log` here, sticking with the name of the elementwise call, but either one is okay...

longjon:
This looks good to me as a counterpart to `ExpLayer`.

As a more general issue affecting both of these, however, I wonder if this is the best way to resolve the tension between granularity and functionality. As far as I can tell, these layers could just implement simple `exp` and `log`, with scale and shift factored out (in-place, if desired), and with base changes factored out, as that's how the implementations work already...

include/caffe/util/mkl_alternate.hpp
jeffdonahue: not my choice -- the MKL function is `vsLn`/`vdLn`.

longjon: Oh. :(

jeffdonahue:
@longjon yeah, I put the shift/scale fields in to match the existing `Exp` and `Power` layers and because in my own usage I often end up wanting them, but, especially with your Python layer specs making it more convenient to do compositions of  simpler layers, it might be time to factor them out (i.e. create `ShiftLayer` and `ScaleLayer` and remove those fields from `Exp`, `Power`, and `Log`).  I guess my preference would be to merge this as it's consistent with what we have now, then follow up with (or let someone else follow up with...) that refactoring later, but maybe it's better to not add to the existing mess -- I'll leave that up to you.

shelhamer:
Looks good to me.

> it might be time to factor them out (i.e. create ShiftLayer and ScaleLayer and remove those fields from Exp, Power, and Log). I guess my preference would be to merge this as it's consistent with what we have now, then follow up with (or let someone else follow up with...) that refactoring later

This is fine. It could make for a nice warm-up PR in the future.

jeffdonahue:
Thanks for the review @shelhamer and @longjon!

PR #2095, opened 2015-03-10
Added param skip_propagate_down to LayerParameter

mtamburrano:
Added `skip_propagate_down` parameter as discussed with @jeffdonahue in #2054.

The parameter can be specified either 0 or N times, where N is the number of bottom blobs.
if `skip_propagate_down[i] == True` then nothing will be backpropagated to `bottom[i]`.
If `force_backward` is specified, then `skip_propagate_down` has no effect.

Tests are included in this PR in `test.net.cpp`

### DIFF
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -137,6 +137,9 @@ class Net {
   inline const vector<Dtype>& blob_loss_weights() const {
     return blob_loss_weights_;
   }
+  inline const vector<bool>& layer_need_backward() const {
+    return layer_need_backward_;
+  }
   /// @brief returns the parameters
   inline const vector<shared_ptr<Blob<Dtype> > >& params() const {
     return params_;
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -79,10 +79,17 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
     }
     // Setup layer.
     const LayerParameter& layer_param = param.layer(layer_id);
+    if (layer_param.propagate_down_size() > 0) {
+      CHECK_EQ(layer_param.propagate_down_size(),
+          layer_param.bottom_size())
+          << "propagate_down param must be specified "
+          << "either 0 or bottom_size times ";
+    }
     layers_.push_back(LayerRegistry<Dtype>::CreateLayer(layer_param));
     layer_names_.push_back(layer_param.name());
     LOG(INFO) << "Creating Layer " << layer_param.name();
     bool need_backward = false;
+
     // Figure out this layer's input and output
     for (int bottom_id = 0; bottom_id < layer_param.bottom_size();
          ++bottom_id) {
@@ -151,15 +158,33 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
   // Go through the net backwards to determine which blobs contribute to the
   // loss.  We can skip backward computation for blobs that don't contribute
   // to the loss.
+  // Also checks if all bottom blobs don't need backward computation (possible
+  // because the skip_propagate_down param) and so we can skip bacward
+  // computation for the entire layer
   set<string> blobs_under_loss;
+  set<string> blobs_skip_backp;
   for (int layer_id = layers_.size() - 1; layer_id >= 0; --layer_id) {
     bool layer_contributes_loss = false;
+    bool layer_skip_propagate_down = true;
     for (int top_id = 0; top_id < top_vecs_[layer_id].size(); ++top_id) {
       const string& blob_name = blob_names_[top_id_vecs_[layer_id][top_id]];
       if (layers_[layer_id]->loss(top_id) ||
           (blobs_under_loss.find(blob_name) != blobs_under_loss.end())) {
         layer_contributes_loss = true;
+      }
+      if (blobs_skip_backp.find(blob_name) == blobs_skip_backp.end()) {
+        layer_skip_propagate_down = false;
+      }
+      if (layer_contributes_loss && !layer_skip_propagate_down)
         break;
+    }
+    // If this layer can skip backward computation, also all his bottom blobs
+    // don't need backpropagation
+    if (layer_need_backward_[layer_id] && layer_skip_propagate_down) {
+      layer_need_backward_[layer_id] = false;
+      for (int bottom_id = 0; bottom_id < bottom_vecs_[layer_id].size();
+               ++bottom_id) {
+        bottom_need_backward_[layer_id][bottom_id] = false;
       }
     }
     if (!layer_contributes_loss) { layer_need_backward_[layer_id] = false; }
@@ -177,6 +202,11 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
         blobs_under_loss.insert(blob_name);
       } else {
         bottom_need_backward_[layer_id][bottom_id] = false;
+      }
+      if (!bottom_need_backward_[layer_id][bottom_id]) {
+        const string& blob_name =
+                   blob_names_[bottom_id_vecs_[layer_id][bottom_id]];
+        blobs_skip_backp.insert(blob_name);
       }
     }
   }
@@ -367,9 +397,9 @@ void Net<Dtype>::AppendTop(const NetParameter& param, const int layer_id,
 
 // Helper for Net::Init: add a new bottom blob to the net.
 template <typename Dtype>
-int Net<Dtype>::AppendBottom(const NetParameter& param,
-    const int layer_id, const int bottom_id,
-    set<string>* available_blobs, map<string, int>* blob_name_to_idx) {
+int Net<Dtype>::AppendBottom(const NetParameter& param, const int layer_id,
+    const int bottom_id, set<string>* available_blobs,
+    map<string, int>* blob_name_to_idx) {
   const LayerParameter& layer_param = param.layer(layer_id);
   const string& blob_name = layer_param.bottom(bottom_id);
   if (available_blobs->find(blob_name) == available_blobs->end()) {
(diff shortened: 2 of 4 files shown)


bhack:

bhack:
@jeffdonahue Can you pass here?

include/caffe/net.hpp
```diff
@@ -192,7 +192,8 @@ class Net {
   /// @brief Append a new bottom blob to the net.
   int AppendBottom(const NetParameter& param, const int layer_id,
                    const int bottom_id, set<string>* available_blobs,
-                   map<string, int>* blob_name_to_idx);
+                   map<string, int>* blob_name_to_idx,
+                   bool skip_propagate = false);
```
jeffdonahue: There's no reason to make `skip_propagate` a new input -- just get it from `layer_param` in `AppendBottom`. (I don't think there should be any changes to `Net::Init` if you do it this way, but I might have overlooked something.)

src/caffe/test/test_net.cpp
```diff
@@ -613,6 +615,103 @@ class NetTest : public MultiDeviceTest<TypeParam> {
...
+      "  bottom: 'fake_labels' "
+      "  top: 'label_argmax' "
+      "  type: 'ArgMax' "
+      "} "
+      "layer { "
+      "  name: 'loss' "
+      "  bottom: 'innerproduct' "
+      "  bottom: 'label_argmax' ";
+    if (test_skip_true)
+      proto += "  skip_propagate_down: [false, true] ";
```
jeffdonahue: Huh, I didn't know you could specify repeated fields this way. Neat.

src/caffe/test/test_net.cpp
```diff
@@ -55,6 +55,8 @@ class NetTest : public MultiDeviceTest<TypeParam> {
     }
   }
 
+
+
```
jeffdonahue: remove whitespace

src/caffe/test/test_net.cpp
```diff
@@ -2224,4 +2323,31 @@ TYPED_TEST(NetTest, TestReshape) {
...
+      // access to bottom_need_backward coresponding to label's blob
+      bool need_back = this->net_->bottom_need_backward()[layer_id][1];
+      // if skip_propagate_down is false, the loss layer will try to
+      // backpropagate on labels
+      CHECK_EQ(need_back, true)
+        << "bottom_need_backward should be True";
+    }
+  }
+  // check bottom_need_backward if skip_propagat_down is true
+  this->InitSkipPropNet(true);
```
jeffdonahue: This should also check that the `Net`'s `layer_need_backward_` (not sure if this is currently exposed; feel free to add a public accessor for it if not) is set to false for all layers below the loss (and the first loop should check that all `layer_need_backward`s are true, except for the data layer).

src/caffe/proto/caffe.proto
```diff
@@ -280,6 +280,10 @@ message LayerParameter {
 
   // The blobs containing the numeric parameters of the layer.
   repeated BlobProto blobs = 7;
+  
+  // Specifies on which bottoms the backpropagation should be skipped.
+  // The size must be either 0 or equals to the number of bottoms.
+  repeated bool skip_propagate_down = 11;
```
jeffdonahue: I know I suggested this name (sorry), but now I think I was wrong and the field name should just be `propagate_down` (thereby inverting the values it takes now) and it should default to the net's inferred settings if unspecified. (`propagate_down` is consistent with the current code and "`false` means skip" is more intuitive than "`true` means skip" to me.)

include/caffe/net.hpp
mtamburrano: I can change it if you want, but in this way [this check](https://github.com/mtamburrano/caffe/blob/fe4dad53e65574573614f600d50aad06c96b8739/src/caffe/net.cpp#L83-L88) needs to be done in the `AppendBottom` function, that it is called `Bottom.size()` times.
I know that this is not a computational relevant part, but don't you think that will be a little bit dirty?

jeffdonahue: Leaving the check in `Net::Init` is fine (though I'd also be fine with moving it to `AppendBottom`).  The rest of the logic is more natural to move to `AppendBottom` (and its signature shouldn't change) since it's associated specifically with a new bottom blob, and `Net::Init` is already longer than a single method should be.

mtamburrano: Ok, I'll move the param retrieval into `AppendBottom` and I'll leave the check into `Net::Init`

src/caffe/test/test_net.cpp
mtamburrano: Examining the code again, I'm not sure `layer_need_backward` is properly updated, although the `bottom_need_backward` is exact for each layers. Tomorrow I'll try to reflect the changes in `layer_need_backward` too

src/caffe/proto/caffe.proto
mtamburrano: I hate you, but ok ;)

mtamburrano:
@jeffdonahue I apported the changes you suggested:
- `AppendBottom` doesn't take anymore additional params
- At the end of `Net::Init`, when the graph is visited top-down, `layer_need_backward_` is updated to be consistent with `bottom_need_backward_`. So if all the top blobs of a Layer don't need backprop, the Layer itself doesn't need back prop and his parents will notice it, to skip backprop themselves if each of their children will not propagate anything.
In this way if a Layer has two children, with one of them able to backprop and the other not, the Layer still  will be correctly needing backprop

mtamburrano:
hi @jeffdonahue, any news?

src/caffe/test/test_net.cpp
```diff
@@ -2224,4 +2321,52 @@ TYPED_TEST(NetTest, TestReshape) {
...
+  this->InitSkipPropNet(false);
+  vector<bool> vec_layer_need_backward = this->net_->layer_need_backward();
+  for (int layer_id = 0; layer_id < this->net_->layers().size(); ++layer_id) {
+    if (this->net_->layer_names()[layer_id] == "loss") {
+      // access to bottom_need_backward coresponding to label's blob
+      bool need_back = this->net_->bottom_need_backward()[layer_id][1];
+      // if propagate_down is true, the loss layer will try to
+      // backpropagate on labels
+      CHECK_EQ(need_back, true)
+        << "bottom_need_backward should be True";
```
jeffdonahue: All the `CHECK_EQ(.., true)`s in this test should be `EXPECT_TRUE` (and `CHECK_EQ(..., false)` -> `EXPECT_FALSE`) -- the tests should fail but not die if the behavior being tested is incorrect.

src/caffe/test/test_net.cpp
```diff
@@ -2224,4 +2321,52 @@ TYPED_TEST(NetTest, TestReshape) {
...
+    }
+    if (this->net_->layer_names()[layer_id] == "ip_fake_labels")
```
jeffdonahue: How about just `EXPECT_TRUE(vec_layer_need_backward[layer_id])` regardless of the layer in this case?

src/caffe/test/test_net.cpp
```diff
@@ -2224,4 +2321,52 @@ TYPED_TEST(NetTest, TestReshape) {
...
+  for (int layer_id = 0; layer_id < this->net_->layers().size(); ++layer_id) {
+    if (this->net_->layer_names()[layer_id] == "loss") {
+      // access to bottom_need_backward coresponding to label's blob
+      bool need_back = this->net_->bottom_need_backward()[layer_id][1];
+      // if propagate_down is false, the loss layer will not try to
+      // backpropagate on labels
+      CHECK_EQ(need_back, false)
+        << "bottom_need_backward should be False";
+    }
+    if (this->net_->layer_names()[layer_id] == "ip_fake_labels")
```
jeffdonahue: and in this case, how about `if (... == "ip_fake_labels") { EXPECT_FALSE(...) } else if { ... == "argmax" } { EXPECT_FALSE(...) } else { EXPECT_TRUE(...) }` to check all the layers?

src/caffe/net.cpp
```diff
@@ -381,7 +411,12 @@ int Net<Dtype>::AppendBottom(const NetParameter& param,
   bottom_vecs_[layer_id].push_back(blobs_[blob_id].get());
   bottom_id_vecs_[layer_id].push_back(blob_id);
   available_blobs->erase(blob_name);
-  const bool need_backward = blob_need_backward_[blob_id];
+  bool propagate_down = true;
+  // Check if the backpropagation on bottom_id should be skipped
+  if (layer_param.propagate_down_size() > 0)
+    propagate_down = layer_param.propagate_down(bottom_id);
+  const bool need_backward = blob_need_backward_[blob_id] &
```
jeffdonahue: nitpick: use double ampersand `&&`

src/caffe/net.cpp
```diff
@@ -79,10 +79,17 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
     }
     // Setup layer.
     const LayerParameter& layer_param = param.layer(layer_id);
+    if (layer_param.propagate_down_size() > 0) {
+      CHECK_EQ(layer_param.propagate_down_size(),
+          layer_param.bottom_size())
+      << "propagate_down param must be specified"
```
jeffdonahue: this and next line should have 4-space indents, and next line string should start with a space (`" either`)

src/caffe/proto/caffe.proto
```diff
@@ -280,6 +280,10 @@ message LayerParameter {
 
   // The blobs containing the numeric parameters of the layer.
   repeated BlobProto blobs = 7;
+  
+  // Specifies on which bottoms the backpropagation should be skipped.
+  // The size must be either 0 or equals to the number of bottoms.
```
jeffdonahue: `equals` -> `equal`

bhack:
@jeffdonahue We want to program our activity to reserve a little bit of time for this. Do you need other changes or is it mergeable?

jeffdonahue:
LGTM -- please squash to a single commit and I'll merge.  Thanks @mtamburrano!

bhack:
Done

jeffdonahue:
Thanks @bhack and @mtamburrano!

bhack:
Another brick in the (merge) wall :)

PR #2096, opened 2015-03-10
XOR hello world

pannous:
Similar to hdf5_classification.ipynb example, but even noob-friendlier and caffe-style.


pannous:
Discard if desired

PR #2097, opened 2015-03-11
Remove obsolete loss layer check

longjon:
With the `n`d blob update, it's no longer true that the `num` dimension has to be the same for both bottoms of every loss layer (e.g., softmax loss layer with `axis = 0`).

I'm not sure if additional checks should be added to other specific loss layers to compensate.

### DIFF
--- a/src/caffe/layers/loss_layer.cpp
+++ b/src/caffe/layers/loss_layer.cpp
@@ -22,8 +22,6 @@ void LossLayer<Dtype>::LayerSetUp(
 template <typename Dtype>
 void LossLayer<Dtype>::Reshape(
     const vector<Blob<Dtype>*>& bottom, const vector<Blob<Dtype>*>& top) {
-  CHECK_EQ(bottom[0]->num(), bottom[1]->num())
-      << "The data and label should have the same number.";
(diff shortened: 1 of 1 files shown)


jeffdonahue:
> (e.g., softmax loss layer with axis = 0)

Good call, didn't think of that case. I guess everything besides AccuracyLayer theoretically should have the check readded since none of those have changed since nd blobs AFAIK?

YutingZhang:
Redirected from #3097 . 
One simple fix is to remove the CHECK statement for num(), but this will ignore the dimension mismatch error. It is probably good to generalize all types of loss layers to arbitrary dimension, and move the "axis" property to loss_param, and let the base loss_layer check the dimension. 
@longjon what do you think?

longjon:
Right, generalizing all loss layers would be nice, but as a simpler first step we should probably (a) remove this check from the base class, and (b) go through each of the loss layers and individually readd this or another check according to whatever makes sense for each layer. You're welcome to upgrade this PR to include (b)!

shelhamer:
Related: #2814

JamesAshwood07:
Does this need to still be open. Was it solved in #2814

PR #2098, opened 2015-03-11
Multiple images per line support in ImageDataLayer

ducksoup:
Add support for two or more images per line in ImageDataLayer. The images are concatenated along the channels dimension and returned in a single top blob.
Backwards compatibility is preserved while making it easy to load n-uples of images (e.g. for siamese networks).

NOTE: the new implementation of the "source" file parsing expects the input to be divided into lines, while the current one can cope with more liberally formatted data.


ducksoup:
Closed: I did not follow the guidelines about the pull request protocol.

PR #2102, opened 2015-03-11
GPU in Python: converting Blob to PyCuda.gpuarray (no copy)

tnarihi:
This changes many things as described in individual commits. First 4 commits are just for utilities. Last 2 commits are main parts of this PR.  I exposed GPU pointers of `data` and `diff` in `Blob` to Python as integer, and use them to initialize `gpuarray` in PyCuda without copying memory. Please see `python/caffe/test/test_pycuda.py` for usage. I think the exposed gpu pointers could be used more other libraries such as gnumpy, Threano, etc.

* Add GetDevice and expose it to Python	 feb8e4d
* Expose functions check_mode_{cpu,gpu} to Python …	 879a450
* Expose Blob.shape to Python	 32cb250
* Expose CUDA_BLOCKS CUDA_NUM_THREADS to Python	 f5c7333
* Expose GPU pointer to Python …	 73e3540
* Add PyCuda utilities and test for Python I/F.	 022dc2c

### DIFF
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -199,6 +199,7 @@ BOOST_PYTHON_MODULE(_caffe) {
   bp::def("set_mode_cpu", &set_mode_cpu);
   bp::def("set_mode_gpu", &set_mode_gpu);
   bp::def("set_device", &Caffe::SetDevice);
+  bp::def("get_device", &Caffe::GetDevice);
 
   bp::class_<Net<Dtype>, shared_ptr<Net<Dtype> >, boost::noncopyable >("Net",
     bp::no_init)
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -35,6 +35,9 @@ const int NPY_DTYPE = NPY_FLOAT32;
 // Selecting mode.
 void set_mode_cpu() { Caffe::set_mode(Caffe::CPU); }
 void set_mode_gpu() { Caffe::set_mode(Caffe::GPU); }
+// Checking current mode.
+bool check_mode_cpu() { return Caffe::mode() == Caffe::CPU; }
+bool check_mode_gpu() { return Caffe::mode() == Caffe::GPU; }
 
 // For convenience, check that input files can be opened, and raise an
 // exception that boost will send to Python if not (caffe could still crash
@@ -198,6 +201,8 @@ BOOST_PYTHON_MODULE(_caffe) {
   // Caffe utility functions
   bp::def("set_mode_cpu", &set_mode_cpu);
   bp::def("set_mode_gpu", &set_mode_gpu);
+  bp::def("check_mode_cpu", &check_mode_cpu);
+  bp::def("check_mode_gpu", &check_mode_gpu);
   bp::def("set_device", &Caffe::SetDevice);
   bp::def("get_device", &Caffe::GetDevice);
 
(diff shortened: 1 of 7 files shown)


tnarihi:
Now we can call cuBlas functions though Scikits.CUDA using cuBlas handle which is hold by Caffe. See 63def1b5f5454 `test_pycuda.py` for usage of `Sgemm`.

bhack:
@tnarihi What do you think about (libgpuarray)[https://github.com/Theano/libgpuarray] approach?

bhack:
See also [motivations](http://deeplearning.net/software/libgpuarray/why.html#motivation)

tnarihi:
@bhack Thanks for sharing it! I didn't know this, and the concept and motivation are quite interesting for me. As far as I look at the motivation page and source code on GitHub, they are developing a C version of numpy.ndarray on GPU (abstracting multiple backends, CUDA and OpenCL so far.), which could be used by writing wrapper regardless languages such as C++, Python, Ruby etc, and they provide Python wrapper already. It seems like they are trying to replace the other existing libraries' array type such as PyCUDA.gpuarray, CudaMat etc, because they are still a subset of numpy.ndarray on GPU.

I reconsidered about this PR. The main point of this PR is exposing GPU (CUDA device) pointers of a Caffe.Blob to python so that we can manipulate them through various existing Python CUDA wrappers. Then, some of examples I implement are PyCUDA and Scikits.CUDA. You can use any other libraries in your own Python scripting. If we see libgpuarray as python interface (pygpu), it could be one of choices of python CUDA wrappers. It seems there is no relation directly from libgpuarray to pycaffe and even C++ Caffe as long as we don't replace Blob with libgpuarray. This is my understanding. What do you think about it?

bhack:
I almost share your view. Under a strategic point of view it could be interesting to share and increase the momentum on libgpuarray together another important deep learning framework like Theano but still in a "vendor" neutral way. On the other hand the library started before [arrayfire](https://github.com/arrayfire/arrayfire) was opensourced under BSD. Is arrayfire another candidate? Have it a too large target?

tnarihi:
Yeah, interesting if we could share core computation and data structures in multiple backends together with other deep learning frameworks but it seems we need much effort to do this.
@jeffdonahue @shelhamer @longjon Any thought? I am still not unclear about how we could share these momentums with Caffe. It looks far bigger discussion than this PR.

bhack:
Yes it is becoming off topic and to hi level for this PR

python/caffe/_caffe.cpp
```diff
@@ -229,29 +385,46 @@ BOOST_PYTHON_MODULE(_caffe) {
...
+      boost::noncopyable>(
+      "Layer", bp::init<const LayerParameter&>())
     .add_property("blobs", bp::make_function(&Layer<Dtype>::blobs,
           bp::return_internal_reference<>()))
     .def("setup", &Layer<Dtype>::LayerSetUp)
+    .def("SetUp", bp::raw_function(&Layer_SetUp))
     .def("reshape", &Layer<Dtype>::Reshape)
+    .def("Reshape", bp::raw_function(&Layer_Reshape))
+    .def("Forward", bp::raw_function(&Layer_Forward))
+    .def("Backward", bp::raw_function(&Layer_Backward))
```
tnarihi: Actually I want to use lower case attribute names, but lower case names are already reserved by PythonLayer.

tnarihi:
Now this PR is too much bigger. I added a lot of new features. Should I split them into individual PRs?

longjon:
@tnarihi, yes, it's always best to split PRs into the smallest sensible units; reviewing effort is superlinear in patch length.

Sorry I haven't had much time lately to look carefully at this and your other Python changes; this seems like a nice direction so hopefully we'll dive into it soon.

tnarihi:
@longjon Sure. I'll consider dependencies among commits and split them into the smallest units later. Thanks for your comment.

tnarihi:
Full version is moved to [my branch: python-gpu-full](https://github.com/tnarihi/caffe/tree/python-gpu-full) for backup.

PR #2107, opened 2015-03-12
Fix OS X installation guide

5kg:
- Remove `--fresh` Homebrew option
- Add missing `brew install` command


shelhamer:
Thanks for doc fixes!

PR #2113, opened 2015-03-12
Making GLog work with (I)Python

philkr:
GLog does currently not print anything in ipython (just the terminal where ipython was called).
This PR implements a LogSink to print any glog message in an (i)python friendly way.


shelhamer:
Closing as I like keeping the glog output in the terminal.

PR #2114, opened 2015-03-12
Multi-GPU

cypof:
Uses CUDA peer-to-peer for communication, and parts of #1148. SGD is now synchronous instead of asynchronous, as @longjon showed bandwidth on one box is actually high enough. We haven’t really benchmarked yet, but it seems to work great. It also gets rid of the momentum coordination problem.

The synchronization code needs to hook into the solver, so it is a bit more invasive than before, but still pretty isolated. I refactored solver.cpp to separate the regularization and gradient compute phases so that they can be invoked at different times by the parallel solver.

One thing still missing is the way to compute the actual number of iterations. For now each solver runs as if it was by itself, so the run is going to take as long as without parallelism. I guess we could adapt the solver to run 1/N steps instead. Also the batch size should be experimented with, as now effectively N times larger. On that, would it be more convenient to switch to the number of images to compute progress, instead of iterations, to be independent of batch size?

To try it, run the samples in example/parallel/

### DIFF
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -435,62 +435,6 @@ void P2PSync<Dtype>::run(shared_ptr<Solver<Dtype> > root,
   }
 }
 
-template<typename Dtype>
-void P2PSync<Dtype>::divide_batch_size(NetParameter* net) {
-  int solver_count = Caffe::solver_count();
-  for (int i = 0; i < net->layer_size(); ++i) {
-    string m = "Batch size must be divisible by the number of solvers (GPUs)";
-    if (net->layer(i).has_data_param()) {
-      if (net->layer(i).data_param().has_batch_size()) {
-        uint32_t total = net->layer(i).data_param().batch_size();
-        uint32_t batch = total / solver_count;
-        CHECK(batch * solver_count == total) << m;
-        net->mutable_layer(i)->mutable_data_param()->set_batch_size(batch);
-
-        // Also adjust the prefetch count, as it is shared by all solvers
-        uint32_t prefetch = net->layer(i).data_param().prefetch();
-        net->mutable_layer(i)->mutable_data_param()->set_prefetch(
-            prefetch * solver_count);
-      }
-    }
-    if (net->layer(i).has_hdf5_data_param()) {
-      if (net->layer(i).hdf5_data_param().has_batch_size()) {
-        uint32_t total = net->layer(i).hdf5_data_param().batch_size();
-        uint32_t batch = total / solver_count;
-        CHECK(batch * solver_count == total) << m;
-        net->mutable_layer(i)->mutable_hdf5_data_param()->set_batch_size(batch);
-      }
-    }
-    if (net->layer(i).has_image_data_param()) {
-      if (net->layer(i).image_data_param().has_batch_size()) {
-        uint32_t total = net->layer(i).image_data_param().batch_size();
-        uint32_t batch = total / solver_count;
-        CHECK(batch * solver_count == total) << m;
-        net->mutable_layer(i)->mutable_image_data_param()->set_batch_size(
-            batch);
-      }
-    }
-    if (net->layer(i).has_memory_data_param()) {
-      if (net->layer(i).memory_data_param().has_batch_size()) {
-        uint32_t total = net->layer(i).memory_data_param().batch_size();
-        uint32_t batch = total / solver_count;
-        CHECK(batch * solver_count == total) << m;
-        net->mutable_layer(i)->mutable_memory_data_param()->set_batch_size(
-            batch);
-      }
-    }
-    if (net->layer(i).has_window_data_param()) {
-      if (net->layer(i).window_data_param().has_batch_size()) {
-        uint32_t total = net->layer(i).window_data_param().batch_size();
-        uint32_t batch = total / solver_count;
-        CHECK(batch * solver_count == total) << m;
-        net->mutable_layer(i)->mutable_window_data_param()->set_batch_size(
-            batch);
-      }
-    }
-  }
-}
-
 INSTANTIATE_CLASS(Params);
 INSTANTIATE_CLASS(GPUParams);
 INSTANTIATE_CLASS(P2PSync);
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -305,7 +305,7 @@ void P2PSync<Dtype>::on_start(Timer* timer, ostringstream* timing) {
   if (children_.size()) {
     timer->Start();
   }
-  for (int i = 0; i < children_.size(); ++i) {
+  for (int i = children_.size() - 1; i >= 0; i--) {
     Dtype* src = data_;
     Dtype* dst = children_[i]->data_;
 
@@ -319,11 +319,7 @@ void P2PSync<Dtype>::on_start(Timer* timer, ostringstream* timing) {
 
     CUDA_CHECK(cudaMemcpyAsync(dst, src, size_ * sizeof(Dtype),  //
         cudaMemcpyDeviceToDevice, cudaStreamDefault));
-  }
-  if (children_.size()) {
     CUDA_CHECK(cudaStreamSynchronize(cudaStreamDefault));
-  }
-  for (int i = 0; i < children_.size(); ++i) {
     children_[i]->queue_.push(this);
   }
   if (children_.size()) {
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -119,18 +119,21 @@ void DevicePair::compute(const vector<int> devices, vector<DevicePair>* pairs) {
 #ifndef CPU_ONLY
   vector<int> remaining(devices);
 
-  // Group GPUs by board
-  for (int i = 0; i < remaining.size(); ++i) {
-    for (int j = i + 1; j < remaining.size(); ++j) {
-      cudaDeviceProp a, b;
-      CUDA_CHECK(cudaGetDeviceProperties(&a, remaining[i]));
-      CUDA_CHECK(cudaGetDeviceProperties(&b, remaining[j]));
-      if (a.isMultiGpuBoard && b.isMultiGpuBoard) {
-        if (a.multiGpuBoardGroupID == b.multiGpuBoardGroupID) {
-          pairs->push_back(DevicePair(remaining[i], remaining[j]));
-          DLOG(INFO) << "GPU board: " << remaining[i] << ":" << remaining[j];
-          remaining.erase(remaining.begin() + j);
-          break;
+  // Group GPUs by board - some boards can have more than 2 ASICs
+  for (int d = 0; d < remaining.size(); ++d) {
+    for (int i = 0; i < remaining.size(); ++i) {
+      for (int j = i + 1; j < remaining.size(); ++j) {
+        cudaDeviceProp a, b;
+        CUDA_CHECK(cudaGetDeviceProperties(&a, remaining[i]));
+        CUDA_CHECK(cudaGetDeviceProperties(&b, remaining[j]));
+        if (a.isMultiGpuBoard && b.isMultiGpuBoard) {
+          if (a.multiGpuBoardGroupID == b.multiGpuBoardGroupID) {
+              pairs->push_back(DevicePair(remaining[i], remaining[j]));
+              DLOG(INFO) << "GPU board: " << remaining[i]
+                         << ":" << remaining[j];
+              remaining.erase(remaining.begin() + j);
+              break;
+          }
         }
       }
     }
@@ -141,32 +144,40 @@ void DevicePair::compute(const vector<int> devices, vector<DevicePair>* pairs) {
   }
   DLOG(INFO) << "GPUs paired by boards, remaining: " << s.str();
 
-  // Group by P2P accessibility
-  for (int i = 0; i < remaining.size(); ++i) {
-    for (int j = i + 1; j < remaining.size(); ++j) {
-      int access;
-      CUDA_CHECK(cudaDeviceCanAccessPeer(&access, remaining[i], remaining[j]));
-      if (access) {
-        pairs->push_back(DevicePair(remaining[i], remaining[j]));
-        DLOG(INFO) << "P2P pair: " << remaining[i] << ":" << remaining[j];
-        remaining.erase(remaining.begin() + j);
-        break;
+  // Group by P2P accessibility - P2P group can be larger than 4 boards
+  for (int d = 0; d < remaining.size(); ++d) {
+    for (int i = 0; i < remaining.size(); ++i) {
+      for (int j = i + 1; j < remaining.size(); ++j) {
+        int access;
+        CUDA_CHECK(cudaDeviceCanAccessPeer(&access,
+                                           remaining[i],
+                                           remaining[j]));
+        if (access) {
+            pairs->push_back(DevicePair(remaining[i], remaining[j]));
+            DLOG(INFO) << "P2P pair: " << remaining[i]
+                       << ":" << remaining[j];
+            remaining.erase(remaining.begin() + j);
+            break;
+        }
       }
     }
   }
   s.str("");
   for (int i = 0; i < remaining.size(); ++i) {
-    s << (i ? ", " : "") << remaining[i];
+      s << (i ? ", " : "") << remaining[i];
   }
   DLOG(INFO) << "GPUs paired by P2P access, remaining: " << s.str();
 
   // Group remaining
-  for (int i = 0; i < remaining.size(); ++i) {
-    for (int j = i + 1; j < remaining.size(); ++j) {
-      pairs->push_back(DevicePair(remaining[i], remaining[j]));
-      DLOG(INFO) << "Remaining pair: " << remaining[i] << ":" << remaining[j];
-      remaining.erase(remaining.begin() + j);
-      break;
+  for (int d = 0; d < remaining.size(); ++d) {  // try to pair everyone
+    for (int i = 0; i < remaining.size(); ++i) {
+      for (int j = i + 1; j < remaining.size(); ++j) {
+          pairs->push_back(DevicePair(remaining[i], remaining[j]));
+          DLOG(INFO) << "Remaining pair: " << remaining[i]
+                     << ":" << remaining[j];
+          remaining.erase(remaining.begin() + j);
+          break;
+      }
     }
   }
   CHECK_EQ(remaining.size(), 1);
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -1,0 +1,499 @@
+#ifndef CPU_ONLY
+#include <cuda_runtime.h>
+#endif
+#include <glog/logging.h>
+#include <stdio.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+
+#include <cstdlib>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "boost/thread.hpp"
+#include "caffe/caffe.hpp"
+#include "caffe/parallel.hpp"
+
+namespace caffe {
+
+enum Op {
+  copy,
+  replace_cpu,
+  replace_gpu,
+  replace_cpu_diff,
+  replace_gpu_diff
+};
+
+template<typename Dtype>
+static void apply_buffers(const vector<shared_ptr<Blob<Dtype> > >& blobs,
+                          Dtype* buffer, size_t total_size, Op op) {
+  Dtype* ptr = buffer;
+  for (int i = 0; i < blobs.size(); ++i) {
+    int size = blobs[i]->count();
+    switch (op) {
+      case copy: {
+        // Init buffer to current values of blobs
+        caffe_copy(size,
+                   reinterpret_cast<const Dtype*>(blobs[i]->data()->cpu_data()),
+                   ptr);
+        break;
+      }
+      case replace_cpu:
+        blobs[i]->data()->set_cpu_data(ptr);
+        break;
+      case replace_gpu:
+        blobs[i]->data()->set_gpu_data(ptr);
+        break;
+      case replace_cpu_diff:
+        blobs[i]->diff()->set_cpu_data(ptr);
+        break;
+      case replace_gpu_diff:
+        blobs[i]->diff()->set_gpu_data(ptr);
+        break;
+    }
+    ptr += size;
+  }
+  CHECK_EQ(total_size, ptr - buffer);
+}
+
+// Buffer size necessary to store given blobs
+template<typename Dtype>
+static size_t total_size(const vector<shared_ptr<Blob<Dtype> > >& params) {
+  size_t size = 0;
+  for (int i = 0; i < params.size(); ++i)
+    size += params[i]->count();
+  return size;
+}
+
+template<typename Dtype>
+Params<Dtype>::Params(shared_ptr<Solver<Dtype> > root_solver)
+    : size_(total_size<Dtype>(root_solver->net()->params())),
+      data_(),
+      diff_() {
+}
+
+template<typename Dtype>
+GPUParams<Dtype>::GPUParams(shared_ptr<Solver<Dtype> > root_solver, int device)
+    : Params<Dtype>(root_solver) {
+#ifndef CPU_ONLY
+  int initial_device;
+  CUDA_CHECK(cudaGetDevice(&initial_device));
+
+  // Allocate device buffers
+  CUDA_CHECK(cudaSetDevice(device));
+  CUDA_CHECK(cudaMalloc(&data_, size_ * sizeof(Dtype)));
+
+  // Copy blob values
+  const vector<shared_ptr<Blob<Dtype> > >& net = root_solver->net()->params();
+  apply_buffers(net, data_, size_, copy);
+
+  CUDA_CHECK(cudaMalloc(&diff_, size_ * sizeof(Dtype)));
+  caffe_gpu_set(size_, Dtype(0), diff_);
+
+  CUDA_CHECK(cudaSetDevice(initial_device));
+#else
+  NO_GPU;
+#endif
+}
+
+template<typename Dtype>
+GPUParams<Dtype>::~GPUParams() {
+#ifndef CPU_ONLY
+  CUDA_CHECK(cudaFree(data_));
+  CUDA_CHECK(cudaFree(diff_));
+#endif
+}
+
+template<typename Dtype>
+void GPUParams<Dtype>::configure(Solver<Dtype>* solver) const {
+  const vector<shared_ptr<Blob<Dtype> > >& net = solver->net()->params();
+  apply_buffers(net, data_, size_, replace_gpu);
+  apply_buffers(net, diff_, size_, replace_gpu_diff);
+}
+
+//
+
+void DevicePair::compute(const vector<int> devices, vector<DevicePair>* pairs) {
+#ifndef CPU_ONLY
+  vector<int> remaining(devices);
+
+  // Group GPUs by board
+  for (int i = 0; i < remaining.size(); ++i) {
+    for (int j = i + 1; j < remaining.size(); ++j) {
+      cudaDeviceProp a, b;
+      CUDA_CHECK(cudaGetDeviceProperties(&a, remaining[i]));
+      CUDA_CHECK(cudaGetDeviceProperties(&b, remaining[j]));
+      if (a.isMultiGpuBoard && b.isMultiGpuBoard) {
+        if (a.multiGpuBoardGroupID == b.multiGpuBoardGroupID) {
+          pairs->push_back(DevicePair(remaining[i], remaining[j]));
+          DLOG(INFO) << "GPU board: " << remaining[i] << ":" << remaining[j];
+          remaining.erase(remaining.begin() + j);
+          break;
+        }
+      }
+    }
+  }
+  ostringstream s;
+  for (int i = 0; i < remaining.size(); ++i) {
+    s << (i ? ", " : "") << remaining[i];
+  }
+  DLOG(INFO) << "GPUs paired by boards, remaining: " << s.str();
+
+  // Group by P2P accessibility
+  for (int i = 0; i < remaining.size(); ++i) {
+    for (int j = i + 1; j < remaining.size(); ++j) {
+      int access;
+      CUDA_CHECK(cudaDeviceCanAccessPeer(&access, remaining[i], remaining[j]));
+      if (access) {
+        pairs->push_back(DevicePair(remaining[i], remaining[j]));
+        DLOG(INFO) << "P2P pair: " << remaining[i] << ":" << remaining[j];
+        remaining.erase(remaining.begin() + j);
+        break;
+      }
+    }
+  }
+  s.str("");
+  for (int i = 0; i < remaining.size(); ++i) {
+    s << (i ? ", " : "") << remaining[i];
+  }
+  DLOG(INFO) << "GPUs paired by P2P access, remaining: " << s.str();
+
+  // Group remaining
+  for (int i = 0; i < remaining.size(); ++i) {
+    for (int j = i + 1; j < remaining.size(); ++j) {
+      pairs->push_back(DevicePair(remaining[i], remaining[j]));
+      DLOG(INFO) << "Remaining pair: " << remaining[i] << ":" << remaining[j];
+      remaining.erase(remaining.begin() + j);
+      break;
+    }
+  }
+  CHECK_EQ(remaining.size(), 1);
+  pairs->insert(pairs->begin(), DevicePair(-1, remaining[0]));
+
+  CHECK(pairs->size() == devices.size());
+  for (int i = 0; i < pairs->size(); ++i) {
+    CHECK((*pairs)[i].parent() != (*pairs)[i].device());
+    for (int j = i + 1; j < pairs->size(); ++j) {
+      CHECK((*pairs)[i].device() != (*pairs)[j].device());
+    }
+  }
+#else
+  NO_GPU;
+#endif
+}
+
+//
+
+template<typename Dtype>
+P2PSync<Dtype>::P2PSync(shared_ptr<Solver<Dtype> > root_solver,
+                        P2PSync<Dtype>* parent, const SolverParameter& param)
+    : GPUParams<Dtype>(root_solver, param.device_id()),
+      parent_(parent),
+      children_(),
+      queue_(),
+      initial_iter_(root_solver->iter()),
+      solver_() {
+#ifndef CPU_ONLY
+  int initial_device;
+  CUDA_CHECK(cudaGetDevice(&initial_device));
+  const int self = param.device_id();
+  CUDA_CHECK(cudaSetDevice(self));
+
+  if (parent == NULL) {
+    solver_ = root_solver;
+  } else {
+    Caffe::set_root_solver(false);
+    solver_.reset(new WorkerSolver<Dtype>(param));
+    Caffe::set_root_solver(true);
+  }
+  this->configure(solver_.get());
+  solver_->add_callback(this);
+
+  if (parent) {
+    // Enable p2p access between devices
+    const int peer = parent->solver_->param().device_id();
+    int access;
+    CUDA_CHECK(cudaDeviceCanAccessPeer(&access, self, peer));
+    if (access) {
+      CUDA_CHECK(cudaDeviceEnablePeerAccess(peer, 0));
+    } else {
+      LOG(INFO)<< "GPU " << self << " does not have p2p access to GPU " << peer;
+    }
+    // Allocate receiving buffer on parent
+    CUDA_CHECK(cudaSetDevice(peer));
+    CUDA_CHECK(cudaMalloc(&parent_grads_, size_ * sizeof(Dtype)));
+    CUDA_CHECK(cudaSetDevice(self));
+  }
+
+  CUDA_CHECK(cudaSetDevice(initial_device));
+#else
+  NO_GPU;
+#endif
+}
+
+template<typename Dtype>
+P2PSync<Dtype>::~P2PSync() {
+#ifndef CPU_ONLY
+  int initial_device;
+  CUDA_CHECK(cudaGetDevice(&initial_device));
+  const int self = solver_->param().device_id();
+  CUDA_CHECK(cudaSetDevice(self));
+
+  if (parent_) {
+    CUDA_CHECK(cudaFree(parent_grads_));
+    const int peer = parent_->solver_->param().device_id();
+    int access;
+    CUDA_CHECK(cudaDeviceCanAccessPeer(&access, self, peer));
+    if (access) {
+      CUDA_CHECK(cudaDeviceDisablePeerAccess(peer));
+    }
+  }
+
+  CUDA_CHECK(cudaSetDevice(initial_device));
+#endif
+}
+
+template<typename Dtype>
+void P2PSync<Dtype>::InternalThreadEntry() {
+  Caffe::SetDevice(solver_->param().device_id());
+  CHECK(Caffe::root_solver());
+  Caffe::set_root_solver(false);
+  // See if there is a defined seed and reset random state if so
+  if (solver_->param().random_seed() >= 0) {
+    // Fetch random seed and modulate by device ID to make sure
+    // everyone doesn't have the same seed.  We seem to have some
+    // solver instability if we have everyone with the same seed
+    Caffe::set_random_seed(
+        solver_->param().random_seed() + solver_->param().device_id());
+  }
+  solver_->Step(solver_->param().max_iter() - initial_iter_);
+}
+
+template<typename Dtype>
+void P2PSync<Dtype>::on_start(Timer* timer, ostringstream* timing) {
+#ifndef CPU_ONLY
+#ifdef DEBUG
+  int device;
+  CUDA_CHECK(cudaGetDevice(&device));
+  CHECK(device == solver_->param().device_id());
+#else
+//  CHECK(false);
+#endif
+
+  // Wait for update from parent
+  if (parent_) {
+    timer->Start();
+    P2PSync<Dtype> *parent = queue_.pop();
+    CHECK(parent == parent_);
+    *timing << " recv_param: " << timer->MilliSeconds();
+  }
+
+  // Update children
+  if (children_.size()) {
+    timer->Start();
+  }
+  for (int i = 0; i < children_.size(); ++i) {
+    Dtype* src = data_;
+    Dtype* dst = children_[i]->data_;
+
+#ifdef DEBUG
+    cudaPointerAttributes attributes;
+    CUDA_CHECK(cudaPointerGetAttributes(&attributes, src));
+    CHECK(attributes.device == device);
+    CUDA_CHECK(cudaPointerGetAttributes(&attributes, dst));
+    CHECK(attributes.device == children_[i]->solver_->param().device_id());
+#endif
+
+    CUDA_CHECK(cudaMemcpyAsync(dst, src, size_ * sizeof(Dtype),  //
+        cudaMemcpyDeviceToDevice, cudaStreamDefault));
+  }
+  if (children_.size()) {
+    CUDA_CHECK(cudaStreamSynchronize(cudaStreamDefault));
+  }
+  for (int i = 0; i < children_.size(); ++i) {
+    children_[i]->queue_.push(this);
+  }
+  if (children_.size()) {
+    *timing << " send_param: " << timer->MilliSeconds();
+  }
+#endif
+}
+
+template<typename Dtype>
+void P2PSync<Dtype>::on_gradients_ready(Timer* timer, ostringstream* timing) {
+#ifndef CPU_ONLY
+#ifdef DEBUG
+  int device;
+  CUDA_CHECK(cudaGetDevice(&device));
+  CHECK(device == solver_->param().device_id());
+#endif
+
+  // Sum children gradients as they appear in the queue
+  for (int i = 0; i < children_.size(); ++i) {
+    timer->Start();
+    P2PSync<Dtype> *child = queue_.pop();
+    Dtype* src = child->parent_grads_;
+    Dtype* dst = diff_;
+
+#ifdef DEBUG
+    bool ok = false;
+    for (int j = 0; j < children_.size(); ++j) {
+      if (child == children_[j]) {
+        ok = true;
+      }
+    }
+    CHECK(ok);
+    cudaPointerAttributes attributes;
+    CUDA_CHECK(cudaPointerGetAttributes(&attributes, src));
+    CHECK(attributes.device == device);
+    CUDA_CHECK(cudaPointerGetAttributes(&attributes, dst));
+    CHECK(attributes.device == device);
+#endif
+
+    caffe_gpu_add(size_, src, dst, dst);
+    *timing << " add_grad: " << timer->MilliSeconds();
+  }
+
+  // Send gradients to parent
+  if (parent_) {
+    timer->Start();
+    Dtype* src = diff_;
+    Dtype* dst = parent_grads_;
+
+#ifdef DEBUG
+    cudaPointerAttributes attributes;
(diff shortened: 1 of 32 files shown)


wkal:
Hi! I had tried this branch, and when I tried to using the gpus.bin work on Mnist, it failed. I input command: 
../../build/examples/parallel/gpus.bin ../../examples/mnist/lenet_solver.prototxt 0:1

the error information like this: 
F0314 14:28:16.295174 34820 p2p_sync.cpp:139] Check failed: error == cudaSuccess (51 vs. 0)  peer access has not been enabled
    @     0x7f7ae5377daa  (unknown)
    @     0x7f7ae5377ce4  (unknown)
    @     0x7f7ae53776e6  (unknown)
    @     0x7f7ae537a687  (unknown)
    @     0x7f7ae57122b2  caffe::P2PSync<>::GPU::~GPU()
    @     0x7f7ae57124a9  caffe::P2PSync<>::GPU::~GPU()
    @           0x40db4e  boost::detail::sp_counted_base::release()
    @           0x40dbd5  caffe::P2PSync<>::~P2PSync()
    @           0x40b6c2  main
    @     0x7f7ae4686ec5  (unknown)
    @           0x40b96e  (unknown)
Aborted (core dumped)

cypof:
That's strange, it seems to be during shutdown when destructors get called. Did it train at all before reaching the end?

wkal:
it didn't print the log information such as network configuration , training/testing loss, and testing accuracy at the screen. But when I call the nvidia-smi after start the gpu.bin, I can watch it really enable the GPU 0 and GPU 1 to work, the GPU monitor information like this:
    0     28026  ../../build/examples/parallel/gpus.bin               122MiB 
    1     28026  ../../build/examples/parallel/gpus.bin               112MiB

wkal:
GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla K40c          Off  | 0000:02:00.0     Off |                  Off |
| 23%   44C    P0   106W / 235W |    148MiB / 12287MiB |     84%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Tesla K40c          Off  | 0000:03:00.0     Off |                    0 |
| 23%   46C    P0   104W / 235W |    137MiB / 11519MiB |     72%      Default

wkal:
So from the above monitor information, I guess the training should started, but I don't know how the error occured and also how to fixed it. Hope you can give me suggestions.  Thanks!

wkal:
When I attempt to add debug code to watch what's happened for this error:
LOG(INFO) << "access: " << access;
      LOG(INFO) << "device: " << device;
      LOG(INFO) << "peer: " << peer;
then error information becomes:
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
*** Aborted at 1426398733 (unix time) try "date -d @1426398733" if you are using GNU date ***
PC: @     0x7fc456963bb9 (unknown)
*** SIGABRT (@0x3e800001e78) received by PID 7800 (TID 0x7dc41ded3700) from PID 7800; stack trace: ***
    @     0x7fc456d03340 (unknown)
    @     0x7fc456963bb9 (unknown)
    @     0x7fc456966fc8 (unknown)
    @     0x7fc4571876b5 (unknown)
    @     0x7fc457185836 (unknown)
    @     0x7fc457185863 (unknown)
    @     0x7fc457185aa2 (unknown)
    @     0x7fc4579a93f1 boost::throw_exception<>()
    @     0x7fc4579a96a5 boost::unique_lock<>::lock()
    @     0x7fc4579a9f48 boost::condition_variable::wait()
    @     0x7fc4579ab238 caffe::blocking_queue<>::pop()
    @     0x7fc457a72f8d caffe::BasePrefetchingDataLayer<>::InternalThreadEntry()
    @     0x7fc44f14fa4a (unknown)
    @     0x7fc456cfb182 start_thread
    @     0x7fc456a27fbd (unknown)
    @                0x0 (unknown)
Aborted (core dumped)

The changed code:
template<typename Dtype>
P2PSync<Dtype>::GPU::~GPU() {
  CUDA_CHECK(cudaStreamDestroy(stream_));
  for (int i = 0; i < sync_.params().size(); ++i) {
    if (i != index_) {
      CUDA_CHECK(cudaFree(targets_[i]));

      const int device = params_.device();
      const int peer = sync_.params()[i]->device();
      int access;
      CUDA_CHECK(cudaDeviceCanAccessPeer(&access, device, peer));
      LOG(INFO) << "access: " << access;
      LOG(INFO) << "device: " << device;
      LOG(INFO) << "peer: " << peer;
      if (access) {
        CUDA_CHECK(cudaDeviceDisablePeerAccess(peer));
      }
    }
  }
}

cypof:
If it does not print any log, it might be glog config:
export GLOG_logtostderr=1

hhuytho:
It crashes if I try more than 4 cards at a time. Any idea?

*** SIGSEGV (@0x30) received by PID 4086 (TID 0x7ffb40f6f720) from PID 48; stack trace: ***
    @       0x3d80c0f710 (unknown)
    @           0x40d534 SolverContext::run()
    @           0x408db2 main
    @       0x3d8081ed5d (unknown)
    @           0x408329 (unknown)
Segmentation fault (core dumped)

hhuytho:
I seem to figure out the problem. Line line 69 of gpus.cpp, "params[gpu[0]]" should be "params[0]". Similarly, in line 76 of gpus.cpp, "params[device]" should be "params[i]". The segmentation fault was caused by the wrong indexing of the solvers.

wkal:
@cypof:"If it does not print any log, it might be glog config:
export GLOG_logtostderr=1"

I tried set the  GLOG_logtostderr to 1 in the /etc/profile, and also GLOG_logtostderr=0,  but I still not seen no output after I did that.  Is my approach correct? If not how should I do that? Thanks!

wkal:
@hhuytho 
I followed your fix, but the error I reported still existed. 
The error:
F0317 11:50:21.572767 29807 p2p_sync.cpp:142] Check failed: error == cudaSuccess (51 vs. 0)  peer access has not been enabled
    @     0x7fdbcffa9daa  (unknown)
    @     0x7fdbcffa9ce4  (unknown)
    @     0x7fdbcffa96e6  (unknown)
    @     0x7fdbcffac687  (unknown)
    @     0x7fdbd0344d79  caffe::P2PSync<>::GPU::~GPU()
    @     0x7fdbd0344fc9  caffe::P2PSync<>::GPU::~GPU()
    @           0x40db4e  boost::detail::sp_counted_base::release()
    @           0x40dbd5  caffe::P2PSync<>::~P2PSync()
    @           0x40b6c2  main
    @     0x7fdbcf2b8ec5  (unknown)
    @           0x40b96e  (unknown)
Aborted (core dumped)
From this error information, I think the error should occured in the p2p_sync.cpp, maybe we used different version of Caffe? The version of Caffe I used is caffe-multi_gpu_db_fixes(https://github.com/cypof/caffe/tree/multi_gpu_db_fixes). Thanks for your sharing!

hhuytho:
I used the "multi_gpu" branch from cypof. Your problem doesn't seem to relate to the bug I listed. Maybe it has something to do with the peer to peer communication between the GPUs?

GoGoStudio:
@cypof Thanks for this PR! I tried it on 2 K80s (4 GPUs) with the multi_gpu branch, it failed to converge on MNIST data. It runs fine, no crashes. caffe.bin in the same branch converges. I think there is something wrong in gpus.cpp codes. BTW, I used mnist on lmdb, cudnn r1 enabled. Am I missing anything?

cypof:
New version, should be more stable and faster. The gradients are now combined using a map-reduce scheme, first the GPUs on the same board, then same CPU socket, then to the root. It has not been tested a lot but seems to work OK.

The batch size and gradients are scaled, so that multi-gpu should now be exactly equivalent to running on a single one. I tried comparing the weights, and they are equal at least to 10e-5. The difference must be numerical approximations. To make runs reproducible it is necessary to set PREFETCH_COUNT to 1 in data_layer.hpp, otherwise examples might be reordered.

I also added a bunch of benchmarking logs that can be enabled through constants in Makefile.config. The code is in caffe.cpp instead of samples, and there is no parallel solver class anymore. Instead it adds a callback on the solver created by caffe.cpp, so it should work with advanced algo like AdaGrad etc., we haven’t tried. I also did some refactoring in solver.cpp, but not changed anything functionally.

We haven’t ran all the way on ImageNet yet but early benchmark seems to show using the default batch sizes and settings:

AlexNet
2 GPUs: 1.8x
4: 2.9x
8: 3.4x

GoogLeNet
2: 1.9x
4: 3.2x
8: 4.5x
8: 6.6x with batch of 128

I don’t know if we can get much more through data parallelism. 16 bits float transfers might help a bit. I hope hybrid models using #2219 can go further. Larger batches get better GPUs utilisation, but lower SGD convergence. On AlexNet the batch is already 256 and increasing seems hurt convergence a lot. On GoogLeNet going from 32 to 128 should be OK but we have no numbers yet.

cypof:
If you get an error about nvidia-ml, install the GDK: https://developer.nvidia.com/gpu-deployment-kit, also this currenly only works on CUDA-7.

thatguymike:
I think you can get around needing the GDK for your specific case of detecting multiGPU boards using direct device queries through the runtime.  Something like:

    if (cudaGetDeviceProperties(&properties, a) != cudaSuccess) {
        return <error>;
    }
    isMulti0 = properties.isMultiGpuBoard;
    boardId0 = properties.multiGpuBoardGroupID;
    
    if (cudaGetDeviceProperties(&properties, whichGPUs[1]) != cudaSuccess) {
        return <error>;
    }
    isMulti1 = properties.isMultiGpuBoard;
    boardId1 = properties.multiGpuBoardGroupID;

    if (boardId0 == boardId1) {
       <same board>;
    }

src/caffe/solver.cpp
```diff
@@ -201,14 +204,33 @@ void Solver<Dtype>::Step(int iters) {
...
+    timing << " apply: " << timer.MilliSeconds();
+    if (callback_) {
+      callback_->finish_iteration(&timer, &timing);
+    }
+#ifdef BENCHMARK_SOLVER
+    LOG(INFO)<< timing.str();
+#endif
 
     // Save a snapshot if needed.
     if (param_.snapshot() && (iter_ + 1) % param_.snapshot() == 0) {
```
sontran: you might need to add  "&& Caffe::root_solver()" here

wkal:
Could we keep the things simpler? I tried the CXXNET (now they are supportting the multi-GPU mode seems work very stable), it did't need install so many additional libraries, and if we have a simpler structure, I believe we can quicker to own a stable Multi-GPUs branch! Could we refer to the multi-GPUs mode of Cuda-Convnet or the CXXNET?   PS: I am not study the source code of this branch very carefully, so I just show my opinions and maybe my opinion is wrong, because it is possible that the current branch require more additional libraries then it can run.

sontran:
Thanks @cypof for this great PR! I've tried it on 2 and 4 gpus machines, local and ec2, with and without p2p access, on different nets, etc... it seems to be pretty stable so far... The speed up factors also seem to match with your numbers....

cypof:
Thanks @thatguymike @sontran for checking this out. I tried the built-in CUDA API before switching to GDK but it seems to return 0 even if the GPUS are on a board. Maybe I was mistaken, I will look again. I hope the GDK will be merged into CUDA, there is lots of useful stuff in there. In terms of architecture, it seems on some machines doing a all-to-all is faster than map-reduce. I'm think it is possible to write a simple code that does both, always reducing on dual-gpu cards then all-to-all. I will probably stop there for single node, unlikely to gain much more this way.

bhack:
@cypof how strictly this is related to cuda? Do you think that will be portable in the future to other backend effort like https://github.com/BVLC/caffe/pull/2195?

cypof:
@bhack as long as you can memcpy between two GPUs it should work. For performance you also want to have a memcpyAsync. Not sure what is provided by OpenCL.

thatguymike:
Getting scaling performance really requires optimal data transfers.  The GDK provides methods to get system topology and information on what GPUs can do direct DMAs.  (I think we can replace those with standard CUDA calls, working on that now).  lspci/hwloc might be usable to get system topology, but I don't believe there is a standard way in OpenCL to setup and manage P2P DMA transfers between devices.

thatguymike:
With respect to needing the GDK, here is work-alike for seeing if 2 ASICs are on the same board.  Also, one can parse the output of 'nvidia-smi topo -m' to get more topology information:


  // Group GPUs by board
  for (int i = 0; i < remaining.size(); ++i) {
      cudaDeviceProp prop_a;
      CUDA_CHECK(cudaGetDeviceProperties(&prop_a, remaining[i]));
      for (int j = i + 1; j < remaining.size(); ++j) {
          cudaDeviceProp prop_b;
          CUDA_CHECK(cudaGetDeviceProperties(&prop_b, remaining[j]));
          int same_board = 0;
          if(prop_a.isMultiGpuBoard && prop_b.isMultiGpuBoard)
              same_board = prop_a.multiGpuBoardGroupID == prop_b.multiGpuBoardGroupID ? 1 : 0;
          if (same_board) {
              pairs->push_back(DevicePair(remaining[i], remaining[j]));
              DLOG(INFO) << "\n\nSame board: " << remaining[i] << ":" << remaining[j] << " - Board ID:" << prop_a.multiGpuBoardGroupID;
              remaining.erase(remaining.begin() + j);
          }
      }
  }

bhack:
I think that p2p will be covered by [hsa] (http://www.theregister.co.uk/2013/10/22/amnd_heterogenous_queueing/)  initiative of hsa foundation chipmakers.

brenton2015:
@cypof Could you show  GPU topology on 4 gpus machines?

thatguymike:
This depends on the general system topology.  The code currently tries to align the tree reduction strategy with the system.  For example, if you have a dual GPU board, the code attempts to do the first reduction on the board, and then one of those ASICs communicates with one ASIC on the other dual GPU board.

If you have 4 independent GPUs, it tries to find P2P pairs, but the current code doesn't go much deeper into the topology trying to find host PCIe bridges (like PLX).   But, it will attempt to avoid crossing PCIe root complexes, where P2P doesn't work, until the top of the reduction tree.

Basically, the code tries to optimize bandwidth use at each level in the reduction tree.  It should do the right thing for most 4 GPU systems.  For more complex server systems with PLX bridges or multiple root complex, there is more tuning that needs to be done.

However, the primary issue in the code at the moment is that there is a race somewhere and even with fixing the incoming random seed, we get different results run to run.  This can lead to things failing to converge.

wkal:
Hello! Everyone. I had download the multi-gpu branch, compiling and run it, but I faced a problem that I specified two gpus 0:1, when I run this branch, I found it only using gpu 0 ( Train net output #0: loss = 0.133571 (* 1 = 0.133571 loss) dev: 0), my running command: ../../build/examples/parallel/gpus.bin lenet_solver.prototxt 0:1, could anyone provide suggestions about how to fix it ? And I also found a strange phenomenon, that when I specify gpu 0 and gpu 3 to run caffe, the program will fail, I had tried CXXNET, it can do this. PS: I am not install the GDK, does this caused by lacking of GDK?

wkal:
Sorry for the previous comment, there something not clear, so I add this comment to correct it. I had 4 gpus to run current multi-gpus branch, however I only could run the gpu 0 and 1 successfully, if I try to run the gpu 2 and 3, caffe will fail, the fail information: 
*** SIGSEGV (@0x0) received by PID 1503 (TID 0x7f40e8d7e9c0) from PID 0; stack trace: ***
    @     0x7f40e79ce340 (unknown)
    @           0x41267b SolverContext::run()
    @           0x40b520 main
    @     0x7f40e7619ec5 (unknown)
    @           0x40b96e (unknown)
    @                0x0 (unknown)
Segmentation fault (core dumped)
and the running command like this: ../../build/examples/parallel/gpus.bin lenet_solver.prototxt 2:3.

I feel it out of my expectation, because I can run the gpu 2 and 3 in single gpu mode and also I can run on the gpu 0 and 1 at the same time, I think there should no reason can't run on gpu 2 and gpu 3. 
 
@sontran  
My system also had 4 gpus, why I can't run the gpu 2 and gpu 3? Need I adjust the topology of the GPUs? Could you help me?

cypof:
@wkal I see you are running the gpus.bin executable. The code is now in caffe.bin itself so you might be running an old version. Try a make clean, and run caffe train instead.

naibaf7:
For forward processing, I'm currently trying to implement SplitNet, which adds the possibility to split up a network DAG on multiple devices and process the network forwarding asynchronously across multiple devices.
This is necessary because our dense labeling network (for microscopy images) does not fit well into the memory of one GPU anymore.

wkal:
@cypof 
Thanks for your advice!  I had seen the caffe.cpp, yeah, it sure different from the previous version. And I want to give a feedback that which version of CuDNN that this multi-gpu branch requires? Because when I compiling the code, it gives me some errors about CuDNN, so I want to confirm whether the errors caused by the version difference.

cypof:
@naibaf7 this PR only deals with data-parallelism, but you might be able to use work from #2219 to split the DAG on multiple devices.

dabilied:
well, I test this version on mnist , however it crash when "Snapshotting to models/" with GPU 0,1 mode
I0421 13:41:12.967577  6304 parallel.cpp:239] GPU 1 does not have p2p access to GPU 0
details:
    Check failed: Caffe::root_solver()
    @     0x7f621a8e842d  google::LogMessage::Fail()
    @     0x7f621a8ea1f3  google::LogMessage::SendToLog()
    @     0x7f621a8e8048  google::LogMessage::Flush()
    @     0x7f621a8eaa4e  google::LogMessageFatal::~LogMessageFatal()
I0421 13:41:35.904013  6304 solver.cpp:359] Snapshotting to models/lenet_iter_500.caffemodel
    @     0x7f6220280ea9  caffe::Solver<>::Snapshot()
    @     0x7f6220283728  caffe::Solver<>::Step()
    @     0x7f6218559822  thread_proxy
    @       0x3ceb807851  (unknown)
    @       0x3ceb4e767d  (unknown)
^C./train_lenet.sh: line 3:  6304 Aborted                 (core dumped) $TOOLS/caffe train --solver=lenet_solver.prototxt

cuda version:6.5
cudnn version : 2
am I need to install cuda-7 and GDK ?

cypof:
New update with fixes from @thatguymike and @sontran, and rebased on the latest data_layer #2351.

sunbaigui:
I've been testing this branch, but I got an error below:

F0427 05:53:55.895048  3297 solver.cpp:347] Check failed: Caffe::root_solver()
    @     0x7f28ec8d004c  google::LogMessage::Fail()
    @     0x7f28ec8cffa5  google::LogMessage::SendToLog()
    @     0x7f28ec8cf983  google::LogMessage::Flush()
    @     0x7f28ec8d290b  google::LogMessageFatal::~LogMessageFatal()
    @     0x7f28f26acb12  caffe::Solver<>::Snapshot()
    @     0x7f28f26b727d  caffe::Solver<>::Step()
    @     0x7f28ea50a283  thread_proxy
    @       0x39aa207851  (unknown)
    @       0x39a9ee767d  (unknown)

does this error related to my solver.prototxt ? @cypof @sontran
this is the command I used, 
GLOG_logtostderr=1 ./build/tools/caffe.bin train -solver ./quick_solver.prototxt

cypof:
This doesn't seem like it can happen with the latest version of this branch. Are you sure you are up to date?

wkal:
I test it again, and also get error about root_solver():
F0428 15:45:58.664613 20024 solver.cpp:347] Check failed: Caffe::root_solver()
F0428 15:45:58.664857 20026 solver.cpp:347] Check failed: Caffe::root_solver() F0428 15:45:58.664923 20025 solver.cpp:347] Check failed: Caffe::root_solver()
F0428 15:45:58.664857 20026 solver.cpp:347] Check failed: Caffe::root_solver() F0428 15:45:58.664923 20025 solver.cpp:347] Check failed: Caffe::root_solver()
    @     0x7f186ca3fdaa  (unknown)
    @     0x7f186ca3fdaa  (unknown)
    @     0x7f186ca3fdaa  (unknown)
    @     0x7f186ca3fce4  (unknown)
    @     0x7f186ca3fce4  (unknown)
    @     0x7f186ca3fce4  (unknown)
I0428 15:45:58.668508 20007 solver.cpp:359] Snapshotting to /home/caffe-multi-gpus/caffe-multi_gpu/examples/mnist/lenet_iter_5000.caffemodel
    @     0x7f186ca3f6e6  (unknown)
    @     0x7f186ca3f6e6  (unknown)
    @     0x7f186ca3f6e6  (unknown)
    @     0x7f186ca42687  (unknown)
    @     0x7f186ca42687  (unknown)
    @     0x7f186ca42687  (unknown)
    @     0x7f186d04993b  caffe::Solver<>::Snapshot()
    @     0x7f186d04993b  caffe::Solver<>::Snapshot()
    @     0x7f186d04993b  caffe::Solver<>::Snapshot()
    @     0x7f186d04a0c9  caffe::Solver<>::Step()
    @     0x7f186d04a0c9  caffe::Solver<>::Step()
    @     0x7f18646d2a4a  (unknown)
    @     0x7f186d04a0c9  caffe::Solver<>::Step()
    @     0x7f18646d2a4a  (unknown)
    @     0x7f18646d2a4a  (unknown)
    @     0x7f1863021182  start_thread
    @     0x7f1863021182  start_thread
    @     0x7f1863021182  start_thread
    @     0x7f186be25fbd  (unknown)
    @     0x7f186be25fbd  (unknown)
    @     0x7f186be25fbd  (unknown)
Aborted (core dumped)

wkal:
And I not installed the nvidia-ml library but I can comiple and run it, only when the program try to save the snapshot then it crashed. Any suggestions to solve this error?

wkal:
I test the single GPU mode, it can save the snapshot greatly, so I guess maybe the saving snapshot rountine not perfectly support multi-gpus mode.

wkal:
Oh! I have to say @cypof, your updating is so fast! Now I switch to the new version then the error is gone!  Thanks for your laborious work!  So please ignore my previous comment and I will delete it to avoid trouble other people who see this page.

cypof:
Rebased on top of the recent PRs. Runs should now be exactly reproducible thanks to the random seed initialization in #2367, and the round-robin data distribution in #2386.

thatguymike:
Still seeing differences run to run with a fixed random seed and convergence issues on CaffeNet training with full (compressed) ImageNet.

tools/caffe.cpp
```diff
@@ -12,13 +12,16 @@ using caffe::Blob;
...
+using caffe::Solver;
 using caffe::shared_ptr;
+using caffe::string;
 using caffe::Timer;
 using caffe::vector;
+using std::ostringstream;
 
-
-DEFINE_int32(gpu, -1,
-    "Run in GPU mode on given device ID.");
```
lukeyeager: Why not leave this flag? There's no sense in breaking backwards-compatibility unnecessarily, is there?

I vote for `-gpu 1` and `-gpus 1` being equivalent.

cypof:
New version with @thatguymike fixes, @lukeyeager suggestion for backward compatibility, and rebased on top of all PRs this time, including optional ones like #2397, #2395, and #2398. This is to make sure we are all testing the same thing.

tools/caffe.cpp
```diff
@@ -12,13 +12,18 @@ using caffe::Blob;
...
 using caffe::shared_ptr;
+using caffe::string;
 using caffe::Timer;
 using caffe::vector;
-
+using std::ostringstream;
 
 DEFINE_int32(gpu, -1,
-    "Run in GPU mode on given device ID.");
+    "Run in GPU mode on given device ID (Legacy switch, use -gpus).");
```
shelhamer: Having both `gpu` and `gpus` seems unecessary. Why introduce a new arg instead of redefining `gpu` as a string and including the logic to handle a single device number, a comma separated list, or "all" ?

lukeyeager:
> This is to make sure we are all testing the same thing

How am I supposed to test the same thing, exactly? Can I just `merge cypof/multi_gpu`, or do I need to merge other PR's first? There are so many branches and pull requests going on, I'm confused what needs to get merged in what order.

cypof:
@lukeyeager yes it was a bit out of control so now this PR contains all the necessary changes. Just checkout this branch and you have the latest.

lukeyeager:
Ok I pulled your code, built it with CUDA (but not with cuDNN), and tried running the test suite. I saw several errors with the `cypof/multi_gpu` branch that I don't see on the `BVLC/master` branch. Each of these errors causes a coredump, so there may be more tests that would fail that I'm not getting to. Randomizing the tests sure makes this confusing. :confused:

First, I'm seeing #2420 with your branch and not with master. In these tests:

* `AccuracyLayerTest/0.TestForwardWithSpatialAxes`
* `MathFunctionsTest/1.TestCopyCPU`

I get this error:
```
*** Aborted at 1432155417 (unix time) try "date -d @1432155417" if you are using GNU date ***
```

And In these tests:

* `MathFunctionsTest/0.TestCopyGPU`
*  `Im2colKernelTest/0.TestGPU`
* `MultinomialLogisticLossLayerTest/1.TestGradientCPU`

I get this error:
```
F0520 13:54:42.989051  9610 syncedmem.hpp:30] Check failed: error == cudaSuccess (11 vs. 0)  invalid argument
```

Sorry if this is just noise and I have a problem with my system. The tests seem to be passing on Travis, but as I understand it none of the GPU tests actually run in Travis. I'll keep trying different things to see if I'm doing something wrong.

thatguymike:
With the patches from @flx42 this looks good.  We continue to test, but races seem to be all squashed.  Perf is solid between 2 GPUs (~1.8x on 2xTitanX) and falls off going to 4 for smaller networks under strong scaling (e.g. not increasing the batch size as GPU count increases).  There is more tuning we can do, but we would like to see the guts of this (now large) PR go in.

There is a LOT of PRs stacked into this and I notice other PRs that touch similar code paths.  @shelhamer and @longjon when do you think you guys can start reviewing and merging?  This code is going to quickly get out of date and become a rebasing mess if integration isn't started soon and the tree continues to shift around.  ETA/commentary on plans to commit or reject?

shelhamer:
@cypof @thatguymike Parallelism is absolutely a priority and we plan to merge. While there are related efforts such as #2219, this PR satisfies the most constraints. Furthermore the layer-wise parallelism of #2219 and solver parallelism of this PR can be reconciled and combined to our advantage in the future.

We have reviewed different stages of this PR, and hope to do a pass at this last edition with fixes soon. Honestly, there is a research deadline that needs our focus but we will review as we can during experiments. I expect to merge #2366 #2367 #2383 #2368 this weekend to get this rolling.

Devs / users from the community: you can help by checking this out and trying it now. From my last review it looked close to ready.

thatguymike:
I completely understand.  This is a very large PR set.   Starting to walk through some of the intermediates that are more stand alone would be helpful.  Then we can deal with the rebase/rewrite of some of the conflicts around other commits that were recently made.

We continue to do numerical testing on our end, we likely have more testing resources than most, but any and all feedback is helpful.  We have a bunch of further optimization work pending, but will require some deeper changes to the code, so we would like to see the initial infrastructure settle.

A quick note on run to run reproducability.  cuDNN v2's dgrad and wgrad functions rely on atomics internally, so there will be some run to run drift because floating point adds will be applied in slightly different orders run to run.  The "numerical" envelope stays consistent though.  For reproducibility testing for now, the cuBLAS path should be used.

dxj19831029:
my training log contains those lines, Is this pr causing this problem?

```
I0529 07:53:26.910881  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.915447  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.920439  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.924780  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.932343  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.936542  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.940038  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.949637  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.958369  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.968183  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.974684  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.981868  5932 blocking_queue.cpp:50] Data layer prefetch queue empty
I0529 07:53:26.987926  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:26.993746  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:27.001727  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:27.012820  5946 blocking_queue.cpp:50] Waiting for data
I0529 07:53:27.021952  5946 blocking_queue.cpp:50] Waiting for data
```

thatguymike:
Your IO device isn't keeping up.  Generally for multi-GPU, we have found you need SSD or RAID arrays.  We should probably reduce the rate at which that message is being output.  What GPU(s)?  cuDNN path enabled?  What type of disk?  Is the database compressed or uncompressed JPEGs?

dxj19831029:
@thatguymike gtx 980 and gtx 780. 
cudnn is enabled.
disk is seagate 7200 64M sata 6Gb/s
data is compressed png format I think.

thatguymike:
And does this only print during the testing phase, or during training?  Do
you also see it running this branch with a single GPU? I have generally
only seen this with multiGPU + cuDNN where the forward pass is quite fast.
The warning is saying that we are not prefetching+decoding fast enough.

Your disk is quite slow at sustained reads (likely ~60MB/s) vs the RAID
(150MB/s) or SSD (500+MB/s) some of us are running.  Just to give you an
idea of data rates: uncompress LMDB running with CaffeNet with cuDNN
requires ~100MB/s per GPU to keep them fed during training.  Compressed
jpgs help a lot to reduce bandwidth requirements, but puts some stress on
the CPU to do decoding.  Compressed PNG is ~2.7:1 on average vs JPG which
is generally 8:1 - 16:1, but is lossy.

dxj19831029:
@thatguymike Indeed, it only prints during testing phase. 

Cool, thanks. :)

thatguymike:
We should debug a bit more or at a minimum work to reduce the verbosity somewhat.

lukeyeager:
Speaking of verbosity, I get about 100 copies of this message during testing:

> I0604 10:50:47.665694  9059 solver.cpp:314] Iteration 938, Testing net (#0)
I0604 10:50:47.670666  9059 blocking_queue.cpp:50] Data layer prefetch queue empty
I0604 10:50:47.672909  9059 blocking_queue.cpp:50] Data layer prefetch queue empty
I0604 10:50:47.675099  9059 blocking_queue.cpp:50] Data layer prefetch queue empty
...

I'm not sure what this means, and I'm not sure that I care. Why add this to the log at all? It's a shame [GLOG doesn't have a debug level](https://google-glog.googlecode.com/svn/trunk/doc/glog.html#severity) for messages like this.

cypof:
This message is useful in that it tells you you don't get the max speed you could be getting. But in practice it should never display, there is a bug on the data reader for the test phase. It's probably just a configuration setting for how much to prefetch that is wrong, but I haven't had time to look at it yet.

mausoomsarkar:
I tried the PR out. I am experiencing a consistent crash at 20th iteration with the following message
F0619 16:21:14.250097  3755 parallel.cpp:375] Check failed: error == cudaSuccess (77 vs. 0)  an illegal memory access was encountered
    @     0x7ff3b4741daa  (unknown)
    @     0x7ff3b4741ce4  (unknown)
    @     0x7ff3b47416e6  (unknown)
    @     0x7ff3b4744687  (unknown)
    @     0x7ff3b4db8390  caffe::P2PSync<>::on_gradients_ready()
    @     0x7ff3b4ddc912  caffe::Solver<>::Step()
    @     0x7ff3b4db6876  caffe::P2PSync<>::InternalThreadEntry()
    @     0x7ff3b4d17225  caffe::InternalThread::entry()
    @     0x7ff3ac6bfa4a  (unknown)
    @     0x7ff3ab2ca182  start_thread
    @     0x7ff3b3d2c47d  (unknown)
F0619 16:21:14.318222  3756 parallel.cpp:375] Check failed: error == cudaSuccess (77 vs. 0)  an illegal memory access was encountered
Aborted (core dumped)

thatguymike:
Can you provide more information on the network you are running, do you have cuDNN enabled or disabled?  We did find a pooling bug in cuDNN v2 when running the model zoo network in network model.  (As a note, your loss rate is off the charts, does this network behave on a single GPU?)

In ./include/caffe/util/cudnn.hpp can you make CUDNN_CHECK look like this and run again to help narrow down the actual error?


#define CUDNN_CHECK(condition) \
  do { \
    cudnnStatus_t status = condition; \
    CHECK_EQ(status, CUDNN_STATUS_SUCCESS) << " "\
      << cudnnGetErrorString(status); \
    cudaError_t error = cudaStreamSynchronize(cudaStreamDefault); \
    CHECK_EQ(error, cudaSuccess) << " " << cudaGetErrorString(error); \
  } while (0)

mausoomsarkar:
I was using cuDNN v2 with 4 titanX when I observed the crash. I compiled without cuDNN and it started working but with smaller batchsize...But it looked to me the memory footprint on gpu increased (used nvidia-smi). To verify I again compiled with cuDNN and ran with this smaller batchsize and it was stable. As for now the network on 4 GPU has run 8000 iterations with cuDNN v2. 
I will try the to reproduce the crash once my training finishes. As for the network, I am training a  siamese with 800x800 resolution.

mausoomsarkar:
Also because I am training such a big network I am having trouble in choosing a good mini batchsize. 
So to use main branch's gradient accumulation, I am trying to remove the conflicts while merging this PR.

mausoomsarkar:
Does the input layer on each GPU  read the same input? As in if my data is in the form of
Image1 Label1
Image2 Label2
Image3 Label3
Image4 Label4
batchsize: 4

And if I am using 4 GPUs then how is the read happening?
Is it like this 
GPU1 reads Image1 Label1
GPU2 reads Image2 Label2
GPU3 reads Image3 Label3
GPU4 reads Image4 Label4
GPU1 reads Image5 Label5
or 
Is it like GPU1 to 4 all read
Image1 Label1
Image2 Label2
Image3 Label3
Image4 Label4
??
I think the code here uses the 2nd option. Then the questions is how does going parallel help?

cypof:
It's the first option. data_reader puts one image/label pair in each queue.
Then each GPU reads its queue to build a batch, so the batch for GPU1 will
contain image1/image5/... Not sure what you mean by how does it help?

mausoomsarkar:
I am a bit confused because ImageDataLayer on every GPU seems to read every file. Yes the batchsize is divided for each ImageDataLayer but still it looked to me that on each GPU every file is uploaded. Am I missing something or is it not implemented for  ImageDataLayer?

cypof:
Ah ok, no it's not implemented for ImageDataLayer. Only LevelDB or LMDB.

mausoomsarkar:
I got another error dump .. see if this makes sense
F0624 22:34:07.300029 17713 cudnn_pooling_layer.cu:39] Check failed: error == cudaSuccess (77 vs. 0)  an illegal memory access was encountered
    @     0x7f32a0a61daa  (unknown)
    @     0x7f32a0a61ce4  (unknown)
    @     0x7f32a0a616e6  (unknown)
    @     0x7f32a0a61daa  (unknown)
    @     0x7f32a0a64687  (unknown)
    @     0x7f32a0a61ce4  (unknown)
    @     0x7f32a0a616e6  (unknown)
    @     0x7f32a114a2a1  caffe::CuDNNPoolingLayer<>::Backward_gpu()
    @     0x7f32a102ee77  caffe::Net<>::BackwardFromTo()
    @     0x7f32a0a64687  (unknown)
    @     0x7f32a102efe1  caffe::Net<>::Backward()
    @     0x7f32a114a2a1  caffe::CuDNNPoolingLayer<>::Backward_gpu()
    @     0x7f32a111c125  caffe::Solver<>::Step()
    @     0x7f32a102ee77  caffe::Net<>::BackwardFromTo()
    @     0x7f32a10f4cd6  caffe::P2PSync<>::InternalThreadEntry()
    @     0x7f32a102efe1  caffe::Net<>::Backward()
    @     0x7f32a10447b5  caffe::InternalThread::entry()
    @     0x7f32a111c125  caffe::Solver<>::Step()
    @     0x7f32987c1a4a  (unknown)
    @     0x7f32989d4182  start_thread
    @     0x7f32a111ca7e  caffe::Solver<>::Solve()
    @     0x7f32a004c47d  (unknown)
Aborted (core dumped)

cypof:
Added commits for the changes we talked about. I will rebase once the dependencies are ready.

shelhamer:
@cypof @thatguymike 

See my rebase of this branch on master is at https://github.com/shelhamer/caffe/tree/parallel. Sorry for the wait! I squashed b554119 and 9b10508 and did a little commit rewording. If this branch looks good to you, mirror it to your fork for a clean merge.

The last TODOs are testing and less aggressive logging.

cypof:
OK, looking at that. The logging change is in #2366. It just logs every 1000 messages.

shelhamer:
@cypof ok, I've pulled in the #2366 and pushed to parallel with latest head at eb2cfcc.

AIROBOTAI:
@cypof Thanks for your excellent and continuous work! One suggestion, could you please offer us some examples, i.e. the prototxt files, solver files to test your codes? I may first try on simple datasets like MNIST and CIFAR10, but I have no idea how to get start using your code. Thx in advance!

AIROBOTAI:
@shelhamer As you may have seen, I have asked @cypof  about the usage of his multi-gpu caffe above . Unluckily I guess he must be very busy now so that I haven't received the response. I see you are also an active developer for this pull request, so could you please tell me how to use the multi-gpu caffe? I have checked your branch of multi-gpu caffe, but I could not find the corresponding solver prototxt and train prototxt. I'm really dying to get help from you. Thanks a lot in advance!

cypof:
@AIROBOTAI actually this code should work with your existing solver and net proto files. You just need to specify a command line flag, e.g. caffe train -gpus 0,1 ...  Let us know how it goes.

AIROBOTAI:
@cypof Thank you so much! I shall use your code to test my dataset to see how it works. @shelhamer Thank you all the same! :)

cypof:
OK, updated and tested with latest updates from @thatguymike.

src/caffe/parallel.cpp
```diff
@@ -0,0 +1,449 @@
...
+    }
+  }
+  s.str("");
+  for (int i = 0; i < remaining.size(); ++i) {
+      s << (i ? ", " : "") << remaining[i];
+  }
+  DLOG(INFO) << "GPUs paired by P2P access, remaining: " << s.str();
+
+  // Group remaining
+  for (int d = 0; d < remaining.size(); ++d) {  // try to pair everyone
```
cdoersch: If remaining.size() is 8 or greater, won't this still fail to pair all GPUs?  It's rather difficult to follow this code, but you can verify that it fails with the following code snippet:

    remaining=[1]*8;
    d=0;
    while(d<len(remaining)):
      i=0;
      while(i<len(remaining)):
        j=i+1;
        while(j<len(remaining)):
          remaining.pop(j);
          j+=1;
          break
        i+=1;
      d+=1;
    print len(remaining)

The inner two loops serve to reduce the size of the list by a factor of two, pairing adjacent devices.  The outer loop index, however, increments by one for each execution of the inner loop; anytime d gets greater than one (i.e. requires more than 2 halvings), the loop will terminate with remaining.size()>1.

I think the outer loop should be a while(remaining.size()>1).  Furthermore, I'm not sure why the innermost loop is written as a for loop, since the break statement means it's guaranteed to execute only once.

thatguymike:
You are correct.  The loop was structured like that to have the same flow as the loops before it, but is pointless because of the basic pairing being done in the final loop.  You are also correct that we need to pair all remaining devices and the outer loop should loop until there is one device left.  I'll take a cleanup pass in the morning.  

Thanks for catching that.

michielschaap:
Hi, thanks a lot for all your work on this!

I just want to let you know that Caffe crashes in this loop on our machine with 6 GPUs. I haven't dived into the code, but I guess it has something to do with what cdoersch mentioned.

Thanks again!

shelhamer:
I've checked this on 2 x Titan X GPUs with the vanilla Caffe backend and see ~1.6x speedup. The losses for 1 vs. 2 GPUs agree to a tolerance of 10^-2 and multi-GPU training is reproducible with a random seed.

This looks good to me once the multi-GPU solver test can be called. It needs a command line arg to the test utility to set the number of devices. I'm fine with it assuming GPUs {0, ..., num-1} are the devices on which to conduct the test.

(For formatting, ideally the lint fix 0d54b11 would be rebased into its respective commits that made the lint errors since its commit noise otherwise.)

Thanks for all the work @cypof and @thatguymike!

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -333,20 +364,37 @@ class GradientBasedSolverTest : public MultiDeviceTest<TypeParam> {
...
-                              &updated_params);
-
-    // Reinitialize the solver and run K+1 solver iterations.
-    RunLeastSquaresSolver(learning_rate, weight_decay, momentum,
-                          iter_to_check + 1);
-
-    // Check that the solver's solution matches ours.
-    CheckLeastSquaresUpdate(updated_params);
+    // Loop for each possible set of devices
+    int available_devices = 1;
```
shelhamer: Instead of hardcoding this, how about a command line arg that sets the number of devices? Then the user can call `test_gradient_based_solver.testbin --num-devices=2` for instance to run on GPUs 0 and 1. Right now multi-device testing requires editing the code and recompiling and that's less than ideal.

cypof: The current setup will run on all possible sets between 1 and all available devices as returned by cuda. It should not be necessary to recompile.

shelhamer: Oh right -- I'm blind and missed line 371. Sorry!

cypof: No pblm, I will see with Mike to get an updated commit for the tree fix and I think we are good.

cdoersch:
I'm seeing about a 3.0x speedup on 4 Grid 520 GPUs vs a single GPU training CaffeNet; this test was run on a g2.8xlarge on EC2.  I did not try to set the seed the same way to ensure identical testing, but overall the training curves look comparable.

cypof:
I just cleaned the commits

shelhamer:
Thanks for the heads-up @cypof. We'll take a last look at this for merge on Wednesday.

AIROBOTAI:
For your reference, I have checked the speed-up ratio of training Imagenet using Googlenet on 4 TITAN X. The speed-up ratio is 2.2x. Because I used a batch size configuration of 32 (8 per GPU) to exactly match the batch size of blvc-googlenet, thus higher speed-up ratio might be expected when using larger batch size. The training result is comparable with the bvlc-googlenet.

naibaf7:
@cypof Would you be in on helping me to get this working with #2610 across OpenCL as well?

zbwglory:
@cypof @AIROBOTAI Hi. Thanks for this great PR. I got a little problem about the speed up ratio. I just want to test on MNIST. Could you tell me how to configure the solver and model to get that speed up ratio? And how to measure the speed up ratio?

thatguymike:
MNIST is a pretty small network to show much speedup, but it will function.  Just run Caffe as you normally do and set the gpus you want to run on via the commandline.  All of the multigpu setup is done in the 'caffe' binary.

cypof:
@naibaf7 I will be happy to help, but we our setup is NVidia right now so I can't spend too much time.

naibaf7:
@cypof OK I will pull the changes soon and get back to you if I have more specific requests - thanks.

bhack:
NVIDIA=CUDA?

naibaf7:
@bhack @cypof 
I think he meant that they just have an nVidia GPU setup right now and thus don't have any practical interest themselves in doing the port to OpenCL, which is fine.
I'd just like to see how the two backends would compare and if multi-GPU could even work across backends (OpenCL to CUDA). It's just for the thrill of it, because flexibility is always a good thing.

thatguymike:
The same techniques should generally work.  However, there are going to be some issues with getting good performance, notably no standard way to do P2P DMA transfers between devinces and being able to walk the topology of the PCIe setup.  Without P2P DMA, you get communication bound VERY quickly having to copy through host.  Walking the topology to figure out who can talk to each other at high speed is also part of this.  Multi-device support in OpenCL is a little trickier generally to deal with, at least if you need tight communication and synchronization between devices.

bhack:
http://streamcomputing.eu/blog/2015-04-18/gpudirect-and-directgma-direct-gpu-gpu-communication/

thatguymike:
Yes, I worked on the AMD extension many years ago and was a member of the committee for many years.  ;-)  My point was there is no standard way in OpenCL to do this.  The memory model is a little tight and you can see some of the issues crop up when using the extension, specifcally around map/unmap and synchronization on buffers between multiple devices and the host.

Again, all of the work @cypof and I did on this PR should functionally port to OpenCL.

bhack:
@thatguymike It is easy to create universal features on hardware on that you are the only vendor ;)

naibaf7:
@thatguymike 
Thanks for the input. Of course I'll start with a functional version first and then try to see which memory transfers can be done with P2P DMA. I have plenty of GPUs here to test with, so I'll see.

shelhamer:
This is very close to merge and our only reservation is how to refactor 346e262d644d0f83ed5d6b42ea87e17ab1a8971e for clarity since the use of member variables and the naming of `InternalThreadEntry()` and `InternalThread::entry()` are not immediately obvious. We should have an alternative commit ready soon.

naibaf7:
Most of the code seems compatible with #2610, except for device initialization and the synced memory. Those will have to be refactored to the DeviceContext abstraction of #2610. I'll do that after the merge.

sri-25:
Does anyone else have the issue of loss blowing up after few thousand iterations? I trained on ImageNet with 2 GeForce 980 GPUs, initially the losses were normal, decreasing from 7.xx then suddenly it blew up to ~87 after ~20k iterations. The batch size was 128 and the learning rate, momentum, stepsize etc were same as the bvlc_reference_caffenet solver. I had to resume from 10k and it trained better again. This is not a one time thing, it happened to me at least 4-5 times, and it has been reproduced in another machine as well (with Titan Xs).

cdoersch:
@sri-25 I haven't seen an issue like this.  Just to be sure (since you didn't mention it in your post), the network you're training works on a single GPU with twice the batch size?

shelhamer:
See #2870 for my packaging of this PR for merge -- the history preserves authorship and I only addressed my concerns about `InternalThread`, a minor TODO, and style fixes.

sri-25:
@cdoersch Yeah, it trains well on a single GPU (with 2x batch size). FYI, I'm training on ImageNet with bvlc_reference_caffenet solver/train_val.

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -82,7 +96,7 @@ class GradientBasedSolverTest : public MultiDeviceTest<TypeParam> {
        "        value: 1.0 "
        "      } "
        "      data_filler { "
-       "        type: 'gaussian' "
+       "        type: '" << (run.const_data ? "constant" : "gaussian") << "' "
```
shelhamer: The `targets` top needs to always be Gaussian or else the test is ineffective. The constant vs. gaussian condition should apply to the filler for `data`. Switching this reveals that the `CheckAccumulation` test is broken since it doesn't set constant data. More ominously, it reveals that all the `TestLeastSquares*` for multi-GPU. Let's follow up on #2870.

shelhamer:
Closing since this is continued on in #2870.

loulansuiye:
@thatguymike I replace the tensor RT 2.x with the version 3.x and cudnn 7.x but the caffe Makefile.config can not find the cudnn now . Did you meet that similarity questions? or other suggestion?,thank you

PR #2115, opened 2015-03-13
Remove bogus implementation of SigmoidCrossEntropyLossLayer's Forward_gpu

longjon:
Not sure what happened here; `Foward_gpu` was a verbatim copy of `Forward_cpu`. Instead gpu should fall back due to lack of a proper forward GPU implementation.

### DIFF
--- a/src/caffe/layers/sigmoid_cross_entropy_loss_layer.cu
+++ b/src/caffe/layers/sigmoid_cross_entropy_loss_layer.cu
@@ -9,26 +9,6 @@
 namespace caffe {
 
 template <typename Dtype>
-void SigmoidCrossEntropyLossLayer<Dtype>::Forward_gpu(
-    const vector<Blob<Dtype>*>& bottom, const vector<Blob<Dtype>*>& top) {
-  // The forward pass computes the sigmoid outputs.
-  sigmoid_bottom_vec_[0] = bottom[0];
-  sigmoid_layer_->Forward(sigmoid_bottom_vec_, sigmoid_top_vec_);
-  // Compute the loss (negative log likelihood)
-  const int count = bottom[0]->count();
-  const int num = bottom[0]->num();
-  // Stable version of loss computation from input data
-  const Dtype* input_data = bottom[0]->cpu_data();
-  const Dtype* target = bottom[1]->cpu_data();
-  Dtype loss = 0;
-  for (int i = 0; i < count; ++i) {
-    loss -= input_data[i] * (target[i] - (input_data[i] >= 0)) -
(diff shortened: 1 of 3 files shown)


src/caffe/layers/sigmoid_cross_entropy_loss_layer.cu
```diff
@@ -11,21 +11,9 @@ namespace caffe {
...
-  const Dtype* input_data = bottom[0]->cpu_data();
-  const Dtype* target = bottom[1]->cpu_data();
-  Dtype loss = 0;
-  for (int i = 0; i < count; ++i) {
-    loss -= input_data[i] * (target[i] - (input_data[i] >= 0)) -
-        log(1 + exp(input_data[i] - 2 * input_data[i] * (input_data[i] >= 0)));
-  }
-  top[0]->mutable_cpu_data()[0] = loss / num;
+  // Since we've only implemented a GPU backward, fall back to CPU
+  // (we need to do this explicitly for the instantiation macro to work).
```
jeffdonahue: You could use `INSTANTIATE_LAYER_GPU_BACKWARD` (I think?). But yeah, good catch -- dunno why I wrote a Forward_gpu...

src/caffe/layers/sigmoid_cross_entropy_loss_layer.cu
```diff
@@ -11,21 +11,9 @@ namespace caffe {
 template <typename Dtype>
 void SigmoidCrossEntropyLossLayer<Dtype>::Forward_gpu(
     const vector<Blob<Dtype>*>& bottom, const vector<Blob<Dtype>*>& top) {
-  // The forward pass computes the sigmoid outputs.
-  sigmoid_bottom_vec_[0] = bottom[0];
-  sigmoid_layer_->Forward(sigmoid_bottom_vec_, sigmoid_top_vec_);
```
sguada: When I looked at this code some time ago, I thought you did it because it will use the GPU forward of the sigmoid layer and be faster.

jeffdonahue: That might have been what I was thinking, but `Forward` will check `Caffe::mode()` and decide whether to call `Forward_cpu` or `gpu` based on that.

longjon:
Updated to use `INSTANTIATE_LAYER_GPU_BACKWARD`... thanks for the tip @jeffdonahue.

longjon:
Note that it's also necessary to use `STUB_GPU_BACKWARD`.

jeffdonahue:
Thanks for the cleanup!

PR #2126, opened 2015-03-15
XCov loss layer

briancheung:
XCov Loss Layer based on our recent paper:
http://arxiv.org/abs/1412.6583

Coded with:
@jackculpepper @JesseLivezey

### DIFF
--- a/include/caffe/filler.hpp
+++ b/include/caffe/filler.hpp
@@ -156,6 +156,20 @@ class XavierFiller : public Filler<Dtype> {
   }
 };
 
+template <typename Dtype>
+class CountFiller : public Filler<Dtype> {
+ public:
+  explicit CountFiller(const FillerParameter& param)
+      : Filler<Dtype>(param) {}
+  virtual void Fill(Blob<Dtype>* blob) {
+    Dtype* data = blob->mutable_cpu_data();
+    for (int i = 0; i < blob->count(); ++i) {
+      data[i] = i/Dtype(blob->count());
+      // std::cout << data[i] << " ";
+    }
+    // std::cout << std::endl;
+  }
+};
 
 /**
  * @brief Get a specific filler from the specification given in FillerParameter.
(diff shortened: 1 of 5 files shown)


koth:
Can we have a mnist example  use this loss layer to train

JesseLivezey:
@koth, that's on our TODO list.

koth:
@JesseLivezey  cool， thanks!

briancheung:
We added MNIST example for training with the XCov cost layer. Generates the transformations described in the paper (sample image below, varying z variables)

index

sunbaigui:
mark

jyegerlehner:
@briancheung nice work; thanks for sharing. 
One comment: the paper says that the implementation uses pylearn2. But there's no Caffe citation, unless I missed it.

briancheung:
The original implementation of this was in Pylearn2 (the very first one was actually in numpy). That library was used to generate the results for the paper on arXiv. Our codebase for that implementation isn't as clean as this Caffe impelementation. Pylearn2 has to be modified quite a bit if you're doing anything besides supervised classification. We decided to release the Caffe version as it doesn't break any of the underlying Caffe design.

shuzhangcasia:
@briancheung Thanks for sharing the code. And I have one question with your paper. In your experiment on Multi-PIE, if you use the identity as the observed variable instead, will your network be able to learn the latent variations, such as pose and illuminations?

briancheung:
@shuzhangcasia We haven't tried this experiment, but my intuition would be yes. But if you only use identity as your supervised variable, I would expect the other latent variations such as pose and illumination to remained 'entangled' with each other albeit disentangled with variations in identity.

shuzhangcasia:
@briancheung Thanks for the comment, I would expect the same. One more thing, would you please sent me a copy of your supplementary material for this paper? I couldn't find it anywhere.

PR #2127, opened 2015-03-15
Fix invalid syntax on classify.py

kentashoji:


longjon:
Thanks @kentashoji, not sure how this happened. Ideally we should have better test coverage of these things...

PR #2136, opened 2015-03-16
Minimal MSVC compatibility adjustments

classner:
Second attempt to submit my minimal changeset for full MSVC compatibility.

These adjustments are as non-invasive and minimal as possible, in the hope to be merged into the main branch. This would enable "external" build systems to build caffe out-of-the box on Windows (for example the [caffe-brewer](https://github.com/ChrislS/caffe-brewer).

### DIFF
--- a/src/caffe/test/test_caffe_main.cpp
+++ b/src/caffe/test/test_caffe_main.cpp
@@ -34,6 +34,7 @@ int main(int argc, char** argv) {
   cudaGetDevice(&device);
   cout << "Current device id: " << device << endl;
   cudaGetDeviceProperties(&CAFFE_TEST_CUDA_PROP, device);
+  cout << "Current device name: " << CAFFE_TEST_CUDA_PROP.name << endl;
 #endif
   // invoke the test.
   return RUN_ALL_TESTS();
--- a/tools/caffe.cpp
+++ b/tools/caffe.cpp
@@ -183,7 +183,13 @@ int train() {
       s << (i ? ", " : "") << gpus[i];
     }
     LOG(INFO) << "Using GPUs " << s.str();
-
+#ifndef CPU_ONLY
+    cudaDeviceProp device_prop;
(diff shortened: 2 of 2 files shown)


Nerei:
@ChrislS Have you tried to execute tests? I suspect it may not work. If I am not mistaken, `__declspec(dllexport)` is required everywhere.

Nerei:
@ChrislS Btw, great work with [caffe-brewer](https://github.com/ChrislS/caffe-brewer)!!! This is exactly what I wanted to create but haven't found time slot yet. No, this is better that what I wanted ;-)

I found answer on my previous question, you use `/OPT:NOREF` option and static linking.

ChrislS:
Hi @Nerei,
all the tests are passing on Windows (take into account the note about having at least 1TB of disk space available)! The system is in productive use and I use the caffe-brewer to link the object files to another project.

Sure, it would be a lot more convenient if Microsoft offered a way to force full static linking. However, as I wrote in the README.MD, /OPT:NOREF did not force the linkage of static global objects, hence resulting in unregistered layers.

willyd:
@ChrislS 

The static linking of caffe object as given me a hard time. The /OPT:NOREF is just useless. The one thing that comes closest too the whole-archive option is the "Use Library Dependency Inputs" which forces a project to link the object files of another instead of the produced library.

CMake does not support this option yet (see this post: https://public.kitware.com/Bug/bug_relationship_graph.php?bug_id=13032&graph=relation) so it's not possible to use this without modifying the VS projects by and after CMake generation. Even then you can use this only within the same solution so you cannot install the library and use it from a consuming project.

The workaround I found is to modify the layer ans class instantiation macros to define additional smybols that can be easily referenced to force symbol linking. So I added a post build script that parses the Caffe sources and generates a header file that when included in the consuming projects forces the linking of the objects. This is suboptimal but works. See my PR for details https://github.com/BVLC/caffe/pull/2538

Issue #462, opened 2014-05-28
How to classify 1-channel inputs with the python wrapper (LeNet deploy)


pjstimac:
Greetings,

I'm evaluating Caffe for a commercial application. I have compiled Caffe and pycaffe and matcaffe and everything appears to be good: the installation passed all tests that are run using make runtest. 

Now I want to use the python wrapper to do a simple image classification. I am trying to use a trained mnist network to do the classification (staying away from the imagenet example due to the commercial restriction associated with obtaining the pretrained network).

For my lenet_deploy.prototxt I changed the very beginning and end of the example lenet_test.prototxt. In the beginning of lenet_deploy.prototxt I have the following before the first convolutional layer for my non-RGB single test image of size [28,28]:

name: "LeNet-deploy"
input: "data"
input_dim: 1
input_dim: 1
input_dim: 28
input_dim: 28

At the end of lenet_deploy.prototxt I have:

layers {
  name: "prob"
  type: SOFTMAX
  bottom: "ip2"
  top: "prob"
}

For the trained model I simply used the lenet_iter_10000 result, which seems very good, and renamed it to lenet_pretrained. Now I want to classify my non-RGB image of size [28 28] by running the python wrapper with the options listed below:

python classify.py 
--model_def='/usr/local/caffe/examples/mnist/lenet_deploy.prototxt' --pretrained_model='/usr/local/caffe/examples/mnist/lenet_pretrained' 
--gpu 
--center_only 
--images_dim='28,28' 
--mean_file='' 
/usr/local/caffe/examples/images/cat2.jpg 
/usr/local/caffe/examples/mnist/looker

Launching the python script results in the network being read in, but it fails in pycaffe.py line 66 in _Net_forward:

self.blobs[in_].data[...] = blob
ValueError: could not broadcast input array from shape (1,3,28,28) into shape (1,1,28,28)

I interpret this error to mean that the code (or the pretrained network?) is expecting a 3 channel image (RGB or the like), but it is only seeing a single channel image. If that is the right interpretation how do I change this? I've spent a lot of time looking at the various posts on this site, but I haven't been able to get past this point. My apologies if I'm missing something super obvious here.

Many thanks for your time.

ToruHironaka:
Same here,  I got rid of the error: "ValueError: could not broadcast input array from shape (64,3,28,28) into shape (64,1,28,28)"  by change io.py file line 27 from (img, (1, 1, 3)) to (img, (1, 1, 1)) but I got another error below. 

[...]

PR #2144, opened 2015-03-17
improved to load RGB image as grayscale image

tishibas:
You can load RGB image as grayscale by using python io function, load_image.

I had some troubles with loading RGB image as grayscale image in python interface.
We need this improvment when trying to classify by models trained with grayscale images such as mnist.
So I got this solution and found a person who has same suggestion in #462.


jeffdonahue:
Thanks @tishibas, LGTM

PR #2145, opened 2015-03-17
Fixed potentially dangerous data offsets bug.

gustavla:
This fixes several instances of a dormant bug related to how data pointers are updated with offset values in the pooling kernels.

The (non-cuDNN) pooling layer updates its memory pointers with the appropriate offsets for each CUDA kernel, e.g.:

```
int offset = ...;
top_diff += offset;
```

The offset is always relative to the beginning of the array, so this could not be applied twice since `top_diff` would no longer be the correct anchor point.

This was not a problem when it was first introduced (65b44b62d206b2fa04c31b1958c46b59cffacdf7), since there were no for loops in the kernels. Then, when loops were added (6372622c6c4977f020195e0b8105f31a24e73cb9), it became theoretically possible for kernels to do more than one job. Currently, the block/threads are set so that no kernel will iterate more than once, thus making this bug inactive. Nevertheless, it is worth fixing for two reasons: (1) It could become an active bug if the calling code is changed, and (2) it is confusing when reading the code.

To fix this I added new offset pointers that are scoped inside the for loops. A few times I introduced new variables to make things clearer and avoid lines getting longer than 80 columns.

Note: This used to be a problem in im2col as well, but was already fixed (75aa8df452894de1107e9321cbd292e00ac2ce89).

### DIFF
--- a/src/caffe/layers/pooling_layer.cu
+++ b/src/caffe/layers/pooling_layer.cu
@@ -28,12 +28,13 @@ __global__ void MaxPoolForward(const int nthreads, const Dtype* bottom_data,
     wstart = max(wstart, 0);
     Dtype maxval = -FLT_MAX;
     int maxidx = -1;
-    bottom_data += (n * channels + c) * height * width;
+    int offset = (n * channels + c) * height * width;
+    const Dtype* offset_bottom_data = bottom_data + offset;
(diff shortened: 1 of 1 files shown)


shelhamer:
Thanks for the report and fix @gustavla! This dormant bug was found in a few more places and fixed by #2493.

PR #2154, opened 2015-03-19
[python] fix an output error in classify.py on tutorial branch

deercoder:
I come across an obvious typo in classify.py on the tutorial branch, which is as this commit shows, and it will give us error when running classify.py, on master branch there will be no problem. So I fixed it and hope it will be fixed in later version.


deercoder:
The build is failed check, but I think this commit have nothing to do with it, as it doesn't needs this file when building, it is caused by a configuration error.

deercoder:
Fixed in another pull request[#2127](https://github.com/BVLC/caffe/pull/2127), close

PR #2163, opened 2015-03-19
CUDA kernels for {Slice,Concat}Layer

jeffdonahue:
This removes the loop over slices/concats in the GPU implementation of Slice and ConcatLayer, which makes them significantly faster depending on the workload (e.g., if slice size is 1, and number of slices is large, the CPU implementation is probably faster than the current GPU implementation, or roughly the same speed).


shelhamer:
Thanks for the kernels Jeff!

PR #2165, opened 2015-03-20
Always call Layer::Reshape in Layer::Forward

longjon:
There are no cases where `Forward` is called without `Reshape`, so we can simplify the call structure. This will also simplify a future PR.


jeffdonahue:
LGTM, thanks for the simplification.

PR #2166, opened 2015-03-20
Add option to take parameters from bottoms

longjon:
After #2165.

This PR is the counterpart to #2079; together they propose a (temporary?) solution to #1474.

An option is added to `LayerParameter` called `param_bottoms` which gives a number of parameter blobs to take from (additional) bottoms. This allows one to break the artificial division between bottoms and parameters. It's implemented in the `Layer` base class, so one can use existing layers unmodified in this way.

One issue with this: while `Forward_*`, `Backward_*`, and `LayerSetUp` calls are wrapped by calls from `Layer`, `Reshape` is not, and is directly passed `bottom`. Thus we cannot perform any option-based finagling of bottoms without changing this call. The somewhat hacky solution here is to add an extra nullary wrapper method, `ReshapeOnly`, that lets the layer decide for itself what its bottoms and tops are. A better (but more aggressive) solution might be to go ahead and change all the method types to reflect the fact that `bottom` and `top` aren't actually allowed to change between set up and forward/backward/reshape calls.

This PR provides an immediate, but perhaps unsatisfying solution to #1474. The issue of extra semantic/naming information between bottoms and params is not addressed (but maybe it's not as important given a higher-level way to write nets, e.g., #2086). Neither is the issue of computing param sizes, which is a little less automatic when using this option. (Further, size checks don't exist for params-from-bottoms).

We could merge this PR, or something like it, as a temporary way of making layers more flexible until a final solution is built. Or we could keep this as a PR until the param/bottom distinction is cleanly removed, and layer code is updated accordingly.

Finally, a usage note: while you can write any number for `param_bottoms`, the only values that will work are zero or the number of parameter blobs expected by the layer (which is not otherwise specified). It's also possible to use a non-zero `param_bottoms` while also loading params, but that's not likely to work and should probably just be disabled.

### DIFF
--- a/include/caffe/layer.hpp
+++ b/include/caffe/layer.hpp
@@ -1,6 +1,9 @@
 #ifndef CAFFE_LAYER_H_
 #define CAFFE_LAYER_H_
 
+// For null_deleter (can use boost/core/null_deleter.hpp for boost >= 1.56).
+#include <boost/serialization/shared_ptr.hpp>
+
 #include <algorithm>
 #include <string>
 #include <vector>
@@ -12,6 +15,8 @@
 #include "caffe/util/device_alternate.hpp"
 
 namespace caffe {
+
+using boost::serialization::null_deleter;
 
 /**
  * @brief An interface for the units of computation which can be composed into a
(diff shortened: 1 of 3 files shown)


sguada:
@longjon why not change the signature of layers to `(params, bottoms, tops)` Otherwise it is going to be a bit confusing to keep the order of param and bottoms aligned.

Another option could be to use a map<string name, Blob bottom> that allows to refer to params and bottoms by name.

shelhamer:
Closing as #2219 was closed.

itkinai:
@sguada can you explain how to set wieghts in InnerProduct as parameter layer with prototxt, or its works only in code with map<string name, Blob bottom> ?

yuanyc06:
@longjon @sguada would you mind give an example about how to set the weight of a layer (e.g. fc) as the bottom of another layer via the Parameter Layer? Thank you very much!

PR #2167, opened 2015-03-20
Blocking queue (from @cypof, with modifications)

longjon:
This is the blocking queue taken from the parallel branch, modified to remove some features and add others. (By the way, I looked for a more "standard" implementation of this, e.g., in boost, but did not find one.)

I'll splitting this out from a future PR for easier review. There's no need to consider it until that PR is also agreed upon, since this serves no other purpose yet.

@cypof, as with #2067, let me know that the author information is as you prefer.

### DIFF
--- a/include/caffe/util/blocking_queue.hpp
+++ b/include/caffe/util/blocking_queue.hpp
@@ -24,10 +24,20 @@ class blocking_queue {
     return queue_.empty();
   }
 
+  void wait_for_empty() {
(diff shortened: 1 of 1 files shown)


cypof:
@longjon author info is fine, thanks. I will fix my config for future commits.

sguada:
@longjon you may want to add a `max_size` and block `push` when the queue is full.

shelhamer:
Closing as replaced by #4563

PR #2168, opened 2015-03-20
Remove spurious inclusions of net.hpp

longjon:
The includes of `net.hpp` removed by this PR appear to be totally spurious; in fact there should be no need for layers to know about `Net` at all.

This will also make it possible for `Net` to use `boost::thread` in the future, without causing any `nvcc` trouble by inclusion in layer CUDA code.

### DIFF
--- a/include/caffe/data_layers.hpp
+++ b/include/caffe/data_layers.hpp
@@ -14,7 +14,6 @@
 #include "caffe/filler.hpp"
 #include "caffe/internal_thread.hpp"
 #include "caffe/layer.hpp"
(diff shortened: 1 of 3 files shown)


shelhamer:
Won't this conflict with #1974 needed for #1976?

longjon:
@shelhamer: This patch should be a reasonable cleanup even with #1974, but yes, the future work that depends on this will conflict with #1974; so be it.

shelhamer:
Fair enough!

jeffdonahue:
Thanks for the cleanup @longjon!

PR #2183, opened 2015-03-23
changes layer name from IMAGE_DATA to IMAGEDATA

SnShine:


jeffdonahue:
Thanks -- we really appreciate your help improving the docs! Since a lot of the current staleness of the docs is due to my PRs (where I should have updated them), I tried to fix most of these in #2201, but please let me know (or send another PR) if you decide to look over it and notice I missed something there.  I'll close this since it should be `ImageData` after #1694, but future docs fixes and improvements are most welcome.

PR #2191, opened 2015-03-24
Bug fix: AutoTop layer can break Python net.outputs

tnarihi:
This PR fixes a bug I faced in `net.outputs` in Python, which is caused by automatic top blobs.
I modified to create a unique automatic blob name for all blobs that are created if  `layer.AutoTopBlobs() == True`. 
Please see the commit message for details.


include/caffe/net.hpp
```diff
@@ -255,6 +258,8 @@ class Net {
   size_t memory_used_;
   /// Whether to compute and display debug info for the net.
   bool debug_info_;
+  /// Auto top blob index to be incremented in GetAutoTopBlobName
+  size_t current_auto_top_blob_index_;
```
jeffdonahue: I think this needs to be explicitly initialized or its value will technically be undefined?

jeffdonahue:
Thanks @tnarihi! LGTM other than initialization issue as commented above.

PR #2192, opened 2015-03-25
Remove scikit-learn dependency

lukeyeager:
As far as I can tell, `scikit-learn` is never used. I get no results when I search for either "scikit-learn" or "sklearn." If it is really not used, can we remove it from requirements.txt? This is the only pip dependency which doesn't have a debian installer in Ubuntu 14.04 (besides `Pillow>=2.7.0`, which I'll address separately).

FWIW, all tests still pass without `scikit-learn` installed, but I don't trust the python test coverage much.

### DIFF
--- a/examples/hdf5_classification.ipynb
+++ b/examples/hdf5_classification.ipynb
@@ -40,6 +40,7 @@
       "import shutil\n",
       "import tempfile\n",
       "\n",
+      "# You may need to 'pip install scikit-learn'\n",
       "import sklearn\n",
       "import sklearn.datasets\n",
       "import sklearn.linear_model"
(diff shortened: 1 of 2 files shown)


shelhamer:
`sklearn` is used in our [vector data / HDF5 example](http://nbviewer.ipython.org/github/BVLC/caffe/blob/master/examples/hdf5_classification.ipynb) for comparison. The dependency could be noted in that example instead I suppose, since Caffe itself never makes use of the module as you noted.

lukeyeager:
> sklearn is used in our vector data / HDF5 example for comparison

Ah, I didn't see that. Strange that it didn't show up in a repository-wide search.

lukeyeager:
I added a comment in the example, is that sufficient?

shelhamer:
Yeah, this seems fine. @longjon how do you feel about not forcing dependencies needed in an example?

longjon:
@shelhamer I'm fine with not forcing dependencies for examples if clearly noted.

PR #2193, opened 2015-03-25
Plain Datumfile input format for minimum memory usage

immars:
When training with leveldb/lmdb, memory increases linearly with iterations and even with Next()s for DataLayer's `rand_skip`.

Here's simple plain datum file format via std::fstream to address this issue.

Pros
- RAM usage basically stay constant (<1G, on googlenet small batches) during training as expected
- datum file size between leveldb and lmdb format
- no noticeable impact on training speed because of prefetch
- concurrent read for multiple process

Cons
- no random read. But caffe does not need random key-value access anyway.

### DIFF
--- a/include/caffe/util/db.hpp
+++ b/include/caffe/util/db.hpp
@@ -191,8 +191,10 @@ class DatumFileCursor : public Cursor {
     SeekToFirst();
   }
   virtual ~DatumFileCursor() {
-    if (in_ && in_->is_open()) {
+    if (in_ != NULL && in_->is_open()) {
       in_->close();
+      delete in_;
+      in_ = NULL;
     }
   }
   virtual void SeekToFirst();
@@ -245,8 +247,9 @@ class DatumFileDB : public DB {
     this->can_write_ = mode != db::READ;
   }
   virtual void Close() {
-    if (out_) {
+    if (out_ != NULL) {
       out_->close();
+      delete out_;
       out_ = NULL;
     }
   }
--- a/include/caffe/util/db.hpp
+++ b/include/caffe/util/db.hpp
@@ -182,84 +182,82 @@ class LMDB : public DB {
 };
 
 
-#define MAX_BUF 10485760  // max entry size
+#define MAX_BUF 104857600  // max entry size
 class DatumFileCursor : public Cursor {
  public:
   explicit DatumFileCursor(const string& path) {
-    this->path = path;
-    in = NULL;
+    this->path_ = path;
+    in_ = NULL;
     SeekToFirst();
   }
   virtual ~DatumFileCursor() {
+    if (in_ && in_->is_open()) {
+      in_->close();
+    }
   }
   virtual void SeekToFirst();
 
   virtual void Next();
 
   virtual string key() {
-    if (!valid()) {
-      LOG(WARNING) << "not valid at key()";
-      return "";
-    }
-    return _key;
+    CHECK(valid()) << "not valid state at key()";
+    return key_;
   }
   virtual string value() {
-    if (!valid()) {
-      LOG(WARNING) << "not valid at value()";
-      return "";
-    }
-    return _value;
+    CHECK(valid()) << "not valid state at value()";
+    return value_;
   }
 
   virtual bool valid() { return valid_; }
 
  private:
-  string path;
-  std::ifstream* in;
+  string path_;
+  std::ifstream* in_;
   bool valid_;
 
-  string _key, _value;
+  string key_, value_;
 };
 
 class DatumFileTransaction : public Transaction {
  public:
   explicit DatumFileTransaction(std::ofstream* out) {
-    this->out = out;
+    this->out_ = out;
   }
 
   virtual void Put(const string& key, const string& value);
 
   virtual void Commit() {
-    out->flush();
+    out_->flush();
   }
 
  private:
-  std::ofstream* out;
+  std::ofstream* out_;
   DISABLE_COPY_AND_ASSIGN(DatumFileTransaction);
 };
 
 
 class DatumFileDB : public DB {
  public:
-  DatumFileDB() { out = NULL; }
+  DatumFileDB() { out_ = NULL; can_write_ = false;}
   virtual ~DatumFileDB() { Close(); }
   virtual void Open(const string& source, Mode mode) {
-    path = source;
-    this->can_write = mode != db::READ;
+    path_ = source;
+    this->can_write_ = mode != db::READ;
   }
   virtual void Close() {
-    if (out) {
-      out->close();
-      out = NULL;
+    if (out_) {
+      out_->close();
+      out_ = NULL;
     }
   }
-  virtual DatumFileCursor* NewCursor() {return new DatumFileCursor(this->path);}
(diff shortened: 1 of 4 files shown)


include/caffe/util/db.hpp
```diff
@@ -181,6 +181,87 @@ class LMDB : public DB {
   MDB_dbi mdb_dbi_;
 };
 
+
+#define MAX_BUF 10485760  // max entry size
```
sguada: Why this value?

include/caffe/util/db.hpp
```diff
@@ -181,6 +181,87 @@ class LMDB : public DB {
...
+class DatumFileCursor : public Cursor {
+ public:
+  explicit DatumFileCursor(const string& path) {
+    this->path = path;
+    in = NULL;
+    SeekToFirst();
+  }
+  virtual ~DatumFileCursor() {
```
sguada: Should it close the input file?

include/caffe/util/db.hpp
```diff
@@ -181,6 +181,87 @@ class LMDB : public DB {
...
+  }
+  virtual void SeekToFirst();
+
+  virtual void Next();
+
+  virtual string key() {
```
sguada: It should fail instead of giving a warning. Ex:
CHECK(valid()) << "not valid state";

include/caffe/util/db.hpp
```diff
@@ -181,6 +181,87 @@ class LMDB : public DB {
...
+    if (!valid()) {
+      LOG(WARNING) << "not valid at key()";
+      return "";
+    }
+    return _key;
+  }
+  virtual string value() {
+    if (!valid()) {
```
sguada: Same as above

include/caffe/util/db.hpp
```diff
@@ -181,6 +181,87 @@ class LMDB : public DB {
...
+  }
+
+  virtual bool valid() { return valid_; }
+
+ private:
+  string path;
+  std::ifstream* in;
+  bool valid_;
+
+  string _key, _value;
```
sguada: Use the convention `key_` `value_` for private variables

src/caffe/util/db.cpp
```diff
@@ -59,12 +59,95 @@ void LMDBTransaction::Put(const string& key, const string& value) {
...
+    }
+    LOG(INFO) << "reset ifstream" << path;
+    in = new std::ifstream(path.c_str(),
+            std::ifstream::in|std::ifstream::binary);
+    Next();
+  }
+
+void DatumFileCursor::Next() {
+  valid_ = false;
+  if (!in->is_open()) {
```
sguada: Should fail 
`CHECK(in->is_open()) << "file is not open!" << path`;

src/caffe/util/db.cpp
```diff
@@ -59,12 +59,95 @@ void LMDBTransaction::Put(const string& key, const string& value) {
...
+    LOG(WARNING) << "file not open!" << path;
+  }
+  uint32_t record_size, key_size, value_size;
+  in->read(reinterpret_cast<char*>(&record_size), sizeof record_size);
+  if (in->gcount() != (sizeof record_size) || record_size > MAX_BUF) {
+    if (!in->eof()) {
```
sguada: These are clear errors from which you cannot recover, so fail in this cases.

src/caffe/util/db.cpp
```diff
@@ -59,12 +59,95 @@ void LMDBTransaction::Put(const string& key, const string& value) {
...
+      LOG(WARNING) << "record_size read error: gcount\t"
+          << in->gcount() << "\trecord_size\t" << record_size;
+    }
+    return;
+  }
+  in->read(reinterpret_cast<char*>(&key_size), sizeof key_size);
+  if (in->gcount() != sizeof key_size || key_size > MAX_BUF) {
+    LOG(WARNING) << "key_size read error: gcount\t"
```
sguada: Same as above

src/caffe/util/db.cpp
```diff
@@ -59,12 +59,95 @@ void LMDBTransaction::Put(const string& key, const string& value) {
...
+        << in->gcount() << "\tkey_size\t" << key_size;
+    return;
+  }
+  _key.resize(key_size);
+  in->read(&_key[0], key_size);
+  if (in->gcount() != key_size) {
+    LOG(WARNING) << "key read error: gcount\t"
```
sguada: Same as above

src/caffe/util/db.cpp
```diff
@@ -59,12 +59,95 @@ void LMDBTransaction::Put(const string& key, const string& value) {
...
+        << in->gcount() << "\tkey_size\t" << key_size;
+    return;
+  }
+  in->read(reinterpret_cast<char*>(&value_size), sizeof value_size);
+  if (in->gcount() != sizeof value_size || value_size > MAX_BUF) {
+    LOG(WARNING) << "value_size read error: gcount\t"
```
sguada: Same as above

src/caffe/util/db.cpp
```diff
@@ -59,12 +59,95 @@ void LMDBTransaction::Put(const string& key, const string& value) {
...
+        << in->gcount() << "\tvalue_size\t" << value_size;
+    return;
+  }
+  _value.resize(value_size);
+  in->read(&_value[0], value_size);
+  if (in->gcount() != value_size) {
+    LOG(WARNING) << "value read error: gcount\t"
```
sguada: Same as above

src/caffe/util/db.cpp
```diff
@@ -59,12 +59,95 @@ void LMDBTransaction::Put(const string& key, const string& value) {
...
+    uint32_t key_size = key.size(), value_size = value.size();
+    uint32_t record_size = key_size + value_size
+        + sizeof key_size + sizeof value_size;
+    out->write(reinterpret_cast<char*>(&record_size), sizeof record_size);
+    out->write(reinterpret_cast<char*>(&key_size), sizeof key_size);
+    out->write(key.data(), key_size);
+    out->write(reinterpret_cast<char*>(&value_size), sizeof value_size);
+    out->write(value.data(), value_size);
+  } catch(std::ios_base::failure& e) {
+    LOG(WARNING) << "exception: "
```
sguada: Use `LOG(FATAL)` in this case

include/caffe/util/db.hpp
immars: It prevents a too large `key_size` or `data_size` read from file, maybe from file corruption.
I thought 10M for a datum is large enough, or maybe should be larger? 100M?

immars:
Thanks for the review @sguada !

weiliu89:
@immars Thanks for the pull! I have been using it, and found that when I start N training jobs accessing the same datumfile, each one only uses 100/N % of CPU. Is it normal? I am not sure if it is going to make the training slower or not.

immars:
@weiliu89 this should not be happening, not according to my test. No locking is used, training process should not be IO bounded either. Are you running N process? what's your `iostat -kx 1` output ?

shelhamer:
Closing as better addressed by the `Python` layer. There are many types of data, and as long as it can be handled in Python it can be handled as a Python layer.

PR #2201, opened 2015-03-26
Update docs for ND blobs (#1970) and layer type is a string (#1694)

jeffdonahue:
A few updates to the docs that I've owed for a while... the process of finding places to fix ended up being somewhat manual so please comment if you know of anything else that needs updating.


shelhamer:
Thanks @jeffdonahue! I amended with a little proofreading.

PR #2211, opened 2015-03-26
Fallback to different cuDNN algorithm when under memory pressure

nsubtil:
CUDNN_CONVOLUTION_FWD_PREFER_FASTEST requires a lot of GPU memory, which may
not always be available. Add a fallback path that uses
CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM when the allocation fails.

### DIFF
--- a/src/caffe/layers/cudnn_conv_layer.cu
+++ b/src/caffe/layers/cudnn_conv_layer.cu
@@ -19,6 +19,11 @@ void CuDNNConvolutionLayer<Dtype>::Forward_gpu(
     Dtype* top_data = top[i]->mutable_gpu_data();
     const Dtype* weight = this->blobs_[0]->gpu_data();
 
+    size_t workspace_limit_bytes = this->kernel_h_ *
+                                   this->kernel_w_ *
+                                   this->channels_ *
+                                   sizeof(int) + 1;
+
     // Forward through cuDNN in parallel over groups.
     for (int g = 0; g < this->group_; g++) {
       cudnnConvolutionFwdAlgo_t algo;
@@ -32,8 +37,8 @@ void CuDNNConvolutionLayer<Dtype>::Forward_gpu(
         filter_desc_,
         conv_descs_[i],
         top_descs_[i],
-        CUDNN_CONVOLUTION_FWD_PREFER_FASTEST,
-        0,  // memoryLimitInBytes,
+        CUDNN_CONVOLUTION_FWD_SPECIFY_WORKSPACE_LIMIT,
+        workspace_limit_bytes,  // memoryLimitInBytes,
         &algo));
 
       // get minimum size of the workspace needed for the desired algorithm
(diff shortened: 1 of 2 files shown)


shelhamer:
I can confirm this works for models like CaffeNet where #2038 runs out of memory.

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -19,6 +19,11 @@ void CuDNNConvolutionLayer<Dtype>::Forward_gpu(
     Dtype* top_data = top[i]->mutable_gpu_data();
     const Dtype* weight = this->blobs_[0]->gpu_data();
 
+    size_t workspace_limit_bytes = this->kernel_h_ *
```
shelhamer: Will this always rule out the cuDNN GEMM convolution? At least in the Caffe GEMM convolution the workspace is the kernel dimensions (kernel_h * kernel_w * channels) * output dimensions (height_out_* width_out_) as in https://github.com/BVLC/caffe/blob/master/src/caffe/layers/base_conv_layer.cpp#L147, although in the cuDNN implementation I suppose the workspace could be just the input data so the `+1` allows it here.

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -19,6 +19,11 @@ void CuDNNConvolutionLayer<Dtype>::Forward_gpu(
...
+                                   this->kernel_w_ *
+                                   this->channels_ *
+                                   sizeof(int) + 1;
```
shelhamer: `sizeof(Dtype)`?

shelhamer:
Thanks for the fix @nsubtil! Please address my inline comments for merge.

src/caffe/layers/cudnn_conv_layer.cu
nsubtil: Yes, my understanding is that the workspace size depends only on the input data size.

The intent is that GEMM convolution will still be chosen if possible. However, in practice, we expect that there won't be enough memory available in many use cases.

src/caffe/layers/cudnn_conv_layer.cu
nsubtil: Good catch! Will fix this.

nsubtil: Actually, sizeof(int) is correct here. The workspace is always an integer buffer.

nsubtil:
It sounds like no further code changes are actually required here. If you're satisfied with my replies, I think this should be ready to merge.

Thanks!

src/caffe/layers/cudnn_conv_layer.cu
shelhamer: It seems odd to me that the workspace can be the same size whether the input and filters are float or double, but if that's how it is in cuDNN then so be it.

PR #2213, opened 2015-03-27
BilinearUpsamplingFiller used with DeconvolutionLayer for upsampling

tnarihi:
This is intended to be used in `DeconvolutionLayer` acting like `UpsamplingLayer` which is not implemented explicitly. You can upsample a feature map by any integer factor using the following proto.

```
layer {
  name: "upsample", type: "Deconvolution"
  bottom: "{{bottom_name}}" top: "{{top_name}}"
  convolution_param {
    kernel_size: {{2 * factor - factor % 2}} stride: {{factor}}
    num_output: {{channels_in}} group: {{channels_in}}
    pad: {{ceil((factor - 1) / 2.)}}
    weight_filler: { type: "bilinear_upsampling" } bias_term: false
  }
  param { lr_mult: 0 decay_mult: 0 }
}
```
Replace `{{}}` with your values. Note that the learning rate and the weight decay are set to 0 in order to keep coefficient values of bilinear interpolation unchanged during training. If you apply this to an image, this operation is equivalent to the following call in Python with Scikit.Image.
```python
out = skimage.transform.rescale(img, factor, mode='constant', cval=0)
```

### DIFF
--- a/include/caffe/filler.hpp
+++ b/include/caffe/filler.hpp
@@ -157,6 +157,60 @@ class XavierFiller : public Filler<Dtype> {
 };
 
 
+/*!
+@brief Fills a Blob with coefficients of bilinear interpolation for upsampling.
+
+This is intended to be used in DeconvolutionLayer acting as UpsamplingLayer.
+You can upsample a feature map with shape of (B, C, H, W) by any integer factor
+using the following proto.
+\code
+layer {
+  name: "upsample", type: "Deconvolution"
+  bottom: "{{bottom_name}}" top: "{{top_name}}"
+  convolution_param {
+    kernel_size: {{2 * factor - factor % 2}} stride: {{factor}}
+    num_output: {{C}} group: {{C}}
+    pad: {{ceil((factor - 1) / 2.)}}
+    weight_filler: { type: "bilinear_upsampling" } bias_term: false
+  }
+  param { lr_mult: 0 decay_mult: 0 }
+}
+\endcode
+Please use this by replacing `{{}}` with your values. By specifying
+`num_output: {{C}} group: {{C}}`, it behaves as
+channel-wise convolution. The filter shape of this deconvolution layer will be
+(C, 1, K, K) where K is `kernel_size`, and this filler will set a (K, K)
+interpolation kernel for every channel of the filter identically. The resulting
+shape of the top feature map will be (B, C, factor * H, factor * W).
+Note that the learning rate and the
+weight decay are set to 0 in order to keep coefficient values of bilinear
+interpolation unchanged during training. If you apply this to an image, this
+operation is equivalent to the following call in Python with Scikit.Image.
+\code{.py}
+out = skimage.transform.rescale(img, factor, mode='constant', cval=0)
+\endcode
+ */
+template <typename Dtype>
+class BilinearUpsamplingFiller : public Filler<Dtype> {
+ public:
+  explicit BilinearUpsamplingFiller(const FillerParameter& param)
+      : Filler<Dtype>(param) {}
+  virtual void Fill(Blob<Dtype>* blob) {
+    CHECK_EQ(blob->num_axes(), 4) << "Blob must be 4 dim.";
+    CHECK_EQ(blob->width(), blob->height()) << "Filter must be square";
+    Dtype* data = blob->mutable_cpu_data();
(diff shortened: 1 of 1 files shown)


weiliu89:
I believe this is wrong. According to https://gist.githubusercontent.com/shelhamer/80667189b218ad570e82/raw/938f470ad91072929cbb9f6c739dc34488ca7d03/solve.py, it uses net.params[l][0].data[range(m), range(k), :, :] = filt. This doesn't fill filt in each channel in each kernel. It only fills the ith channel in the ith kernel, all others are 0.

Bilinear interpolation is doing channels-wise. It could be easy to fix it.

tnarihi:
Thanks for commenting @weiliu89, but this is not wrong. Please note that I assume you specify `group: {{num_in}}` in `convolution_param`. That results channel-wise convolution, and it is more efficient for both memory and computation. We have conovolunal weights with size of `(num_in, 1, kernel_size, kernel_size)` instead of `(num_in, num_In, kernel_size, kernel_size)`. I don't do diagonal convolution explicitly. I updated my [IPython notebook example](http://nbviewer.ipython.org/gist/tnarihi/54744612d35776f53278) for a RGB image other than gray one. This example shows you this implementation gives you result images that are equivalent to results from a well-known imaging library in Python (Scikits.Image).

weiliu89:
@tnarihi Thanks for pointing out my mistake. Yours is indeed more efficient than the original python code. I thought group is not important in the setting and thus ignored it. Probably it is better to emphasize this more in the code, otherwise the filler is doing completely the wrong thing.

shelhamer:
@tnarihi @weiliu89 setting group is more obvious and makes use of less memory. The posted semantic segmentation FCNs will likely be switched to group deconvolution at some point for this reason. However the current Python filler is correct:

    [range(m), range(k), :, :]

indexes each `(idx, idx)` pair for channelwise interpolation while all "off-diagonal" filter coefficients are left to their zero initialization.

tnarihi:
@weiliu89 Your point is right. I updated doc.

@shelhamer Thanks for your attention to this PR. Yeah, I once called it `bilinear`, but as you pointed, it is  mainly dedicated for upsampling and I couldn't find other purpose of this filler besides upsamling, so I now call it `bilinear_upsampling` for clarity. Anyway, I don't have any strong preference about the name. Either is fine for me.

longjon:
@shelhamer @tnarihi, I agree that "upsampling" is a bit overly specific, e.g., you can use this to initialize a convolution layer and do *down*sampling. Either name is okay with me though. Also consider `bilinear_filter`.

tnarihi:
@longjon I am still unclear how we could create a downsampling layer which is equivalent to common image downsampling implementation by using this filler though, `bilinear_filter` also sounds good me.
@shelhamer @longjon Please choose what you prefer. Is `bilinear` simpler? I don't have any preference among three now. Once naming is fixed, I think this PR is ready to merge.

naibaf7:
Are you sure on the correctness of this filler?
shouldn't it rather be:
```
    int f = ceil(blob->width() / 2.);
    float c = (2 * f - 1 - f % 2) / (2. * f);
    for (int i = 0; i < blob->count(); ++i) {
      float x = i % blob->width();
      float y = (i / blob->width()) % blob->height();
      data[i] = ((1.0 - fabs(x / f - c)) * (1.0 - fabs(y / f - c))) / (float)(f * f);
    }
```
then the kernel will, in sum, add up to 1. Otherwise the kernel output will be f^2 too high in energy, I believe...

shelhamer:
Merged in 805a995. @tnarihi I changed the filler name to `bilinear` but kept your commit details and authorship.

tnarihi:
Thanks for merging, @shelhamer!

xyy19920105:
how to use this to downsampling with bilinear?

qingzew:
@tnarihi what‘s the idea for initializing the bilinear filter

thigi:
This does not work for 5D blobs. Does anyone know a workaround to use the `Deconvolution-Layer` with `bilinear` filler for 5D blobs? [5056](https://github.com/BVLC/caffe/issues/5056)

zmlmanly:
Hi, I have a problem that the result of tensorflow.image.resize_bilinear() is different from Deconvolution-Layer with bilinear in Caffe, could you give me a help? Thank you very much.

PR #2217, opened 2015-03-27
Rebase @ssafar's ReshapeLayer

jeffdonahue:
This is a rebase of #1263 by @ssafar and replaces my simpler version in #2088.   Thanks to @sguada's original suggestions and @ssafar's implementation in #1263, it supports all the fancy `dim: 0` (for copying input dimensions) and `dim: -1` (to infer a dimension) logic. I also added the fields `axis` and `num_axes` to specify only reshaping a portion of the input blob (equivalent to setting `dim: 0` for the remaining axes). e.g., `reshape_param { shape { dim: 1 } axis: 0 num_axes: 0 }` would simply insert a singleton axis at the beginning of the blob.


jyegerlehner:
+1 for this PR. I'm running it and it's working as advertised in my case.

dzhwinter:
how to get this pr? or get this timestamp mirror of this repo? @jyegerlehner

jyegerlehner:
Hmm, I think all I did was 
1) checkout some branch that was up-to-date with master.
2) add jeffdonahue's git to my remotes. 
3) git pull jeffdonahue ssafar-reshape-rebase
Or something similar to that. Whatever it was it wasn't very complicated.

dzhwinter:
@jyegerlehner thanks for your reply. I'm confused with what jeff does. not the git work flow. ssafar's repo has been stoped mantained for six month. jeff says his implementation is rebase of ssafar's #1263 。Now here is the question. i  checkout the 2217 and jeffdonahue ssafar-reshape-rebase is enough?

jyegerlehner:
@dzhwinter 
> I'm confused with what jeff does. not the git work flow.

Oh, sorry. I have a talent for misunderstanding people. 

> jeffdonahue ssafar-reshape-rebase is enough? 

I tried it, and caffe still built and passed tests. And then I added a layer with type "Reshape" to my prototxt and it solved my problem.  I just set the reshape_param according to the comments for ReshapeParameter in caffe.proto. They look like this after the pull:
```
// Message that stores parameters used by ReshapeLayer
message ReshapeParameter {
  // Specify the output dimensions. If some of the dimensions are set to 0,
  // the corresponding dimension from the bottom layer is used (unchanged).
  // Exactly one dimension may be set to -1, in which case its value is
  // inferred from the count of the bottom blob and the remaining dimensions.
  // For example, suppose we want to reshape a 2D blob "input" with shape 2 x 8:
  //
  //   layer {
  //     type: "Reshape" bottom: "input" top: "output"
  //     reshape_param { ... }
  //   }
  //
  // If "input" is 2D with shape 2 x 8, then the following reshape_param
  // specifications are all equivalent, producing a 3D blob "output" with shape
  // 2 x 2 x 4:
  //
  //   reshape_param { shape { dim:  2  dim: 2  dim:  4 } }
  //   reshape_param { shape { dim:  0  dim: 2  dim:  4 } }
  //   reshape_param { shape { dim:  0  dim: 2  dim: -1 } }
  //   reshape_param { shape { dim: -1  dim: 0  dim:  2 } }
  //
  optional BlobShape shape = 1;

  // axis and num_axes control the portion of the bottom blob's shape that are
  // replaced by (included in) the reshape. By default (axis == 0 and
  // num_axes == -1), the entire bottom blob shape is included in the reshape,
  // and hence the shape field must specify the entire output shape.
  //
  // axis may be non-zero to retain some portion of the beginning of the input
  // shape (and may be negative to index from the end; e.g., -1 to begin the
  // reshape after the last axis, including nothing in the reshape,
  // -2 to include only the last axis, etc.).
  //
  // For example, suppose "input" is a 2D blob with shape 2 x 8.
  // Then the following ReshapeLayer specifications are all equivalent,
  // producing a blob "output" with shape 2 x 2 x 4:
  //
  //   reshape_param { shape { dim: 2  dim: 2  dim: 4 } }
  //   reshape_param { shape { dim: 2  dim: 4 } axis:  1 }
  //   reshape_param { shape { dim: 2  dim: 4 } axis: -3 }
  //
  // num_axes specifies the extent of the reshape.
  // If num_axes >= 0 (and axis >= 0), the reshape will be performed only on
  // input axes in the range [axis, axis+num_axes].
  // num_axes may also be -1, the default, to include all remaining axes
  // (starting from axis).
  //
  // For example, suppose "input" is a 2D blob with shape 2 x 8.
  // Then the following ReshapeLayer specifications are equivalent,
  // producing a blob "output" with shape 1 x 2 x 8.
  //
  //   reshape_param { shape { dim:  1  dim: 2  dim:  8 } }
  //   reshape_param { shape { dim:  1  dim: 2  }  num_axes: 1 }
  //   reshape_param { shape { dim:  1  }  num_axes: 0 }
  //
  // On the other hand, these would produce output blob shape 2 x 1 x 8:
  //
  //   reshape_param { shape { dim: 2  dim: 1  dim: 8  }  }
  //   reshape_param { shape { dim: 1 }  axis: 1  num_axes: 0 }
  //
  optional int32 axis = 2 [default = 0];
  optional int32 num_axes = 3 [default = -1];
}
```
Pleasantly explicit, I would say.

include/caffe/common_layers.hpp
```diff
@@ -297,6 +297,44 @@ class MVNLayer : public Layer<Dtype> {
...
+      const vector<Blob<Dtype>*>& top) {}
+  virtual void Backward_gpu(const vector<Blob<Dtype>*>& top,
+      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom) {}
+
+  /// @brief vector of axes indices whose dimensions we'll copy from the bottom
+  vector<int> copy_axes_;
+  /// @brief the index of the axis whose dimension we infer, or -1 if none
+  int inferred_axis_;
+  /// @brief the product of the "constant" output dimensions
+  int64_t constant_count_;
```
jeffdonahue: TODO: change `int64_t` to `int` and merge (after discussion with @shelhamer and @longjon)

shelhamer:
LGTM. @jeffdonahue will swap out the `int64_t` and merge.

PR #2230, opened 2015-03-30
[Python] release/acquire Global Interpreter Lock (GIL) 

tnarihi:
Dependencies: #2149 #2001

The final commit only mentions this PR. This releases GIL in CPU-intensive operation in C++ side such as `Forward` and `Backward`. This will allow us, for example, to prefetch in data layers in Python using a different thread (not process) while other layers are running.


moi90:
Any progress on this?

PR #2240, opened 2015-04-02
Wrangle (some) Caffe dependencies through CMake

nsubtil:
Caffe has many dependencies, some of which seem quite esoteric. In order to alleviate the problem, I propose to modify the CMake build to fetch and build some dependencies as part of the Caffe build itself.

This can be done through the ExternalProject module in CMake. At build time, if a system-wide version of a given library is not found, CMake can be instructed to download and build it. Any dependencies built this way are always built as static libraries, to avoid having to pollute the user's system with extra shared libraries at install time.

Right now, this PR only handles glog and gflags. The same method can be extended to pretty much anything, although I'd recommend only doing this for libraries not commonly installed on a user's system. I've used this technique successfully in the past to tackle software that depends on, e.g., specific versions of Boost, or libraries that are not packaged yet and therefore hard to install for end users.

There are a few advantages to doing this: less dependencies to worry about when compiling Caffe, plus we can ensure that the build uses known-good versions of the libraries we depend on.

This method is not without it's drawbacks, though, and there are a few open questions: is it OK to depend on GitHub being accessible during the build? Are we concerned about bloating binary sizes due to static vs. dynamic linking? This will increase build times when our dependencies are not installed, would that be a concern for automated build systems?

Even though this is technically ready to merge, this PR is expected to be more of a starting point for discussion. Does this sound like a viable/interesting option to pursue?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,8 @@ project(Caffe C CXX)
 # ---[ Using cmake scripts and modules
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/Modules)
 
+include(ExternalProject)
+
 include(cmake/Utils.cmake)
 include(cmake/Targets.cmake)
 include(cmake/Misc.cmake)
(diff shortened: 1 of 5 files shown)


shelhamer:
glog and gflags are both available as packages on common platforms, but I imagine they aren't so easily acquired elsewhere and this would be helpful. Since this falls back to system glog and gflags this seems only positive, so I'll merge.

> Are we concerned about bloating binary sizes due to static vs. dynamic linking?

I prefer dynamic linking for build size and time.

Thanks Nuno!

Issue #2221, opened 2015-03-28
Homebrew includes and library paths for OS X build


dwt:
Hi there,

building caffe on Mac OS X Yosemite with the dependencies provided by home-brew I noticed that the build didn't pick up my build locations as they are non standard (i.e. I have homebrew installed in my home directory).

The problem was fixed by handing in the CXXFLAGS at three locations in the build script, so that they also apply when building the test tool and the python bindings.

```patch
diff --git a/Makefile b/Makefile
index cb39679..24173b6 100644
--- a/Makefile
+++ b/Makefile
@@ -431,7 +431,7 @@ py: $(PY$(PROJECT)_SO) $(PROTO_GEN_PY)
 $(PY$(PROJECT)_SO): $(PY$(PROJECT)_SRC) $(PY$(PROJECT)_HXX) | $(DYNAMIC_NAME)
 	@ echo CXX/LD -o $@ $<
 	$(Q)$(CXX) -shared -o $@ $(PY$(PROJECT)_SRC) \
-		-o $@ $(LINKFLAGS) -l$(PROJECT) $(PYTHON_LDFLAGS) \
+		-o $@ $(CXXFLAGS) $(LINKFLAGS) -l$(PROJECT) $(PYTHON_LDFLAGS) \
 		-Wl,-rpath,$(ORIGIN)/../../build/lib
 
 mat$(PROJECT): mat
@@ -520,7 +520,7 @@ $(TEST_ALL_BIN): $(TEST_MAIN_SRC) $(TEST_OBJS) $(GTEST_OBJ) \
 		| $(DYNAMIC_NAME) $(TEST_BIN_DIR)
 	@ echo CXX/LD -o $@ $<
 	$(Q)$(CXX) $(TEST_MAIN_SRC) $(TEST_OBJS) $(GTEST_OBJ) \
-		-o $@ $(LINKFLAGS) $(LDFLAGS) -l$(PROJECT) -Wl,-rpath,$(ORIGIN)/../lib
+		-o $@ $(CXXFLAGS) $(LINKFLAGS) $(LDFLAGS) -l$(PROJECT) -Wl,-rpath,$(ORIGIN)/../lib
 
 $(TEST_CU_BINS): $(TEST_BIN_DIR)/%.testbin: $(TEST_CU_BUILD_DIR)/%.o \
 	$(GTEST_OBJ) | $(DYNAMIC_NAME) $(TEST_BIN_DIR)
[...]

PR #2245, opened 2015-04-02
Add commented out helpers for homebrew users

dwt:
As requested in #2221 here's a pull request.


shelhamer:
Thanks. I think having these comments is fine since they might save someone time.

Issue #2247, opened 2015-04-03
PyCaffe: in examples/hdf5_classification.ipynb: name 'pd' is not defined


tiangolo:
I was testing the HDF5 IPython netbook (in examples/hdf5_classification.ipynb), but when I ran the second cell I got this error:

```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-2-13a258632ba9> in <module>()
      9 # Visualize sample of the data
     10 ind = np.random.permutation(X.shape[0])[:1000]
---> 11 df = pd.DataFrame(X[ind])
     12 _ = pd.scatter_matrix(df, figsize=(9, 9), diagonal='kde', marker='o', s=40, alpha=.4, c=y[ind])

NameError: name 'pd' is not defined
```

It's because Pandas is not being imported in the first cell. I will create a pull request to solve it.

shelhamer:
Fixed by #2249.

PR #2249, opened 2015-04-03
Import Pandas in HDF5 IPython notebook. Fix for issue BVLC/caffe#2247

tiangolo:
This is a fix for issue #2247 : PyCaffe: in examples/hdf5_classification.ipynb: name 'pd' is not defined

Also, a comment, when I opened the IPython notebook it said that:

>>This notebook has been converted from an older notebook format (v3) to the current notebook format (v4). The next time you save this notebook, the current notebook format will be used. Older versions of IPython may not be able to read the new format. To preserve the original version, close the notebook without saving it.

I didn't save it in the new format, I don't know if you want it converted, so, I updated the notebook with vim, just to import Pandas and fix the issue, using the old (v3) notebook format.

### DIFF
--- a/examples/hdf5_classification.ipynb
+++ b/examples/hdf5_classification.ipynb
@@ -25,6 +25,7 @@
      "collapsed": false,
      "input": [
       "import numpy as np\n",
+      "import pandas as pd\n",
       "import matplotlib.pyplot as plt\n",
       "%matplotlib inline\n",
       "\n",


shelhamer:
Thanks for the fix. We'll likely update all the notebooks to the new version in one go soon.  I cherry-picked since you had a stray merge in this PR.

Issue #2012, opened 2015-03-02
Snapshot on signal


shelhamer:
Caffe should attempt to snapshot on SIGINT. Suggested by @longjon.

seanbell:
I like this idea, though I think it would be helpful to have a way of easily disabling this behavior (e.g. command-line flag).  Otherwise, management scripts that call caffe now have to worry about extra disk-space consequences of stopping training early.

I think it's also worth letting users configure which signal code will trigger a snapshot (e.g. in solver.prototxt).  The default could be SIGINT (2), but other codes such as SIGHUP (1) would let you signal a snapshot without interrupting training.

shelhamer:
@seanbell agreed. All good points.

beniz:
Is there a ticket to focus all thoughts on error handling within Caffe already ? If not, maybe it could make sense to create one with bullet points for sub-tasks. I would happily work on this after it's all been discussed and agreed upon so that no effort is wasted.

jyegerlehner:
Ref [PR 2253](https://github.com/BVLC/caffe/pull/2253).

shelhamer:
Solved by #2253

PR #2253, opened 2015-04-03
Snapshot on signal

jyegerlehner:
This an implementation of the feature discussed in [Issue 2012](https://github.com/BVLC/caffe/issues/2012).

When you hit Ctrl-C to kill caffe (while training), it will now save a snapshot before exiting. Actually, the Solver just stops training, and a snapshot is only saved if `snapshot_after_train` is true.

This is the default behavior which is configurable via the sigint_effect and sighup_effect command line options. Also by default, SIGHUP signal causes caffe to save a snapshot and continue training.  So you can make caffe save a snapshot by sending it SIGHUP signal, e.g.:

`kill -SIGHUP PID `

where PID is the process id of caffe, which you can find by doing `ps -ef | grep caffe`.

The design has two pieces to it. 1. `Solver` is modified slightly so that after each iteration it checks to see if its client wants it to either snapshot or exit. It does this via a callback function that a client can set on the Solver instance. If the callback hasn't been set, it just carries on as usual. So there is no breaking change to the Solver interface and the behavior of existing code shouldn't change. 2. The caffe executable provides the callback function to the Solver, and the callback is implemented on a SignalHandler that intercepts SIGINT and SIGHUP.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -1,12 +1,34 @@
 #ifndef CAFFE_OPTIMIZATION_SOLVER_HPP_
 #define CAFFE_OPTIMIZATION_SOLVER_HPP_
-
+#include <boost/function.hpp>
 #include <string>
 #include <vector>
 
 #include "caffe/net.hpp"
 
 namespace caffe {
+
+/**
+  * @brief Enumeration of actions that a client of the Solver may request by
+  * implementing the Solver's action request function, which a
+  * a client may optionally provide in order to request early termination
+  * or saving a snapshot without exiting. In the executable caffe, this
+  * mechanism is used to allow the snapshot to be saved when stopping
+  * execution with a SIGINT (Ctrl-C).
+  */
+  namespace SolverAction {
+    enum Enum {
+      NONE = 0,  // Take no special action.
+      STOP = 1,  // Stop training. snapshot_after_train controls whether a
+                 // snapshot is created.
+      SNAPSHOT = 2  // Take a snapshot, and keep training.
+    };
+  }
+
+/**
+ * @brief Type of a function that returns a Solver Action enumeration.
+ */
+typedef boost::function<SolverAction::Enum()> ActionCallback;
 
 /**
  * @brief An interface for classes that perform optimization on Net%s.
@@ -23,6 +45,12 @@ class Solver {
   void Init(const SolverParameter& param);
   void InitTrainNet();
   void InitTestNets();
+
+  // Client of the Solver optionally may call this in order to set the function
+  // that the solver uses to see what action it should take (e.g. snapshot or
+  // exit training early).
+  void SetActionFunction(ActionCallback func);
+  SolverAction::Enum GetRequestedAction();
   // The main entry of the solver function. In default, iter will be zero. Pass
   // in a non-zero iter number to resume training for a pre-trained net.
   virtual void Solve(const char* resume_file = NULL);
@@ -83,6 +111,13 @@ class Solver {
   // The root solver that holds root nets (actually containing shared layers)
   // in data parallelism
   const Solver* const root_solver_;
+
+  // A function that can be set by a client of the Solver to provide indication
+  // that it wants a snapshot saved and/or to exit early.
+  ActionCallback action_request_function_;
+
+  // True iff a request to stop early was received.
+  bool requested_early_exit_;
 
   DISABLE_COPY_AND_ASSIGN(Solver);
 };
--- a/include/caffe/util/signal_handler.h
+++ b/include/caffe/util/signal_handler.h
@@ -1,0 +1,24 @@
+#ifndef INCLUDE_CAFFE_UTIL_SIGNAL_HANDLER_H_
+#define INCLUDE_CAFFE_UTIL_SIGNAL_HANDLER_H_
+
+#include "caffe/proto/caffe.pb.h"
+#include "caffe/solver.hpp"
+
+namespace caffe {
+
+class SignalHandler {
+ public:
+  // Contructor. Specify what action to take when a signal is received.
+  SignalHandler(SolverAction::Enum SIGINT_action,
+                SolverAction::Enum SIGHUP_action);
+  ~SignalHandler();
+  ActionCallback GetActionFunction();
+ private:
+  SolverAction::Enum CheckForSignals() const;
+  SolverAction::Enum SIGINT_action_;
+  SolverAction::Enum SIGHUP_action_;
+};
+
+}  // namespace caffe
+
+#endif  // INCLUDE_CAFFE_UTIL_SIGNAL_HANDLER_H_
--- a/src/caffe/util/signal_handler.cpp
+++ b/src/caffe/util/signal_handler.cpp
@@ -1,0 +1,115 @@
+#include <boost/bind.hpp>
+#include <glog/logging.h>
+
+#include <signal.h>
+#include <csignal>
+
+#include "caffe/util/signal_handler.h"
+
+namespace {
+  static volatile sig_atomic_t got_sigint = false;
+  static volatile sig_atomic_t got_sighup = false;
+  static bool already_hooked_up = false;
+
+  void handle_signal(int signal) {
+    switch (signal) {
+    case SIGHUP:
+      got_sighup = true;
+      break;
+    case SIGINT:
+      got_sigint = true;
+      break;
+    }
+  }
+
+  void HookupHandler() {
+    if (already_hooked_up) {
+      LOG(FATAL) << "Tried to hookup signal handlers more than once.";
+    }
+    already_hooked_up = true;
+
+    struct sigaction sa;
+    // Setup the handler
+    sa.sa_handler = &handle_signal;
+    // Restart the system call, if at all possible
+    sa.sa_flags = SA_RESTART;
+    // Block every signal during the handler
+    sigfillset(&sa.sa_mask);
+    // Intercept SIGHUP and SIGINT
+    if (sigaction(SIGHUP, &sa, NULL) == -1) {
+      LOG(FATAL) << "Cannot install SIGHUP handler.";
+    }
+    if (sigaction(SIGINT, &sa, NULL) == -1) {
+      LOG(FATAL) << "Cannot install SIGINT handler.";
+    }
+  }
+
+  // Set the signal handlers to the default.
+  void UnhookHandler() {
+    if (already_hooked_up) {
+      struct sigaction sa;
+      // Setup the sighub handler
+      sa.sa_handler = SIG_DFL;
+      // Restart the system call, if at all possible
+      sa.sa_flags = SA_RESTART;
+      // Block every signal during the handler
+      sigfillset(&sa.sa_mask);
+      // Intercept SIGHUP and SIGINT
+      if (sigaction(SIGHUP, &sa, NULL) == -1) {
+        LOG(FATAL) << "Cannot uninstall SIGHUP handler.";
+      }
+      if (sigaction(SIGINT, &sa, NULL) == -1) {
+        LOG(FATAL) << "Cannot uninstall SIGINT handler.";
+      }
+
+      already_hooked_up = false;
+    }
+  }
+
+  // Return true iff a SIGINT has been received since the last time this
+  // function was called.
+  bool GotSIGINT() {
+    bool result = got_sigint;
+    got_sigint = false;
+    return result;
+  }
+
+  // Return true iff a SIGHUP has been received since the last time this
+  // function was called.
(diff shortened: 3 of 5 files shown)


flx42:
Looks cool, but I don't think you can safely lock a mutex in a signal handler, nor try to perform I/O.

jyegerlehner:
Thanks @flx42, I should have researched signals a bit more before I implemented this.

jyegerlehner:
@flx42 do you see any problem with this new implementation?

flx42:
The type should be "sig_atomic_t volatile".
Also, since the variables are no longer of type "bool", it might be better to use "1" and "0" instead of "true" and "false".

jyegerlehner:
Thanks for the reply.
> The type should be "sig_atomic_t volatile".

Hrmm.. all the examples I see, the qualifier precedes the type name, same as with "const" or "mutable". Please let me know if I'm missing something.

> since the variables are no longer of type "bool", it might be better to use "1" and "0" instead of "true" and "false".

true and false are promoted to 1 and 0 per the language spec I'm pretty sure. Since they better convey the meaning, I'm inclined to leave the true and false in there.

flx42:
On Sat, Apr 18, 2015 at 12:48 PM, jyegerlehner <notifications@github.com>
wrote:

> Thanks for the reply.
>
> The type should be "sig_atomic_t volatile".
>
> Hrmm.. all the examples I see, the qualifier precedes the type name, same
> as with "const" or "mutable". Please let me know if I'm missing something.
>
No, you're right. That's what I meant.


>  since the variables are no longer of type "bool", it might be better to
> use "1" and "0" instead of "true" and "false".
>
> true and false are promoted to 1 and 0 per the language spec I'm pretty
> sure. Since they better convey the meaning, I'm inclined to leave the true
> and false in there.
>
>  It was really a nitpicking comment about style. But both ways will be
fine.

flx42:
Another nitpick: you should avoid unnecessary empty lines changes in your patch.
Like this: https://github.com/jyegerlehner/caffe/commit/c0df90916cccf083a7793615e3b0ddac32588b57#diff-ed792b89816c46f109918495c88421bcL3

lukeyeager:
Any update on this? I want it.

jyegerlehner:
Hi Luke, I'm not aware of anything deficient about it.

jyegerlehner:
Well there is one thing about this that is debatable. Solver doesn't check to see if it should exit when it is doing test, only when it is training. This means if you send it SIGINT and it happens to be in the middle of test (not train), caffe keeps going until the test is finished, and only then does it exit (or snapshot). So you might have to wait a bit if you try to request it to stop when it's testing. I thought that was a good thing since it allows the test to run to completion and produce a good result. However, one might prefer that it be more responsive and stop right away regardless of whether it is testing, and throw away the test that is in-progress. We could make it behave that way with a bit of extra complexity.

lukeyeager:
> one might prefer that it be more responsive and stop right away regardless of whether it is testing, and throw away the test that is in-progress

That sounds better to me. I would expect Ctrl+C to kill the process within a second or two. If you wait for  testing to finish, you might have to wait several minutes.

jyegerlehner:
> I would expect Ctrl+C to kill the process within a second or two.

If anyone objects to Luke's preferred behaviour please speak up. Otherwise I'll plan to make that change.

shelhamer:
I can see an argument for either finishing the testing or quitting
immediately, but to better fit expectations it should likely quit right
away no matter the action.

Thanks for the signal handling!

flx42:
It would be nice to have both, if that's not too much complexity for you.
It's not uncommon to have different shutdown behaviors depending on which signal was received. For instance, nginx will do a fast shutdown on SIGTERM/SIGINT, and a graceful shutdown on SIGQUIT (wait for all current connections to finish). Apache has a similar behavior.

jyegerlehner:
Whoopsiedaisy. WIP. Let me rebase this on master.

jyegerlehner:
> to better fit expectations it should likely quit right away

Makes sense to me.

> Thanks for the signal handling!

Anything for the cause, comrade!

jyegerlehner:
> It would be nice to have both, if that's not too much complexity for you.

If we find we need or want that extra sophistication, I think my preference is to add it in a separate PR so that we may thereby proceed more incrementally. The changes to Solver started out very simple, and got more complex with the latest change and feels to me like pushing the limits of added risk in one PR.

That said, if everyone really really wants it right now, I can do it.

flx42:
No, I agree, this should be done in a separate patch.

jyegerlehner:
I think this could use a "Ready for Review" label, if those make any difference.

jyegerlehner:
Hmmm. Was trying to rebase off master. Will correct if I can figure out what went wrong.

include/caffe/solver.hpp
```diff
@@ -66,6 +77,8 @@ class Solver {
   string SnapshotToBinaryProto();
   string SnapshotToHDF5();
   // The test routine
+  // stop_was_requested will be set to true iff a request to stop training
+  // was received whilst testing.
```
jyegerlehner: Delete this.

src/caffe/util/signal_handler.cpp
```diff
@@ -0,0 +1,91 @@
+#include <boost/bind.hpp>
+#include <boost/thread/mutex.hpp>
+#include <boost/thread/thread.hpp>
```
jyegerlehner: Don't think we need boost/thread anymore. Delete if not needed.

tools/caffe.cpp
```diff
@@ -126,6 +133,20 @@ void CopyLayers(caffe::Solver<float>* solver, const std::string& model_list) {
   }
 }
 
+caffe::SolverParameter_Action GetRequestedAction(
```
jyegerlehner: Line break not needed for 80 cols? Remove if not needed.

ronghanghu:
LGTM. I shall review this PR within next week.

jyegerlehner:
OK thanks. I will squash the commits once the review is done and we know there aren't any more changes required.

ronghanghu:
@jyegerlehner I just did a quick review today, and tested on my machine. Looks good to me! I'll finish reviewing this PR within tomorrow.

(As far as I know, this won't PR apply to windows. But since we are not planning to support windows)

One related stuff to this PR is that we can also support learning rate adjustment on signaling, so that one may adjust it during training based on e.g. learning curve from log, similar to some other deep learning tools. @jeffdonahue what do you think?

include/caffe/util/signal_handler.h
```diff
@@ -0,0 +1,24 @@
...
+namespace caffe {
+
+class SignalHandler {
+ public:
+  // Contructor. Specify what action to take when a signal is received.
+  SignalHandler(SolverParameter_Action SIGINT_action,
+                SolverParameter_Action SIGHUP_action);
+  ActionCallback GetActionFunction();
+ private:
+  SignalHandler();  // Not implemented.
```
ronghanghu: I'm sort of confused here... Why do we need a private constructor declaration in SignalHandler class?

src/caffe/proto/caffe.proto
```diff
@@ -234,6 +234,19 @@ message SolverParameter {
...
+  // a client may optionally provide in order to request early termination
+  // or saving a snapshot without exiting. In the executable caffe, this
+  // mechanism is used to allow the snapshot to be saved when stopping
+  // execution with a SIGINT (Ctrl-C).
+  enum Action {
+    NONE = 0; // Take no special action.
+    STOP = 1; // Stop training. snapshot_after_train controls whether a snapshot
+              // is created.
+    SNAPSHOT = 2; // Take a snapshot, and keep training.
+  }
```
ronghanghu: Personally, I am not very in favor of putting this enum in `caffe.proto`, since it is enum never used in our net or solver prototxt but only in C++ code. @jeffdonahue what do you think?

bhack:
@willyd How this will impact with your windows plans?

willyd:
@bhack On windows we would need to call [SetConsoleCtrlHandler](https://msdn.microsoft.com/en-us/library/windows/desktop/ms685049%28v=vs.85%29.aspx) to handle SIGINT but don't think there is an equivalent to SIGHUP.

A cross-plaform implementation is available in [boost.asio] (http://www.boost.org/doc/libs/1_59_0/doc/html/boost_asio/overview/signals.html).

bhack:
If there is still interest in https://github.com/BVLC/caffe/pull/2537 I will avoid asio solution.

include/caffe/util/signal_handler.h
jyegerlehner: If the default constructor is not private, the compiler will provide it implicitly. So by preventing default construction we force the use of the constructor taking explicitly-specified signal actions. We could provide a default constructor that, say, initializes the actions to sensible defaults. But it would be unused as it is and thus unnecessary.

ronghanghu: >If the default constructor is not declared private, the compiler will provide it implicitly.

C++ does generate a default constructor but **only if you don't provide one of your own**. Since you've already provided one: `SignalHandler(SolverParameter_Action SIGINT_action, SolverParameter_Action SIGHUP_action);`, a default constructor won't be generated by compiler implicitly. So a private constructor declaration here is unnecessary.

You may verify with this simple program:

    class C {
     public:
      C(int) {}
    };
    int main() { C c; } // compilation fails because there's no default constructor in class C.

Anyway, this is only a small issue. The PR is great!

src/caffe/util/signal_handler.cpp
```diff
@@ -0,0 +1,89 @@
...
+    // Block every signal during the handler
+    sigfillset(&sa.sa_mask);
+    // Intercept SIGHUP and SIGINT
+    if (sigaction(SIGHUP, &sa, NULL) == -1) {
+      LOG(FATAL) << "Cannot install SIGHUP handler.";
+    }
+    if (sigaction(SIGINT, &sa, NULL) == -1) {
+      LOG(FATAL) << "Cannot install SIGINT handler.";
+    }
+  }
```
ronghanghu: Maybe release these signal handlers in `SignalHandler::~SignalHandler()`?

ronghanghu:
Completed a thorough pass today. This PR seems in good shape to me, handles signals via POSIX sigaction and address actions from solver's train & test loop in a call-back fashion. A side effect: this is platform-specific and may impact community windows ports.

include/caffe/util/signal_handler.h
jyegerlehner: OK thanks for the  correction and explanation. I will remove the private dec.

src/caffe/proto/caffe.proto
jyegerlehner: I'm happy to move these out of proto into regular enum declaration. Wait for word from JD?

src/caffe/util/signal_handler.cpp
jyegerlehner: OK will do.

jyegerlehner:
@ronghanghu The latest commits are intended to resolve the review issues you raised. Please let us know if they are not satisfactory.

As far as the Travis build failing: this looks like it happened due to an error installing a package. Does anyone know: should I make a dummy commit to provoke it to try building again?  Or am I missing something that's an actual problem I need to fix?.

ronghanghu:
I restarted Travis CI and now all tests pass. I'll try to take a look today.

ronghanghu:
Seems ready to me :)

ronghanghu:
@jeffdonahue @longjon I would like to merge this PR if you don't have other concerns.

Community windows ports can perhaps simply strip this feature e.g. via removing the anonymous namespace section in `caffe/util/signal_handler.cpp` and always returning `SolverAction::NONE` in  `SignalHandler::CheckForSignals()` (and the default Ctrl+C still terminates the program on windows).

jyegerlehner:
OK, do you prefer commits to be squashed?

ronghanghu:
Yes, please squash into one commit, so I can merge in this weekend.

erogol:
this is great. thanks for the effort @jyegerlehner

jyegerlehner:
Sure thing @erogol. Glad to hear it's helpful.

Coderx7:
Can you add an option to the solver so that users can take snap-shots at any given time by pressing a key combination like Ctrl-S for example?

ajtulloch:
@Coderx7 I think that'd add to much complexity to the feature. If you want to snapshot at an arbitrary time without stopping training, just send SIGHUP (`kill -SIGHUP $(pidof caffe)`)

Coderx7:
@ajtulloch yes, but at the same time it provides a very convenient and good feature to have, 
in linux thats a very good idea, but how does one do the exact same thing on windows?

PR #2271, opened 2015-04-07
fix #1362. prefetch HDF5DataLayer

pclove1:
@jeffdonahue , @shelhamer 

This PR attempts to fix #1362.
I highly recommend you to read #1362's description first to see what this PR tries to achieve.
To be short, we want to prefetch HDF5 files and avoid excessive memory usage by reading HDF5 files partially.

The below is the summary of modifications made on top of #1362.
- override `DataLayerSetUp()` instead of `LayerSetUp()` following [the existing design](https://github.com/BVLC/caffe/blob/249aba49ab3100d0409d93ef187ebcd5a53865b9/include/caffe/data_layers.hpp#L35)
- Blobs are N-D arrays now since #1970
- wait until a thread is joined in when destructing `HDF5DataLayer`
- preserve the ability of shuffling files
- add a unit test case where a batch is loaded from interleaving HDF5 files to cover the found error on the index math.
- plus a few minor bug fixes

# Caveats
- don't shuffle rows in a HDF5 file due to the introduction of partial reading
  - I believe this is also true for LMDB/LevelDB cases where the users are supposed to shuffle the data when they construct them.
- If someone needs to access the same HDF5 file concurrently, they need to make sure to have a thread-safe version of hdf5 library. (See [this](http://www.hdfgroup.org/hdf5-quest.html#gconc))

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -27,6 +27,7 @@ Caffe has several dependencies.
 * [OpenCV](http://opencv.org/) >= 2.4 including 3.0
 * `protobuf`, `glog`, `gflags`
 * IO libraries `hdf5`, `leveldb`, `snappy`, `lmdb`
+    * hdf5 needs to be built with thread-safe enabled, if you need to access the same HDF5 file concurrently. (See [this](http://www.hdfgroup.org/hdf5-quest.html#gconc))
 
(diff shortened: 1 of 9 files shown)


ronghanghu:
Rebased and cleaned up in #2892. Thanks the original authors @jeffdonahue and @pclove1

PR #2274, opened 2015-04-07
Update generate_sample_data.py

Ashwani001:
Changed os.path.dirname(__file__) as it was returning empty...

### DIFF
--- a/src/caffe/test/test_data/generate_sample_data.py
+++ b/src/caffe/test/test_data/generate_sample_data.py
@@ -27,12 +27,12 @@ label2 = label + 1
 print data
 print label
 
-with h5py.File(os.path.dirname(__file__) + '/sample_data.h5', 'w') as f:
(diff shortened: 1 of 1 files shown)


jeffdonahue:
I think it's meant to be run from the Caffe root in which case the `dirname` result should not be empty.  But making it runnable from anywhere seems fine.  Maybe declare it at the top with a name like `script_dir = os.path.dirname(os.path.abspath(__file__)` to make the code clearer and more concise though?

Ashwani001:
That could be the reason. I probably ran it from elsewhere. Changed it to what you suggested. Does look better.

jeffdonahue:
Thanks! Please squash to a single commit for merge.

jeffdonahue:
Never mind -- just noticed that they each had specific and clean commit messages so I'll go ahead and merge.  Thanks @Ashwani001!

Ashwani001:
Thanks @jeffdonahue

PR #2278, opened 2015-04-07
Db split

sguada:
Split db.hpp and db.cpp by different backends, this allows easier extensions.
Changed DataParameter_DB to string for easier extensions.

@longjon please take a quick look, it is very innocuous but useful change so I plan to merge soon.


longjon:
I can review carefully after the 22nd... is there a reason this must be merged earlier?

The refactor seems pretty reasonable; what's the use case you have in mind?

The main discussion-worthy issue I see is that the type change to the `backend` parameter will break existing prototxt files, and it doesn't look like there's an automatic upgrade path. As far as I know we've never made a breaking change of that form before.

sguada:
@longjon The idea is to be able to separate nicely the different DB_backends, and make easier to add new ones and also to be able to use in other projects. The current approach of having everything depending of "almost" everything else is not sustainable, I think we should start separating dependencies into libraries and add options to the compilation.

I will try to devise a way to auto-upgrade so it will not break old prototxt (maybe just use a new name `backend_type`). However in the long run we may want to stop supporting really old versions, and encourage people to use the upgrade tool to upgrade their prototxt instead of magically upgrading, I have encounter problems with that when different versions are upgraded automatically.

sguada:
@longjon any suggestions?
I see two options, create a new param `string backend_type` which replaces `backend` and deprecate `backend` or leave `backend` as it is but then every time we want to add a new backend we need to add it to caffe.proto and change the enum.

shelhamer:
Everything that's polymorphic or has several types should be a string. I
think this for solver type, engine, and backend type. The ENUMs are pain.

sguada:
@shelhamer 
Totally agree, we should replace all ENUMs with strings for anything that can change overtime. I would also vouch for using registry factory for all this cases too. 
The question to me is if we should do a prototxt v3 and do that once and for all or do step by step.

sguada:
@longjon have you had a chance to look into this PR? I will be happy to change it if needed.

sguada:
@longjon I have simplify this PR, now only splits db.hpp leave for later registry or other changes.

Once Travis pass, I will merge, since these changes don't affect current behavior.

Issue #2078, opened 2015-03-09
cmake with -DBUILD_matlab=ON fails


wookayin:
Hello, I suspect that this might be a bug introduced after the commit 3d30510 (PR #2059). 

The environment with which I've tested is: Mac OS X 10.10. I haven't tested with Ubuntu, since I don't have any Linux machine with matlab installed as of now.

```
$ cd build
$ cmake .. -DBUILD_matlab=ON && make -j4
```

However, building matcaffe fails with the message `ld: library not found for -lpython2`.

```
[100%] Building Matlab interface: ${CAFFE_ROOT}/matlab/caffe/caffe.mexmaci64
Building with 'Xcode Clang++'.
ld: library not found for -lpython2
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Let's see the command line argument passed (absolute paths were replaced by `${CAFFE_ROOT}`).
There are python-related link flags, `-lpython2` and `-lboost_python-mt`, which were added from `${PYTHON_LIBRARIES}`

```
/Applications/MATLAB_R2014b.app/bin/mex -output ${CAFFE_ROOT}/matlab/caffe/caffe.mexmaci64 ${CAFFE_ROOT}/matlab/caffe/matcaffe.cpp -DCPU_ONLY -DWITH_PYTHON_LAYER -DGTEST_USE_OWN_TR1_TUPLE -I${CAFFE_ROOT}/src -I/usr/local/include -I${CAFFE_ROOT}/build/include -I/usr/local/include/opencv -I/System/Library/Frameworks/Accelerate.framework/Versions/Current/Frameworks/vecLib.framework/Versions/Current/Headers -I/System/Library/Frameworks/Python.framework/Headers -I/usr/local/lib/python2.7/site-packages/numpy/core/include -I${CAFFE_ROOT}/include -I${CAFFE_ROOT}/build -L${CAFFE_ROOT}/build/lib -L/usr/local/lib -L/usr/local/lib -L/usr/local/lib -L/usr/local/lib -L/usr/lib -lcaffe -lboost_system-mt -lboost_thread-mt -lglog -lgflags -lhdf5_hl -lhdf5 -llmdb -lleveldb -lsnappy -lopencv_core -lopencv_highgui -lopencv_imgproc -lcblas -lpython2  -lboost_python-mt -lprotobuf -v
```

[...]

PR #2279, opened 2015-04-08
Remove python2 from matlab CMakeLists.txt

eshelman:
This is a proposed fix for Issue #2078 

Builds with MATLAB support are complaining `library not found for -lpython2` when Python support is enabled. Linking to Python is not necessary for the MATLAB build, so that library should be stripped out.

### DIFF
--- a/matlab/CMakeLists.txt
+++ b/matlab/CMakeLists.txt
@@ -45,9 +45,11 @@ string(REPLACE ";" ":"  rpath_folders   "${folders}")
 if(build_using MATCHES "Matlab")
   set(libflags -lcaffe${CAffe_POSTFIX} ${libflags}) # Matlab R2014a complans for -Wl,--whole-archive
 
+  string(REPLACE ";-lpython2" "" matlab_libflags "${libflags}")
+
   caffe_fetch_and_set_proper_mexext(Matlab_caffe_mex)
   add_custom_command(OUTPUT ${Matlab_caffe_mex} COMMAND ${Matlab_mex}
-      ARGS -output ${Matlab_caffe_mex} ${Matlab_srcs} ${cflags} ${link_folders} ${libflags}
+      ARGS -output ${Matlab_caffe_mex} ${Matlab_srcs} ${cflags} ${link_folders} ${matlab_libflags}
       DEPENDS caffe COMMENT "Building Matlab interface: ${Matlab_caffe_mex}" VERBATIM)
   add_custom_target(matlab ALL DEPENDS ${Matlab_caffe_mex} SOURCES ${Matlab_srcs})
 


shelhamer:
@Nerei do you know why Python is included in the MATLAB build? This looks like a fix, but please comment + / - for the merge.

Thanks @eshelman.

ronghanghu:
@eshelman @shelhamer this was not the right fix after PythonLayer is introduced. Python lib could be necessary if you want to use a net in MatCaffe that has PythonLayer. I think the correct solution to #2078 would be changing `-lpython2` to specific python library such as `-lpython2.7`.

Nerei:
@eshelman @shelhamer 
Yes, agree that this looks like a fix. Python was included into link just because, for simplicity, I link all libraries that are linked to Caffe (they were accumulated in Caffe_LINKER_LIBS list). For static caffe + octave it is what is exactly required. For Matlab, only shared Caffe is supported now, so suppose linking Caffe dependencies can be omitted.

@ronghanghu For PythonLayer, is python linked to Caffe directly now? If so, than it shouldn't be a an issue, right?  Or, have I missed something? I can't check this now, I am here from tablet device.

Nerei:
@ronghanghu Just checked, Python was added to Caffe link list in 3d30510a50007e03067a028d4585bb1c358d49dc, when PythonLayer was introduced. It seems after that, this issue with Matlab appeared. From my view, just removing python form the list is ok.

ronghanghu:
> For static caffe + octave it is what is exactly required. For Matlab, only shared Caffe is supported now, so suppose linking Caffe dependencies can be omitted.
Forgive me if I am wrong, but I think Dynamic linking #1842 did not cover matcaffe, and matcaffe is still using static linking?

Nerei:
@ronghanghu  By default CMake builds shared caffe (it doesn't build both). With static caffe Matlab doesn't work at all.  For Matlab, I didn't find flags equivalent to `-Wl,--whole-archive` for gcc/octave. Such hack is required for static caffe use, otherwise linker doesn't link required caffe layers from archive.

ronghanghu:
@Nerei I am using the Makefile and use make instead of cmake, and I believe matcaffe is still statically linked against libcaffe.a instead of libcaffe.so (and it works). To verify this, I deleted ./build/lib/libcaffe.so after building. After that, although the command line tools became broken, I can still use matcaffe without any problem.

> By default CMake builds shared caffe (it doesn't build both). With static caffe Matlab doesn't work at all.
I am not very familiar with CMake and those CMakeLists.txt in Caffe. Not sure why it is.

intbots:
The addition of `-lpython2` instead of `-lpython2.7` is due to wrong behaviour in the `caffe_parse_linker_libs` function in cmake/Utils.cmake. The pull request #3650 solves this.

However if python is indeed not required for matlab wrapper, then it should be removed. 

However the fix proposed in this pull request of simply removing python from the linkerflags by string matching is very hacky.

eshelman:
Let's not hack things. I'll close this, as it's my hack ;-)

PR #2049, opened 2015-03-06
ND convolution with im2col

jeffdonahue:
This PR extends convolution to N spatial axes, where Caffe's current convolution supports only 2D convolution (with 2 spatial axes: height and width).  For 2D convolution, this implementation doesn't compare favorably with the existing one -- I haven't done much benchmarking, but I believe it's 25-75% slower on both CPU and GPU.  So before this could be merged, I'd need to restore the existing implementation and use it as the default "engine" for 2D convolutions (but this more destructive version makes it easier to tell what I was thinking from looking at the diff).  If anyone has any suggestions on improving the performance or thoughts on why it might be so much slower, I'd love to hear them.

Edit: benchmarking this on alexnet, it's about 33% slower:

@ master:
```
I0305 21:07:25.042047 22060 caffe.cpp:271] Average Forward pass: 486.327 ms.
I0305 21:07:25.042064 22060 caffe.cpp:273] Average Backward pass: 824.147 ms.
I0305 21:07:25.042079 22060 caffe.cpp:275] Average Forward-Backward: 1310.68 ms.
```

@ nd-convolution:
```
I0305 21:02:03.827594 12909 caffe.cpp:271] Average Forward pass: 681.38 ms.
I0305 21:02:03.827608 12909 caffe.cpp:273] Average Backward pass: 1068.98 ms.
I0305 21:02:03.827623 12909 caffe.cpp:275] Average Forward-Backward: 1750.56 ms.
```

PR #2283, opened 2015-04-08
3D Convolution Layer

ddetone:
Here is a layer that implements 3D convolution. I essentially used a lot of the 2D convolution code as a guideline, and extended it to 3 dimensions. My implementation uses a vol2col utility function that is very similar the im2col (wasteful on memory but simple and pretty fast).

The layer, "Convolution3DLayer" inherits directly from "Layer", rather than "BaseConvolutionLayer".

Feedback appreciated! I'd be happy to move stuff around to have it fit best how everyone sees fit.

Next up:
-3D Max Pooling


shelhamer:
@ddetone have a look at the N-D convolution in #2049 and compare -- is there an advantage to your layer in the 3-D case? Thanks for sending a PR.

There isn't any 3-D or N-D pooling yet, so that will be a welcome improvement.

ddetone:
I didn't know about the N-D convolution branch. Cool! The implementation in #2049 is certainly a more elegant than mine.  I'll experiment with the N-D convolution branch and see if I can compare performance.

shelhamer:
Closed in favor of #2049, further tuned in #3452.

PR #2285, opened 2015-04-08
Log the average gradient norm

kevinbache:
This PR logs the average L2 norm of the minibatch gradient at each display step.  The average is taken across all the minibatches since the last time the gradient norm was logged. This is useful for helping to track the convergence of a solver; gradient norms should go to zero as we approach a local optima on the loss surface induced by the training dataset.

Tracking the gradient norm incurs a small additional cost per iteration (O(d) where d is the number of parameters in the gradient), but this will be dominated by the gradient computation itself.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -80,10 +80,16 @@ class SGDSolver : public Solver<Dtype> {
  protected:
   void PreSolve();
   Dtype GetLearningRate();
+  Dtype GetGradNorm();
+  void TrackAvgGradNorm();
+  Dtype ResetAvgGradNorm();
+  Dtype grad_norm;
+  int n_grad_norm_iters;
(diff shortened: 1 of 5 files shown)


kevinbache:
Also, I forgot to mention that this functionality can be disabled by setting the new solver parameter log_avg_gradient_norm to false (it defaults to true).

PiranjaF:
Great work, thanks! Seems very useful for finding a correct gradient clipping threshold (Pascanu et. al., 'On the diffculty of training recurrent neural networks').

shelhamer:
Closing as handled by `debug_info` https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto#L88 in `Net` https://github.com/BVLC/caffe/blob/master/src/caffe/net.cpp#L585-L636

Thanks for considering the monitoring of training.

PR #2294, opened 2015-04-10
Repair the path to bvlc_caffenet_full_conv.prototxt in net_surgery.ipynb

TorosFanny:
The original path to `bvlc_caffenet_full_conv.prototxt` is not valid. It is under `net_surgery`


shelhamer:
Manually merged in 7e98145. Thanks for catching this.

Issue #2308, opened 2015-04-13
Contrastive loss layer implementation issue


slevinkelevra:
Hi,
I am a little bit confused of an implementation of contrastive loss function. As pointed in LeCun's paper
http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf the equation of this loss function should be: `L = 0.5 * (1 - Y) * D^2 + 0.5 * Y * {max(0, margin - D)}^2`. It's an equation 4 in the original paper. As far as I understood, source code of contrastive loss layer implements this loss function in other way. Here is a piece of code (lines 48-52 of contrastive_loss_layer.cpp file):
```
if (static_cast<int>(bottom[2]->cpu_data()[i])) {  // similar pairs
      loss += dist_sq_.cpu_data()[i];
    } else {  // dissimilar pairs
      loss += std::max(margin-dist_sq_.cpu_data()[i], Dtype(0.0));
    }
```
Is it a bug in implementation?

melgor:
@slevinkelevra I have beeen analysing it too recently. I was thinning about bug too, but  then I understand it :)

1. LeCun function does averege between similar pairs and not. In Caffe there in only averege between all the pairs. <code>loss = loss / static_cast<Dtype>(bottom[0]->num()) / Dtype(2); </code> . This is not exacly the same, it depend on your input data.
2. About D^2, it it done by <code>ist_sq_.mutable_cpu_data()[i] = caffe_cpu_dot(channels,
        diff_.cpu_data() + (i*channels), diff_.cpu_data() + (i*channels)); </code>
3. (1 - Y) * D^2 is done by <code> loss += dist_sq_.cpu_data()[i]; </code>
   Y * {max(0, margin - D)}^2 is done by <code> std::max(margin-dist_sq_.cpu_data()[i], Dtype(0.0)); </code>

So, except the average,  everything is ok.

slevinkelevra:
Hm, I am not pretty sure... Look at your third point. 
The code `std::max(margin - dist_sq_.cpu_data()[i], Dtype(0.0));` 
doesn't implement `{max(0, margin - D)}^2` equation for sure! It implements this one
`{max(0, margin - D^2)}` isn't in? I might be wrong, but I think that's the main problem, not the average.

melgor:
You are right, there is the difference, good catch. 
Briefly analyzing it, the main impacts of it are:
- you need to setup other "margin" value
[...]

PR #2321, opened 2015-04-17
Fixed contrastive loss layer to be the same as proposed in Hadsell et al 2006

nickcarlevaris:
The current contrastive loss layer implements a slightly different loss than that proposed in Hadsell et all 2006.  This PR updates it so that matches the original paper. This is in reference to issue #2308.

If `d` is the distance between two vectors describing a non-matching pair, the current code implements `max(margin - d^2, 0)` while the loss proposed by Hadsell et al is `max(margin - d, 0)^2`.

@slevinkelevra and @melgor, you guys can give this PR a try and see if it works better than the current version.

### DIFF
--- a/src/caffe/layers/contrastive_loss_layer.cpp
+++ b/src/caffe/layers/contrastive_loss_layer.cpp
@@ -48,7 +48,8 @@ void ContrastiveLossLayer<Dtype>::Forward_cpu(
     if (static_cast<int>(bottom[2]->cpu_data()[i])) {  // similar pairs
       loss += dist_sq_.cpu_data()[i];
     } else {  // dissimilar pairs
-      loss += std::max(margin-dist_sq_.cpu_data()[i], Dtype(0.0));
+      Dtype dist = std::max(margin - sqrt(dist_sq_.cpu_data()[i]), 0.0);
+      loss += dist*dist;
     }
   }
   loss = loss / static_cast<Dtype>(bottom[0]->num()) / Dtype(2);
@@ -76,10 +77,12 @@ void ContrastiveLossLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
               Dtype(0.0),
               bout + (j*channels));
         } else {  // dissimilar pairs
-          if ((margin-dist_sq_.cpu_data()[j]) > Dtype(0.0)) {
+          Dtype dist = sqrt(dist_sq_.cpu_data()[j]);
+          Dtype mdist = (margin - dist);
+          if (mdist > Dtype(0.0)) {
             caffe_cpu_axpby(
                 channels,
-                -alpha,
+                -alpha * mdist / dist,
                 diff_.cpu_data() + (j*channels),
                 Dtype(0.0),
                 bout + (j*channels));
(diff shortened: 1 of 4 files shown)


seanbell:
Rather than overwriting what's already there (which breaks already-saved models), why not add a parameter option to choose which version gets used?

melgor:
It works similar like current version. But I think that @nickcarlevaris version should be merged and the old one deleted. Mainly because there is no reference about current loss function, what may cause a lot of question: is this version better, who create it?
So, I think that only one version should exist, from Hadsell et all 2006

seanbell:
@melgor I understand that the two losses are similar.  However, I've already trained networks with the current version, and I would rather not have to maintain a difference with the master branch just to avoid having old functionality deleted.  I'm okay with the new default becoming [Hadsell 2006], but I think the old functionality should be obtainable with an option in the prototxt.

norouzi:
Thanks @nickcarlevaris for the update!
I think there is a small problem in computing the gradient though, when squared distance (dist) becomes very close to zero. This causes the gradient to explode. I suggest adding a small value e.g. 1e-4 the denominator in the gradient, e.g., dividing by (dist + 1e-4).

nickcarlevaris:
@norouzi, I pushed a commit that adds the epsilon to prevent the possibility of dividing by zero.

@seanbell, my inclination would be to overwrite the current cost function with the correct one. When I originally implemented this layer (#959), my intention was to make it match the Hadsell et al paper. It is only because I didn't double check the original paper that I ended up implementing this slightly different cost function. I see this more as fixing an error. 

As far as breaking existing models goes. changing the loss layer shouldn't break the deployed version of any models, because it would only affect the training net. Also, it should be pretty easy to fine tune an existing model from the old cost to this new cost, since a network that is good for one should be pretty good for the other.

That being said, I am not strongly against keeping both. Lets see what the caffe maintainers think would be better.

seanbell:
@nickcarlevaris Ah, I didn't realize you were the one who contributed the original layer as well.  I understand that this feels like a "bugfix", but the old code is a valid loss function -- just different.  They might be equivalent in performance, maybe not.  At this point, the layer has been around through an entire publication cycle.  So I think we should treat this as deprecation and not "bugfix".  That means keeping the old version and making the new version the default.

For example, I already have a public SIGGRAPH preprint on visual similarity ( http://www.seanbell.ca/tmp/siggraph2015-preprint-bell.pdf -- section 3) that uses the old layer definition.  It would be nice if others can reproduce our results.  I bet there will be publications at CVPR this year that use the old layer.

Anyway, sure, let's see what the caffe maintainers think.

shelhamer:
@nickcarlevaris @seanbell while this is a little tricky since the layer has already been adopted by existing work, I think it is best to

1. switch the loss function to that published in Hadsell et al. 2006 as that was the original intention
2. add an option to revert to the old behavior for "legacy" papers that made use of the other loss
3. encourage authors to reproduce their results with the established Hadsell et al. 2006

if you agree. @nickcarlevaris could you add a field to `contrastive_loss_param` to pick the original or your concave variant? I defer the naming to you, since you created the loss.

Sorry for the wait and more so my apologies for not spotting the discrepancy in the equation the first time around.

nickcarlevaris:
@seanbell @shelhamer, I updated the PR as suggested. You can now get at the old behavior through a "legacy_version" parameter. Let me know if everything looks OK. I can rebase and squash if needed.

shelhamer:
Thanks @nickcarlevaris!

cancan101:
I think docs need to be updated: https://github.com/BVLC/caffe/blob/50ab52cbbf738cc92b0eda042661b8c0172774c1/include/caffe/loss_layers.hpp#L128-L151

shelhamer:
@cancan101 thanks for the report -- fixed in 7f70854.

PR #2327, opened 2015-04-17
error and weight trace

cseward:
Added functionality that allows a trace of the weights, training errors and test errors to be created. This allows the user to then create plots of how the net learns like the plots here

plot1

plot2

In addition, if the net is restored from a snapshot, the saved weights and errors are also restored so that the trace is continuous


cseward:
This PR is obsolete because it has been replaced by PR #3569

Issue #2328, opened 2015-04-17
make clean without Makefile.config tries to delete /bin and /lib


flx42:
If you do make clean without having a Makefile.config file, the command will try to "rm -r" /bin and /lib on your system. This will totally nuke your system if you have sufficient privileges. This bug is obviously the highest priority you can get.
Since Makefile fails to import Makefile.config,  $(DISTRIBUTE_DIR)/bin $(DISTRIBUTE_DIR)/lib will resolve to /bin and /lib. This also works if you do have a Makefile.config file but DISTRIBUTE_DIR is empty:
DISTRIBUTE_DIR :=  

Simple repro:
$ git clone https://github.com/BVLC/caffe.git
$ cd caffe
$ make clean
Makefile:4: Makefile.config: No such file or directory
make: *** No rule to make target `Makefile.config'.
make: Failed to remake makefile `Makefile.config'.
rm: cannot remove '/bin/setfacl': Permission denied
rm: cannot remove '/bin/sed': Permission denied
rm: cannot remove '/bin/bzegrep': Permission denied
rm: cannot remove '/bin/true': Permission denied
rm: cannot remove '/bin/mt-gnu': Permission denied
rm: cannot remove '/bin/bzfgrep': Permission denied
rm: cannot remove '/bin/dbus-cleanup-sockets': Permission denied
rm: cannot remove '/bin/uncompress': Permission denied
rm: cannot remove '/bin/open': Permission denied
rm: cannot remove '/bin/kbd_mode': Permission denied
rm: cannot remove '/bin/rbash': Permission denied
rm: cannot remove '/bin/setfont': Permission denied
rm: cannot remove '/bin/gzip': Permission denied
rm: cannot remove '/bin/findmnt': Permission denied
rm: cannot remove '/bin/zegrep': Permission denied
rm: cannot remove '/bin/lesskey': Permission denied
rm: cannot remove '/bin/busybox': Permission denied
rm: cannot remove '/bin/umount': Permission denied
rm: cannot remove '/bin/lessfile': Permission denied
rm: cannot remove '/bin/keyctl': Permission denied
rm: cannot remove '/bin/bzcat': Permission denied
[...]
rm: cannot remove '/lib/libip6tc.so.0.1.0': Permission denied
rm: cannot remove '/lib/libiptc.so.0.0.0': Permission denied
rm: cannot remove '/lib/ufw/ufw-init': Permission denied
[...]

PR #2330, opened 2015-04-17
Abort Makefile parsing if the configuration file cannot be found.

flx42:
Related to #2328

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 PROJECT := caffe
 
 CONFIG_FILE := Makefile.config
+ifeq ($(wildcard $(CONFIG_FILE)),)
+$(error $(CONFIG_FILE): file not found.)
+endif
 include $(CONFIG_FILE)
 
 BUILD_DIR_LINK := $(BUILD_DIR)


longjon:
Erm, doesn't `make` already do this? This is what I get with no `Makefile.config`:
```
$ make
Makefile:4: Makefile.config: No such file or directory
make: *** No rule to make target `Makefile.config'.  Stop.
```
(This is also why the `-include` directive exists.)

I could see an argument for a friendlier message explaining `Makefile.config`, but that's not in this PR right now.

longjon:
In fact, I can't reproduce #2328 (before the patch) unless I `make -k clean`. @flx42, do you have an alias?

flx42:
Yes, I do have an alias for make. But this barely changes the problem IMO. The make process should be aborted asap or this can potentially lead to many troubles.

longjon:
Okay, I agree. `Makefile.config` is effectively an implicit dependency for everything else, but right now `make -k` will happily cruise along with it.

flx42:
Yes, and I think adding `Makefile.config` as an explicit dependency would clutter the Makefile and create cryptic error messages. Unless you have another idea, I think a solution like this one is the simpler.

longjon:
Merged (I modified the text a little and added a comment to remind future readers why the check is necessary in caaf321a1dc6d51adb1ee0c80f76d4d172145884). Thanks @flx42!

PR #2352, opened 2015-04-22
Fix RCNN model fetching script

pgao:
I think the R-CNN example has an incorrect script. The current example fetches the ImageNet model without R-CNN (bvlc_reference_caffenet), but to run the example we would need bvlc_reference_rcnn_ilsvrc13, as is referenced in the command-line args to detect.py later in the example.


longjon:
Looks right, thanks @pgao. Your patch also adds a newline at the of the file; can you clean that up (ideally `git commit --amend`)?

pgao:
@longjon Fixed the newline addition.

PR #2363, opened 2015-04-25
Die on NaN loss

jeffdonahue:
This will make training die on a NaN loss, unless turned off by setting `SolverParameter.die_on_nan == false`.  It's been saving the GPUs I've been using many pointless cycles...

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -88,7 +88,7 @@ message NetParameter {
 // NOTE
 // Update the next available ID when you add a new SolverParameter field.
 //
-// SolverParameter next available ID: 36 (last added: clip_gradients)
+// SolverParameter next available ID: 37 (last added: die_on_nan)
 message SolverParameter {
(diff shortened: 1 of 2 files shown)


jeffdonahue:
(Of course, this adds to the existing list of annoyances associated with Caffe's checks -- e.g. in PyCaffe -- but that's kind of a separate issue... Nonetheless I can see how it probably shouldn't be merged until those things are resolved.)

longjon:
Makes sense. Yes, it might be a bad idea to crash pycaffe in this normal situation. See also #1349.

jeffdonahue:
Whoops, thanks -- should have searched... Also missed #1479 which did the same thing, sorry @sguada -- I'll leave this open, at least for now, in case anyone wants a current version of this to use privately, but if any version of this is merged it should be a rebase of #1479 rather than this one.

jeffdonahue:
Closing this as it's a duplicate and no longer mergable.

PR #2365, opened 2015-04-25
Typo fix: the label titles in plot_log.gnuplot.example

wakamori:
Fixed a tiny mistake that the titles of xlabel (Training loss) and ylabel (Training iterations) were swapped in the output image of `gnuplot plot_log.gnuplot.example`.


jeffdonahue:
Thanks @wakamori!

PR #2366, opened 2015-04-25
blocking_queue

cypof:
Part of #2351, I could not switch to #2167 because I needed boost::thread forward declared. Also instead of having a max size, I prefer to use queues in pairs: free and full, so that items can be reused when the consumer is done with them. It is particularly useful for GPU memory which is costly to allocate and destroy.

### DIFF
--- a/include/caffe/util/blocking_queue.hpp
+++ b/include/caffe/util/blocking_queue.hpp
@@ -1,0 +1,47 @@
+#ifndef CAFFE_UTIL_BLOCKING_QUEUE_HPP_
+#define CAFFE_UTIL_BLOCKING_QUEUE_HPP_
+
+#include <queue>
+#include <string>
+
+#include "caffe/common.hpp"
+
+namespace caffe {
+
+template<typename T>
+class BlockingQueue {
+ public:
+  explicit BlockingQueue();
+
+  void push(const T& t);
+
+  bool try_pop(T* t);
+
+  // This logs a message if the threads needs to be blocked
+  // useful for detecting e.g. when data feeding is too slow
+  T pop(const string& log_on_wait = "");
+
+  bool try_peek(T* t);
+
+  // Return element without removing it
+  T peek();
+
+  size_t size() const;
+
+ protected:
+  /**
+   Move synchronization fields out instead of including boost/thread.hpp
+   to avoid a boost/NVCC issues (#1009, #1010) on OSX. Also fails on
+   Linux CUDA 7.0.18.
+   */
+  class sync;
+
+  std::queue<T> queue_;
+  shared_ptr<sync> sync_;
+
+DISABLE_COPY_AND_ASSIGN(BlockingQueue);
+};
+
+}  // namespace caffe
+
+#endif
--- a/src/caffe/util/blocking_queue.cpp
+++ b/src/caffe/util/blocking_queue.cpp
@@ -1,0 +1,86 @@
+#include <boost/thread.hpp>
+#include <string>
+
+#include "caffe/util/blocking_queue.hpp"
+
+namespace caffe {
+
+template<typename T>
+class BlockingQueue<T>::sync {
+ public:
+  mutable boost::mutex mutex_;
+  boost::condition_variable condition_;
+};
+
+template<typename T>
+BlockingQueue<T>::BlockingQueue()
+    : sync_(new sync()) {
+}
+
+template<typename T>
+void BlockingQueue<T>::push(const T& t) {
+  boost::mutex::scoped_lock lock(sync_->mutex_);
+  queue_.push(t);
+  lock.unlock();
+  sync_->condition_.notify_one();
+}
+
+template<typename T>
+bool BlockingQueue<T>::try_pop(T* t) {
+  boost::mutex::scoped_lock lock(sync_->mutex_);
+
+  if (queue_.empty()) {
+    return false;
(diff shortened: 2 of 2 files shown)


include/caffe/util/blocking_queue.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <queue>
+#include <string>
+
+#include "caffe/common.hpp"
+
+namespace caffe {
+
+template<typename T>
+class blocking_queue {
```
jeffdonahue: should be called `BlockingQueue` (https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Type_Names)

jeffdonahue:
I glanced over this and it LGTM other than my capitalization nitpick, but someone who's used `boost::mutex` (not me) should probably take a look.

src/caffe/util/blocking_queue.cpp
```diff
@@ -0,0 +1,90 @@
+#include <boost/thread.hpp>
+#include <string>
+
+#include "caffe/util/blocking_queue.hpp"
```
flx42: This should be the first `#include`.

include/caffe/util/blocking_queue.hpp
```diff
@@ -0,0 +1,48 @@
...
+ public:
+  explicit blocking_queue();
+  virtual ~blocking_queue();
```
flx42: Why `virtual`? Do you plan to use it as a base class later?

src/caffe/util/blocking_queue.cpp
```diff
@@ -0,0 +1,90 @@
...
+    : sync_(new sync()) {
+}
+
+template<typename T>
+blocking_queue<T>::~blocking_queue() {
+}
+
+template<typename T>
+void blocking_queue<T>::push(const T& t) {
+  boost::mutex::scoped_lock lock(sync_.get()->mutex_);
```
flx42: Why not `sync_->mutex_`?

include/caffe/util/blocking_queue.hpp
cypof: OK will change

include/caffe/util/blocking_queue.hpp
cypof: Unlikely, to be removed

src/caffe/util/blocking_queue.cpp
cypof: Lint wants the system imports first

src/caffe/util/blocking_queue.cpp
cypof: Mistake, thks

src/caffe/util/blocking_queue.cpp
flx42: Oh, sorry, my bad.

flx42:
LGTM once you add a proper commit message :)

cypof:
Good point, fixed

src/caffe/util/blocking_queue.cpp
```diff
@@ -0,0 +1,86 @@
...
+template<typename T>
+BlockingQueue<T>::BlockingQueue()
+    : sync_(new sync()) {
+}
+
+template<typename T>
+void BlockingQueue<T>::push(const T& t) {
+  boost::mutex::scoped_lock lock(sync_->mutex_);
+  queue_.push(t);
+  lock.unlock();
```
vimalthilak: Does the push method need to make an explicit unlock() call? Have you performed any timing experiments that suggest that unlocking helps in reducing lock contention?

src/caffe/util/blocking_queue.cpp
```diff
@@ -0,0 +1,86 @@
...
+  sync_->condition_.notify_one();
+}
```
vimalthilak: In this model you are assuming that all consumer threads are created equal. Will this assumption change in the future?

src/caffe/util/blocking_queue.cpp
cypof: Calling notify_one is best done outside the locked section. It can be done inside, but this way allows the notified thread to run immediately.

src/caffe/util/blocking_queue.cpp
cypof: Right now there is always a single consumer thread, but we might find other uses for this class.

src/caffe/util/blocking_queue.cpp
flx42: Some reference:
http://en.cppreference.com/w/cpp/thread/condition_variable/notify_one

> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock. However, some implementations (in particular many implementations of pthreads) recognize this situation and avoid this "hurry up and wait" scenario by transferring the waiting thread from the condition variable's queue directly to the queue of the mutex within the notify call, without waking it up.

src/caffe/util/blocking_queue.cpp
longjon: Yeah, I think lint is wrong here with respect to the Google style guide; ideally we'll fix that at some point, but we've been following lint on this point in other places so best just to stay consistent here.

shelhamer:

PR #2368, opened 2015-04-25
Persistent prefetch thread

cypof:
Part of #2351. Makes sure the prefetch thread for data_layer is created only once, instead of for each batch. We discussed about switching to vector<Blob> instead of Batch, which would probably be easy but I haven't looked at it.

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -98,12 +98,12 @@ void GlobalInit(int* pargc, char*** pargv);
 class Caffe {
  public:
   ~Caffe();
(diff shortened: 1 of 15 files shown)


amuetzel:
Regarding the vector: do you mean using a vector<Blob> instead of the Batch with the data and label members?

If so, I would favor that option. An example use case for this would be custom data layers that expose more than 2 tops. I'm currently working on such a data layer for my research, and might be able to help with the change. (Though I'm still new to caffe.)

shelhamer:

PR #2369, opened 2015-04-26
Makefile bugfix: debug/release dir link not toggled properly

jeffdonahue:
This fixes a bug that was causing the `build` link to not properly toggle between `.build_release` and `.build_debug`.  Due to the lazy variable expansion when using the `?=` assignment operator, `$OTHER_BUILD_DIR` was getting assigned to `..build_release_debug` rather than `.build_debug` (or `..build_debug_release` rather than `.build_release`) and the `.linked` file wasn't getting removed to tell make to move the link when the `DEBUG` option is toggled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,12 @@ CONFIG_FILE := Makefile.config
 include $(CONFIG_FILE)
 
 BUILD_DIR_LINK := $(BUILD_DIR)
-RELEASE_BUILD_DIR ?= .$(BUILD_DIR)_release
-DEBUG_BUILD_DIR ?= .$(BUILD_DIR)_debug
+ifeq ($(RELEASE_BUILD_DIR),)
+	RELEASE_BUILD_DIR := .$(BUILD_DIR)_release
+endif
+ifeq ($(DEBUG_BUILD_DIR),)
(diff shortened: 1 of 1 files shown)


longjon:
I guess the semantics of `BUILD_DIR` change throughout the file (which is a little disorienting by the way, it seems like `Makefile.config` variables should retain their values in `Makefile`...). Anyway, you could alternatively:
* just use `BUILD_DIR_LINK`
* just use simple assignment (is this actually useful functionality?)

jeffdonahue:
I don't have any preference (so I guess I prefer not to do anything except remove the extraneous comma I just noticed, but am happy to take either suggestion). @shelhamer switched it from simple assignment in 6d183710, any thoughts Evan?

jeffdonahue:
Merging after removing extraneous comma and discussion with @longjon and @shelhamer.

alfredtofu:
It is still not fixed. But I am not familiar with Makefile, I have no idea how to solve it. 
Do as the following step to reproduce the bug:
1. make DEBUG=0
2. make DEBUG=1
3. make DEBUG=0, this time, the link target of build is not changed

PR #2370, opened 2015-04-26
move CPU_ONLY to right place

gdh1995:
At other places, `CPU_ONLY` are used on such a way:
```
    case Caffe::GPU:
#ifndef CPU_ONLY
      ...
#else
    NO_GPU;
#endif
      break;
```

A typo in comment is also fixed.

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -19,7 +19,7 @@
 #include "caffe/util/device_alternate.hpp"
 
 // gflags 2.1 issue: namespace google was changed to gflags without warning.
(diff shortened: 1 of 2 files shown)


longjon:
Yes, this is equivalent but seems like a slightly better style; `case` blocks should generally end with an explicit control flow statement. Thanks for the update @gdh1995! I'm going to go ahead and merge this small change, but two comments for the future:
* we'd appreciate separate PRs for unrelated changes (even small typo fixes; that way we can always merge those without delay)
* please try to be precise with your commit messages so we can read history easily (e.g., here it's `break` rather than `CPU_ONLY` that's in the wrong place)

Thanks!

PR #2383, opened 2015-04-28
Thread start and stop revamp

cypof:
Split off #2367. Now that threads can be waiting on locks or condition variables, it can be necessary to interrupt them to stop the app. This PR makes sure stopping a Caffe thread always calls interrupt before waiting on join. It also provides a method 'must_stop' that can be tested when running in a loop to exit on demand.

Start and stop methods now fail a CHECK instead of returning a boolean error, given that the application doesn't really have a way to recover from this type of error.

### DIFF
--- a/include/caffe/internal_thread.hpp
+++ b/include/caffe/internal_thread.hpp
@@ -25,12 +25,11 @@ class InternalThread {
    * Caffe's thread local state will be initialized using the current
    * thread values, e.g. device id, solver index etc. The random seed
    * is initialized using caffe_rng_rand.
-   * Will not return until the internal thread has exited.
    */
-  bool StartInternalThread();
+  void StartInternalThread();
 
   /** Will not return until the internal thread has exited. */
-  bool WaitForInternalThreadToExit();
+  void StopInternalThread();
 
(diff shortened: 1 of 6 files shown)


flx42:
Looks OK, but using CHECK instead of the boolean return value is debatable I think.

include/caffe/internal_thread.hpp
```diff
@@ -14,18 +14,22 @@ namespace caffe {
...
-  bool StartInternalThread();
+  /**
+   * Caffe's thread local state will be initialized using the current
+   * thread values, e.g. device id, solver index etc. The random seed
+   * is initialized using caffe_rng_rand.
+   */
+  void StartInternalThread();
 
   /** Will not return until the internal thread has exited. */
-  bool WaitForInternalThreadToExit();
```
futurely: Should a public method be deleted without labeling it as `@deprecated` for some time?

flx42: I think it's supposed to be implementation details of the thread mechanism used internally, so I would say yes.

shelhamer:
No longer needed with the switch in #4563.

PR #2386, opened 2015-04-29
Data Reader

cypof:
Part of #2351, but switched to a round-robin way of distributing data to solvers, instead of the shared queue that was not deterministic. Combined with random seeds initialization on threads in #2367, it should make parallel training reproducible.

The data reader sits between a database and each solver's prefetch thread. It makes sure each solver processes a different subset of the database. It also prefetches data, but to host memory, and the amount is configurable. Solvers prefetch threads instead only prefetch a fixed small amount of data, since it is stored in GPU memory.

* A single reading thread is created per source, even if multiple solvers are running in parallel.
* Sources are identified by layer name + source path, in case net has multiple data layers on same DB.
* Databases are read sequentially, for better performance.
* Each solver sees a different subset of the database.
* Data is distributed to solvers in a round-robin way to keep parallel training deterministic.

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -98,12 +98,12 @@ void GlobalInit(int* pargc, char*** pargv);
 class Caffe {
  public:
   ~Caffe();
-  inline static Caffe& Get() {
-    if (!singleton_.get()) {
-      singleton_.reset(new Caffe());
-    }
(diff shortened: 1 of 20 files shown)


shelhamer:
Merged with revisions in #2903, thanks.

PR #2393, opened 2015-04-29
Little CMake fix (can be reviewed and merged quickly)

Nerei:
* After the fix Caffe will complain if glog/gflags are not found at CMake time (without waiting compilation error). The bug was in `FindGlog/FindGFlag.cmake` scripts due to case sensitiveness of CMake.


longjon:
Looks good, thanks.

PR #2394, opened 2015-04-30
cmake detects Atlas BLAS and Lapack under Fedora/RHEL/CentOS

rajish:
Currently Fedora Core package for ATLAS doesn't contain neither `cblas` nor `lapack` libraries included. That's why a separate steps have to be made by installing `blas-devel` and `lapack-devel`.

Adjustments in `FindAtlas.cmake` help to find the libraries and headers under said systems family.


bkanuka:
I also needed these changes in order to cmake on CentOS 7

PR #2395, opened 2015-04-30
Added timing info for iterations, data loading, and multi-GPU

cypof:
The data loading and multi-GPU ones are optional through a Makefile.config variable.


shelhamer:
Excluded from #2903 since the timing was a bit intrusive, but thanks for including this while we benchmarked the parallel branch.

PR #2396, opened 2015-04-30
Add virtual destuctors to virtual classes

cypof:


longjon:
If I'm not mistaken this change has no effect... all of these classes are derived from classes with `virtual` destructors, and therefore their destructors are automatically `virtual`. (Indeed, base classes should always have `virtual` destructors.)

See, for example, https://isocpp.org/wiki/faq/virtual-functions#virtual-dtors; in particular:

> Note: in a derived class, if your base class has a virtual destructor, your own destructor is automatically virtual. You might need an explicitly defined destructor for other reasons, but there’s no need to redeclare a destructor simply to make sure it is virtual. No matter whether you declare it with the virtual keyword, declare it without the virtual keyword, or don’t declare it at all, it’s still virtual.

Is there some other reason for making this change?

flx42:
I agree with @longjon

cypof:
Ah OK, good to know, closing.

PR #2397, opened 2015-04-30
Refactor solvers regularization and logging code

cypof:
Update: changed description to "refactor"


flx42:
Maybe "refactor" instead of "deduplicate"?

cypof:
More precise type of refactoring? :)

flx42:
I think "deduplication" is usually used for data, not for code. But I might be wrong :)

shelhamer:
Adapted and merged in #2518. Thanks for this @cypof!

PR #2399, opened 2015-04-30
Makefile use env variables for config and test filters

cypof:
Handy to build different configs without updating files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROJECT := caffe
 
-CONFIG_FILE := Makefile.config
+CONFIG_FILE ?= Makefile.config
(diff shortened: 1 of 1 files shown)


longjon:
I don't see any reason not to make this change (e.g., these don't seem likely to cause namespace conflicts with the environment). Does anyone else?

flx42:
Those variable names are quite generic, especially `CONFIG_FILE`. Adding a `CAFFE_` prefix might be better, something like this maybe?
```
CONFIG_FILE := $(CAFFE_CONFIG_FILE)
CONFIG_FILE ?= Makefile.config
```

PR #2401, opened 2015-04-30
Test for Snapshot and Restore in Solver

prateekbs:
I have written some preliminary code to test Snapshot and Restore in the Solver. However, it wrecks the current interfaces by converting Snapshot, SnapshotSolverState and RestoreSolverState from protected members to public members. 

Is testing non public code frowned upon? If the answer to the previous question is no, then how do I go about testing protected members?

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -31,6 +31,12 @@ class Solver {
   // previously snapshotted state. You should implement the RestoreSolverState()
   // function that restores the state from a SolverState protocol buffer.
   void Restore(const char* resume_file);
+  // The Solver::Snapshot function implements the basic snapshotting utility
+    // that stores the learned net. You should implement the SnapshotSolverState()
+    // function that produces a SolverState protocol buffer that needs to be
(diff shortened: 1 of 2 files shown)


flx42:
Be careful, your indentation is totally messed up. Don't use tabs, use 2 spaces.

prateekbs:
@flx42: Thanks for the heads up. I fixed the indentation.

jyegerlehner:
> Is testing non public code frowned upon? If the answer to the previous question is no, then how do I go about testing protected members?

When there is a need to access protected or private members for test purposes, the technique I've usually seen used is to create a trivial subclass of the class you are trying to test, e.g. SolverForTest, in the test code which exposes public accessors to the protected members of the base class. Then run the test against that subclass. That way the "real" class keeps its members protected, but they can still be accessed in the test.

prateekbs:
@jyegerlehner: Thanks.

shelhamer:
Closing for the interference in the interface, but thanks for the testing effort.

Issue #1738, opened 2015-01-15
Separate dependencies for configurable installation


shelhamer:
Caffe has dependencies that aren't necessarily universally needed especially when it comes to IO. The installation could be streamlined by making

- leveldb
- lmdb
- hdf5
- and even OpenCV

optional so that the Python or MATLAB interfaces can take care of IO or a client C++ library could handle input.

Of course there is an advantage to having a standard all-in build since input formats and portability can be counted on, but there are cases when it's overkill.

This would need a fleet of build flags and ifdef guards and is perhaps best left until the build has been standardized by work like #1667.

Note that this was first suggested by @kloudkl in #1074 but it was too early and at the time and the build needed to improve before turning our attention to separation.

bhack:
As I stated some month ago when cmake was introduced in caffe it is better to support only one build system to avoid divergence between make tools and cmake and to not duplicate community and core members effort to maintain two build systems. Especially now that we have https://github.com/BVLC/caffe/pull/1667

Nerei:
Excellent idea! Once cmake standardized and all switch it, I need one weekend to prepare PR and test all these features.

NB: nothing will bring so much popularity to your library as porting it to iOS. Optional dependencies with cmake is just one of steps required to reach this.

bhack:
My original comment was here https://github.com/BVLC/caffe/pull/943. Also with only one build system we drastically reduce travis testing time of the build matrix.

futurely:
#1074

shelhamer:
Closed by #2523 for optional IO dependencies -- HDF5 was made a hard dependency by #2836.

PR #2402, opened 2015-04-30
Make IO dependencies optional (leveldb lmdb snappy hdf5 opencv)

cypof:
It's probably not what you want to do in the long run, but it's simple and fits the use case of embedding Caffe in another framework. That's by far the most important aspect of dependency management for us and probably a lot of people. As you suggested last week data can be processed externally and provided through the memory_data_layer.

The cmake build would need to get updated, I can look at it if you are interested to merge this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
-# Use a build matrix to do two builds in parallel:
-# one using CMake, and one using make.
+# Use a build matrix to do builds in parallel:
 env:
   matrix:
     - WITH_CUDA=false WITH_CMAKE=false
     - WITH_CUDA=false WITH_CMAKE=true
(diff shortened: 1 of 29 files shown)


futurely:
The omnipotent `NO_IO_DEPENDENCIES` is simpler than but not as flexible as the macros used in #1074 to manage each of the libraries independently. #1738 proposed to also improve CMake scripts.

cypof:
Closing, probably not that useful outside of our use case.

shelhamer:
There is actually interest in having a barebones Caffe without all the IO, so I'll leave this open for now so that it's more visible. While granular switches are the ideal, this all-or-nothing switch might be useful for some in the meantime.

shelhamer:
Closing as replaced by #2523 but thanks for pioneering this.

PR #2408, opened 2015-05-04
Add skip and rewind to data layer.

jnhwkim:
skip var to skip a few data points (< N).
rewind var to rewind the data cursor for overlapping learning.


jnhwkim:
I did notice that it requires build check through Travis CI.

PR #2410, opened 2015-05-04
Datum transform

sguada:
Refactor Data_Transform to be able to estimate the Blob sizes after transforming Datum. This simplify the Data Layers and will allow to do more transformations when using encoded Datum.

Having replicated code across Data Layers and Data_Transform to estimate Blob sizes makes hard to add new transformation, like resize the image to have an specific small size.

@longjon @shelhamer Please take a look at let me know 
@cypof I think this should help with the IO

### DIFF
--- a/include/caffe/data_transformer.hpp
+++ b/include/caffe/data_transformer.hpp
@@ -62,6 +62,7 @@ class DataTransformer {
    */
   void Transform(const vector<cv::Mat> & mat_vector,
                 Blob<Dtype>* transformed_blob);
+
   /**
    * @brief Applies the transformation defined in the data layer's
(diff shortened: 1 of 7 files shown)


jeffdonahue:
Thanks for the refactor @sguada! LGTM.

jeffdonahue:
Travis' [merge build](https://travis-ci.org/BVLC/caffe/builds/64670190) failed due to the LMDB download failure, as reported by @ronghanghu in https://github.com/BVLC/caffe/pull/2527#issuecomment-106978985.  We'll have to find a more reliable download URL for LMDB...I think we changed to the current one because the previous one was causing our build to fail as well.

PR #2426, opened 2015-05-07
Check that count does not overflow in Blob::Reshape

longjon:
`master`/nd blob edition of #1584, using the method suggested by @jeffdonahue.

As far as I can tell, it's still possible to overflow `count`, leading to bad allocations and other incorrect behavior; this should prevent that.


jeffdonahue:
cool, thanks Jon!

PR #2441, opened 2015-05-11
Fixed wrong io import in Python 3

gustavla:
The Python standard lib has a module called `io`, so instead of Python 3
throwing an error at `import io`, it imports the wrong module without complaining.


longjon:
Looks good, thanks. From a glance at the Google style guide and related advice, it seems like we should probably just switch all of these to absolute imports; if that seems right, PRs are welcome.

PR #2456, opened 2015-05-14
Use bp::object instead of PyObject* for self in Python layer

longjon:
From @tnarihi in #2001. Rather than couple it to the other changes in that PR, I've extracted this little cleanup for immediate review and merge.

@tnarihi, I've used your author information here; let me know if there are any issues with that.


jeffdonahue:
LGTM.  Thanks @tnarihi and thanks for breaking off this piece @longjon.

PR #2462, opened 2015-05-15
Handle Python layer exceptions correctly

longjon:
(This is a simplified alternative implementation of part of the first commit of #2001, plus a test.)

As noted by @tnarihi, Python layer exceptions are not handled correctly in pycaffe. `PyErr_Print` is used to display exception text on stderr, but this has the side effect of clearing the exception, which results in an additional `SystemError: NULL return without exception set`.

The exceptions get passed through correctly if we simply don't try to catch and print them; that's the first commit. (Note that the Python interpreter _will_ print these exceptions, if uncaught; it's only the `caffe` tool that won't.) Since we'd still like to see the exception text when using the `caffe` tool, the second commit adds a single, `ifdef`ed, `catch`/`PyErr_Print` block to the tool code. (There's no harm in calling `PyErr_Print` since we know we're not in the interpreter.)

I prefer this approach to the one in #2001, for these reasons:
* it simplifies the Python wrapper code, removing the duplicate error catching blocks
* it removes the need to think about where exceptions might occur in future wrapper code
* it avoids the business of embedding Python calls to read (and restore) the exception
* calling code (from C++) has the option of avoiding the exception printout (e.g., and dealing with the exception)

The disadvantage of this approach is that:
* Python exceptions won't be automatically printed when invoking nets with Python layers from C++ code (outside the Python interpreter) other than the `caffe` tool

Let me know what you think @tnarihi and others, and thanks @tnarihi for pointing out this issue. (Note also that this does not include _all_ of the exception-fixing functionality from #2001, which I'll plan to check out separately.)

### DIFF
--- a/include/caffe/python_layer.hpp
+++ b/include/caffe/python_layer.hpp
@@ -18,22 +18,11 @@ class PythonLayer : public Layer<Dtype> {
 
   virtual void LayerSetUp(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
-    try {
-      self_.attr("setup")(bottom, top);
-    } catch (bp::error_already_set) {
-      PyErr_Print();
-      throw;
-    }
+    self_.attr("setup")(bottom, top);
   }
-
   virtual void Reshape(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
-    try {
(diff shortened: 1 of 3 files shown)


shelhamer:
Merging as I agree that this is simpler than the solution in #2001. Thanks Jon and thanks Takuya for raising the issue!

Issue #2440, opened 2015-05-11
Why is caffes Python binding not a package (hosted on PyPI)?


MartinThoma:
I am wondering why caffe has [these installation instructions](http://caffe.berkeleyvision.org/installation.html#python) instead of simply creating a Python package and allowing the user to install it via

```bash
$ pip install caffe
```

If the only reason is time / knowledge how to create such a package, I could do that.

lukeyeager:
:+1: 

But, you can't really do this without choosing a version number (#486).

MartinThoma:
Hmmm ... I thought this might be as easy as creating a `setup.py` like

```python
from setuptools import find_packages
from setuptools import setup
import pip.download
from pip.req import parse_requirements

# parse_requirements() returns generator of pip.req.InstallRequirement objects
install_reqs = parse_requirements('requirements.txt',
                                  session=pip.download.PipSession())

# reqs is a list of requirement
# e.g. ['django==1.5.1', 'mezzanine==1.4.6']
reqs = [str(ir.req) for ir in install_reqs]

config = {
    'name': 'caffe',
    'version': '0.1.0',
    'author': 'BVLC members and the open-source community',
    'packages': find_packages(),
    'scripts': ['bin/classify.py', 'bin/detect.py',
                'bin/draw_net.py'],
    'platforms': ['Linux', 'MacOS X', 'Windows'],
    'url': 'https://github.com/BVLC/caffe',
    'license': 'BSD',
    'description': ('Caffe is a deep learning framework made with expression, '
                    'speed, and modularity in mind.'),
    'long_description': ('Caffe is a deep learning framework made with '
                         'expression,  speed, and modularity in mind. It is '
                         'developed by the Berkeley Vision and Learning '
                         'Center (BVLC) and community contributors.'),
    'install_requires': reqs,
    'keywords': ['caffe', 'deep learning'],
    'download_url': 'https://github.com/BVLC/caffe',
    'classifiers': ['Development Status :: 3 - Alpha',
                    'Environment :: Console',
                    'Intended Audience :: Developers',
                    'Intended Audience :: Science/Research',
                    'License :: OSI Approved :: BSD License',
                    'Natural Language :: English',
                    'Programming Language :: Python :: 2.7',
                    'Programming Language :: Python :: 3',
                    'Programming Language :: Python :: 3.3',
                    'Programming Language :: Python :: 3.4',
                    'Topic :: Scientific/Engineering :: Artificial Intelligence',
[...]

PR #2467, opened 2015-05-15
Python: Formatted docstrings to numpydoc (Take, Give -> Parameters, Returns)

MartinThoma:
Again a very simple, documentation-only pull request.
(I did not see those the first time. Now I `grep`ed through the Python code to make sure the I got all "Give" and "Take")

### DIFF
--- a/python/caffe/classifier.py
+++ b/python/caffe/classifier.py
@@ -12,17 +12,17 @@ class Classifier(caffe.Net):
     """
     Classifier extends Net for image class prediction
     by scaling, center cropping, or oversampling.
+
+    Parameters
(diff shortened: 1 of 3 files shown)


longjon:
Nice, thank you @MartinThoma!

We should look into putting these on the web at some point...

MartinThoma:
@longjon It would be much easier to put them on the web (e.g. pythonhosted.org or readthedocs.org; both are well-known free (without advertising) hosting services for Python package documentation) if the Python part was a package. However, I am not sure how this package should deal with the `_caffe.so` file which needs to be generated. See issue #2440

PR #2473, opened 2015-05-16
int->size_t to support large datasets more than 2G instances

buaaliyi:
When I was trying to use Caffe to train my large dataset (billions of instances), I found the class 'SyncedMemory' uses data type 'size_t' to alloc memories, while blob.count_ and blob.capacity_ is of type 'int'. As a result, this have cut off the alloc size to less than 2GB, and my experiment was failed due to the pointer overflow.
This patch fixed the data size related types from int to size_t that guarantee to use the correct size on 64bit machine, even though dataset size is over 2 billions.

Thanks for the review.

### DIFF
--- a/src/caffe/util/io.cpp
+++ b/src/caffe/util/io.cpp
@@ -41,6 +41,7 @@ bool ReadProtoFromTextFile(const char* filename, Message* proto) {
 
 void WriteProtoToTextFile(const Message& proto, const char* filename) {
   int fd = open(filename, O_WRONLY | O_CREAT | O_TRUNC, 0644);
+  CHECK_NE(fd, -1) << "Filei path not exists: " << filename;
   FileOutputStream* output = new FileOutputStream(fd);
   CHECK(google::protobuf::TextFormat::Print(proto, output));
   delete output;
@@ -64,6 +65,7 @@ bool ReadProtoFromBinaryFile(const char* filename, Message* proto) {
 
 void WriteProtoToBinaryFile(const Message& proto, const char* filename) {
   fstream output(filename, ios::out | ios::trunc | ios::binary);
+  CHECK(output) << "File path not exists: " << filename;
   CHECK(proto.SerializeToOstream(&output));
 }
 


mynameisfiber:
Excited to see this patch pass the Travis tests... I'm running into the same issue!

.gitignore
```diff
@@ -32,6 +32,10 @@
 # Editor temporaries
 *.swp
 *~
+tags
+TAGS
+*.tags
+cscope.*
```
flx42: This should not be part of this patch.

include/caffe/data_layers.hpp
```diff
@@ -264,7 +264,8 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
 
   // Reset should accept const pointers, but can't, because the memory
   //  will be given to Blob, which is mutable
-  void Reset(Dtype* data, Dtype* label, int n);
+  void Reset(Dtype* data, Dtype* label, size_t n);
+
   void set_batch_size(int new_size);
```
flx42: No need to add a new empty line here.

src/caffe/blob.cpp
```diff
@@ -24,9 +24,11 @@ void Blob<Dtype>::Reshape(const vector<int>& shape) {
   CHECK_LE(shape.size(), kMaxBlobAxes);
   count_ = 1;
   shape_.resize(shape.size());
+  bool use_64bit = (sizeof(count_) == 8);
   for (int i = 0; i < shape.size(); ++i) {
     CHECK_GE(shape[i], 0);
-    CHECK_LE(shape[i], INT_MAX / count_) << "blob size exceeds INT_MAX";
+    CHECK_LE(shape[i], (use_64bit ? ULONG_MAX : UINT_MAX) / count_)
+      << "blob size exceeds " << use_64bit ? "ULONG_MAX" : "UINT_MAX";
```
flx42: That's a bit ugly. What about using std::numeric_limits instead?

jeffdonahue:
Thanks for the PR @buaaliyi, and for the reviewing efforts @flx42. I think we do eventually want to increase the blob size limit.  A couple of comments:

* In https://github.com/BVLC/caffe/pull/1970#discussion_r25580035 we effectively set a ceiling of `INT64_MAX` on future blob size -- see the discussion there for reasoning on not using unsigned ints (in short, there are places in the code that use negative dimensions as special values) . So unless there is further discussion with good reasoning otherwise, this should use `int64_t` rather than `size_t`. (Maybe typedef'ing is the way to go? `blob_size_t` is so verbose though...)

* There are a *lot* more places in the codebase that need to be updated before a PR increasing the blob size limit could be merged -- I think almost every C source file currently has `int`s based on the current max size of blobs lurking about.

flx42:
Why not use `size_t`? Functions `malloc` and `cudaMalloc*` take `size_t` as the size argument. If `int64_t` is used instead it would require additional range checks before allocating memory, especially on a 32-bit architecture where we likely have `sizeof(size_t) != sizeof(int64_t)`

buaaliyi:
This patch has been updated base on @flx42 's comments

jmcq:
I need this same change.  I had just filed #3159 in error because I did not search properly, and will close it if I can.

I would resolve this by using ssize_t (signed size_t) everywhere you use int to hold a size but are willing to forgo the highest bit in order to get special negative values.  Wherever you are using unsigned int, you could use size_t.   This should be a straightforward change, at least for g++ on linux, though it will probably touch most files.

I can prepare a CPU-tested change set for a pull request if one of the developers is willing to consider it.

Issue #2068, opened 2015-03-08
set_mean(self, in_, mean) in python/io.py: mean has dimension issue when using the default "ilsvrc_2012_mean.npy"


lxduan:
When passing the mean image matrix in the default mean file, i.e. *ilsvrc_2012_mean.npy*, as `mean` to `caffe.Classifier` in *https://github.com/BVLC/caffe/blob/master/python/classify.py* (master branch), `set_mean(self, in_, mean)` in *python/io.py* detects dimension error, because `mean` is of size (3, 256, 256) while `self.inputs['data'][1:]` is of (3, 227, 227).

Any plan to fix this issue, possibly by also cropping the mean matrix into (3, 227, 277) ones? Currently, a workaround is to pass `mean=np.load(caffe_root + 'python/caffe/imagenet/ilsvrc_2012_mean.npy').mean(1).mean(1)` to `caffe.Classifier`. However, I feel it's more appropriate to pass the full mean matrix, considering it's more meaningful to do a pixel-wise subtraction.

longjon:
It looks like this is a preprocessing bug in `classify.py`; the mean should be subtracted before crops are taken, instead of the other way around. In general it's more robust to interface directly with `caffe.Net` rather than using `classify.py`.

rasmuspjohansson:
I can confirm that the workaround (mean=np.load(caffe_root + 'python/caffe/imagenet/ilsvrc_2012_mean.npy').mean(1).mean(1)) not is good enough.
On a net trained on my own dataset, accuracy drops from ~90% during bash script training to ~50%  when I use the classifer.py method with the mean(1).mean(1) workaround for classifying the same images one by one. 
In my own dataset the position in the image correlates strongly with pixel values, this must be the reason for the drastic change in accuracy when using means not sensitive to position.

This needs to be fixed in order for the python interface to be usable. 
Is anyone working on a fix? or maybe someone can point me in the right direction for fixing it on my own?

PR #2476, opened 2015-05-17
Fix classify.py mean shape and refactor oversample to crop_image

Hsinfu:


Hsinfu:
Sorry, didn't see the discussion #2068

PR #2482, opened 2015-05-18
Clean up redundant protobuf message comments

longjon:
`caffe.proto` has a series of comments that feel pretty redundant to me, in lines of the form
```
// Message that stores parameters used by XLayer
message XParameter {
```

This PR simply removes those comments, in favor of a single explanatory comment before all the layer parameter messages.


flx42:
Yeah, those comments seems to have little value.

shelhamer:
Agreed, thanks for stripping these.

PR #2477, opened 2015-05-17
fix bug and add only forward mode in testing time

alfredtofu:
1. fix bug for getting the index of  output_blob_indices
2. one can chooses to test time of forward only

PR #2484, opened 2015-05-18
[bug] fix blob_loss_weights index in test() in caffe.cpp

ronghanghu:
This PR fix a simple yet annoying bug in test() in caffe.cpp.
Currently `caffe test` crashes when there are many outputs in an output blob

A simple example to reproduce this bug:
`caffe test -model sample_net.prototxt -weights /dev/null`

where `sample_net.prototxt` is

    layer {
      name: 'data'
      type: 'DummyData'
      top: 'data'
      dummy_data_param {
        # it crashes when output blob has count >= 2, because of the bug in caffe.cpp
        shape { dim: 3 }
      }
    }
    layer {
      name: 'relu'
      type: 'ReLU'
      bottom: 'data'
      top: 'relu'
    }

### DIFF
--- a/tools/caffe.cpp
+++ b/tools/caffe.cpp
@@ -187,8 +187,8 @@ int test() {
   for (int i = 0; i < test_score.size(); ++i) {
     const std::string& output_name = caffe_net.blob_names()[
         caffe_net.output_blob_indices()[test_score_output_id[i]]];
-    const float loss_weight =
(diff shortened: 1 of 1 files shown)


ronghanghu:
Sorry, didn't see there is already a PR #2477 containing a fix for this bug. Close

shelhamer:
Thanks for the fix Ronghang! Since #2477 isn't quite ready and includes other changes I would rather merge this now.

PR #2493, opened 2015-05-20
Fix dangerous state in pooling and LRN CUDA kernels

longjon:
Several of Caffe's CUDA kernels (in particular, in the pooling and LRN layers) take pointers to input/output arrays, but deal only with slices of those arrays, selected by an offset computed using the thread index.

The code that performs this selection is a bit lazy by modifying arguments in place (with code that resembles `data += offset`). This worked fine since that code was only executed once. With the addition of `CUDA_KERNEL_LOOP` by #239, these increments end up inside a loop, and produce incorrect behavior if executed multiple times. (This is not normally noticed because all of the `CUDA_KERNEL_LOOP`s used in Caffe so far actually only execute once.)

Aside from the kernel code being incorrect in this case, I'm in general not a big fan of this style of modifying things in place that don't need to be modified in place; it makes the code harder to read by transferring more state to the reader's head, it changes the usual meaning of variable names, and it can lead to subtle bugs like this one.

This PR cleans up this dangerous state by creating new names for the offset/slice pointers. As a follow on, it beefs up the `const`-ness of these kernels to mitigate future issues of this nature. (In particular, while kernels were taking non-pointer arguments as `const`, they were not taking pointer arguments as such).

To my knowledge, this exhaustively fixes the issues of this nature (bad increment inside `CUDA_KERNEL_LOOP`).

### DIFF
--- a/src/caffe/layers/lrn_layer.cu
+++ b/src/caffe/layers/lrn_layer.cu
@@ -18,8 +18,8 @@ __global__ void LRNFillScale(const int nthreads, const Dtype* in,
     int n = index / width / height;
     int offset = (n * channels * height + h) * width + w;
     int step = height * width;
-    in += offset;
-    scale += offset;
+    const Dtype* const in_off = in + offset;
+    Dtype* const scale_off = scale + offset;
     int head = 0;
     int pre_pad = (size - 1) / 2;
     int post_pad = size - pre_pad - 1;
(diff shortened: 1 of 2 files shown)


shelhamer:
#2145 addressed this for pooling kernels, but this PR is more complete since it fixes the LRN kernels and makes kernels more const.

PR #2498, opened 2015-05-21
Add a simple C++ classification example

### LINKED ISSUE #2487: Simple C++ prediction example
Hi,

I want to contribute a simple C++ prediction example to the codebase, is this something you are interested in adding? If yes, what kind of features would you like to see in this sample app?

From my point of view, I think the sample application should take as input the following files: deploy file, trained weights, labels, mean file and input image to classify.
It would be nice to be able to use file `models/bvlc_reference_caffenet/deploy.prototxt` as-is, without requiring to add an extra layer for preprocessing (like cropping, resizing, mean subtraction), so I think the processing should be done from the sample app after probing the network for the dimensions of its input layer. This way, the app should hopefully work with caffenet and MNIST by just changing the input files (and those files are either already present in the repo, or are downloadable through a provided script).

What do you think? Any other feature request? I'm already excluding features like image batching or oversampling from the list, for the sake of readability.

flx42:
Closes #2487

Example usage:
./build/examples/cpp_classification/classification.bin \
  models/bvlc_reference_caffenet/deploy.prototxt \
  models/bvlc_reference_caffenet/bvlc_reference_caffenet.caffemodel \
  data/ilsvrc12/imagenet_mean.binaryproto \
  data/ilsvrc12/synset_words.txt \
  examples/images/cat.jpg

### DIFF
--- a/examples/cpp_classification/classification.cpp
+++ b/examples/cpp_classification/classification.cpp
@@ -1,0 +1,255 @@
+#include <caffe/caffe.hpp>
+#include <opencv2/core/core.hpp>
+#include <opencv2/highgui/highgui.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+#include <iosfwd>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+using namespace caffe;  // NOLINT(build/namespaces)
+using std::string;
+
+/* Pair (label, confidence) representing a prediction. */
+typedef std::pair<string, float> Prediction;
+
+class Classifier {
+ public:
+  Classifier(const string& model_file,
+             const string& trained_file,
+             const string& mean_file,
+             const string& label_file);
+
+  std::vector<Prediction> Classify(const cv::Mat& img, int N = 5);
+
+ private:
+  void SetMean(const string& mean_file);
+
+  std::vector<float> Predict(const cv::Mat& img);
+
+  void WrapInputLayer(std::vector<cv::Mat>* input_channels);
+
+  void Preprocess(const cv::Mat& img,
+                  std::vector<cv::Mat>* input_channels);
+
+ private:
+  shared_ptr<Net<float> > net_;
+  cv::Size input_geometry_;
+  int num_channels_;
+  cv::Mat mean_;
+  std::vector<string> labels_;
+};
+
+Classifier::Classifier(const string& model_file,
+                       const string& trained_file,
+                       const string& mean_file,
+                       const string& label_file) {
+#ifdef CPU_ONLY
+  Caffe::set_mode(Caffe::CPU);
+#else
+  Caffe::set_mode(Caffe::GPU);
+#endif
+
+  /* Load the network. */
+  net_.reset(new Net<float>(model_file, TEST));
+  net_->CopyTrainedLayersFrom(trained_file);
+
+  CHECK_EQ(net_->num_inputs(), 1) << "Network should have exactly one input.";
+  CHECK_EQ(net_->num_outputs(), 1) << "Network should have exactly one output.";
+
+  Blob<float>* input_layer = net_->input_blobs()[0];
+  num_channels_ = input_layer->channels();
+  CHECK(num_channels_ == 3 || num_channels_ == 1)
+    << "Input layer should have 1 or 3 channels.";
+  input_geometry_ = cv::Size(input_layer->width(), input_layer->height());
+
+  /* Load the binaryproto mean file. */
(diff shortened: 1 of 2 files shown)


bhack:
Why you don't use Memorydatalayer that has an opencv interface?

lukeyeager:
The example builds for me and I like the format.

But I don't think the documentation for a C++ example belongs in an iPython notebook. It looks like the proper way to document examples is by adding a `readme.md` to `examples/cpp_classification/readme.md`). [This script](https://github.com/BVLC/caffe/blob/rc2/scripts/gather_examples.sh) automatically adds them to this list: http://caffe.berkeleyvision.org/#examples.

flx42:
> Why you don't use Memorydatalayer that has an opencv interface?

For flexibility during preprocessing of the sample. Also, if you use MemoryDataLayer you have to modify the deploy file and hardcode the dimensions IIRC.
With my approach, I simply probe the dimensions of the input layer and resize to this size.

flx42:
> But I don't think the documentation for a C++ example belongs in an iPython notebook. It looks like the proper way to document examples is by adding a readme.md to examples/cpp_classification/readme.md). This script automatically adds them to this list: http://caffe.berkeleyvision.org/#examples.

That's a good point, I will do that and modify the iPython notebook to only mention this example.

flx42:
Updated to use a `readme.md` file like @lukeyeager suggested.

examples/cpp_classification/classification.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+  std::vector<float> Predict(const cv::Mat& img);
+
+  void WrapInputLayer(std::vector<cv::Mat>* input_channels);
+
+  void Preprocess(const cv::Mat& img,
+                  std::vector<cv::Mat>* input_channels);
+
+ private:
+  std::auto_ptr<Net<float> > net_;
```
vimalthilak: Could you change this to using boost's smart pointer or perhaps opencv's smart pointer? If C++11 support is available this should be a unique_ptr. 

 auto_ptr hopefully never makes it into caffe

flx42: > Could you change this to using boost's smart pointer or perhaps opencv's smart pointer?

This sample is not using Boost at all, I would prefer to keep it this way especially if it's only to use a simple feature, even if Caffe already depends on Boost. Regarding OpenCV, it's uncommon to see code using their smart pointer class. Also, in this sample OpenCV should only be used for image storage and processing IMO.

>  If C++11 support is available this should be a unique_ptr. 

What do you mean? Caffe is not compiled with C++11 so this is not possible, if the project ever switches to C++11, I would be happy to use `unique_ptr` indeed.

> auto_ptr hopefully never makes it into caffe

What is wrong with auto_ptr? I can disable the copy constructor of the `Classifier` class if you are concerned about the copy semantic of auto_ptr.

vimalthilak: auto_ptr is deprecated in C++11. It's quite plausible that caffe at some point would move to using C++11/C++14 (and hopefully before C++17 implementations are available to us!)

I'm using a caffe fork that's CPU only that I'm building with C++11. I'd like to use your code as my "application" layer, hence my request to use boost's smart pointers instead of auto_ptr.  

I understand that you don't want to build out dependencies but given that boost is being used anyway, my preference would be to use one of its smart pointer. It's almost tragic that C++11

flx42: OK, I understand, you made a good point. The patch has been updated. Thanks!

examples/cpp_classification/readme.md
```diff
@@ -0,0 +1,75 @@
...
+include_in_docs: true
+priority: 10
+---
+
+# Classifying ImageNet: using the C++ API
+
+Caffe, at its core, is written in C++. It is possible to use
+the C++ API of Caffe to implement an image classification application
+similar to the Python code presented in one of the Notebook example.
+
```
shelhamer: Thanks for the example. As a finishing touch, could you include a reference to the `caffe` [command line tool source](https://github.com/BVLC/caffe/blob/master/tools/caffe.cpp) as a further example of C++ code? Although I might hope this is an obvious place to check, it seems it has been overlooked in the past.

Adding a pointer after this intro paragraph could help as a stepping stone to seeing more C++ code.

flx42: Done, thanks for the review!

shelhamer:
Thanks for the example! (Ignore the travis failure -- it was an unrelated server issue.)

Issue #858, opened 2014-08-05
finetune_net.bin don't speak~


dabilied:
I use finetune_net.bin to retrain a net ,however these is not any debug information;and the program is runing; "finetune_net.bin img_solver.prototxt caffe_reference_imagenet_model";How chould this be?

PR #2504, opened 2015-05-23
Suppress output for python bindings

lukeyeager:
I'm tired of pycaffe printing hundreds of lines to stderr. See #789, #858 and #872 for discussions on why this was done by default for C++.

Is suppressing the output when using python a bad idea for any reason? Is there a better way to suppress the output than this (it's kinda hacky)? GLOG flags explained [here](https://google-glog.googlecode.com/svn/trunk/doc/glog.html#flags).

### DIFF
--- a/python/caffe/__init__.py
+++ b/python/caffe/__init__.py
@@ -1,6 +1,19 @@
+# Suppress GLOG output for python bindings
+# unless explicitly requested in environment
+import os
+if 'GLOG_minloglevel' not in os.environ:
+    # Hide INFO and WARNING, show ERROR and FATAL
+    os.environ['GLOG_minloglevel'] = '2'
+    _unset_glog_level = True
+else:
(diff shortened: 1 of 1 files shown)


lukeyeager:
Edited to allow ERROR and FATAL messages to propagate through.

lukeyeager:
Based on comments in [this thread](https://groups.google.com/forum/#!topic/caffe-users/4-HtwKqE5xM), it seems like people still want to see the output during training. That makes sense - I've never used pycaffe for training, only for working with trained models.

Can anyone who uses pycaffe for training comment on whether any version of this PR could be useful? Does the solver store the training updates somewhere, or is the log the only way to access that data?

IvyGongoogle:
see [here](https://stackoverflow.com/questions/29788075/setting-glog-minloglevel-1-to-prevent-output-in-shell-from-caffe).

Issue #1500, opened 2014-11-29
Make Net master of its phase and device


shelhamer:
Leaving phase and device to the Caffe singleton is awkward and spells trouble for parallelization and device abstraction.

Net should own phase and device, which are to be set at instantiation, so that nets can co-exist without the delicate coordination of phase and device for parallel execution within and across nets.
- [x] give Phase to Net #1790
- [ ] give mode / device to Net
- [ ] diffuse mode / device to Layer

~~At the same time phase issues like #192 #1250  #1430 should be settled at the same time.~~ Phase has been settled in #1790.

@longjon @jeffdonahue

PR #2511, opened 2015-05-26
Fix invalid mode changes during tests

flx42:
Some existing tests are modifying the Caffe mode halfway through the execution, this is documented to be invalid:
https://github.com/BVLC/caffe/blob/8df472/include/caffe/common.hpp#L140-L143

If, for performance reasons, host memory is allocated through `cudaMallocHost`, changing the mode halfway can cause a pointer returned by `cudaMallocHost` to be freed by `free(2)`, resulting in undefined behavior. The reciprocal is also possible. Another possible issue is that if some tests incorrectly assume that the default mode is CPU, the test could actually run on the GPU if the previous test clobbered the global mode. See the full analysis of this issue in #2398 

The solution is, IMHO, to forbid calls to `Caffe::set_mode()` in individual test cases, this function should only be called by the test framework in order to limit the risks of a misuse. To achieve this, the following patch set reuses the existing `MultiDeviceTest` class and similarly add new classes `GPUDeviceTest` and `CPUDeviceTest`. In the case where we need to share code between CPU and GPU tests, the shared test code can directly derive from class `MultiDeviceTest` but derived classes needs to be defined for CPU and GPU.

### DIFF
--- a/include/caffe/test/test_caffe_main.hpp
+++ b/include/caffe/test/test_caffe_main.hpp
@@ -40,34 +40,28 @@ class MultiDeviceTest : public ::testing::Test {
 
 typedef ::testing::Types<float, double> TestDtypes;
 
-struct FloatCPU {
-  typedef float Dtype;
-  static const Caffe::Brew device = Caffe::CPU;
-};
-
-struct DoubleCPU {
-  typedef double Dtype;
+template <typename TypeParam>
+struct CPUDevice {
+  typedef TypeParam Dtype;
   static const Caffe::Brew device = Caffe::CPU;
 };
 
 #ifdef CPU_ONLY
 
-typedef ::testing::Types<FloatCPU, DoubleCPU> TestDtypesAndDevices;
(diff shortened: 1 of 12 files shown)


shelhamer:
This looks good to me. Independent of deciding what's right for mode and device in general, templated tests seem like a more robust approach to making sure the mode is right.

Exactly what to do with mode and device is an ongoing conversation, but I think diffusing mode + device to `Net`s, `Layer`s, and `Solver`s and making it immutable is reasonable. The dismantling of the singleton / diffusing of the handles is on the charts as #1500 (at least for `Net`).

I don't know that we ever fully converged on this however. @longjon @jeffdonahue comment if you remember any threads.

jeffdonahue:
LGTM too.  Not sure what we'll end up doing with `mode` but this looks like it could only make any transition away from what we have now smoother (by centralizing/decreasing the number of references to `mode` in the code base).

PR #2514, opened 2015-05-26
Add leading zeros (up to 10 digits) to keys in feature DB files + Bug fix regarding multiple features to be extracted

norouzi:
The feature extraction code, stores the features in a DB file using string keys generated from integer ids. The string keys are sorted alphabetically, and therefore, the order of features in the feature DB file becomes different from the order of images in the input DB file. (My experience is based on LMDB database.)

I have added some leading zeros to fix this.

There is also a bug with regard to the db_type when the number of features to be extracted is larger than 1. This bug happens because ++arg_pos is run multiple times.

I added a bug fix fir this to the PR too.


shelhamer:
Thanks for the fixes.

PR #2522, opened 2015-05-27
python/draw_net.py and python/caffe/draw.py

MartinThoma:
Simplified code; added more docstrings; adjusted code according to PEP8


shelhamer:
Thanks for the Python grooming!

Issue #2006, opened 2015-02-28
Can't serialize when protobuf size > 2gb


Russell91:
I'm getting segfaults from write proto with large models. It seems that protobuf simply doesn't support writing messages greater than 2gb (https://github.com/marbl/harvest-tools/issues/3). Does anyone have a workaround they would suggest? It would be nice if we added a check on the message size to give a nicer error message if nothing else.

Here are the details:

I0228 14:22:15.306717 23672 solver.cpp:355] Snapshotting to /snapshots/caffe_iter_10.caffemodel

Program received signal SIGSEGV, Segmentation fault.
0x000000000047fa74 in caffe::LayerParameter::GetCachedSize (
    this=0x903d621f263c2284) at .build_debug/src/caffe/proto/caffe.pb.h:2063

The backtrace from gdb:

#0  0x000000000047fa74 in caffe::LayerParameter::GetCachedSize (
    this=0x903d621f263c2284) at .build_debug/src/caffe/proto/caffe.pb.h:2063
#1  0x000000000048cedd in google::protobuf::internal::WireFormatLite::WriteMessageNoVirtualToArray<caffe::LayerParameter> (field_number=2, value=..., 
    target=0x2305c18b "<\v\a\237\274>\025\254\274r\376\245<g̪<\003`\332<n\345\061=\241\374\202\273>ֶ\274J2|=\201\300\326\067\203\211.=\212\035\371\274\224\337B\275\025&ɼ") at /usr/include/google/protobuf/wire_format_lite_inl.h:708
#2  0x0000000000433cbe in caffe::NetParameter::SerializeWithCachedSizesToArray
    (this=0x7fffffffd160, 
    target=0x2305c18a "\022<\v\a\237\274>\025\254\274r\376\245<g̪<\003`\332<n\345\061=\241\374\202\273>ֶ\274J2|=\201\300\326\067\203\211.=\212\035\371\274\224\337B\275\025&ɼ") at .build_debug/src/caffe/proto/caffe.pb.cc:3530
#3  0x00007ffff2094d0c in google::protobuf::MessageLite::SerializePartialToCodedStream(google::protobuf::io::CodedOutputStream*) const ()
   from /usr/lib/x86_64-linux-gnu/libprotobuf.so.8
#4  0x00007ffff2094dc5 in google::protobuf::MessageLite::SerializeToCodedStream(google::protobuf::io::CodedOutputStream*) const ()
[...]

PR #2523, opened 2015-05-28
Separate dependencies for configurable installation

eelstork:
Addresses #1738.
The PR benefited earlier attempts by @cypof (#2402) and @kloudkl (#1074).
Options provided include USE_OPENCV, USE_LMDB, USE_LEVELDB, USE_HDF5, USE_SNAPPY, with both make and CMake flags available.
This allows selecting which IO libraries to build with. Wherever possible functions that would behave incorrectly as a result of disabling a library are excluded before runtime.

Added two items to the test matrix:
    - WITH_CUDA=false WITH_CMAKE=false WITH_IO=false
    - WITH_CUDA=false WITH_CMAKE=true WITH_IO=false

Without bloating the test matrix, it is essential to have at least one of these, otherwise correct separation of these dependencies would be unmaintainable.


flx42:
Instead of adding patches fixing your own mistakes, squash multiple patches in one patch and force push to your branch to update the PR.

cmake/Summary.cmake
```diff
@@ -114,17 +114,30 @@ function(caffe_print_configuration_summary)
...
+  caffe_status("  gflags            :   Yes")
   caffe_status("  protobuf          : " PROTOBUF_FOUND THEN "Yes (ver. ${PROTOBUF_VERSION})" ELSE "No" )
-  caffe_status("  lmdb              : " LMDB_FOUND THEN "Yes (ver. ${LMDB_VERSION})" ELSE "No")
-  caffe_status("  Snappy            : " SNAPPY_FOUND THEN "Yes (ver. ${Snappy_VERSION})" ELSE "No" )
-  caffe_status("  LevelDB           : " LEVELDB_FOUND THEN  "Yes (ver. ${LEVELDB_VERSION})" ELSE "No")
-  caffe_status("  OpenCV            :   Yes (ver. ${OpenCV_VERSION})")
+  if(USE_LMDB)
+    caffe_status("  lmdb              : " LMDB_FOUND THEN "Yes (ver. ${LMDB_VERSION})" ELSE "No")
+  endif()
+  if(USE-SNAPPY)
```
Nerei: Misprint

cmake/Dependencies.cmake
```diff
@@ -24,24 +24,36 @@ list(APPEND Caffe_LINKER_LIBS ${GFLAGS_LIBRARIES})
...
 
 # ---[ HDF5
-find_package(HDF5 COMPONENTS HL REQUIRED)
-include_directories(SYSTEM ${HDF5_INCLUDE_DIRS} ${HDF5_HL_INCLUDE_DIR})
-list(APPEND Caffe_LINKER_LIBS ${HDF5_LIBRARIES})
+if(USE_HDF5)
+  find_package(HDF5 COMPONENTS HL REQUIRED)
+  include_directories(SYSTEM ${HDF5_INCLUDE_DIRS} ${HDF5_HL_INCLUDE_DIR})
+  list(APPEND Caffe_LINKER_LIBS ${HDF5_LIBRARIES})
+  add_definitions(-DUSE_HDF5)
```
Nerei: Don't use definitions. Use caffe_config.h instead

include/caffe/common.hpp
```diff
@@ -65,8 +65,10 @@ private:\
 // is executed we will see a fatal log.
 #define NOT_IMPLEMENTED LOG(FATAL) << "Not Implemented Yet"
 
+#ifdef USE_OPENCV
```
Nerei: Suppose this is not required

shelhamer:
@eelstork do keep hacking at this! It will be great to have separate dependencies in line with #1738.

cmake/Dependencies.cmake
Nerei: Hm.. in this case this have to be installed

Nerei:
This PR doesn't generate proper CaffeConfig.cmake. All such definitions should be propagated to client code too.

eelstork:
@Nerei thanks for the many helpful suggestions. Working on this.

eelstork:
@Nerei the forward declaration of cv::Mat would not be helpful since cv:Mat will not be available at runtime?

Nerei:
@eelstork Agree. If user try to call opencv dependent method, it will get link errors. Well this is all up to code organization. In general, It's preferable to hide as much as possible `USE_*` defines into cpp, but this might require more significant caffe code change. That's IMHO.

Regarding commit c5ebe85, I think it should be reverted. Such config should be included into each header (otherwise that defines are always undefined) but it is not generated by Make. I put my comment above and then disproved it.

eelstork:
@Nerei referring c5ebe858b0f9f027d67fa8293a2e594365140f5e, If add_definitions() is not portable this may need fixing later but for now caffe_config.h appears to be somewhat ineffective. I would suggest that we leave things as-is (consistent with handling of 'CPU_ONLY' flag).

eelstork:
@Nerei please have a look at the generated CaffeConfig.cmake; if this is not appropriate, I welcome suggestions to improve it.

eelstork:
Pending further feedback, standing by.

flx42:
Please squash the patches, it's difficult to review right now.

Nerei:
Fine with CaffeConfig.cmake

bhack:
@eelstork Can you rebase and squash?

eelstork:
@bhack @flx42 done.

keenbrowne:
@shelhamer @bhack @flx42 @eelstork 
What work is left to get this merged? We would like to depend on these changes in our caffe based application.

eelstork:
@shelhamer @keenbrowne ready for review/merge.

Makefile.config.example
```diff
@@ -7,6 +7,13 @@
 # CPU-only switch (uncomment to build without GPU support).
 # CPU_ONLY := 1
 
+# Comment out to disable IO dependencies
+USE_LEVELDB := 1 # {IO-flag}
+USE_LMDB := 1    # {IO-flag}
+USE_HDF5 := 1    # {IO-flag}
+USE_OPENCV := 1  # {IO-flag}
+USE_SNAPPY := 1  # {IO-flag}
+
```
sh1r0: The trailing comments (# {IO-flag}) might cause problems for value evaluation.

bhack:
@eelstork Have you tested if with this new modularity can build with androd-cmake: https://github.com/taka-no-me/android-cmake/blob/master/README.md

bhack:
@sh1r0 This add also apk generation capabilties: https://github.com/Discordia/android-cmake

eelstork:
Removed trailing comments in makefile.config. Thanks to @sh1r0 for the suggestion.

eelstork:
@bhack No, I haven't; nor do I know whether it was possible before? Although we hope that removing the dependencies will help porting Caffe to other platforms, I believe that compatibility with a specific platform/config should be addressed separately.

bhack:
@eelstork Yes of course. I've asked only if somebody tested it with this modularization to know now if could be any other modularization that could improve the process with android NDK.

bhack:
I really hope that we are not start to incoming in a [preprocessor abuse](http://jhnet.co.uk/articles/cpp_magic) syndrome caused by bad code modularization design.

bhack:
Android effort started at https://github.com/BVLC/caffe/pull/2619

eelstork:
@bhack We need more information (and more time/resources) to improve modularisation design. Imho, directives can help transition towards a better design later.

eelstork:
Rebased, squashed. Still awaiting review/merge.

bhack:
@eelstork You need to wait. The status of core devs It is still unkown. The last know one was CVPR 2015 and there is no community management other than core devs and occasional BVLC traiger students.

src/caffe/layers/data_layer.cpp
```diff
@@ -1,3 +1,4 @@
+#ifdef USE_OPENCV
```
shelhamer: This isn't strictly needed since lmdb / leveldb data that is unencoded does not require OpenCV IO. I think `DataLayer` should only require lmdb / leveldb. If one tries to use encoded data without OpenCV the `DataTransformer` guards will report the dependency failure.

src/caffe/test/test_upgrade_proto.cpp
```diff
@@ -2889,6 +2889,7 @@ TEST_F(NetUpgradeTest, TestImageNet) {
   this->RunV1UpgradeTest(expected_v1_proto, expected_v2_proto);
 }  // NOLINT(readability/fn_size)
 
+#if defined (USE_OPENCV) && defined(USE_HDF5)
```
shelhamer: Although ugly, it would be safer to always run this for the layers that are instantiated and have guards within the loop to skip the disabled layer types.

tools/compute_image_mean.cpp
```diff
@@ -115,5 +116,8 @@ int main(int argc, char** argv) {
     }
     LOG(INFO) << "mean_value channel [" << c << "]:" << mean_values[c] / dim;
   }
+#else
+  LOG(FATAL) << "Compile with OpenCV to use this.";
```
shelhamer: "This tool requires OpenCV." would match the text for the examples although it might be best if the message everywhere were "This [whatever] requires [dependency]: compile with [dependency] to use."

tools/convert_imageset.cpp
```diff
@@ -148,5 +149,8 @@ int main(int argc, char** argv) {
     txn->Commit();
     LOG(ERROR) << "Processed " << count << " files.";
   }
+#else
+  LOG(FATAL) << "Compile with OpenCV to use this.";
```
shelhamer: Match this to other messages.

shelhamer:
@eelstork thanks for splitting the dependencies! @longjon @jeffdonahue and I agree that it is better to separate IO. I defer to @Nerei regarding the CMake changes; I only know the Makefile (and that looks fine).

hdf5 can be left bundled since we plan to switch to it for serializing weights and solverstates to fix #2006.

Since the Makefile is more simple, the test matrix could include the CMake == true, IO == false case to catch more issues if the two make testing too slow.

Nerei:
CMake changes look fine for me.

eelstork:
@shelhamer about hdf5: currently, default config is with all dependencies enabled to maintain backwards compatibility. 
So, if I understand correctly, you suggest that the option to build without HDF5 should be removed.

@shelhamer @Nerei thanks for your feedback, I will update, resolve conflicts and resubmit for review/merge.

bhack:
Actually hdf5 it is the new default model format. So it is an hard dependency for Caffe and cannot be modularized anymore if we not want to fallback to proto. So actually if we follow the default the minimal dependencies of caffe it is increased.

bhack:
Is this mergiable?

eelstork:
@bhack I have completed several updates reflecting Shelhamer's suggestions; I will merge these back into the PR branch momently; in the meantime please refer https://github.com/eelstork/caffe/tree/separate-io-update 
Still need to remove USE_HDF5 but current commit works.

eelstork:
@shelhamer actioned all suggested changes; I would suggest keeping the test matrix as is since it doesn't appear to be slowing things down much and will help maintaining this feature.
Ready for review/merge.

docs/installation.md
```diff
@@ -24,9 +24,9 @@ Caffe has several dependencies.
     * 5.5, and 5.0 are compatible but considered legacy
 * [BLAS](http://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms) via ATLAS, MKL, or OpenBLAS.
 * [Boost](http://www.boost.org/) >= 1.55
-* [OpenCV](http://opencv.org/) >= 2.4 including 3.0
+* [OpenCV](http://opencv.org/) >= 2.4 including 3.0 (optional)
 * `protobuf`, `glog`, `gflags`
-* IO libraries `hdf5`, `leveldb`, `snappy`, `lmdb`
+* IO libraries `hdf5`, `leveldb`, `snappy`, `lmdb` (optional)
```
shelhamer: hdf5 should be listed with protobuf, glog, and gflags as required. Could you place these required dependencies under boost in this list so that all the hard requirements come before the optional?

scripts/travis/travis_build_and_test.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Script called by Travis to do a CPU-only build of and test Caffe.
+# Script called by Travis to build and test Caffe.
```
shelhamer: To be clear the Travis CI tests are CPU-only and this should stay in the comments. The GPU code is compiled but not tested for lack of hardware.

shelhamer:
Thanks @eelstork! I made last comments on minor docs + comment changes so this looks good to me for merge once those are taken care of.

Second opinions @longjon @jeffdonahue ?

src/caffe/util/db.cpp
```diff
@@ -8,24 +8,32 @@ namespace caffe { namespace db {
...
+#ifdef USE_LMDB
+if (backend == "lmdb") {
     return new LMDB();
-  } else {
-    LOG(FATAL) << "Unknown database backend";
   }
-}
+#endif
+  LOG(FATAL) << "Unknown database backend";
+  }
```
jeffdonahue: fix indentation

jeffdonahue:
See above nit, otherwise mostly LGTM.  I'd prefer if the `#endif` guard comments matched the `#if` ones like the current header guards, like so:

```
#ifdef USE_LEVELDB
...
#endif  // USE_LEVELDB
```

But that's minor and a lot of work to change; not really necessary IMO.

bhack:
@jeffdonahue @shelhamer What do you think of the related https://github.com/BVLC/caffe/pull/2619?

eelstork:
@shelhamer and @jeffdonahue thanks for these suggestions, I have made changes accordingly (see relevant commits); squash? 
Pending review/merge.

shelhamer:
Thanks @eelstork! This looks good to me, so please squash for merge. However 526372a can stay its own commit to explain Travis CI is CPU-only although the message should change to reflect that.

eelstork:
@shelhamer Thanks! Before letting go, I suggest we drop the option to explicitly enable/disable snappy. This is because caffe does not directly depend on Snappy, instead, it is LevelDB that requires it. Where build scripts relied on associate flag, USE_LEVELDB would be used instead. Intending to push a commit to illustrate proposed solution.

eelstork:
@shelhamer @jeffdonahue @bhack @Nerei, opinions about  58f017e8c11dfbe1db63d8b8e34af266e48e6b93 ?

shelhamer:
@eelstork 584f017e is good, but please squash with a5bdc06 and add an explanatory commit message that lists the new flags and explain that leveldb + snappy are coupled since snappy is just a leveldb dependency. Thanks.

eelstork:
@shelhamer - updated; squashed.

eelstork:
@Shelhamer Anything missing to get this merged?

jeffdonahue:
@eelstork I took another look -- sorry for the last minute comments.  Otherwise looks good for merge to me.

eelstork:
@jeffdonahue updated. I would suggest leaving 2349c6de69bf5043508cde41bb1d337fdb78e188 as a separate commit.

jeffdonahue:
Thanks @eelstork!

bhack:
@jeffdonahue Can you take a look at related https://github.com/BVLC/caffe/pull/2523

shelhamer:
@mtngld thanks for pointing this out. 

@jeffdonahue thoughts on making all IO on by default for backwards compatibility? The Makefile.config / CMake can then explicitly disable dependencies as desired.

jeffdonahue:
Oops, I didn't realize this changed the default to no IO dependencies. Yes, I agree the default should be to have them all enabled as before.

danst18:
Currently it is not possible to disable leveldb and lmdb (at least on Windows) which would be useful to use Caffe for deployment only.

eelstork:
@danst18 I believe that disabling LevelDB and LMDB is possible on Windows (via CMake), granted that building caffe on Windows does require tweaks. What motivates your comment?

danst18:
@eelstork I mean when both of them are disabled at the same time. In db.cpp I get an error saying that GetDB should return a value. If I remove db.h and db.cpp then I have to add guards to other parts of the code where they are referenced, for example in data_reader.cpp

eelstork:
@danst18  sorry for the late reply. I'm looking into it.

PR #2528, opened 2015-05-30
Fix lmdb install for travis

shelhamer:
switch to git mirror

- ~~risky since lmdb might change upstream~~
- download the latest tag... which looks safe enough given tags from 2013 are still present
- safe b.c. the static download urls go stale

### DIFF
--- a/scripts/travis/travis_install.sh
+++ b/scripts/travis/travis_install.sh
@@ -47,12 +47,12 @@ if $WITH_CUDA; then
 fi
 
 # Install LMDB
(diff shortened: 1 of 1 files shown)


jeffdonahue:
> risky since lmdb might change upstream

How about checking out the latest tagged release? (LMDB_0.9.14 it looks like?)

shelhamer:
Amended to just download the tarball... let's see if this works.

ronghanghu:
The current link seems to resurrected a few minutes ago, so my PR now passes Travis. But indeed a stable download link is needed and I agree lmdb github release should be a stable one.

PR #2532, opened 2015-06-01
PReLU accumulates grad

shelhamer:
Adapt the PReLU gradient computation to accumulation now that #1977 is in.

@tnarihi could you double-check this? I cherry-picked your commit and cleaned up comments. Thanks.

### DIFF
--- a/src/caffe/layers/prelu_layer.cpp
+++ b/src/caffe/layers/prelu_layer.cpp
@@ -113,7 +113,6 @@ void PReLULayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
   // keep top_diff unchanged.
   if (this->param_propagate_down_[0]) {
     Dtype* slope_diff = this->blobs_[0]->mutable_cpu_diff();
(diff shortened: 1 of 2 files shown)


tnarihi:
Looks good. Thanks Evan!
Also, in order to test if every layer is implemented for gradient accumulation I made the following chagne in my branch: https://github.com/tnarihi/caffe/commit/7d455263dc66b7d663340c12458795dc86f7466d
Please consider this as well.

shelhamer:
@jeffdonahue could you decide what to do with  tnarihi@7d45526 too since you and Takuya talked about the gradient checker? Thanks.

jeffdonahue:
LGTM, thanks @tnahari and @shelhamer.   I'm also not sure of the best way to go with changing the gradient checker -- always checking accumulation is safer, but anyone who's developed a layer of their own with parameters may be confused as to why their implementation is suddenly failing the gradient checker.  I guess ideally there would be a first pass that does the current gradient check and, only if that passes, also checks accumulation afterwards and fails with an error message pointing to failed accumulation if applicable.

PR #2536, opened 2015-06-02
Remove unnecessary filler parameter in the sample model

kibum14:
Unnecessary filler parameters, 'std', were used with Xavier weight fillers in bvlc_googlenet model.


jeffdonahue:
Thanks @kibum14!

PR #2537, opened 2015-06-02
First attemp to remove boost dependecies.

bhack:
This is an initial effort to generally remove boost dependencies and leave it only for python wrapper.
I think that c++11 support it is quite stable now, enabled by default in gcc 5.x series, full supported by LLVM and almost supported in the android NDK.
With https://github.com/BVLC/caffe/pull/2523 modularity we could really strip caffe dependencies on embedded systems and especially for android NDK builds.
@shelhamer @longjon Can you take a look at this?
@Nerei Generally the Travis build matrix of caffe doesn't seems coherent to me between Cmake and Make (Cmake really build cuda?).
I think that @shelhamer can make a run on OSX because we don't have it in Travis.
Generally this seems in a good shape, CPU and GPU tests pass fine. 
There is still a problem with a python test as you can see in Travis logs. 
Can you give me some feedback on this?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ include(cmake/Dependencies.cmake)
 
 # ---[ Flags
 if(UNIX OR APPLE)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall -std=c++11")
 endif()
 
 if(USE_libstdcpp)
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ ifneq ($(CPU_ONLY), 1)
 	LIBRARIES := cudart cublas curand
 endif
 LIBRARIES += glog gflags protobuf leveldb snappy \
-	lmdb boost_system hdf5_hl hdf5 m \
+	lmdb hdf5_hl hdf5 m \
 	opencv_core opencv_highgui opencv_imgproc
 PYTHON_LIBRARIES := boost_python python2.7
 WARNINGS := -Wall -Wno-sign-compare
@@ -233,7 +233,7 @@ ifeq ($(LINUX), 1)
 	endif
 	# boost::thread is reasonably called boost_thread (compare OS X)
 	# We will also explicitly add stdc++ to the link target.
-	LIBRARIES += boost_thread stdc++
+	LIBRARIES += stdc++
 endif
 
 # OS X:
@@ -253,7 +253,7 @@ ifeq ($(OSX), 1)
 	# gtest needs to use its own tuple to not conflict with clang
 	COMMON_FLAGS += -DGTEST_USE_OWN_TR1_TUPLE=1
 	# boost::thread is called boost_thread-mt to mark multithreading on OS X
-	LIBRARIES += boost_thread-mt
+	#LIBRARIES += boost_thread-mt
 	# we need to explicitly ask for the rpath to be obeyed
 	DYNAMIC_FLAGS := -install_name @rpath/libcaffe.so
 	ORIGIN := @loader_path
@@ -344,16 +344,18 @@ LIBRARY_DIRS += $(BLAS_LIB)
 
 LIBRARY_DIRS += $(LIB_BUILD_DIR)
 
+INCLUDE_DIRS += /usr/include/hdf5/serial
+
 # Automatic dependency generation (nvcc is handled separately)
 CXXFLAGS += -MMD -MP
 
 # Complete build flags.
 COMMON_FLAGS += $(foreach includedir,$(INCLUDE_DIRS),-I$(includedir))
-CXXFLAGS += -pthread -fPIC $(COMMON_FLAGS) $(WARNINGS)
-NVCCFLAGS += -ccbin=$(CXX) -Xcompiler -fPIC $(COMMON_FLAGS)
+CXXFLAGS += -pthread -fPIC -std=c++11 $(COMMON_FLAGS) $(WARNINGS)
+NVCCFLAGS += -ccbin=$(CXX) -Xcompiler -fPIC -std=c++11 $(COMMON_FLAGS)
 # mex may invoke an older gcc that is too liberal with -Wuninitalized
 MATLAB_CXXFLAGS := $(CXXFLAGS) -Wno-uninitialized
-LINKFLAGS += -pthread -fPIC $(COMMON_FLAGS) $(WARNINGS)
+LINKFLAGS += -pthread -fPIC -std=c++11 $(COMMON_FLAGS) $(WARNINGS)
 
 USE_PKG_CONFIG ?= 0
 ifeq ($(USE_PKG_CONFIG), 1)
@@ -441,7 +443,7 @@ py: $(PY$(PROJECT)_SO) $(PROTO_GEN_PY)
 
 $(PY$(PROJECT)_SO): $(PY$(PROJECT)_SRC) $(PY$(PROJECT)_HXX) | $(DYNAMIC_NAME)
 	@ echo CXX/LD -o $@ $<
-	$(Q)$(CXX) -shared -o $@ $(PY$(PROJECT)_SRC) \
+	$(Q)$(CXX) -std=c++11 -shared -o $@ $(PY$(PROJECT)_SRC) \
 		-o $@ $(LINKFLAGS) -l$(PROJECT) $(PYTHON_LDFLAGS) \
 		-Wl,-rpath,$(ORIGIN)/../../build/lib
 
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -2,9 +2,9 @@
 set(Caffe_LINKER_LIBS "")
 
 # ---[ Boost
-find_package(Boost 1.46 REQUIRED COMPONENTS system thread)
-include_directories(SYSTEM ${Boost_INCLUDE_DIR})
-list(APPEND Caffe_LINKER_LIBS ${Boost_LIBRARIES})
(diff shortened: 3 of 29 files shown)


thatguymike:
While I agree with the general direction, until C++11 is supported by default compilers on distributions with long term support, this is going to be difficult for many users to contend with.  For example, for Ubuntu 14.04, GCC version is 4.8.2 still by default.  CentOS 7 appears to also be < 5.0.

bhack:
@thatguymike This is actually building with g++4.8 on Travis. I've cited GCC 5.X series only for the "default flag".

thatguymike:
That makes me more comfortable, but I thought with 4.8, C++11 was "touchy", but maybe that is limited to regex and misc utilities.  So perhaps this just needs a LOT of testing across platforms.  I am still worried that some users will get pushed out of Caffe because they are stuck on something old like RedHat 6.x (we still see a lot of that...) or an embedded platform that doesn't quite have all the needed compiler support available.

flx42:
gcc 4.8 C++11 language support:
https://gcc.gnu.org/gcc-4.8/cxx0x_status.html
gcc 4.8 C++11 standard library support:
https://gcc.gnu.org/onlinedocs/gcc-4.8.4/libstdc++/manual/manual/status.html#status.iso.2011

Regex support is indeed broken with gcc 4.8, the problem is that it will fail at runtime. It's fixed in 4.9:
https://gcc.gnu.org/onlinedocs/gcc-4.9.2/libstdc++/manual/manual/status.html#status.iso.2011

Caffe is not using regexes in C++ right now, so it should be fine. But I agree with @thatguymike, this will require a lot of testing on all platforms, including Android. For RedHat 6.x or embedded platforms support, this is really a strategical/philosophical decision you have to make about the project.

bhack:
@flx42 All the features we are actually using in Caffe/boost it is supported in gcc 4.8. In fact this PR it is set to build with gcc4.8 in Travis. I think that the 4.8 it is quite available in long term support distro.

flx42:
> All the features we are actually using in Caffe/boost it is supported in gcc 4.8. 

That's exactly what I said.

> In fact this PR it is set to build with gcc4.8 in Travis.

It could build but fail at runtime in subtle ways. Travis doesn't run GPU tests and doesn't support all the platforms we might be interested in. So I think this requires further testing.

bhack:
@flx42 Yes I've published here to collect feebacks. People can get this PR and try to run it on his own environment. Probably you and @thatguymike could be the first two :smile:. 
I don't think that @thatguymike has problem to find a nvidia GPU to run gpu test :wink:

vimalthilak:
@bhack: Do you need an extra pair of hands to test this PR on OS X/ubuntu bistro

PS: Please add iOS to your list of "have to support" OSes.

bhack:
@vimalthilak Yes please test.
Forma IOS you can  caffe-preset trougth robovm. See https://github.com/bytedeco/javacpp-presets/issues/34

bhack:
@longjon Do you have any hint on why this python test is failing?

bhack:
@thatguymike @flx42 Have you tested on your environment?

flx42:
It worked for me!
The Travis error you are getting is puzzling, did you manage to get a repro locally?

bhack:
@flx42 Have you runner python test? make pycaffe then make pytest

vimalthilak:
@bhack: On on OS X laptop, I ran into a static assert (build error) with your patch.

Please change ```return std::nextafter<Dtype>(...``` to ``` return std::nextafter(...```

std::nextafter is not a templated function.

flx42:
@bhack My python install is busted so it doesn't work, but for different reasons :)

bhack:
@vimalthilak Nice catch. Done. But seems that a ReductionLayerTest fail. I've not touched ReductionLayerTest that was merged to master after this PR

vimalthilak:
@bhack How do I run the ReductionlayerTest? Seems like that got added last November. Can you specify the steps needed to repro the problem on my side? Thanks.

bhack:
@vimalthilak Was merge on 3 June in master. You can run that test with:
```.build_release/test/test_all.testbin --gtest_filter='*Reduction*'```

**Comment by bhack** [comment]
@sh1r0 Are you interested to test this with you NDK tests?

**Comment by vimalthilak** [comment]
@bhack: Just to clear, did you rebase your branch in order to run the
ReductionLayer test? I will give this a try once I have some breathing
space at work.

**Comment by bhack** [comment]
@vimalthilak Yes I've rebased to test it. I can push to this branch tomorrow.

**Comment by bhack** [comment]
@vimalthilak It is rebased now.

**Comment by sh1r0** [comment]
@bhack Thanks. I removed some boost dependencies (https://github.com/sh1r0/caffe-android-lib/commit/c021b34f34bf74f8e29cbd7f5358decd22833615), and it seems to be fine with some naive tests.

**Comment by bhack** [comment]
@sh1r0 Do you compile also the cuda part? Some android tablet like Nvidia shield have cuda support.

**Comment by sh1r0** [comment]
@bhack Nope. I hope I can but it's a shame that I don't have such a device to test.

**Comment by bhack** [comment]
@sh1r0  I have one. I can make some test in the next days. Can you reabase your submodule?
I think that with this and https://github.com/BVLC/caffe/pull/2523 you can really simplify you commit on caffe submodules.

**Comment by sh1r0** [comment]
ReductionLayer tests failed.

**Comment by bhack** [comment]
@sh1r0 I know was merged some days after this PR. But I don't know why it is failing. Do you catch something?

**Comment by vimalthilak** [comment]
@bhack, @sh1r0 I understand why the ReductionLayer test fails. It's got to do with switching void caffe_rng_uniform(const int n, const Dtype a, const Dtype b, Dtype* r) to use std::uniform_real_distribution< instead of boost:uniform_real<>

If you switch back to using boost for Uniform Filler then you will notice that the tests will pass. 

I don't have the time to complete debugging and writing up a solution today but if one of you wants to take it forward I'm more than happy to provide any help you need to complete this patch.

**Comment by bhack** [comment]
I've tried to see the behavior with 64bit rng and now only one test is failing with a very little margin

```
src/caffe/test/test_reduction_layer.cpp:75: Failure

Value of: computed_result

  Actual: 98.398926

Expected: expected_result

Which is: 98.398964

Incorrect result computed with op SUMSQ, coeff 2.2999999523162842
```
Probably there is something different between std::mt19937 and boost::mt19937

**Comment by jeffdonahue** [comment]
You can change `EXPECT_FLOAT_EQ(a, b)` to `EXPECT_NEAR(a, b, 1e-5)` at `src/caffe/test/test_reduction_layer.cpp:75`.  I can review/merge a separate PR that does that, if anyone wants to create one.

**Comment by vimalthilak** [comment]
@jeffdonahue 

The failure with "std::ring" occurs for the following input:
> 0.00332564, 2 ,0.665283

The failure occurs because the element-wise gradient comparison operation fails for the above iput
The problem is that the estimated gradient is way off compared to the computed gradient. I noticed that with boost's ring, the input data doesn't have any values that "close" to zero. I haven't had a chance to dig in more but does the above allow you to gain any insight into what might be causing the error/failure? Is the step-size for CheckGradient too large?

**Comment by jeffdonahue** [comment]
@vimalthilak at least in the [Travis build](https://travis-ci.org/BVLC/caffe/builds/66957842) of @bhack's latest commit on this PR (b86afe3), there was only the single test failure as @bhack described.

**Comment by bhack** [comment]
@jeffdonahue I've fixed the tollerace here.
Now alla cpp tests pass again.
We still have heare a problem on python. Any hint?

**Comment by bhack** [comment]
Ok actually the python issue is: https://svn.boost.org/trac/boost/ticket/6545

**Comment by bhack** [comment]
@longjon I think that you are the main developer of the boost::python wrapper. I'm not practical with boost::python because I've generally wrapped rarely and only with Cython. I think that we need to register a new converter that can support std::shared_ptr as boost ticket [6545](https://svn.boost.org/trac/boost/ticket/6545). Do you have any hint? Can you contribute? 
```
/test/test_python_layer.py", line 15, in reshape
    top[0].reshape(*bottom[0].data.shape)
TypeError: No registered converter was able to produce a C++ rvalue of type std::shared_ptr<caffe::Blob<float> > from this Python object of type Blob
```

stefanseefeld:
Let me try to have a look...

bhack:
@stefanseefeld Thanks.

bhack:
@stefanseefeld Any news on boost side? @longjon?

haosdent:
+1

bhack:
@thatguymike Do you still has interest in this?

bhack:
@stefanseefeld Are there any progress on https://svn.boost.org/trac/boost/ticket/6545? Is there any workaround that we could adopt here?

stefanseefeld:
No workarounds that I'm aware of. Someone simply needs to work on a patch to add support for std::shared_ptr and submit that. (THat could be me, if no-one beats me to it.)

bhack:
@stefanseefeld It seem still assigned to @jhasse but I don't know if he is still active on this. The ticket was 4 years old.

jhasse:
Why do you think it's assigned to me? Never have been working on it (sorry), but I'm interested as well ;)

bhack:
@jhasse Sorry I've taken the wrong field. You are in cc. Seems assigned to @TallJimbo

TallJimbo:
... and I've been trying to find someone else to assign it to for years now.  I just don't have any time to work on Boost.Python.

stefanseefeld:
I'll try to look into it. Don't hold your breath, though.

bhack:
Referencing https://github.com/boostorg/python/issues/29

bhack:
Caffe recently merged more PRs that rely on BOOST instead of using c++11 counterparts. The absence of comments by core devs here and new boost code integrated let me tough that there is no interest to switch and remove boost. If nobody is interested anymore I will close this in few days.

bhack:
For python bindings we could evaluate https://github.com/wjakob/pybind11/blob/master/README.md

PR #2538, opened 2015-06-02
Added support for building with Visual Studio

willyd:
This is a first attempt to add support for Visual Studio via the CMake build. 

Only static libraries are supported now, but it should not be too hard to add support for shared libraries. Since Visual Studio has no direct equivalent of the --whole-archive option I have included a post build script to generate a header that forces symbol inclusion in caffe.bin and other executable targets.

Dependencies can be hard to build under Windows. To make the dependency compilation easier I provide a super-build project where almost all Caffe dependencies can be built with Visual Studio and a CMake cache file is produced so that Caffe configuration is lot less tedious on Windows (see [caffe-windows-dependencies](https://github.com/willyd/caffe-windows-dependencies)). 

I only tested theses changes with Visual Studio 2013 64 bit and CMake 3.2.


eelstork:
C.I. issues with this look easy to fix; one of the configs passes tests.

willyd:
@eelstork 
Yes I noticed the CI issues. I will resubmit with a fix for them soon.

Nerei:
That's cool idea to generate reference header automatically! ;-)

Nerei:
In my opinion, it's good idea to make user explicitly enable external projects, i.e. make it off by default and use have to manually set to on in order to run automatic download  compilation of glog/gflags. Otherwise it might be a surprise for many users who integrate caffe into own large projects.

cmake/Dependencies.cmake
```diff
@@ -10,21 +18,30 @@ list(APPEND Caffe_LINKER_LIBS ${Boost_LIBRARIES})
...
 # ---[ Google-gflags
 include("cmake/External/gflags.cmake")
 include_directories(SYSTEM ${GFLAGS_INCLUDE_DIRS})
 list(APPEND Caffe_LINKER_LIBS ${GFLAGS_LIBRARIES})
 
+# ---[ Google-glog
+if(MSVC)
+    add_definitions(-DGOOGLE_GLOG_DLL_DECL=)
+endif()
+include("cmake/External/glog.cmake")
```
Nerei: I think this should be done only if user explicitly set special option for this.

willyd: Agreed. Some settings of my dependencies made it over in the Caffe CMake files. I will submit again with these settings set through a cache file instead.

ChrislS:
Good approach! So finally there will even be more than one build system work on Windows! If you stay within CMake already, did you have a look at [biicode](http://www.biicode.com)? It should do the dependency management automatically, and already offer quite some of the dependencies either as blocks or precompiled (including boost) in a well-structured way.

willyd:
@ChrislS I stumbled upon biicode some time ago but they were missing binary blocks at that time. Maybe it is worth to reconsider using this now that they support it (although it seems a bit like a hack: http://blog.biicode.com/upload-to-biicode-precompiled-binaries-sfml/).

willyd:
Closing this one in favor of my second attempt (PR  #2816).

Issue #1550, opened 2014-12-10
PR: add different resize and noise-adding policies to input data


ducha-aiki:
Is anyone interested in PR, which allows to flexible set different resize (warp, aspect-keeping, different padding, random rotation etc.) and noise policies each with given probability? 

The prototxt looks like following

```
transform_param {
    crop_size: 227
    resize_param {
      prob: 0.5
      resize_mode: WARP
      height: 256
      width: 256
      interp_mode: LINEAR
    }
    resize_param {
      prob: 0.25
      resize_mode: BY_LARGE_SIZE
      height: 256
      width: 256
      interp_mode: AREA
    }
      resize_param {
      prob: 0.2
      resize_mode: BY_SMALL_SIZE_AND_PAD
      height: 256
      width: 256
      pad_mode: MIRRORED
      interp_mode: NEAREST
    }
      resize_param {
      prob: 0.05
      resize_mode: BY_SMALL_SIZE_AND_PAD
      height: 256
      width: 256
      pad_mode: CONSTANT
      pad_value: 104
      pad_value: 117
      pad_value: 124
      interp_mode: NEAREST
    }
    noise_param {
    prob: 0.6
    }
    noise_param {
    hist_eq: true
    prob: 0.2
    }  
    noise_param {
    decolorize: true
    gauss_blur: true
    prob: 0.2
    }  
```

RalphMao:
Thanks!

ducha-aiki:
Done in https://github.com/ducha-aiki/caffe/tree/more-transforms (works only for cv::Mat version of transform)

I plan to clean it up and enhance after #1900 would be merged.

RalphMao:
Can you please offer a prototxt example? The original prototxtx meets the error below:
Error parsing text-format caffe.NetParameter: 12:19: Message type "caffe.LayerParameter" has no field named "transform_param".

ducha-aiki:
@RalphMao are you using my branch or caffe-master? And it is not net parameter, you just need to replace default transform parameter in caffenet proto.

RalphMao:
@ducha-aiki I'm using your branch, and I am still not sure which one should I modify.
Here is part of my prototxt file, could you help point out where is wrong?
layers {
  name: "data"
  type: DATA
  top: "data"
  top: "label"
  data_param{
     source: "examples/imagenet/ilsvrc12_train_lmdb"
    backend: LMDB
    batch_size: 20
  }
  transform_param {
    crop_size: 224
[...]

PR #2539, opened 2015-06-02
Multiple scale training at the mini-batch level via random resizing

waldol1:
While traditional network architectures expect a fixed input size, not all of them do.  For example, FCNs or networks with spatial pyramid or global pooling can handle a wide range of image sizes.  This PR attempts to add that functionality via additional transform parameters to specify the upper and lower bounds on the image resizing dimensions, and whether the dimensions should be square.  For each mini-batch, image dimensions are chosen at random from this range and resizing takes place *after* image cropping.  Note that this is similar to PR #1900, but it proposes to do resizing before cropping to a fixed input size.  Perhaps both could be controlled by separate parameters.

I've only made some simple changes, and only for DataLayer reading in encoded images because those use a cv::img intermediary which is easily resized.  When the image data is not encoded, it is transformed directly from the Datum bytes to the Blob.  That would have to be changed which if done unconditionally would unnecessarily slow down the input pipeline.

Given that PRs #1900 and #1550 (and possibly others) are also adding similar features, how would it be best to proceed?


waldol1:
I started using this fork for personal use, so its no longer in a condition to merge to the main repo.

PR #2540, opened 2015-06-02
ImageData layer image resize that preserves aspect ratio

samarth-robo:
This adds a parameter `new_dim` in the ImageData layer parameters that allows the layer to resize the image without destroying the aspect ratio.

The image is resized such that `smallest side = new_dim`.


ronghanghu:
I just skimmed through your code. It seems that in your implementation the images inside a batch will have different sizes. So how do you decide the top blob size (for each batch)?

Issue #2543, opened 2015-06-03
LogLayer instantiated twice when compiling using GPU


ShaggO:
I've just pulled the newest changes this morning, at it seems the log_layer.cpp is generating compiler errors.
I'm on an OS X 10.10 machine using clang for compiling and using GPU.

The error I'm getting when running make all is:
"src/caffe/layers/log_layer.cpp:133:19: error: duplicate explicit instantiation of 'Forward_gpu'

INSTANTIATE_CLASS(LogLayer);

./include/caffe/common.hpp:39:18: note: expanded from macro 'INSTANTIATE_CLASS'
  template class classname<float>; \
                 ^
src/caffe/layers/log_layer.cpp:129:1: note: previous explicit instantiation is here
INSTANTIATE_LAYER_GPU_FUNCS(LogLayer);
^
./include/caffe/common.hpp:61:3: note: expanded from macro 'INSTANTIATE_LAYER_GPU_FUNCS'
  INSTANTIATE_LAYER_GPU_FORWARD(classname); \
  ^
./include/caffe/common.hpp:43:35: note: expanded from macro 'INSTANTIATE_LAYER_GPU_FORWARD'
  template void classname<float>::Forward_gpu( \"

The error is repeated and is likewise given for 'Backward_gpu'

It seems the 'INSTANTIATE_CLASS' macro uses the 'INSTANTIATE_LAYER_GPU_FUNCS', which is likewise used in line 129 of log_layer.cpp, and thus the LogLayer is instantiated twice when compiling for GPU. Everything seems to work when out-commenting line 129 though (make all, make test, make runtest). Since I'm new to the development of caffe, I'm not 100% sure this is correct and thus I've refrained from simply creating a pull request. Should the GPU functions perhaps be moved to a new log_layer.cu file?

Best,
Malte Stær Nissen

shelhamer:
Thanks for the report -- this is fixed in master@72d7089.

shelhamer:
> Should the GPU functions perhaps be moved to a new log_layer.cu file?

Yeah, these should only be instantiated in the `.cu`. Thanks for offering to PR a fix!

ShaggO:
PR for moving the gpu-functionality to log_layer.cu should be ready for merging

PR #2551, opened 2015-06-04
LogLayer gpu functionality moved to .cu file

ShaggO:
Additional "fix" of #2543
Splitting of log_layer.cpp by moving Forward_gpu and Backward_gpu into the new log_layer.cu-file.


shelhamer:
Oh, thanks for the organization. Not sure how I missed it was all in one `.cpp` before.

PR #2552, opened 2015-06-04
Triplet training data generation and layer modification

Wangyida:
This triplet sample based on MNIST database is modified from siamese network, The LMDB data is generated correctly for the triplet training. But the Loss layer used in siamese is ContrastiveLoss, such function allows 2 input feature as a pair output feature.
I need to modify the loss layer in Caffe to further suit for triplet network training.


cbalint13:
@Wangyida ,

* Allow me one question: Does this implementation intention to use the "triplet" loss function like in [1] (see "triplet cost function") using a single path of CNN and no weight shared siamese (dual) network path ? I am sure there are other papers on the "triplet" schema out there, i just cited one of them.

* You mentioned that still WIP on loss-layer, but also there is need to modify the batch set construction too, as the paper [1] say it is trivial to "arrange" data in early batches to get the network start learning.

* I would be very interested to train such way it e.g. on Brown dataset [2].


[1] http://lrs.icg.tugraz.at/pubs/wohlhart_cvpr15.pdf
[2] http://www.cs.ubc.ca/~mbrown/patchdata/patchdata.html

With Respect,
~cristian.

Wangyida:
@cbalint13 
I will write a "triplet" function mainly reference to [1] in the end.

It seems that you are also interested on 3D information jointed pose estimation and classification using CNN, as deicribed in [1], I‘ｍ going to combine triplet loss and pair-wise loss a last, it is indeed also necessary to arrange data properly as in paper[1]. I will modify the data generation code after the loss code is finished. It will be great for you to shows some triplet loss layer code for further improvement if there are any works~

bhack:
I think that @pwohlhart that is one of the authors of the paper wil also release a theano version soon.

Wangyida:
This PR is integrated to #2603

PR #2553, opened 2015-06-04
Add gpu device id option in python command lines

kibum14:
Add gpu device id option in python command lines.

In python classification and detection examples, one can select a specific gpu to run a model if there are more than 1 gpu.

### DIFF
--- a/python/classify.py
+++ b/python/classify.py
@@ -42,7 +42,7 @@ def main(argv):
     )
     parser.add_argument(
         "--gpu",
-        action='store_true',
+        default='0',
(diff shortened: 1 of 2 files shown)


cdoersch:
A minor issue is that the help message now no longer really describes what the --gpu flag does; it should say in there that you can specify the device id.

More importantly, as I understand it, the 'default' in argparse applies even when the --gpu flag is not present.  Hence it seems like this change would make it impossible to run in cpu mode.

shelhamer:
Closing as these Python scripts are soon to be deprecated.

PR #2572, opened 2015-06-08
Optimize inner product layer for special case M == 1

flx42:
Using gemm with one of the operand being a vector or a scalar is
suboptimal. Functions gemv and axpy should be prefered in these cases.

This optimization provided a 10-15% speedup for inference on a GeForce GTX 980.
The speedup should be larger on bandwidth-bound devices.

### DIFF
--- a/src/caffe/layers/inner_product_layer.cu
+++ b/src/caffe/layers/inner_product_layer.cu
@@ -15,12 +15,19 @@ void InnerProductLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
   const Dtype* bottom_data = bottom[0]->gpu_data();
   Dtype* top_data = top[0]->mutable_gpu_data();
   const Dtype* weight = this->blobs_[0]->gpu_data();
-  caffe_gpu_gemm<Dtype>(CblasNoTrans, CblasTrans, M_, N_, K_, (Dtype)1.,
-      bottom_data, weight, (Dtype)0., top_data);
-  if (bias_term_) {
-    caffe_gpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, M_, N_, 1, (Dtype)1.,
-        bias_multiplier_.gpu_data(),
-        this->blobs_[1]->gpu_data(), (Dtype)1., top_data);
+  if (M_ == 1) {
+    caffe_gpu_gemv<Dtype>(CblasNoTrans, N_, K_, (Dtype)1.,
+                         weight, bottom_data, (Dtype)0., top_data);
+    if (bias_term_)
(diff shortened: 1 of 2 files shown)


seanbell:
It would be a good idea to add a test case for `M == 1` in `src/caffe/test/test_inner_product_layer.cpp` to go along with your proposed change.

flx42:
@seanbell I did, thanks for the review.

If you have a better idea to avoid duplicating code for the Forward test, I'm all for it.

src/caffe/layers/inner_product_layer.cu
```diff
@@ -15,12 +15,19 @@ void InnerProductLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
...
-      bottom_data, weight, (Dtype)0., top_data);
-  if (bias_term_) {
-    caffe_gpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, M_, N_, 1, (Dtype)1.,
-        bias_multiplier_.gpu_data(),
-        this->blobs_[1]->gpu_data(), (Dtype)1., top_data);
+  if (M_ == 1) {
+    caffe_gpu_gemv<Dtype>(CblasNoTrans, N_, K_, (Dtype)1.,
+                         weight, bottom_data, (Dtype)0., top_data);
+    if (bias_term_)
+      caffe_gpu_axpy<Dtype>(N_, *bias_multiplier_.cpu_data(),
```
shelhamer: This should be `bias_multiplier_.gpu_data()`.

flx42: I'm dereferencing the pointer, I don't think this is going to end well on the host :)

seanbell: `bias_multiplier_.cpu_data()[0]` might be more readable and compile to the same result.

Also, `bias_multiplier_.cpu_data()` is an array of all 1s, so you could replace it with just `Dtype(1)`.  I suppose if anyone ever wants to do something more clever with `bias_multiplier_`, then they will need to also update this case.

flx42: I don't really think it's really more readable with `bias_multiplier_.cpu_data()[0]`, especially since there is only one element.

Also, for symmetry with the general case, I prefer using `bias_multiplier_` instead of `Dtype(1)`.

ronghanghu: agree with @seanbell that `bias_multiplier_.cpu_data()[0]` is more readable, even if there is only one element.

flx42:
Friendly ping. Reviewers, do you have other comments on this PR?

jeffdonahue:
LGTM.  Thanks for this optimization @flx42! (FWIW I like `*bias_multiplier_.cpu_data()` too.)

PR #2576, opened 2015-06-09
Extend Neuron layers to handle input Blob vectors.

BlGene:
At the moment all Neuron Layers ( ReLU etc. ) only process the first Blob vector entry. It would be good to extend the Layers to handle bigger input vectors, like the convolution layer does.

The pull request is incomplete, I only changed the ReLU code as an example, in case this is a desired change please drop a note and I will do the other layers + test too.

BR. Max


shelhamer:
Closing as the need for multiple tops and bottoms as done in the convolution layer is dulled by the rise of net spec. Thanks for offering to further adapt the activation layers though.

PR #2577, opened 2015-06-09
Deduplicate the CPU and GPU versions of Forward and Backward methods

futurely:
To avoid the complicated state management issue in #610, this PR doesn't use the global Caffe mode or layerwise device. Instead, the Caffe::Brew must always be specified explicitly while the original APIs call the Brew aware methods. The design is reflected in the AbsValLayer.


shelhamer:
Closing as device abstraction never took hold.

PR #2579, opened 2015-06-09
Small platform-specific bugfix for draw.py

### LINKED ISSUE #2376: bad escape in draw.py
Per [this Stack Overflow thread](https://stackoverflow.com/questions/29055715/drawing-network-in-caffe-causes-pydot-to-throw-end-of-line-errors), there is probably an error in [line 58 of draw.py](https://github.com/BVLC/caffe/blob/master/python/caffe/draw.py#L58). On my system, replacing `'\n'` with `r'\n'`, as that thread suggests, fixed the crash. I'm not sure if this is a configuration problem or indeed a bug.

My config is:

    $ uname -r
    3.18.11-gentoo
    $ python --version
    Python 2.7.9
    $ eix graphviz | grep -i installed
         Installed versions:  2.26.3-r4^t(02:54:36 21/04/15)(cairo gtk nls python -doc -examples -java -lasi -perl -ruby -static-libs -tcl ELIBC="-FreeBSD")

[Discussion at NVIDIA/DIGITS](https://github.com/NVIDIA/DIGITS/issues/90)

lukeyeager:
Close #2376

On Gentoo and CentOS (and others?), you get this error:

    Warning: /tmp/tmpjqPQBC:6: string ran past end of line

This issue was reported by @Pastafarianist here: NVIDIA/DIGITS#90 and then here: #2376, and by @Chamunsu here: https://github.com/NVIDIA/DIGITS/issues/112#issuecomment-103149490. Solution suggested by @bahaugen [here on SO](http://stackoverflow.com/a/29173396/2404152).

### DIFF
--- a/python/caffe/draw.py
+++ b/python/caffe/draw.py
@@ -72,7 +72,7 @@ def get_layer_label(layer, rankdir):
     else:
         # If graph orientation is horizontal, vertical space is free and
         # horizontal space is not; separate words with newlines
-        separator = '\n'
+        separator = '\\n'
 
     if layer.type == 'Convolution':
         # Outer double quotes needed or else colon characters don't parse


lukeyeager:
This was also necessary to pass a test on [this TravisCI build](https://travis-ci.org/NVIDIA/DIGITS/builds/66290017#L3504), meaning the fix is needed for Ubuntu 12 as well.

BlGene:
I can confirm that this is needed and fixes the problem.

BR, Ty

shelhamer:
Thanks for the fix @lukeyeager and confirmation @BlGene.

PR #2580, opened 2015-06-09
One of n layer

cxy7452:
In light of the one_of_n encoding being a popular way to represent integer labeled data (i.e. characters for text data), we have written a common layer that takes one-channeled data with integer values and expands into n-channeled one-of-n encoding vectors.

The protobuf is updated with a new type called OneOfN, that takes a single optional parameter called option_n (default = 2) that specifies the number of distinct possible values of the input data, which will be expanded into one-of-n vectors. 

contributing authors:
Nathan Hurst  @njhurst 
Chen-Ping Yu @cpu90

### DIFF
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -181,69 +181,6 @@ class EltwiseLayer : public Layer<Dtype> {
 };
 
 /**
- * @brief Takes two+ Blobs, interprets last Blob as a selector and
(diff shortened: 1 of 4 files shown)


shaibagon:
Closing:
1. Seems redundant: [`"Embed"`](http://caffe.help/manual/layers/embed.html) layer generalizes this functionality.
2. Too old PR: too much effort to make the changes into new file structure of different headers for layers etc.

PR #2583, opened 2015-06-10
Change log levels in upgrade_proto

lukeyeager:
See discussion here: https://github.com/NVIDIA/DIGITS/issues/140

If you finetune on an out-of-date model and the upgrade is successful, then there shouldn't be any `ERROR` messages in the logs.

### DIFF
--- a/src/caffe/util/upgrade_proto.cpp
+++ b/src/caffe/util/upgrade_proto.cpp
@@ -588,8 +588,8 @@ bool UpgradeNetAsNeeded(const string& param_file, NetParameter* param) {
   if (NetNeedsV0ToV1Upgrade(*param)) {
     // NetParameter was specified using the old style (V0LayerParameter); try to
     // upgrade it.
-    LOG(ERROR) << "Attempting to upgrade input file specified using deprecated "
-               << "V0LayerParameter: " << param_file;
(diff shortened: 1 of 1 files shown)


lukeyeager:
Any update on this? I haven't run into any other problems like this, so it seems that most of the Caffe log message levels are reasonable. Can we set this one section right?

Note that an error message still occurs if the upgrade fails.

src/caffe/util/upgrade_proto.cpp
```diff
@@ -588,7 +588,7 @@ bool UpgradeNetAsNeeded(const string& param_file, NetParameter* param) {
   if (NetNeedsV0ToV1Upgrade(*param)) {
     // NetParameter was specified using the old style (V0LayerParameter); try to
     // upgrade it.
-    LOG(ERROR) << "Attempting to upgrade input file specified using deprecated "
+    LOG(INFO) << "Attempting to upgrade input file specified using deprecated "
                << "V0LayerParameter: " << param_file;
```
jeffdonahue: please adjust alignment here and at 609+614

jeffdonahue:
see alignment nitpick, otherwise LGTM

lukeyeager:
No problem. Fixed alignment and rebased to master.

jeffdonahue:
Thanks @lukeyeager!

PR #2592, opened 2015-06-12
Implemented SpatialDropoutLayer

amuetzel:
This layer implements the SpatialDropout proposed by Tompson et al [1] to improve training for object localization tasks. SpatialDropout drops out whole channels instead of working on single pixels and is reported to improve network performance when the activations of a network are spatially coherent, i.e. similar features are activated in close locations.

Please note that I haven't tested its performance in a real world application, but the tests pass and it does what it is supposed to do.

[1]http://www.cims.nyu.edu/~tompson/others/tompson_cvpr15.pdf


bhack:
Please make lint

amuetzel:
Just noticed that, too. Fixed the lint errors, rebased and will push in a minute :-)

PR #2599, opened 2015-06-13
Expose NesterovSolver and AdaGradSolver in pycaffe.py and __init__.py…

dbolotin:
… as they they are already exposed in cpp binding file.


jeffdonahue:
Thanks for this earlier PR -- looks like this was done for all current solver types in #3024.

PR #2603, opened 2015-06-15
CNN Triplet Tranining 

Wangyida:
* Update loss layers 
* Prepare to add test
* LMDB data transfer added


sunbaigui:
mark

bhack:
Have you change log layer? I see loglayer diffs in commits.

Wangyida:
I just added a loss layer called triplet_loss_layer and defined a parameter of this layer called 'margin' in upgrade_proto, other changes are all obeyed on the rules on Caffe new layer contribute instruction. 
By the way, my whole code is working well now, an example on MNIST dataset with the format of LMDB is also included in examples/triplet folder. When I finished the GPU version of this triplet loss layer, all build will be passed. @bhack @mtamburrano

bhack:
I still see log layers in the diffs.

Wangyida:
Yes, I do see the diffs in log_layer, it's caused by the version updating from a newly released version, some files which I don't care in details changes.
The macro definition is always changing as new layer added and I must change some macro definition of triplet loss layer as the index being used by new accepted layer, how to deal with it as the triplet layer not merged in Caffe? @bhack

bhack:
You need to rebase regularly. See in particular the section ["When you are ready to develop a feature or fixing a bug, follow this protocol"] (http://caffe.berkeleyvision.org/development.html)

Wangyida:
@pwohlhart My code for an easier loss function which produce the similar result as described in paper[1] is done in mnist database, but as I have also rendered some images using OpenCV in LINE-MOD database, the tranining samples are not easily arranged, will your codes containing the data generation process?

@FlorisGaisser Your concern about the label could be solved in this paper~

http://lrs.icg.tugraz.at/pubs/wohlhart_cvpr15.pdf

bhack:
@shelhamer Can you make a first pass on this? This activity is part of opencv google summer of code 2015

src/caffe/layers/triplet_loss_layer.cpp
```diff
@@ -0,0 +1,191 @@
...
+    // Loss component calculated from d and e
+    dist_sq_par.mutable_cpu_data()[i] = caffe_cpu_dot(channels,
+        diff_par.cpu_data() + (i*channels), diff_par.cpu_data() + (i*channels));
+    // d e is a similar pair for pair wise
+    // loss accumulated by the pair wise part
+    loss += dist_sq_par.cpu_data()[i];
+  }
+  loss = loss / static_cast<Dtype>(bottom[0]->num()) / Dtype(2);
+  top[0]->mutable_cpu_data()[0] = loss;
+}
```
mtamburrano: I see you implemented the cost function as described in ["Learning Fine-grained Image Similarity with Deep Ranking"](http://arxiv.org/abs/1404.4661).
You could add a param to the loss layer to switch between this cost function and the one described in ["
Learning Descriptors for Object Recognition and 3D Pose Estimation"](http://arxiv.org/abs/1502.05908).
The function is similar but the authors report that theirs works better in special cases. So because the code could be very similar, could be worth implement also the newest cost  function

src/caffe/layers/triplet_loss_layer.cpp
```diff
@@ -0,0 +1,191 @@
...
+    loss += std::max(margin + dist_sq_.cpu_data()[i], Dtype(0.0));
+    // Pair wise loss accumulation
+    // Loss component calculated from d and e
+    dist_sq_par.mutable_cpu_data()[i] = caffe_cpu_dot(channels,
+        diff_par.cpu_data() + (i*channels), diff_par.cpu_data() + (i*channels));
+    // d e is a similar pair for pair wise
+    // loss accumulated by the pair wise part
+    loss += dist_sq_par.cpu_data()[i];
+  }
+  loss = loss / static_cast<Dtype>(bottom[0]->num()) / Dtype(2);
```
mtamburrano: why divide by 2?
It is because you sum triplet and pair costs? If it so, I don't think it is correct, the final loss should just be a composition of the two single losses

src/caffe/layers/triplet_loss_layer.cpp
Wangyida: OK， I will change it to the new one for triplet loss. The GPU version was modified for the triplet only loss rather than the triplet and pair-wise jointed loss. The GPU version will be rewrite later when the triplet is updated to the new one.

src/caffe/layers/triplet_loss_layer.cpp
Wangyida: Here I assume that the triplet and pair-wise part is like this form:margin + 0.5*|feat1 - feat2|^2 - 0.5*|feat1 - feat3|^2 and 0.5*|feat4 - feat5|^2, so dividing 2 in the loss is just a simplification for the gradient writing, it is not important in the loss form cause the aim is finding the gradient, I could also remove the '2' and add a factor '2' before the gradient. If I use the new loss function in "
Learning Descriptors for Object Recognition and 3D Pose Estimation", the factor '2' will only be used in the pair-wise loss.

src/caffe/layers/triplet_loss_layer.cpp
pwohlhart: Hi, actually the difference in the formulation is more of theoretic nature, at least for the case I ended up using it. In practice they both work nearly the same. I can think of situations where one is better suited than the other though, so probably an option to choose the loss formulation would be interesting.

bhack:
@Wangyida Please do always a "make lint" pass locally before pushing or Travis tests will fail.

src/caffe/layers/triplet_loss_layer.cpp
Wangyida: OK, I will carry out the new softer one as in your paper, I haven't done experiment on that one, I will have tests afterwards. @pwohlhart

examples/triplet/convert_3d_triplet_data.cpp
```diff
@@ -0,0 +1,163 @@
...
+  datum.set_height(rows);
+  datum.set_width(cols);
+  LOG(INFO) << "A total of " << num_items << " items.";
+  LOG(INFO) << "Rows: " << rows << " Cols: " << cols;
+  for (unsigned int itemid = 0; itemid < 10 * num_items; ++itemid) {\
+    int i = caffe::caffe_rng_rand() % num_items;  // pick triplet groups
+    int j = caffe::caffe_rng_rand() % num_items;
+    int k = caffe::caffe_rng_rand() % num_items;
+    int l = caffe::caffe_rng_rand() % num_items;  // pick pair wise groups
+    int m = caffe::caffe_rng_rand() % num_items;
```
mtamburrano: I don't think this is the best way to generate the triplets. In this way you can't be sure to use a good portion (if I'm not wrong, if you are unlucky enough, you could obtain always the same triplets and pairs) of the dataset.
Probably you should iteratively take an image, then build a random triplet and a pair, then choose the next image and so on.

examples/triplet/convert_3d_triplet_data.cpp
```diff
@@ -0,0 +1,163 @@
...
+    int ik_x = ik_diff_x*ik_diff_x;
+    int ik_y = ik_diff_y*ik_diff_y;
+    int ik_z = ik_diff_z*ik_diff_z;
+
+    int dist_ij = ij_x + ij_y + ij_z;
+    int dist_ik = ik_x + ik_y + ik_z;
+    if ((*label_i  == *label_j) && (*label_i  == *label_k))
+      triplet_class_same = true;
+    if ((dist_ij < dist_ik) && (triplet_class_same))
+      triplet_pose_pass = true;
```
mtamburrano: I'm not sure it is enough to check if the distance between the first image and the second is less than the distance between the first image and the third.
Maybe you should apply two different threshold to the distances when `triplet_class_same == true` and accept the triplet if `dist_ij < small_thresh` and `dist_ik > larger_thresh`.
The small threshold can be the distance between two points connected by an edge into the icosahedron (or maybe 2 points distant 2 edges apart), the larger threshold could be the distance between two points distant at least 6-7 edges apart. The issue could be to calculate the minimum edge's lenght because is iter_depth dependant.
@pwohlhart what do you think about it? These thresholds can help or it is enough that the pose between first and second image is more similar than first and third ones (in this way all the three poses can be very different)?

examples/triplet/convert_3d_triplet_data.cpp
```diff
@@ -0,0 +1,163 @@
...
+    if ((*label_i  == *label_j) && (*label_i  != *label_k))
+      triplet_class_pass = true;
+    if (*label_l == *label_m)
```
mtamburrano: Here the distance threshold, as in the previous comment, is mandatory.
The pair should consist of two images with very similar poses, so you should add this check

examples/triplet/convert_3d_triplet_data.cpp
```diff
@@ -0,0 +1,163 @@
...
+      pair_class_pass = true;
+    if ((triplet_class_pass || triplet_pose_pass) && pair_class_pass) {
+      datum.set_label(1);
+      datum.SerializeToString(&value);
+      snprintf(key, kMaxKeyLength, "%08d", itemid);
+      db->Put(leveldb::WriteOptions(), std::string(key), value);
+    } else {
+      itemid--;
+      datum.set_label(0);
```
mtamburrano: why to set the datum label to 0? It's not useless?

examples/triplet/convert_3d_triplet_data.cpp
pwohlhart: For pairs, it's always (training sample, its closest template).

For triplets we had a two stage approach: 
Initially, 
Type 1: i is a training sample, j is its most similar template, k is any other template from the same object
Type 2: i is a training sample, j is its most similar template, k is any other template from a different object

After a couple of epochs of training we collected additional triplets that most violate the constraint (ie have the highest cost).

We did this bootstrapping 3 times.

pwohlhart: Another remark: I think its beneficial to make sure the classes are evenly distributed over the minibatches. We did so by looping through the classes/instances while taking the training samples.

mtamburrano: thank you @pwohlhart, just another question:
sampling the icosahedron equally, if the distance is euclidean, we got 4 "most similar templates" for each training sample.
Do you simply choose randomly among them? Or maybe do you have some addittionally similarity metric than the euclidean distance to choose the most similar?

examples/triplet/convert_3d_triplet_data.cpp
```diff
@@ -0,0 +1,163 @@
...
+  signed char* label_k = new signed char[4];
+  signed char* label_l = new signed char[4];  // label for pair wise
+  signed char* label_m = new signed char[4];
+  char* pixels = new char[5 * rows * cols];
+  const int kMaxKeyLength = 10;
+  char key[kMaxKeyLength];
+  std::string value;
+
+  caffe::Datum datum;
+  datum.set_channels(5);  // one channel for each image in the triplet and pair
```
pwohlhart: (I'm sorry I did not yet have a look at the full code, so if I am wrong here, ignore me.)

Do I understand right that you plan to let each minibatch consist of samples, and each sample consists of a triplet and a pair? If that's the case, that's probably not the most efficient way, because descriptors for each sample get calculated for each pair and triplet they are part of, instead of just once per minibatch (as mentioned in Sec 3.3 of the paper).
The idea would be to assemble a minibatch of samples (training samples and templates) for which descriptors are calculated once and the several pairs and triplets can be constructed over this set, by just defining lists of triplets and pairs of indices within each minibatch.

I have not worked with caffe yet, but I assume it would need adjustments in the cost layer and I have no idea how to feed it the extra set of indices for triplets and pairs associated with each minibatch. Overall it should however be possible, no? Did you intentionally decide against that?

examples/triplet/convert_3d_triplet_data.cpp
pwohlhart: I actually measure the angular distance between the two viewing vectors. Makes more sense intuitively, but the relative ordering/ranking is the same with your choice of Euclidean distance.
Also, it is still the same for four "most similar templates", so I randomly choose one.

examples/triplet/convert_3d_triplet_data.cpp
pwohlhart: Actually the tripletIdx and pairIdx could just be blobs passed to the loss layer just like the labels, no?

bhack:
@shelhamer @jeffdonahue @longjon The GSOC firm pencil down it is 21 August. If you have any other feedback to collect we could let the student @Wangyida to fix that before the deadline. @pwohlhart is one of the original authors of the paper and It is giving here some precious hints.

examples/triplet/convert_3d_triplet_data.cpp
Wangyida: Already fixed. Now I choose the first sample in a triplet in sequence to cover all samples which different 1st sample leading to different triplet. I have also arranged the labels of 1st sample in triplet of adjacent triplet sets in different classes one by one, so the classes are evenly distributed over the minibatches.
So there I mainly arrange the 1st sample in each triplet to suit the demand for distribution and covering more data.

examples/triplet/convert_3d_triplet_data.cpp
Wangyida: I modified the codes and now it will use different distance as threshold to verify the similarity. As my sphere all have a radius of 100, I set the similar distance as less than 30 degree in view point which means a nearly 100 * sin(pi/60) = 50 distance on sphere and the dissimilar distance as more than 90 degree in view point which means a 100 * sqrt(2) = 141 on sphere.
I haven't have a count on the edges as you have described because the edges is not enough in a simple sphere and the edges are also not perfect as expected because of the method to eliminating duplex points generated in different iterations also eliminated some useful points, maybe it will be solved later.

examples/triplet/convert_3d_triplet_data.cpp
Wangyida: The pair distance is also put into consideration. Method is similar to the triplet one, I use distance rather than number of esges.

examples/triplet/convert_3d_triplet_data.cpp
Wangyida: The label of tripet training network is not useful now, Maybe the label will be useful in other data. The 3d object classification and pose estimation only need the data with label 1, it means the set of triplet is OK to be used.

examples/triplet/convert_3d_triplet_data.cpp
Wangyida: I had a try on Caffe as you and Manuele described and it could pass the index about class label and pose label for use in the triplet loss layer indeed, so I think it is also able to modify the codes as forwarding a set of samples and their labels of a batch, but I think the problem is that although the feature extarction process is more efficient than ever because some duplex samples will not be forwards many times, but there will be a time consuming in the triplet and pair regrouping process in the triplet and this grouping process could be set at first before we use Caffe for training. In the training process I set each batch used for accumulating loss including all classes and the poses also differs much, so in fact there are not many samples which is duplex with others.  So there I use another idea which Manuele informed that as we use 1 pair wise set and 3 triplet set as a basic unit to calculate a single loss and accumulate each single loss together to get the loss for a turn of forward passing process with a batch. My previous codes will use 11 blobs for this process(1 pair and 3 triplet), but in the new process there will be 5 blobs instead(1 reference sample, 1 with the closest pose, 1 with larger pose and 2 different class), so the pair and 3 triplet will be regrouped as reference+closest pose; reference+closest pose+different class1; .reference+closest pose+different class2; reference+closest pose+larger pose;  And I have also a question that is a loss could be represented as the sum of the loss from the 1 pair and the 3 triplet? or they should be calculated seperately? I think if I use them as a unit it will be both considering class and the pose at the same time. @pwohlhart 

And I find that if I use the 'number' channel to assemble all the samples(5 samples as a unit), the memory will be the same as I use the 'channel' channel to assemble all the samples and still use the slice layer to set them apart in the forward process and collect them again in the loss layer in training stage, I think they will show the same performence on memory, the codes till now is based on the channel seperation method on my codes. What do you think about it Manuele? @mtamburrano

mtamburrano: It's not memory related the choice to use the "num" or "channel" shape, the usage is the same with both method (with exception of the overhead due to the layers splitted) . But actually you need to slice each image before forwarding them and write a branch for each image into the prototxt. Leaving aside the extra computation time needed for slicing and successive concatenation before the loss layer, this approach is not scalable when you want for example to use more triplets, in this case you need to slice more images and add more branches into your net (one for each image).
Concatenating the images on the "num" channel lets you to forward them on the same branch of the net without the need of a slice and concat layer, and you if you need more triplets, you can just change a parameter of the loss_layer and increase the batch_size.
For further discussion on implementation details, we can continue on private channels

cbalint13:
@pwohlhart , @Wangyida ,

 Gentlemans,

 * I would ask if would be possible to train using triplets within a single cnn network Brown's 2d image patch [1] dataset ? Not sure, but i have the impression that work done here would work only (is tuned) on 3d dataset, Brown's one would be 2d and is related only to gray images.

 * I believe a reasonably small cnn trained with triplet-loss  could break world record on 2d image patch "descriptor" field, it could be learned to be very compact and even in binary hamming space.

[1] http://www.cs.ubc.ca/~mbrown/patchdata/patchdata.html

pwohlhart:
@cbalint13 
Hi,

1) I dont see why this should not be applicable to gray scale. After all the depth image is technically the same.

2) We are currently working on binarization (end-to-end training with binary output). We have something running that works approximately equally well in hamming space as the original float vectors in Euclidean space (with about the same number of bits). We will release this soon(ish).

bhack:
@pwohlhart Nice! Binarization was the original final step of this GSOC. But there is not enough time before deadline now.

cbalint13:
@pwohlhart , @Wangyida , @bhack,

Paul,

  Thank you for the information ! By a chance will try that, and for positive results will try post the model used. Also, I guess you are already aware of binarisation work in the filed, I would share such relevant paper [1]. Code is also available from author website [2], seems to be based on caffe. I believe results there are state-of-art but unsure.

[1] http://www.iis.sinica.edu.tw/~kevinlin311.tw/cvprw15.pdf
[2] http://kevinlintw.weebly.com

pwohlhart:
No, haven't seen that. There was a lot of work on binarization/hashing with CNNs in the main track of cvpr, I'm currently working through. Thanks for the pointer.

kevinlin311tw:
Hi all,
Thanks for your interest in our work. 
Yes. our work is based on caffe. We have recently made the caffe we used and source code available, you can download it and train your own model :)

cbalint13:
@Wangyida ,

* GPU training seems to not work. It fails like below:

I0808 19:29:22.932333 15456 net.cpp:685] Copying source layer feat_p2
I0808 19:29:22.932346 15456 net.cpp:685] Copying source layer loss
I0808 19:29:22.937747 15464 data_layer.cpp:118] Prefetch batch: 3 ms.
I0808 19:29:22.937789 15464 data_layer.cpp:119]      Read time: 1.307 ms.
I0808 19:29:22.937806 15464 data_layer.cpp:120] Transform time: 1.896 ms.
[New Thread 0x7fffdf28f700 (LWP 15461)]
[New Thread 0x7fffe0a71700 (LWP 15460)]

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7e15b09 in caffe::TripletLossLayer<float>::Forward_gpu (this=0x4cd1dc0, 
    bottom=std::vector of length 6, capacity 8 = {...}, top=std::vector of length 1, capacity 1 = {...})
    at /home/cbalint/rpmbuild/BUILD/caffe/src/caffe/layers/triplet_loss_layer.cu:101
101	    dist_sq_.mutable_gpu_data()[i] = dist_sq_pos.gpu_data()[i];

Wangyida:
The codes is being modified to suit for the slice layer eliminated version which is able to fill in more than 1 triplet in a loss, the GPU code is an old version which is not suit for the training data, I will rewrite it after two forms of the triplet loss layer is finished. @cbalint13

cbalint13:
@Wangyida ,

- Realized that last commit removed the mnist example. Would be nice to keep it as reference, linemod might be not accessible for users.
- Also not sure if work is finished yet as you mentioned regarding GPU version, i keep looking forward.

With Respect,
~cristian.

bhack:
@ronghanghu This is officially the last week of Google summer of code 2015. I hope that @Wangyida could continue to support this integration after the deadline but if you can do a first pass on this it is surely useful.

Wangyida:
OK， I remove the MNIST demo from triplet examples cause the label of MNIST is too simple, it is better using multi label for a training sample for data arrangement in triplet training, so we can fully use the information of training data in constructing a model, I will add it back any way only using the 10 labels. As for the GPU version, the codes are implemented again for the new using style of training data, but it is not tested, I will test later. @cbalint13

bhack:
The GSOC 2015 it is official over. Thank you @Wangyida and @mtamburrano for the efforts. A special thanks to @pwohlhart for some detailed and precious hints. This could be used in companion with https://github.com/Itseez/opencv_contrib/pull/276. I hope that we can collect from @BVLC some feedbacks or generalization requirements to let this to be merged.

Wangyida:
Now the GPU codes for triplet loss on GPU version  is fixed. It could be used producing the same result as CPU codes.
For the training data rendering process, you could use this demo for image generation from .ply files: https://github.com/Wangyida/opencv_contrib/tree/cnn_3dobj/modules/cnn_3dobj
@imbinwang
Linemode data set doesn't provide all .ply files for every model, some models are not included but you could still get the readered images.

shelhamer:
@Wangyida thanks for picking Caffe for your GSoC and happy hacking still even if the summer is over!
One step for review is to squash fixup commits and merges for an orderly history that can be reviewed commit by commit.

Wangyida:
@shelhamer I will squash these commits.

bhack:
@Wangyida You have reabsed now. But you need to squash. See http://programmers.stackexchange.com/questions/263164/why-squash-git-commits-for-pull-requests

Wangyida:
Alright, I have done some fixup and squash on the newly rebased codes for the latest few commits but not all because commits in the last few months differs a lot from codes shown as now. I will have a check on those commits. @bhack

Wangyida:
I have done 'fixup' and 'squash' on all my previous commits, since I have done rebasing several times before, there are some commit carried out by other contributors, I think these commits are neater for reviewing. @bhack

bhack:
@Wangyida I think you need to rabase your commit. All your commit need to be ahead of the last commit in caffe master branch. See [Shelhamer’s “life of a branch in four acts”](http://caffe.berkeleyvision.org/development.html). Please also squash some trivial commits the are still in the history.

Wangyida:
I have made some fixup and squash on my commits except for a few initial commits, I think this modification is more clear for a review, the codes are also rebased on the latest codes @bhack

bhack:
@Wangyida  Ok now the rebase seems fine but not the squash. @shelhamer I think that the history need to be more simply. There are still too many work in progress and fix commits that could be squashed.

Wangyida:
@bhack I have tried to quash and fixup early commits before commit 40d8aa11850d1d6dd8571e54f86a83a3d51e2bb8 which are not neat enough, they could be squashed properly but when I rebase again, the squashed commits will be restored. This might be caused be some inappropriate rebasing acts in early stage of developing this codes. On the other hand I have made it neat after commit 40d8aa11850d1d6dd8571e54f86a83a3d51e2bb8 and I think you could have a review just after this commit, this including all important modification and is well squashed and rebased. Is this able to suit the need for reviewing:)

bhack:
I've some doubt that this will be reviewed in this status. @shelhamer what do you think if @Wangyida close this after opening two new separated pull requests: one for solver + tests and the other one layer + test + utility both squashed in a single commit?

cbalint13:
@Wangyida , i believe you could open a new fresh PR, and reference in that PR this closed PR.

Wangyida:
@cbalint13 I think this is OK, I should have a new PR instead, there some bad operation in the first stage of coding on Github in the beginning.

bhack:
It is reopened as https://github.com/BVLC/caffe/pull/3103

bhack:
Now at https://github.com/BVLC/caffe/pull/3123

dzwallkilled:
Thanks for the great work. I met a problem with the installation. 
I succeed in $make all and $make test, but failed in $make runtest (The 4 tests of TripletLossLayerTest all failed). Could anyone help?

PR #2602, opened 2015-06-15
typo: "a fixed steps" to "at fixed steps"

ajschumacher:
I'm not sure if these are generated from some other source; just noticed the typo and found this first.

PR #2605, opened 2015-06-15
typo: "a fixed steps" to "at fixed steps"

ajschumacher:
fixing in the correct place as per @shelhamer's advice from #2602


jeffdonahue:
Thanks @ajschumacher!

PR #2608, opened 2015-06-16
Feature load image data sets

FlorisGaisser:
Changed the ImageDataLayer to have more than 1 image output.
By adding more 'top' params in the data layer in the .prototxt file while keeping the label last, more images can be loaded.
This can be used to load for example siamese image pairs from a textfile as created by 'create_imageset' in 'examples/siamese/'.

The test for the ImageDataLayer has been extended to include multiple inputs to test. Therefore 3 images have been added to the 'examples/images/' folder.

It has to be noted that since this required changes in the BasePrefetchingDataLayer and also other
classes are inherited from it, for now a new class: BasePrefetchingMulitDataLayer has been made.
After thorough testing, this code can be moved into the BasePrefetchingDataLayer class as it should be compatable.


FlorisGaisser:
I just noticed that this pull request also includes commits from unrelated work. I'll close this PR and make a new one.

PR #2609, opened 2015-06-16
Fix HTML display for docs

lukeyeager:
See the section "Shelhamer’s life of a branch in four acts” on this page:
http://caffe.berkeleyvision.org/development.html#shelhamershttpsgithubcomshelhamer-life-of-a-branch-in-four-acts
The git commands are not separated from each other.

Then, see the formatting of that section when rendered with Markdown (intended behavior):
https://github.com/BVLC/caffe/blob/0d7c6cb4b2286980c3d5b47cfddb2457202c93cf/docs/development.md#shelhamers-life-of-a-branch-in-four-acts

This little fix ought to fix the HTML generation without breaking the Markdown generation. I think it'll work based on the fact that [this section](https://github.com/BVLC/caffe/blob/0d7c6cb4b2286980c3d5b47cfddb2457202c93cf/docs/development.md#testing) displays correctly, but I don't actually know how the HTML generation works.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -62,28 +62,25 @@ The following is a poetic presentation of the protocol in code form.
 #### [Shelhamer's](https://github.com/shelhamer) “life of a branch in four acts”
 
 Make the `feature` branch off of the latest `bvlc/master`
-```
-git checkout master
-git pull upstream master
-git checkout -b feature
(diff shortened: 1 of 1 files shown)


shelhamer:
Thanks for the formatting fix!

PR #2611, opened 2015-06-17
fixed two compatibility bugs for prototext format

dpaiton:
The first bug was in InitUnsharedWeightsNet. The bias_term var
was passed as an int and was breaking earlier versions of protobuffer.
I wrote an inline conditional to convert the term into a string.
This allows backwards compatibility with earlier protobuf
versions (e.g. version 2.3.0 on Redhat was failing without this).

The second bug was in the syntax for repeated bool parameters,
assigned to the propagate_down parameter. The style used for e.g.
propagate_down: [true,true] does not work with earlier prototext
versions (failed with version 2.3.0 on Redhat). New syntax works
for all versions.


shelhamer:
Thanks for the backward compatibility!

@jeffdonahue I like the concise syntax of `[true, true]` but I suppose we should keep ourselves to the lowest common denominator protobuf for the tests.

PR #2612, opened 2015-06-17
register a dummy reducer to prevent mincepie runtime error

ih4cku:
mincepie need a default reducer, otherwise error would occur during runtime. In this script, a dummy `NoPassReducer` is sufficient.


shelhamer:
Thanks for catching this—that tool isn't used so frequently so it was missed.

PR #2613, opened 2015-06-17
Feature layer image data load set

FlorisGaisser:
Changed the ImageDataLayer to have more than 1 image output.
By adding more 'top' params in the data layer in the .prototxt file while keeping the label last, more images can be loaded.
This can be used to load for example siamese image pairs from a textfile as created by 'create_imageset' in 'examples/siamese/'.

The test for the ImageDataLayer has been extended to include multiple inputs to test. Therefore 3 images have been added to the 'examples/images/' folder.

It has to be noted that since this required changes in the BasePrefetchingDataLayer and also other
classes are inherited from it, for now a new class: BasePrefetchingMulitDataLayer has been made.
After thorough testing, this code can be moved into the BasePrefetchingDataLayer class as it should be compatable.

P.S. This replaces https://github.com/BVLC/caffe/pull/2608 which contained commits from unrelated work.


mtmarsh2:
Is this just for grayscale images with 1 channel or can it be used for 3 channel images?

FlorisGaisser:
I haven't tested it, but I think it should work with colour images too. 
Could you test it and message me again? If it's not working, I'll try to make it work with colour images from Tuesday, but atm I don't have excess to my pc.

mtmarsh2:
Hmm not sure if its related to the color channel, but I'm getting the following error when I run the "examples/siamese/create_siamese_from_imageset.sh" script. I have two folder: train_images and test_images, and each have the siamese_set.txt inside. I also checked the paths, and both files exist on disk

ERROR!	 one or both input images are empty:
1: examples/101/Motorbikes/image_0344.jpg is empty: 0
2: examples/101/Leopards/image_0178.jpg is empty: 0

FlorisGaisser:
This error means that loading of the image failed (there are different reasons to this). But this probably means the path is incorrect. I think you didn't define a root_folder param. Try defining this param from root: /path/to/caffe/ and probably it would work.

-------- Original message --------

nassarofficial:
I see that in the files changed that there are RGB images, is it working with RGB?

PR #2614, opened 2015-06-17
Reword help message of python training log plotting script for clarity.

shawnLeeZX:
Seems that original author just copy the help message from shell script version of plotting script. The minor rename makes it easier to learn to use this script.


jeffdonahue:
I'm not sure if `plot_log.sh` is the correct name of the script, but I am pretty sure the new version isn't correct as the name of the documentation file rather than a script filename.  Can reopen if I'm missing something (and new PRs are welcome if `plot_log.sh` is indeed wrong).

PR #2505, opened 2015-05-24
MatCaffe3

ronghanghu:
This PR creates a powerful Matlab interface for Caffe. It has been an issue for a long time in Caffe that Matlab interface is not as powerful as Python interface. #501 (followed by #1913) attempted to resolve this issue by adding more functions. However features in #1913 are still limited (e.g. not allowing creating multiple nets, not allowing training a net), and development of #1913 seems to have been stalled.

This Matlab interface of Caffe implements *more features* when compared with #1913, with *comparable or even less code*, by creating wrappers for `caffe::Solver`, `caffe::Net`, `caffe::Layer` and `caffe::Blob` in Matlab interface. In this interface, **almost everything that can be done in PyCaffe can also be done here** (except for a MATLAB_LAYER, which can be a future functionality).

~~This PR is add-only and non-invasive, which means the old Matlab interface is not removed in case someone still wants to use it. Note: the old Matlab wrapper doesn't work right now due to #1970.~~ This PR removes the old Matlab wrapper as suggested by @shelhamer , but keeps the HDF5 example and modified the image classification example using BVLC CaffeNet.

Since #1913 is already called MatCaffe2, I'll call this PR MatCaffe3 :)

## MATLAB

The MATLAB interface -- matcaffe -- is the `caffe` package in caffe/matlab in which you can integrate Caffe in your Matlab code.

In MatCaffe, you can

* Creating multiple Nets in Matlab
* Do forward and backward computation
* Access any layer within a network, and any parameter blob in a layer
* Get and set data or diff to any blob within a network, not restricting to input blobs or output blobs
* Save a network's parameters to file, and load parameters from file
* Reshape a blob and reshape a network
* Edit network parameter and do network surgery
* Create multiple Solvers in Matlab for training
* Resume training from solver snapshots
* Access train net and test nets in a solver
* Run for a certain number of iterations and give back control to Matlab
* Intermingle arbitrary Matlab code with gradient steps

An ILSVRC image classification demo is in caffe/matlab/demo/classification_demo.m (you need to download BVLC CaffeNet from [Model Zoo](http://caffe.berkeleyvision.org/model_zoo.html) to run it).

### Build MatCaffe

Build MatCaffe with `make all matcaffe`. After that, you may test it using `make mattest`.

[...]

PR #2629, opened 2015-06-20
improve matcaffe to support  input data of arbitrary size

qinhongwei:
This PR brings the arbitrary input size support back.
In the previous matcaffe, the input image can be of arbitrary size, while the current version does not support that. When dealing with fully convolutional networks forward for response map, arbitrary image size support is kind of necessary. Besides, the data size match check is not necessary in matcaffe because Caffe itself will do.

### DIFF
--- a/matlab/+caffe/private/caffe_.cpp
+++ b/matlab/+caffe/private/caffe_.cpp
@@ -56,8 +56,25 @@ enum WhichMemory { DATA, DIFF };
 // Copy matlab array to Blob data or diff
 static void mx_mat_to_blob(const mxArray* mx_mat, Blob<float>* blob,
     WhichMemory data_or_diff) {
-  mxCHECK(blob->count() == mxGetNumberOfElements(mx_mat),
-      "number of elements in target blob doesn't match that in input mxArray");
+  // Reshape blob dims to match matlab data if needed
+  if (mxGetNumberOfElements(mx_mat) != blob->count()) {
+    const mwSize* dims = mxGetDimensions(mx_mat);
+    const mwSize num_dims = mxGetNumberOfDimensions(mx_mat);
+    const int input_dims = blob->shape().size();
+    const int pad = input_dims - num_dims;
+    vector<int> shape(input_dims);
+    // In matlab, you cannot have trailing singleton dimensions
+    // So an input batch of size 1 (W x H x C x 1) will come in with 3 dims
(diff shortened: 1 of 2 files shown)


qinhongwei:
After double check, I found the current matcaffe #2505 can perform arbitrary size input at the cost that you reshape the net in matlab every time you forward.

e.g.
net.blobs('data').reshape([W H C N]);

So it's user's choice. I'll close this PR.

ronghanghu:
Personally I prefer the current way since it is consistent with pycaffe.

**Note: after calling `net.blobs('data').reshape([W H C N]);`, you also need to call `net.reshape()` to reshape all subsequent layers.**

> In the previous matcaffe, the input image can be of arbitrary size
I don't remember that. Maybe it is supported in a custom branch?

matlab/+caffe/private/caffe_.cpp
```diff
@@ -56,8 +56,25 @@ enum WhichMemory { DATA, DIFF };
...
+    // In matlab, you cannot have trailing singleton dimensions
+    // So an input batch of size 1 (W x H x C x 1) will come in with 3 dims
+    // like (W x H x C). Therefore, we need to pad with singleton dims.
+    for (int d = 0; d < pad; ++d) {
+      shape[d] = 1;
+    }
+    for (int d = pad; d < input_dims; ++d) {
+      shape[d] = dims[num_dims - (d - pad) - 1];
+    }
+    blob->Reshape(shape);
```
ronghanghu: It is incorrect here as one also needs to `net->Reshape()` to reshape subsequent layers;

qinhongwei:
When we want to get a response map of the input data, matcaffe send the input data to Caffe and receive output data from Caffe, so
`net.blobs('data').reshape([W H C N])` is enough. `net.reshape()` is not necessary.

So it is in `matlab/+caffe/private/caffe_.cpp `

> I don't remember that. Maybe it is supported in a custom branch?

nope, it's the BVLC master branch

ronghanghu:
> When we want to get a response map of the input data, matcaffe sends the input data to Caffe and receives output data from Caffe, so net.blobs('data').reshape([W H C N]) is enough. net.reshape() is not necessary.

Yes, you are right. I see that all layers get reshaped automatically when you call net.forward(), since in https://github.com/BVLC/caffe/blob/master/include/caffe/layer.hpp#L409 every layer reshape itself before running forward.

PR #2636, opened 2015-06-22
fixed _force_color check, fixes #2635

### LINKED ISSUE #2635: Check failed: !param_.force_color() && !param_.force_gray()
When specifying 
```
  transform_param {
    force_gray: true
  }
```

I get the following error:  `data_transformer.cpp:130] Check failed: !param_.force_color() && !param_.force_gray() cannot set both force_color and force_gray`

But I only set `force_gray` and not `force_color`. The logic check in data_transformer.cpp:130 is wrong:
`CHECK(!param_.force_color() && !param_.force_gray())`
This expression is only true if both are false.

berleon:
Fixes the logic error, described in issue #2635


sguada:
@berleon thanks for the catch

PR #2640, opened 2015-06-23
Update classify.py to force sorting on the input folder

beatlevic:
glob.glob does not guarantee that the list of files in the folder is returned in sorted order (depends on the OS).
Therefore force the files to be ordered (adding "sorted"), so you know the predictions are in sorted order.


shelhamer:

PR #2644, opened 2015-06-24
fix bug of ipython corruption

jingpengwu:
the filename of io.py is the same with one of ipython script in system and caused an issue of system ipython corruption. Simply rename it fix this issue. Another bug is the proto folder name is not right between the import location and the folder name.


seanbell:
The only reason why io.py would conflict with system libraries is if the caffe directory itself gets added to the python path.  Instead, the *parent* directory of the caffe directory should be added to the python path.  This is why you are getting conflicts.

jingpengwu:
Thanks for the explanation. That makes a lot of sense!

PR #2651, opened 2015-06-26
Make the two separate build systems clearer in the documentation

keir:


bhack:
Or probably remove Makefile. Two build system are quite impossible to be maintained in sync. Also Travis builds are not aligned between Make and Cmake.

keir:
I had some trouble getting Caffe built since the installation instructions were not easily scannable, hence this patch. Is the documentation out of date with practice (e.g. is CMake the new official system)? I suggest merging this as-is since it's a strict improvement. If the Make system is deprecated, then that is a separate change.

bhack:
Yes is it ok. But it is not really clear how this two build systems are in sync.

keir:
OK. So can you merge it? Or should I abandon this?

bhack:
@keir All core devs are MIA. We need to wait some celestial sign.

keir:
Ah, I didn't realize you were not a core dev :) They may be finishing off their ICCV reviews.

jeffdonahue:
Thanks for improving these docs @keir -- merged in 6d11108 with a typo fix.

PR #2663, opened 2015-06-30
Change to download script to download proper model

tobyhijzen:


shelhamer:
Closing since the current model is the one chosen for the demo (which is not identical to the full R-CNN!) and since the dev branch is deprecated. Please send PRs to master.

PR #2665, opened 2015-06-30
Check for invalid cursor in data_layer

aleph7:
If your solver has more iterations than you have data, training will silently fail. Added checks to verify that the cursor is valid in data_layer. This would have saved a lot of time.

### DIFF
--- a/src/caffe/layers/data_layer.cpp
+++ b/src/caffe/layers/data_layer.cpp
@@ -70,6 +70,7 @@ void DataLayer<Dtype>::InternalThreadEntry() {
   CPUTimer timer;
   CHECK(this->prefetch_data_.count());
   CHECK(this->transformed_data_.count());
+  CHECK(cursor_->valid());
 
   // Reshape according to the first datum of each batch
   // on single input batches allows for inputs of varying dimension.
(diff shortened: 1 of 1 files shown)


shelhamer:
No, the layer checks if it has reached the end of the data and loop around: https://github.com/BVLC/caffe/blob/master/src/caffe/layers/data_layer.cpp#L109-L114. This makes sure the cursor is always valid.

If not, please explain further.

aleph7:
For some reason I didn't have that, my mistake.

karthikeyanchenniappan:
Hi ,

1.I download source from https://github.com/aleph7/caffe/ and builded the caffe as static library.
2.Created sample demo code and Linked Caffe static lib(.a) and execute the code.
3.Now I got Run time error

F0519 14:54:12.494139 14504 layer_factory.hpp:77] Check failed: registry.count(t ype) == 1 (0 vs. 1) Unknown layer type: Convolution (known types: MemoryData)

4.I searched a lot and found one solution from below link

http://stackoverflow.com/questions/30325108/caffe-layer-creation-failure

5.If I create dynamic library instead of static library. It will work.

6.I tried to convert static library into Dynamic library.I got Error on Xcode that is cannot open the project I referred below link.

http://stackoverflow.com/questions/1349115/how-do-i-change-an-existing-xcode-target-from-dynamic-to-static

Can you help how to solve this..?

aleph7:
I ended up just including the source files directly in my project. If you still have questions please post on my fork: https://github.com/aleph7/caffe/issues

PR #2676, opened 2015-07-01
Update installation docs for boost - close #2454

### LINKED ISSUE #2454: Add --no-install-recommends libboost-all-dev to the apt-get install commands on Ubuntu
Helps avoid issues like: https://answers.launchpad.net/ubuntu/+question/261752

lukeyeager:
See discussion at #2454.

This avoids whacking a previous CUDA install.

### DIFF
--- a/docs/install_apt.md
+++ b/docs/install_apt.md
@@ -6,7 +6,8 @@ title: Installation: Ubuntu
 
 **General dependencies**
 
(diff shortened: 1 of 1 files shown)


shelhamer:
Ok, I've never run into this but thanks for the docs update to help those affected.

PR #1950, opened 2015-02-24
Easier installation of python requirements.

pgao:
Minor change, but this is the recommended way of installing requirements files.

https://pip.pypa.io/en/latest/user_guide.html#requirements-files

PR #2679, opened 2015-07-02
Update Python section in installation.md

mekza:
Better way to use pip https://pip.pypa.io/en/stable/user_guide.html#requirements-files


seanbell:
See #1950 for why this method is not being used.

stuaxo:
On #1950 it's suggested that moving h5py before Cython in the requirements makes this work, if this can be verified, then it makes sense to do this instead ?

shelhamer:
Closing according to #1950

PR #2693, opened 2015-07-04
Deprecated OpenCV consts leading to compilation error

AdamStelmaszczyk:
When compiling Caffe I got the same compilation error as here:
https://groups.google.com/forum/#!topic/caffe-users/mZDi02lgcj0

This is because constants such as CV_BGRA2GRAY are deprecated in OpenCV 2 and 3.
The replacements are constants such as cv::COLOR_BGRA2GRAY.

### DIFF
--- a/examples/cpp_classification/classification.cpp
+++ b/examples/cpp_classification/classification.cpp
@@ -186,13 +186,13 @@ void Classifier::Preprocess(const cv::Mat& img,
   /* Convert the input image to the input image format of the network. */
   cv::Mat sample;
   if (img.channels() == 3 && num_channels_ == 1)
-    cv::cvtColor(img, sample, CV_BGR2GRAY);
+    cv::cvtColor(img, sample, cv::COLOR_BGR2GRAY);
   else if (img.channels() == 4 && num_channels_ == 1)
-    cv::cvtColor(img, sample, CV_BGRA2GRAY);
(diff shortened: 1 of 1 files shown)


flx42:
LGTM

flx42:
... aside from the commit message where there is a typo in the title and you should remove the first paragraph.

AdamStelmaszczyk:
@flx42 I edited the commit message and title of this pull request.

flx42:
You should have kept the message about OpenCV versions

But actually, I installed OpenCV 3.0.0 and I couldn't repro the issue, which version of OpenCV are you using exactly?

AdamStelmaszczyk:
I used OpenCV 3.0.0. 

Wesley from earlier mentioned Google Group thread also used 3.0.0.

These consts come from this OpenCV file: https://github.com/Itseez/opencv/blob/master/modules/imgproc/include/opencv2/imgproc.hpp

Why you can't reproduce this issue: maybe you haven't correctly uninstalled previous version of OpenCV or haven't done `make clean`.

Not sure if this is relevant, but me and Wesley are building with `CPU_ONLY := 1` in `Makefile.config:8`.

flx42:
Sure, it's deprecated, but I don't understand why it's not compiling for you, the constants still exist:
https://github.com/Itseez/opencv/blob/master/modules/imgproc/include/opencv2/imgproc/types_c.h#L103-L128

Issue #2478, opened 2015-05-17
ImportError: dynamic module does not define init function (PyInit__caffe)


sendit2me:
I have, and am still trying to get caffe running under python (python3.4)

during the "import caffe" statement I get.
is there a make pycaffe3 for python 3 vs 2 init ? 
or am I missing a step.

Thanks
Peter

longjon:
Please note that Python 3 is not (yet) tested, and therefore we can't guarantee that it will work (though we try not to break it, and accept PRs that fix it). See also #2464.

In addition, please ask usage questions on the [caffe-users](https://groups.google.com/forum/#!forum/caffe-users) list. Thanks!

monajalal:
Is Caffe not still supported for Python3?

Coderx7:
@monajalal : You should be able to compile with python3 in linux. in windows it is not supported afaik. just change the make.config file to use python3.

Issue #2429, opened 2015-05-07
problem with pycaffe and boost


kayhan-batmanghelich:
Dear All,

I built caffe and all test pass. I need to use caffe with python (anaconda python). but unfortunately I  have a issue with python interface. 

After build I did this:
```
make  pycaffe
```

which seemed fine but then when I want to instantiate the caffe inside of the python I get this error. Would you test if you get the same error. Apparently it is related to this issue [1]:
```
In [1]: import caffe
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-1cca3aa1f8c5> in <module>()
----> 1 import caffe

/home/batmangh/caffe/python/caffe/__init__.py in <module>()
----> 1 from .pycaffe import Net, SGDSolver
      2 from ._caffe import set_mode_cpu, set_mode_gpu, set_device, Layer, get_solver
      3 from .proto.caffe_pb2 import TRAIN, TEST
      4 from .classifier import Classifier
      5 from .detector import Detector

/home/batmangh/caffe/python/caffe/pycaffe.py in <module>()
     11 import numpy as np
     12
---> 13 from ._caffe import Net, SGDSolver
     14 import caffe.io
     15

ImportError: /opt/boost/gnu/lib/libboost_python.so.1.55.0: undefined symbol: PyUnicodeUCS2_FromEncodedObject
```

Any idea?

Thanks,

longjon:
Something is up with your build. Please ask usage questions on the [caffe-users](https://groups.google.com/forum/#!forum/caffe-users) list. Thanks!

Issue #1098, opened 2014-09-17
Error while executing python files


crazymuse:
Traceback (most recent call last):
  File "classify.py", line 14, in <module>
    import caffe
  File "/home/jayasimha/caffe/python/caffe/**init**.py", line 1, in <module>
    from .pycaffe import Net, SGDSolver
  File "/home/jayasimha/caffe/python/caffe/pycaffe.py", line 10, in <module>
    from ._caffe import Net, SGDSolver
ImportError: /home/jayasimha/caffe/python/caffe/_caffe.so: undefined symbol: _ZN6gflags21ParseCommandLineFlagsEPiPPPcb

Error while trying to running draw_net.py,
problem is persistant for other files too . . . 

I do have _caffe.so in /caffe/python/caffe folder,
I have declared environment variable PYTHONPATH, with value the location of library
also tried make pycaffe, and make distribute, but seems like no help. I don't know how to debug the problem further.

Issue #938, opened 2014-08-16
Boost Python Problem on import caffe


nsauder:
Hello,

Thanks for the great library. I have trying to run filter_visualization.py however importing caffe module has been raising problems. Boost has been built, pycaffe compiled properly and all other examples are running smoothly. However, I am getting the following error related to boostpython:

ImportError: ../python/caffe/_caffe.so: undefined symbol: _ZN5boost6python6detail12gcc_demangleEPKc

Thanks
Nate

PR #2698, opened 2015-07-05
Add makefile config option for linking Python 3 libraries

cooperra:
Existing makefile always links to boost_python and python2.7, but this breaks Python 3 support which needs boost_python3 and python3.4m instead.

This PR enables overriding the default via Makefile.config and adds a commented suggestion in Makefile.config.example.

Fixed my `import caffe` issue:
`ImportError: /usr/lib/python3.4/site-packages/caffe/_caffe.so: undefined symbol: _ZN5boost6python6detail11init_moduleER11PyModuleDefPFvvE`

Looks like this happened because the Python/C API is different in Python 3, and the linked boost library was compiled for Python 2.

Might be related to #2478, #2429, #1098 (except with gflags instead of boost), #938. 

I'm using Arch Linux where python3 is default.

### DIFF
--- a/Makefile.config.example
+++ b/Makefile.config.example
@@ -68,6 +68,10 @@ PYTHON_LIB := /usr/lib
 # Uncomment to support layers written in Python (will link against Python libs)
 # WITH_PYTHON_LAYER := 1
 
+# Uncomment to use Python 3 (default is Python 2)
+# PYTHON_LIBRARIES := boost_python3 python3.4m
+# Also, you'll need to edit PYTHON_INCLUDE above.
+
 # Whatever else you find you need goes here.
 INCLUDE_DIRS := $(PYTHON_INCLUDE) /usr/local/include
 LIBRARY_DIRS := $(PYTHON_LIB) /usr/local/lib /usr/lib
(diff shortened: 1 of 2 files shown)


graphaelli:
+1

Makefile.config.example
```diff
@@ -68,6 +68,10 @@ PYTHON_LIB := /usr/lib
 # Uncomment to support layers written in Python (will link against Python libs)
 # WITH_PYTHON_LAYER := 1
 
+# Uncomment to use Python 3 (default is Python 2)
+# PYTHON_LIBRARIES := boost_python3 python3.4m
+# Also, you'll need to edit PYTHON_INCLUDE above.
+
```
philkr: One small suggestion: it might be worth moving these lines a bit up in the code, right below the anaconda example, so that we can also have a PYTHON_INCLUDE example and just comment out both PYTHON_LIBRARIES and PYTHON_INCLUDE if we switch to python3.

philkr:
Apart from the small comment above this looks good.

graphaelli:
rebased and squashed in: #3584

BlGene:
@cooperra @philkr 

It looks like this PR should be closed.

PR #2712, opened 2015-07-07
List protobuf-compiler dependency in the correct place (it is in the …

semitrivial:
…package managers for both 14.04 and 12.04)

### DIFF
--- a/docs/install_apt.md
+++ b/docs/install_apt.md
@@ -6,7 +6,7 @@ title: Installation: Ubuntu
 
 **General dependencies**
 
-    sudo apt-get install libprotobuf-dev libleveldb-dev libsnappy-dev libopencv-dev libhdf5-serial-dev
+    sudo apt-get install libprotobuf-dev libleveldb-dev libsnappy-dev libopencv-dev libhdf5-serial-dev protobuf-compiler
     sudo apt-get install --no-install-recommends libboost-all-dev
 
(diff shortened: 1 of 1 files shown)


lukeyeager:
[This comment](https://github.com/BVLC/caffe/blob/77d66dfc907dd875d69bb9fc12dd950b531e464f/cmake/ProtoBuf.cmake#L8-L9) in the CMake files leads me to believe that the `protobuf-compiler` package is already included in the `libprotobuf-dev` package for 12.04:
```cmake
# As of Ubuntu 14.04 protoc is no longer a part of libprotobuf-dev package
# and should be installed  separately as in: sudo apt-get install protobuf-compiler
```
Is it not?

semitrivial:
Doesn't appear to be [1], unless it's part of the basic GNU C/C++ libraries or zlib.  Just to make sure, I uninstalled protobuf-compiler, libprotobuf-dev and their dependencies from my 12.04 machine, re-installed libprotobuf-dev, and this did not cause protobuf-compiler to be re-installed.

[1] http://packages.ubuntu.com/precise/libprotobuf-dev

jeffdonahue:
Thanks for the addition! (I verified that the [Travis install script](https://github.com/BVLC/caffe/blob/master/scripts/travis/travis_install.sh#L18) does this as well.)

PR #2714, opened 2015-07-07
Making the net_spec python3 compatible

philkr:
Fixes a few python3 bugs caused by iteritem and input redirection.

### DIFF
--- a/examples/pycaffe/caffenet.py
+++ b/examples/pycaffe/caffenet.py
@@ -1,5 +1,6 @@
 from caffe import layers as L, params as P, to_proto
 from caffe.proto import caffe_pb2
+from __future__ import print_function
 
 # helper function for common structures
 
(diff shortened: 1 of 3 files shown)


examples/pycaffe/caffenet.py
```diff
@@ -45,10 +46,10 @@ def caffenet(lmdb, batch_size=256, include_acc=False):
 
 def make_net():
     with open('train.prototxt', 'w') as f:
-        print >>f, caffenet('/path/to/caffe-train-lmdb')
+        print( caffenet('/path/to/caffe-train-lmdb'), file=f )
```
longjon: Please follow PEP8 here and below: http://legacy.python.org/dev/peps/pep-0008/#pet-peeves.

longjon:
I guess this adds `six` as a dependency, which has been suggested before. That seems pretty lightweight and useful, so I have no problem with that, but this PR should update https://github.com/BVLC/caffe/blob/master/python/requirements.txt.

Other than that and the style issue noted, this looks good to go. Thanks for keeping us up with the times @philkr!

philkr:
I updated the PR.

longjon:
Looks good, thanks.

PR #2586, opened 2015-06-10
High level tensor and math API to unify the devices

futurely:
Torch which is endorsed by [Facebook AI Research](https://research.facebook.com/blog/879898285375829/fair-open-sources-deep-learning-modules-for-torch/) and [Google DeepMind](https://github.com/deepmind) provides a productive API for fast research prototyping. Compared to Torch and Theano, Caffe was deemed to be [not targeted toward the research community](http://fastml.com/torch-vs-theano/). One of most followed DL researcher @karpathy [switched from Caffe and Theano to Torch recently](http://karpathy.github.io/2015/05/21/rnn-effectiveness/) because the former two didn't get right with regards to the abstraction level. At the high level, it takes Torch several times less lines of code to define a complex model. At the low level, Torch has concise [tensor](https://github.com/torch/torch7/blob/master/doc/tensor.md) and [maths](https://github.com/torch/torch7/blob/master/doc/maths.md) APIs like those of MATLAB while Caffe sticks to the raw BLAS. This PR attempts to at least partly fill the abstraction level gap between BLAS and tensor maths to improve the development productivity of advanced neural network models and speed up the turnaround time of doing research with Caffe.

Three types of API syntax for matrix-matrix multiplication:

``` C++
Tensor A = ...;
Tensor B = ...;
```
1. `Tensor C = A \* B; // Zero copy with the help of move semantics in Boost.Move
2. `Tensor C = Tensor::mm(A, B); // Zero copy too` or `Tensor C; Tensor::mm(A, B, &C);` or `Tensor C; Tensor::mm(&C, A, B);`
3. `Tensor C; C.mm(A, B); // Store the result of A * B in C`
   Seamlessly compute on GPU:

``` C++
A.cuda(); // or A.gpu();
B.cuda();
C.cuda();
C.mm(A, B);
```

Issue #2585, opened 2015-06-10
Does classification.bin work fine?


RafaRuiz:
Hello, I'm trying to call this code:

I'm calling the binary stored in caffe/build/examples/cpp_classification/

`./classification.bin ../../../examples/imagenet/imagenet_deploy.prototxt ../../../caffe_mycaffe_train_iter_10000.caffemodel ../../../data/mycaffe/mycaffe_train_mean.binaryproto ../../../examples/imagenet/labels.txt ../../../data/mycaffe/testingset/alm/alm263.png`

and with more images. This is what I get with almost all results:

---------- Prediction for ../../../data/mycaffe/testingset/alm/alm263.png ----------
Segmentation fault (core dumped)

for others, I get this:

---------- Prediction for ../../../data/mycaffe/testingset/alm/alm265.png ----------
0.0000 - "0 almweird"
0.0000 - "b(5S?M????????Accuracy?AS`(?+?I????4S?JAccuracyI?AS?(?p+??I????4S 1

                                                                             InnerProductASpܩ`?PA??I????4S1

                                                                                                           InnerProductASpN????A?hJ????4S?J?1SplitpH?AS?9???J???4SICr/lib1SplitAS?8?p?HK???4SAReshape@!X8??G?3S1Pj(5S???ameteconstantat(5S?3?-S-???3?-S?3?-S

but for some of them, I get binary code for a long time like this previous which seems that I cracked Matrix.

Is this well compiled? what's wrong ?

flx42:
Hello, thanks for reporting the issue, can you try the following?
1) Is it working with the command given in the docs (on image `cat.jpg`)?
2) Compile in debug mode: set `DEBUG := 1` in `Makefile.config` and then prefix your command with `gdb --args` to get a nice backtrace.

StrongK:
In classification.cpp,its output is accuracy ranked in the top five.if your types less than five,you will get "Segmentation fault".Two ways can solve your problem:
1.Increase your object types to more than five;
2.modify classification.cpp where:
   std::vector<Prediction> Classify(const cv::Mat& img, int N = 5);
 let N=1 or others.Then you should recompile it

(Forgive me for my poor English，I hope it can help you)

flx42:
@StrongK thank you for reporting! I finally had the time to submit a simple fix.
Since it's just an example program and not a "tool", I'm silently modifying the number of classes to display in this situation.

PR #2715, opened 2015-07-08
Cuda/Blas array

philkr:
This is a basic implementation of CPU/GPU array class for caffe. It overloads basic element-wise operations (+,-,/,*,minimum,maximum,log,exp,...) it an easier to use interface that the current blas/cublas functions. For most parts it completely hides the underlying GPU/CPU interface, so you no longer have to write duplicate code for the two. It also works for both C++ and Python (see test_array.py for a python example).

Currently it is based on the SyncedMemory and can bind to any existing SyncedMemory, such as Blob data or diff. In terms of speed it's not much slower than the current blas functions, but not as fast as custom kernels, such as thrust lambda functions. If there is an interest in merging this, there are some more features we should add:
- [ ] Reductions such as (sum, min/max, ...) both along single dimensions and over the entire array
- [ ] Specialized expression evaluation for common functions (such as axpy, ...)

Sorry, this is kind of a large PR, and would probably take some work to merge. Please let me know if there is a general interest in the community to add this feature.

If you want to see the array class in action, you can look at my array based solver branch: https://github.com/philkr/caffe/tree/array_solver . It makes writing new solvers much easier without having to worry about CPU or GPU functions.


futurely:
Would you like to have a look at the design in #2586? By replacing the underlying data structure of Blob with SyncedTensor which has hight level math API, developers will be able to get rid of calling low level blas functions directly. A bigger benefit is that most layers will only have to be implemented in a single version instead of both CPU and GPU versions right now.

philkr:
I completely agree that caffe should have an array or tensor class, with
basic math primitives. I'm not sure if replacing SyncedMemory is a bit too
radical. I feel if someone really wants to use blas they should. It would
also be nice to have the array interfere minimally with the rest of caffe,
in the beginning.
Another thing which might help win people over is that and array might
actually have the potential of being faster than blas for operations like
a*x+b*y, as those could be optimised and executed in a single kernel
instead of 2 blas calls...

futurely:
Both this PR and #2585 could be parts of the [refactoring blueprints](http://tutorial.caffe.berkeleyvision.org/caffe-cvpr15-improving.pdf) of @Yangqing.

The number one priority of the design of #2585 is backward compatibiity and not to break any code that already works. Please note that all the existing APIs are kept intact and all the unit tests pass as usual which means that all the existing applications built with Caffe have to do nothing to adopt the new API. The developers are free to choose the old plain BLAS and the new tensor API.

futurely:
Speaking of operators, [DMLC's Minerva provides some examples](https://github.com/dmlc/minerva/tree/master/minerva/op).

shelhamer:
Closing since Caffe never took the all-knowing, all-doing array/tensor direction.

PR #2737, opened 2015-07-10
Basic cuDNN v3 support

slayton58:
This PR implements basic support for new features in cuDNN v3. Summarized these are:

* Adding versions of LCN and LRN layers
* Adding support for new algorithms for both forward and now backwards convolution.
* Move choice of algorithm + allocation of convolution workspace(s) from Forward_gpu to Reshape

### DIFF
--- a/include/caffe/vision_layers.hpp
+++ b/include/caffe/vision_layers.hpp
@@ -304,13 +304,24 @@ class CuDNNConvolutionLayer : public ConvolutionLayer<Dtype> {
   bool handles_setup_;
   cudnnHandle_t* handle_;
   cudaStream_t*  stream_;
+
+  // algorithms for forward and backwards convolutions
+  cudnnConvolutionFwdAlgo_t *fwd_algo_;
+  cudnnConvolutionBwdFilterAlgo_t *bwd_filter_algo_;
+  cudnnConvolutionBwdDataAlgo_t *bwd_data_algo_;
+
   vector<cudnnTensorDescriptor_t> bottom_descs_, top_descs_;
   cudnnTensorDescriptor_t    bias_desc_;
   cudnnFilterDescriptor_t      filter_desc_;
   vector<cudnnConvolutionDescriptor_t> conv_descs_;
-  int bottom_offset_, top_offset_, bias_offset_;
-  size_t workspaceSizeInBytes;
-  void *workspace;
+  int bottom_offset_, top_offset_, weight_offset_, bias_offset_;
+
+  size_t *workspace_fwd_sizes_;
+  size_t *workspace_bwd_data_sizes_;
+  size_t *workspace_bwd_filter_sizes_;
+  size_t workspaceSizeInBytes;  // size of underlying storage
+  void *workspaceData;  // underlying storage
+  void **workspace;  // aliases into workspaceData
 };
 #endif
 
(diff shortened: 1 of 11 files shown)


philkr:
Just out of curiosity where did you get cuDNN v3 from?

slayton58:
@philkr Sorry, I missed this question - I'm an engineer for NVIDIA

stas-sl:
I tried this PR with CUDNN v3 RC on my customised GoogleNet model and got only marginal performance increase (16ms vs 17ms with CUDNN v2 per image). While memory usage increased 1.5x times from ~2GB to 3GB on Amazon g2.2xlarge instance.

talda:
I also tried this PR and got similar results to what @stas-sl describes.
cudnnGetConvolutionForwardAlgorithm always selects either CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM or CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM.

When I comment out the call to cudnnGetConvolutionForwardAlgorithm and force CUDNN_CONVOLUTION_FWD_ALGO_FFT as the algorithm the call to cudnnGetConvolutionForwardWorkspaceSize returns CUDNN_STATUS_NOT_SUPPORTED, which means:
"The combination of the tensor descriptors, filter
descriptor and convolution descriptor is not
supported for the specified algorithm."

@slayton58 , Do you know do you know what are the limitations on these params for selecting the FFT algorithm?

slayton58:
@stas-sl  Most of the potential performance increases in cuDNNv3 come from either tunings for Maxwell, or using new FFT-based convolution algorithms. Given the GPUs you're using I suspect you're currently not using either of those features.

@talda There are a few restrictions: H and W must be packed (wStride = 1, hStride = W), the stride must be 1 (default for Caffe's convolution layers) and the padded size (H + 2*pad_h, W + 2*pad_w) <= (256, 256)

ducha-aiki:
@slayton58, on Titan X, which is Maxwell, speed gain for GoogLeNet is ~5%. Is it maximum I can get for such architecture?

xjtuljy:
@slayton58  I tried this with my own data and networks (1 convolutional + 2 fully connected), finding using gpu+cuDNN v3 is much slower than using gpu only (Time: CPU>>GPU+cuDNN>GPU only). I'm quite new in caffe and currently working with C++ (vs2013) on windows, do you have any idea of the possible reasons? Many thanks!

xjtuljy:
@slayton58   I tried this with my own data and networks (1 convolutional + 2 fully connected), finding using gpu+cuDNN v3 is much slower than using gpu only (Time: CPU>>GPU+cuDNN>GPU only). I tried different batch sizes but the results showed that cuDNN is at least twice slower than gpu only, especially on the backward pass (at least 3 times slower).  I'm quite new in caffe and currently working with C++ (vs2013) on windows, do you have any idea of the possible reasons? Many thanks!

talda:
@slayton58 My problem was that in order to save some memory I changed the stride of one of the convolution layers to 3.

Now that I fixed this I don't get this error but the amount of memory needed as workspace just for the fft forward convolution layers is so big that I can't really train my network on a K40.

Can you post any example of a reasonably sized net that can be trained using the fft algorithm?

liuyipei:
Ping: are there plans to get this merged soon? Or is this the start of a longer term project?

seanbell:
What's the status on this?

hyojinie:
Is cuDNN v3 currently supported?

shelhamer:
Merging as #3160 to include a few last details.

Thanks for this integration @slayton58 and sorry for the earlier holdup! I appreciate having the TODO for choosing the algo on `Reshape()` solved.

shelhamer:
@slayton58 this is failing `CuDNNConvolutionLayerTest/0.TestSimpleConvolutionGroupCuDNN` on a Titan X with CUDA 7.0 and driver version 346.46 for me. Please follow-up at #3160.

slayton58:
@shelhamer Fixed the convolution groups test failing - weight_offset_ wasn't being set correctly

include/caffe/vision_layers.hpp
```diff
@@ -304,13 +304,24 @@ class CuDNNConvolutionLayer : public ConvolutionLayer<Dtype> {
...
+  cudnnConvolutionBwdDataAlgo_t *bwd_data_algo_;
+
   vector<cudnnTensorDescriptor_t> bottom_descs_, top_descs_;
   cudnnTensorDescriptor_t    bias_desc_;
   cudnnFilterDescriptor_t      filter_desc_;
   vector<cudnnConvolutionDescriptor_t> conv_descs_;
-  int bottom_offset_, top_offset_, bias_offset_;
-  size_t workspaceSizeInBytes;
-  void *workspace;
+  int bottom_offset_, top_offset_, weight_offset_, bias_offset_;
```
shelhamer: Introducing `weight_offset_` caused the issue in https://github.com/BVLC/caffe/pull/2737#issuecomment-145994304 since it was shadowing the `BaseConvolutionLayer` member that is set in `BaseConvolutionLayer::LayerSetup()`.

shelhamer:
Once again merging as #3160 to include the little details mentioned there and the minimal fix for the `weight_offset_`. Thanks for taking another look @slayton58.

PR #2738, opened 2015-07-10
Sample implementation for memory pool by CNMeM support

slayton58:
Example implementation of a memory pool (in this case CNMeM, github.com/NVIDIA/cnmem) to Caffe through the singleton.

Implements all GPU allocations and frees through static methods in the singleton -- these resolve to either a call to CNMeM or straight to cudaMalloc / cudaFree depending on the USE_CNMEM flag in Makefile.config.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,6 +290,12 @@ ifeq ($(USE_CUDNN), 1)
 	COMMON_FLAGS += -DUSE_CUDNN
 endif
 
+# CNMeM configuration
+ifeq ($(USE_CNMEM), 1)
+	LIBRARIES += cnmem
+	COMMON_FLAGS += -DUSE_CNMEM
+endif
+
 # CPU-only configuration
 ifeq ($(CPU_ONLY), 1)
 	OBJS := $(PROTO_OBJS) $(CXX_OBJS)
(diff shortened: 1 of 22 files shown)


Yangqing:
Thanks Simon for the implementation! I have one little question: from the cnMem implementation, each GPU will reserve half its memory in default for the memory pool (https://github.com/NVIDIA/cnmem/blob/f27d3ee2783580ef9bf03611a915828fbe8ef9ba/src/cnmem.cpp#L1080). This would make that GPU somewhat exclusive to the running binary. Is my understanding correct?

This is not an issue, since most of time we will be using the GPU heavily enough that not sharing the GPU to another process is totally reasonable. Just want to raise this in case we have some other proposals of growing memory.

slayton58:
Your understanding is correct, but you can specify an amount of each GPU to hold at initialization (or allow the pool to grow - behaviour is set with a flag when cnmemInit is called)  - I believe I use 80% in my sample implementation, but in reality I run locally with ~95% of each device reserved.

The overall point of the implementation was to allow cuDNN to have enough scratch space to use it’s FFT-based convolution routines, which can give a massive speedup, so we willingly take the downside that we are effectively always running in process exclusive mode

shelhamer:
Thanks for the sample integration, but after further discussion CUB was chosen for a memory pool (until that direction stalled).

Issue #2722, opened 2015-07-08
Travis: WITH_CMAKE=true, WITH_CUDA=true does not use cuda.


philkr:
Travis with WITH_CMAKE=true, WITH_CUDA=true still has the CPU_ONLY=ON flag set, so it never builds the cuda code with cmake.
See https://github.com/BVLC/caffe/blob/master/scripts/travis/travis_build_and_test.sh line 10. Is there a reason for this or just oversight?

longjon:
It doesn't look like there's a good reason for this; the commits are @jeffdonahue's, so maybe he can comment further. I think we'd accept a PR that fixes this.

jeffdonahue:
Yup, my bad. Fix welcome.

shelhamer:
Fixed in #2740

PR #2740, opened 2015-07-10
Travis scripts for python3 and pytest for cmake.

philkr:
This PR adds python tests for CMake, and allows python3 to be tested on Travis. Later took a bit of hacking to get working:
 * The python tests are updated to be python3 compatible
 * We now have two miniconda directories (```$HOME/miniconda``` for python 2.7 and ```$HOME/miniconda3``` for python3).
 * Boost is installed through miniconda to make sure we have a boost_python that is compatible with the python version used.
 * bc was replaced by awk in the Makefile, as miniconda installs a new libreadline.so that breaks bc.
 * Remove the CPU_ONLY flag for cmake and WITH_CUDA (fixes issue #2722)
 * cmake 3 is required to ensure that the python interpreter and library versions match (the binary is downloaded directory from cmake.org)
 * For python3 protobuf 3 is compiled directly from source


scripts/travis/travis_build_and_test.sh
```diff
@@ -7,8 +7,17 @@ MAKE="make --jobs=$NUM_THREADS --keep-going"
 if $WITH_CMAKE; then
   mkdir build
   cd build
-  cmake -DBUILD_python=ON -DCMAKE_BUILD_TYPE=Release -DCPU_ONLY=ON ..
```
shelhamer: Can you split this fix into another commit or amend the message to note the fix?

python/caffe/test/test_net.py
```diff
@@ -2,6 +2,7 @@
 import tempfile
 import os
 import numpy as np
+import six
```
shelhamer: Please add six to the requirements.txt since it isn't always bundled as far as I know.

philkr: already done in #2714

scripts/travis/travis_build_and_test.sh
philkr: amended the message.

python/caffe/test/test_net.py
shelhamer: Oh right, thanks for the reminder.

PR #2741, opened 2015-07-10
Allowing netspec to have empty top (ntop=0)

philkr:
Just as it says, allows specification of layers with empty tops (eg Silence, HDF5Output).

### DIFF
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -90,6 +90,16 @@ class Top(object):
         return to_proto(self)
 
 
+class NoTop:
+    """ Special class for layers without a top """
+
+    def __init__(self, fn):
+        self.fn = fn
+
+    def to_proto(self):
+        return to_proto(self)
+
+
 class Function(object):
     """A Function specifies a layer, its parameters, and its inputs (which
     are Tops from other layers)."""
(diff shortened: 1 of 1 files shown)


python/caffe/net_spec.py
```diff
@@ -87,6 +87,12 @@ def to_proto(self):
         this top."""
 
         return to_proto(self)
+class NoTop:
```
shelhamer: Add two newlines for pep8itude.

shelhamer:
Looks good to me but @longjon should check.

python/caffe/net_spec.py
philkr: hmm that pep8... done.

longjon:
Thanks for PRing this. I have a rather simpler implementation myself, so stay tuned.

philkr:
Simpler than 20 line changes? :)

longjon:
See https://github.com/longjon/caffe/commit/b464a6e5d2e381185bd8aa0b9cbcba53c9819585, and preparations in https://github.com/longjon/caffe/commit/96c2fe1de80c9752b992c4578a3ce46028d21fc5.

philkr:
#2813 seems to be functionally the same, at least for what I'm using it for. So I don't have any objections going with the other solution, even though it's 50 line changes :)

shelhamer:
Closing in favor of #2813. Thanks for getting this rolling @philkr

PR #2742, opened 2015-07-11
Removes a unused variable warning

philkr:
Fixes a unused variable warning in AbsValLayer


jeffdonahue:
Thanks for the fix @philkr!  Odd that the warning has never shown up for me or in Travis.  Is `-Wall` not being applied to the nvcc calls for some reason?

Issue #2717, opened 2015-07-08
"You can install with" instruction is a bit odd


stuaxo:
You can install the dependencies with

```
for req in $(cat requirements.txt); do pip install $req; done
```

Is a bit strange, when 

```pip install -r requirements.txt```

Will do

ronghanghu:
No, `pip install -r requirements.txt` won't work.

Take a look at #1950 for details.

stuaxo:
Ok, closing this as a duplicate of #1950 .... it's a horrible thing to try and workaround, I'm going to bash pip a bit and see what happens ...

PR #2489, opened 2015-05-19
Python: update installation doc to use pip with `-r` option.

devxoul:
It's pretty much easier than for loop.

PR #2753, opened 2015-07-12
simplier pip command

sevaivanov:


seanbell:
See #1950, #2489, #2717

sevaivanov:
Thank you, also I forgot the install command as in "pip install -r requirements.txt". Also, you should not install as root user with sudo but use the --user to install it to your user profile as in "pip install -r requirements.txt --user".

PR #2762, opened 2015-07-14
tiny fix in Layer::Backward documentation

kashefy:
I realize this edit is not very significant, but thought it might as well get corrected. I suspect it came from copy-pasting the documentation of the Forward method.


longjon:
Thanks!

Issue #1376, opened 2014-10-29
Supercharge Caffe's relationship with Python


longjon:
I'm making this issue to track some ongoing improvements to pycaffe that I don't have time to PR just yet. #1020 made it possible for Caffe to call Python, in addition to the converse, and it's now becoming possible to combine Caffe C++ and Python code in essentially arbitrary ways.

Some of the coming changes are already PRs, some are already written locally, and some are not done yet. This list is to help me remember to PR everything when ready, to let you know what I have in mind, and provide a point of discussion.

I don't expect to be making more PRs before mid-November, but I do intend to do (or have done) everything on this list. There are a few other improvements that I have in mind, but that I'll leave off until they become more concrete.
- [x] Test the Python wrapper, starting at least with the core functionality (#173). [Now #1473.]
- [x] Rewrite the C++ wrapper code to use boost::python the way it was intended, removing the existing wrapper classes, and enabling some of the following changes. [Now #1703.]
- [x] Allow stepping of the solver, so that arbitrary Python code can be intermingled with gradient steps (#1228).
- [x] Make Python layer work by subclassing `caffe.Layer` in Python. This will make all the layer functionality (e.g., learnable parameters) exposable (in theory) from Python. (Updates #1020). [Now #1703.]
- [ ] Expose layer parameters (from protobuf) to Python by bridging C++ and Python protobuf messages.
- [x] Write examples exercising the capability of pycaffe, old and new.

BlGene:
Here are two additional ease-of-use features that I would suggest adding to this to-do list.

- [ ] loading models directly from netspec 
```ns = caffe.NetSpec()
...
n = caffe.Net(ns, caffe.TEST)```
- [ ] Solver specification in python.

PR #2767, opened 2015-07-14
Add arguments for PythonLayer

semisight:
This pull request adds a way for PythonLayers (#1703) to accept arguments through prototxt files.

Instead of loading files or other processing in `setup()`, this allows one to pass parameters through the 'arguments' string in PythonParameter. The receiver can then process the string however they choose: load through `json.loads()`, comma separated values, etc.

This pull request is a simpler alternative to #2001, without exception handling changes.

Example class:

```python
class MultLayer(caffe.Layer):
    def setup(self, bottom, top):
        self.mult_factor = int(self.arguments)

        top[0].reshape(bottom[0].num, bottom[0].channels, bottom[0].height,
            bottom[0].width)

        top[0].data[...] = self.mult_factor * bottom[0].data

        if propagate_down[0]:
            bottom[0].diff[...] = self.mult_factor * top[0].diff
```

Example addition to prototxt:
```protobuf
layer {
  name: "python1"
  type: "Python"
  bottom: "prev1"
  top: "curr1"
  python_param {
    module: "multlayer"
    layer: "MultLayer"
    arguments: "10"
  }
}
```

### DIFF
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -190,6 +190,10 @@ bp::object Blob_Reshape(bp::tuple args, bp::dict kwargs) {
   return bp::object();
 }
 
+string Layer_arguments(const Layer<Dtype>& layer) {
+  return layer.layer_param().python_param().arguments();
+}
+
 BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(SolveOverloads, Solve, 0, 1);
 
 BOOST_PYTHON_MODULE(_caffe) {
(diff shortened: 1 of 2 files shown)


BlGene:
A simple parameter string seems to be a good solution.  I would still add a comment recommending a particular way of passing parameters ( JSON seems suitable ) as the default, just to minimize unnecessary work of dealing with different formats. 

@longjon  does this sufficiently solve the second last point of #1376?

python/caffe/_caffe.cpp
```diff
@@ -248,6 +252,7 @@ BOOST_PYTHON_MODULE(_caffe) {
           bp::return_internal_reference<>()))
     .def("setup", &Layer<Dtype>::LayerSetUp)
     .def("reshape", &Layer<Dtype>::Reshape)
+    .add_property("arguments", &Layer_arguments)
```
shelhamer: This is adding an `arguments` property to all `Layer`s and not just `PythonLayer`.

shelhamer:
Thanks for contributing your own version of parameter passing to Python layers, but I'm closing this in favor of the original change by @tnarihi. The change here incorrectly introduces a new property for all layers.

See #2871 for the cherry-picking of the `PythonLayer` parameter string part of #2001 alone.

PR #2204, opened 2015-03-26
AdaDelta v2

kevinbache:
This is a minor update to #1122.  It brings @mohomran's AdaDelta solver into sync with modern Caffe and fixes the slow tests mentioned in #1122 by reducing the number of iterations for multi-iter tests from 500 down to 4 (as with the Nesterov tests).  All tests pass and take a normal amount of time.

PR #2782, opened 2015-07-18
AdaDelta Solver (v3)

matthiasplappert:
Picked up @kevinbache's branch (https://github.com/BVLC/caffe/pull/2204), merged it with master, resolved merge conflicts and fixed a couple of issues due to API changes. All tests pass.

However, I need input on one change, please see comment directly in the diff.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -82,12 +82,12 @@ class SGDSolver : public Solver<Dtype> {
   const vector<shared_ptr<Blob<Dtype> > >& history() { return history_; }
 
  protected:
+  void PreSolve();
   Dtype GetLearningRate();
   virtual void ApplyUpdate();
   virtual void Normalize(int param_id);
   virtual void Regularize(int param_id);
   virtual void ComputeUpdateValue(int param_id, Dtype rate);
-  virtual void PreSolve();
   virtual void ClipGradients();
   virtual void SnapshotSolverState(const string& model_filename);
   virtual void SnapshotSolverStateToBinaryProto(const string& model_filename);
@@ -162,19 +162,13 @@ template <typename Dtype>
 class AdaDeltaSolver : public SGDSolver<Dtype> {
  public:
   explicit AdaDeltaSolver(const SolverParameter& param)
-      : SGDSolver<Dtype>(param) { PreSolve(); constructor_sanity_check(); }
+      : SGDSolver<Dtype>(param) { AdaDeltaPreSolve(); }
   explicit AdaDeltaSolver(const string& param_file)
-      : SGDSolver<Dtype>(param_file) { PreSolve(); constructor_sanity_check(); }
+      : SGDSolver<Dtype>(param_file) { AdaDeltaPreSolve(); }
 
  protected:
-  virtual void PreSolve();
-  virtual void ComputeUpdateValue();
-  void constructor_sanity_check() {
-    CHECK_EQ(0, this->param_.base_lr())
-        << "Learning rate cannot be used with AdaDelta.";
-    CHECK_EQ("", this->param_.lr_policy())
-        << "Learning rate policy cannot be applied to AdaDelta.";
-  }
+  void AdaDeltaPreSolve();
+  virtual void ComputeUpdateValue(int param_id, Dtype rate);
 
   DISABLE_COPY_AND_ASSIGN(AdaDeltaSolver);
 };
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -82,12 +82,12 @@ class SGDSolver : public Solver<Dtype> {
   const vector<shared_ptr<Blob<Dtype> > >& history() { return history_; }
 
  protected:
-  void PreSolve();
   Dtype GetLearningRate();
   virtual void ApplyUpdate();
   virtual void Normalize(int param_id);
   virtual void Regularize(int param_id);
   virtual void ComputeUpdateValue(int param_id, Dtype rate);
+  virtual void PreSolve();
   virtual void ClipGradients();
   virtual void SnapshotSolverState(const string& model_filename);
   virtual void SnapshotSolverStateToBinaryProto(const string& model_filename);
@@ -162,9 +162,9 @@ template <typename Dtype>
 class AdaDeltaSolver : public SGDSolver<Dtype> {
  public:
   explicit AdaDeltaSolver(const SolverParameter& param)
-      : SGDSolver<Dtype>(param) { constructor_sanity_check(); }
+      : SGDSolver<Dtype>(param) { PreSolve(); constructor_sanity_check(); }
   explicit AdaDeltaSolver(const string& param_file)
-      : SGDSolver<Dtype>(param_file) { constructor_sanity_check(); }
+      : SGDSolver<Dtype>(param_file) { PreSolve(); constructor_sanity_check(); }
 
  protected:
   virtual void PreSolve();
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -159,6 +159,27 @@ class RMSPropSolver : public SGDSolver<Dtype> {
 };
 
 template <typename Dtype>
+class AdaDeltaSolver : public SGDSolver<Dtype> {
+ public:
+  explicit AdaDeltaSolver(const SolverParameter& param)
+      : SGDSolver<Dtype>(param) { constructor_sanity_check(); }
+  explicit AdaDeltaSolver(const string& param_file)
+      : SGDSolver<Dtype>(param_file) { constructor_sanity_check(); }
+
+ protected:
+  virtual void PreSolve();
+  virtual void ComputeUpdateValue();
+  void constructor_sanity_check() {
+    CHECK_EQ(0, this->param_.base_lr())
+        << "Learning rate cannot be used with AdaDelta.";
+    CHECK_EQ("", this->param_.lr_policy())
+        << "Learning rate policy cannot be applied to AdaDelta.";
+  }
+
+  DISABLE_COPY_AND_ASSIGN(AdaDeltaSolver);
+};
+
+template <typename Dtype>
 Solver<Dtype>* GetSolver(const SolverParameter& param) {
   SolverParameter_SolverType type = param.solver_type();
 
@@ -171,6 +192,8 @@ Solver<Dtype>* GetSolver(const SolverParameter& param) {
       return new AdaGradSolver<Dtype>(param);
   case SolverParameter_SolverType_RMSPROP:
       return new RMSPropSolver<Dtype>(param);
+  case SolverParameter_SolverType_ADADELTA:
+      return new AdaDeltaSolver<Dtype>(param);
   default:
       LOG(FATAL) << "Unknown SolverType: " << type;
   }
--- a/src/caffe/solver.cpp
+++ b/src/caffe/solver.cpp
@@ -934,10 +934,209 @@ void RMSPropSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
   }
 }
 
+template <typename Dtype>
+void AdaDeltaSolver<Dtype>::PreSolve() {
+  // Initialize the history
+  vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
+  this->history_.clear();
+  this->update_.clear();
+  this->temp_.clear();
+  for (int i = 0; i < net_params.size(); ++i) {
+    const Blob<Dtype>* net_param = net_params[i].get();
+    this->history_.push_back(shared_ptr<Blob<Dtype> >(new Blob<Dtype>(
+        net_param->num(), net_param->channels(), net_param->height(),
+        net_param->width())));
+    this->update_.push_back(shared_ptr<Blob<Dtype> >(new Blob<Dtype>(
+        net_param->num(), net_param->channels(), net_param->height(),
+        net_param->width())));
+    this->temp_.push_back(shared_ptr<Blob<Dtype> >(new Blob<Dtype>(
+        net_param->num(), net_param->channels(), net_param->height(),
+        net_param->width())));
+  }
+  for (int i = 0; i < net_params.size(); ++i) {
+    const Blob<Dtype>* net_param = net_params[i].get();
+    this->history_.push_back(shared_ptr<Blob<Dtype> >(new Blob<Dtype>(
+        net_param->num(), net_param->channels(), net_param->height(),
+        net_param->width())));
+  }
+}
+
+template <typename Dtype>
+void AdaDeltaSolver<Dtype>::ComputeUpdateValue() {
+  vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
+  vector<float>& net_params_weight_decay = this->net_->params_weight_decay();
+  Dtype delta = this->param_.delta();
+  Dtype momentum = this->param_.momentum();
+  Dtype weight_decay = this->param_.weight_decay();
+  string regularization_type = this->param_.regularization_type();
+  size_t update_history_offset = net_params.size();
+  switch (Caffe::mode()) {
+  case Caffe::CPU:
+    for (int param_id = 0; param_id < net_params.size(); ++param_id) {
+      Dtype local_decay = weight_decay * net_params_weight_decay[param_id];
+
+      if (local_decay) {
+        if (regularization_type == "L2") {
+          // add weight decay
+          caffe_axpy(net_params[param_id]->count(),
+              local_decay,
+              net_params[param_id]->cpu_data(),
+              net_params[param_id]->mutable_cpu_diff());
+        } else if (regularization_type == "L1") {
+          caffe_cpu_sign(net_params[param_id]->count(),
+              net_params[param_id]->cpu_data(),
+              this->temp_[param_id]->mutable_cpu_data());
+          caffe_axpy(net_params[param_id]->count(),
+              local_decay,
+              this->temp_[param_id]->cpu_data(),
+              net_params[param_id]->mutable_cpu_diff());
+        } else {
+          LOG(FATAL) << "Unknown regularization type: " << regularization_type;
+        }
+      }
+
+      // compute square of gradient in update
+      caffe_powx(net_params[param_id]->count(),
+          net_params[param_id]->cpu_diff(), Dtype(2),
+          this->update_[param_id]->mutable_cpu_data());
+
+      // update history of gradients
+      caffe_cpu_axpby(net_params[param_id]->count(), Dtype(1) - momentum,
+          this->update_[param_id]->cpu_data(), momentum,
+          this->history_[param_id]->mutable_cpu_data());
+
+      // add delta to history to guard against dividing by zero later
+      caffe_set(net_params[param_id]->count(), delta,
+          this->temp_[param_id]->mutable_cpu_data());
+
+      caffe_add(net_params[param_id]->count(),
+          this->temp_[param_id]->cpu_data(),
+          this->history_[update_history_offset + param_id]->cpu_data(),
+          this->update_[param_id]->mutable_cpu_data());
+
+      caffe_add(net_params[param_id]->count(),
+          this->temp_[param_id]->cpu_data(),
+          this->history_[param_id]->cpu_data(),
+          this->temp_[param_id]->mutable_cpu_data());
+
+      // divide history of updates by history of gradients
+      caffe_div(net_params[param_id]->count(),
+          this->update_[param_id]->cpu_data(),
+          this->temp_[param_id]->cpu_data(),
+          this->update_[param_id]->mutable_cpu_data());
+
+      // jointly compute the RMS of both for update and gradient history
+      caffe_powx(net_params[param_id]->count(),
+          this->update_[param_id]->cpu_data(), Dtype(0.5),
+          this->update_[param_id]->mutable_cpu_data());
+
+      // compute the update
+      caffe_mul(net_params[param_id]->count(),
+          net_params[param_id]->cpu_diff(),
+          this->update_[param_id]->cpu_data(),
+          net_params[param_id]->mutable_cpu_diff());
+
+      // compute square of update
+      caffe_powx(net_params[param_id]->count(),
+          net_params[param_id]->cpu_diff(), Dtype(2),
+          this->update_[param_id]->mutable_cpu_data());
+
+      // update history of updates
+      caffe_cpu_axpby(net_params[param_id]->count(), Dtype(1) - momentum,
+          this->update_[param_id]->cpu_data(), momentum,
+          this->history_[update_history_offset + param_id]->mutable_cpu_data());
+    }
+    break;
+  case Caffe::GPU:
+#ifndef CPU_ONLY
+    for (int param_id = 0; param_id < net_params.size(); ++param_id) {
+      Dtype local_decay = weight_decay * net_params_weight_decay[param_id];
+
+      if (local_decay) {
+        if (regularization_type == "L2") {
+          // add weight decay
+          caffe_gpu_axpy(net_params[param_id]->count(),
+              local_decay,
+              net_params[param_id]->gpu_data(),
+              net_params[param_id]->mutable_gpu_diff());
+        } else if (regularization_type == "L1") {
+          caffe_gpu_sign(net_params[param_id]->count(),
+              net_params[param_id]->gpu_data(),
+              this->temp_[param_id]->mutable_gpu_data());
+          caffe_gpu_axpy(net_params[param_id]->count(),
+              local_decay,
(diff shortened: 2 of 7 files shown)


src/caffe/solver.cpp
```diff
@@ -434,7 +434,8 @@ Dtype SGDSolver<Dtype>::GetLearningRate() {
         (Dtype(1.) + exp(-this->param_.gamma() * (Dtype(this->iter_) -
           Dtype(this->param_.stepsize())))));
   } else {
-    LOG(FATAL) << "Unknown learning rate policy: " << lr_policy;
+    rate = Dtype(0.);
```
matthiasplappert: I'm unsure what the best way to solve this is. The problem here is that AdaDelta solver does not support a learning rate. However, since AdaDelta inherits from SGD, and SGD calls ApplyUpdates which, in turn, calls this method, we trigger the default case and therefore the fatal log. Returning a rate of `0.0` works fine, but is probably likely to cause errors in other areas of the code base where a valid learning rate is expected. Any input on this is greatly appreciated!

matthiasplappert:
Travis failed b/c of lint error (the commented-out LOG is causing the error, which will go away before merging this anyway, see comment above).

src/caffe/solver.cpp
seanbell: One possible idea: keep the learning rate schedule, and treat it as a multiplier on the AdaDelta update step size.  The only ugly part of this solution is that it would require the user to specify `base_lr: 1 lr_policy: 'fixed'` in order to get the default behavior.

matthiasplappert: That would be a possible solution. Before going any further with this, is adding AdaDelta even of interest for caffe? I don't want to invest time into this if it's not likely to land in master eventually.

PatWie: I would strongly argue for AdaDelta shipped within the Caffe-Framework. I was surprised that it isn't already in the master-branch.

shelhamer:
@matthiasplappert thanks for making the update, but take another look at #2397 and #1977 and see how the regularization and logging code was pulled out into `SGDSolver`.

src/caffe/solver.cpp
ronghanghu: I am also strongly in favor of having AdaDelta in Caffe. I'll go over this PR today and during weekend.

ronghanghu: For the learning rate issue, I suggest using `base_lr: 1` and `lr_policy: 'fixed'` I suppose learning rate specification is still sometimes needed, even if you use AdaDelta. Take fine-tuning as an example, you may still want to have a smaller learning rate on pre-trained layers than on random-initialized layers even if you use AdaDelta.

include/caffe/solver.hpp
```diff
@@ -129,6 +129,27 @@ class AdaGradSolver : public SGDSolver<Dtype> {
 };
 
 template <typename Dtype>
+class AdaDeltaSolver : public SGDSolver<Dtype> {
+ public:
+  explicit AdaDeltaSolver(const SolverParameter& param)
+      : SGDSolver<Dtype>(param) { PreSolve(); constructor_sanity_check(); }
+  explicit AdaDeltaSolver(const string& param_file)
+      : SGDSolver<Dtype>(param_file) { PreSolve(); constructor_sanity_check(); }
```
ronghanghu: I suppose you have something wrong here. Now you are calling `PreSolve()` in constructor of both `AdaDeltaSolver` and `SGDSolver`, and since it is a virtual method, you are calling `AdaDeltaSolver::PreSolve()` twice when constructing a AdaDeltaSolver instance. Is that the desired behavior?

include/caffe/solver.hpp
```diff
@@ -129,6 +129,27 @@ class AdaGradSolver : public SGDSolver<Dtype> {
...
+
+ protected:
+  virtual void PreSolve();
+  virtual void ComputeUpdateValue(int param_id, Dtype rate);
+  void constructor_sanity_check() {
+    CHECK_EQ(0, this->param_.base_lr())
+        << "Learning rate cannot be used with AdaDelta.";
+    CHECK_EQ("", this->param_.lr_policy())
+        << "Learning rate policy cannot be applied to AdaDelta.";
+  }
```
ronghanghu: Let's still keep the learning rate policy as discussed.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
   }
 }
 
+template <typename Dtype>
+void AdaDeltaSolver<Dtype>::PreSolve() {
+  // Add the extra history entries for AdaDelta after those from
+  // SGDSolver::PreSolve
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
```
ronghanghu: Are you trying to call `AdaDeltaSolver<Dtype>::PreSolve()` after calling `SGDSolver<Dtype>::Presolve()`?

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+            net_params[param_id]->cpu_data(),
+            this->temp_[param_id]->mutable_cpu_data());
+        caffe_axpy(net_params[param_id]->count(),
+            local_decay,
+            this->temp_[param_id]->cpu_data(),
+            net_params[param_id]->mutable_cpu_diff());
+      } else {
+        LOG(FATAL) << "Unknown regularization type: " << regularization_type;
+      }
+    }
```
ronghanghu: Remove the regularization code. Regularization should be handled in `SGDSolver<Dtype>::Regularize`. Take AdaGradSolver as an example.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+            net_params[param_id]->gpu_data(),
+            this->temp_[param_id]->mutable_gpu_data());
+        caffe_gpu_axpy(net_params[param_id]->count(),
+            local_decay,
+            this->temp_[param_id]->gpu_data(),
+            net_params[param_id]->mutable_gpu_diff());
+      } else {
+        LOG(FATAL) << "Unknown regularization type: " << regularization_type;
+      }
+    }
```
ronghanghu: remove regularization code here

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -61,8 +62,6 @@ class GradientBasedSolverTest : public MultiDeviceTest<TypeParam> {
     ostringstream proto;
     proto <<
        "max_iter: " << num_iters << " "
-       "base_lr: " << learning_rate << " "
-       "lr_policy: 'fixed' "
        "iter_size: " << iter_size << " "
```
ronghanghu: Let's keep the `base_lr` and `lr_policy` as discussed above.

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -558,4 +578,78 @@ TYPED_TEST(NesterovSolverTest, TestLeastSquaresUpdateWithEverythingAccum) {
...
+  typedef typename TypeParam::Dtype Dtype;
+  const Dtype kLearningRate = 0.0;
+  const Dtype kWeightDecay = 0.1;
+  const Dtype kMomentum = 0.95;
+  const int kNumIters = 4;
+  for (int i = 0; i <= kNumIters; ++i) {
+      this->TestLeastSquaresUpdate(kLearningRate, kWeightDecay, kMomentum, i);
+  }
+}
+
```
ronghanghu: Please add one more test case:

TestLeastSquaresUpdateWithEverythingAccum (where you may set kNumIters = 4 and kIterSize = 2 and use CheckAccumulation)
You may take a look at those AdaGradSolverTest for details.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+}
+
+template <typename Dtype>
+void AdaDeltaSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
+  const vector<float>& net_params_weight_decay =
+          this->net_->params_weight_decay();
+  Dtype delta = this->param_.delta();
+  Dtype momentum = this->param_.momentum();
+  Dtype weight_decay = this->param_.weight_decay();
```
ronghanghu: let's add `Dtype local_rate = rate * net_params_lr[param_id];` here

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+        this->update_[param_id]->cpu_data(),
+        net_params[param_id]->mutable_cpu_diff());
+
+    // compute square of update
+    caffe_powx(net_params[param_id]->count(),
+        net_params[param_id]->cpu_diff(), Dtype(2),
+        this->update_[param_id]->mutable_cpu_data());
+
+    // update history of updates
+    caffe_cpu_axpby(net_params[param_id]->count(), Dtype(1) - momentum,
```
ronghanghu: let's add `local_rate` multiplication here.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+        this->update_[param_id]->gpu_data(),
+        net_params[param_id]->mutable_gpu_diff());
+
+    // compute square of update
+    caffe_gpu_powx(net_params[param_id]->count(),
+        net_params[param_id]->gpu_diff(), Dtype(2),
+        this->update_[param_id]->mutable_gpu_data());
+
+    // update history of updates
+    caffe_gpu_axpby(net_params[param_id]->count(), Dtype(1) - momentum,
```
ronghanghu: let's add `local_rate` multiplication here.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+    // jointly compute the RMS of both for update and gradient history
+    caffe_powx(net_params[param_id]->count(),
+        this->update_[param_id]->cpu_data(), Dtype(0.5),
+        this->update_[param_id]->mutable_cpu_data());
+
+    // compute the update
+    caffe_mul(net_params[param_id]->count(),
+        net_params[param_id]->cpu_diff(),
+        this->update_[param_id]->cpu_data(),
+        net_params[param_id]->mutable_cpu_diff());
```
ronghanghu: let's add local_rate multiplication here.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+    // jointly compute the RMS of both for update and gradient history
+    caffe_gpu_powx(net_params[param_id]->count(),
+        this->update_[param_id]->gpu_data(), Dtype(0.5),
+        this->update_[param_id]->mutable_gpu_data());
+
+    // compute the update and copy to net_diff
+    caffe_gpu_mul(net_params[param_id]->count(),
+        net_params[param_id]->gpu_diff(),
+        this->update_[param_id]->gpu_data(),
+        net_params[param_id]->mutable_gpu_diff());
```
ronghanghu: let's add local_rate multiplication here.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+
+    // compute square of update
+    caffe_gpu_powx(net_params[param_id]->count(),
+        net_params[param_id]->gpu_diff(), Dtype(2),
+        this->update_[param_id]->mutable_gpu_data());
+
+    // update history of updates
+    caffe_gpu_axpby(net_params[param_id]->count(), Dtype(1) - momentum,
+        this->update_[param_id]->gpu_data(), momentum,
+        this->history_[update_history_offset + param_id]->mutable_gpu_data());
```
ronghanghu: let's add `local_rate` multiplication at the end.

    caffe_gpu_scale(net_params[param_id]->count(), local_rate,
        net_params[param_id]->gpu_diff(),
        net_params[param_id]->gpu_diff());

src/caffe/solver.cpp
```diff
@@ -775,9 +776,192 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+
+    // compute square of update
+    caffe_powx(net_params[param_id]->count(),
+        net_params[param_id]->cpu_diff(), Dtype(2),
+        this->update_[param_id]->mutable_cpu_data());
+
+    // update history of updates
+    caffe_cpu_axpby(net_params[param_id]->count(), Dtype(1) - momentum,
+        this->update_[param_id]->cpu_data(), momentum,
+        this->history_[update_history_offset + param_id]->mutable_cpu_data());
```
ronghanghu: let's add `local_rate` multiplication at the end.

    caffe_cpu_scale(net_params[param_id]->count(), local_rate,
        net_params[param_id]->cpu_diff(),
        net_params[param_id]->mutable_cpu_diff());

ronghanghu:
@matthiasplappert Thanks for your great PR to introduce AdaDelta solver into Caffe!

Please modify and update according to the reviews

src/caffe/solver.cpp
ronghanghu: Remove the regularization code here.

ronghanghu:
@matthiasplappert Thanks for your great PR to introduce AdaDelta solver into Caffe!

The remaining work include:

* Add learning rate.
* Remove regularization.
* Add more test cases.
* Change back `Presolve()` to be non-virtual.

Please modify and update according to the reviews.

matthiasplappert:
@ronghanghu I'll try to find some time over the weekend to get all of this done. We should also thank @kevinbache and especially @mohomran (who wrote the original code), since I just carried on with their work.

ronghanghu:
#2836 and #2887 introduced new conflicts to be resolved.

matthiasplappert:
I'll resolve the conflict later today and (hopefully) address the reaming issues as well.

src/caffe/solver.cpp
matthiasplappert: Taken care of in https://github.com/matthiasplappert/caffe/commit/fc3551812353966e2e0dccf2db530e94bcf407dc

src/caffe/solver.cpp
matthiasplappert: Taken care of in https://github.com/matthiasplappert/caffe/commit/fc3551812353966e2e0dccf2db530e94bcf407dc

include/caffe/solver.hpp
matthiasplappert: `virtual` issue addressed in https://github.com/BVLC/caffe/commit/aedff906f6b990c8d23ef2db646bc94ffb6e3497

src/caffe/solver.cpp
matthiasplappert: Addressed in https://github.com/matthiasplappert/caffe/commit/fc3551812353966e2e0dccf2db530e94bcf407dc

src/caffe/solver.cpp
matthiasplappert: Addressed in https://github.com/matthiasplappert/caffe/commit/322a9de77b6d56f4dea0fa060bc02e7b2c7b97d5

src/caffe/test/test_gradient_based_solver.cpp
matthiasplappert: Addressed in https://github.com/matthiasplappert/caffe/commit/322a9de77b6d56f4dea0fa060bc02e7b2c7b97d5

src/caffe/solver.cpp
matthiasplappert: Addressed in https://github.com/matthiasplappert/caffe/commit/322a9de77b6d56f4dea0fa060bc02e7b2c7b97d5

src/caffe/solver.cpp
matthiasplappert: Addressed in https://github.com/matthiasplappert/caffe/commit/322a9de77b6d56f4dea0fa060bc02e7b2c7b97d5

src/caffe/test/test_gradient_based_solver.cpp
matthiasplappert: Addressed in https://github.com/BVLC/caffe/commit/e4eb50bef42561507792f0a6601048cfc55e0790

matthiasplappert:
Update on this: This branch is now up-to-date with master and all feedback has been addressed. The tests pass locally and I expect them to also pass on the CI.

Please review my changes and let me know if everything else is required on my end, e.g. cleaning up the commit history (not sure how you usually handle this). I've also pointed out the relevant commits in each feedback discussion to hopefully help with reviewing the changes.

Finally, I have one suggestion to make: having all solvers in one relatively big file (`solver.cpp`) proved to be a really big pain while resolving the merge conflicts. The problem there was that RMSProb and AdaDelta were completely mixed up since they share a lot of similar code. I would propose to eventually split out the individual solvers into separate files to avoid this in the future. Should I open an issue for that?

ronghanghu:
@matthiasplappert Thanks a lot for the update. I will review the changes today.

>>Finally, I have one suggestion to make: having all solvers in one relatively big file (solver.cpp) proved to be a really big pain while resolving the merge conflicts. The problem there was that RMSProb and AdaDelta were completely mixed up since they share a lot of similar code. I would propose to eventually split out the individual solvers into separate files to avoid this in the future. Should I open an issue for that?

Yes, this is quite a problem. I expect to send a solver refactor PR to split solver.cpp and extract common code for these adaptive gradient solvers, after merging AdaDelta and Adam (#2856).

include/caffe/solver.hpp
```diff
@@ -159,6 +158,22 @@ class RMSPropSolver : public SGDSolver<Dtype> {
...
 template <typename Dtype>
+class AdaDeltaSolver : public SGDSolver<Dtype> {
+ public:
+  explicit AdaDeltaSolver(const SolverParameter& param)
+      : SGDSolver<Dtype>(param) { PreSolve(); }
+  explicit AdaDeltaSolver(const string& param_file)
+      : SGDSolver<Dtype>(param_file) { PreSolve(); }
+
+ protected:
+  void PreSolve();
```
ronghanghu: I think it is better to rename `AdaDeltaSolver::PreSolve()` into `AdaDeltaSolver::AdaDeltaPreSolve()`. Since you are going to call AdaDeltaSolver's presolve function after SGDSolver's presolve function, it is better to avoid a name conflict with `SGDSolver::PreSolve()`.

include/caffe/solver.hpp
```diff
@@ -159,6 +158,22 @@ class RMSPropSolver : public SGDSolver<Dtype> {
...
+  virtual void Regularize(int param_id);
```
ronghanghu: I didn't find a significant difference between your `AdaDeltaSolver::Regularize` and the original `SGDSolver::Regularize` (the former is almost duplicated from the latter).

We can just use `SGDSolver::Regularize` and avoid introducing a new regularization function in AdaDeltaSolver.

src/caffe/solver.cpp
```diff
@@ -860,6 +860,214 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+  for (int i = 0; i < net_params.size(); ++i) {
+        const vector<int>& shape = net_params[i]->shape();
+        this->history_.push_back(
+                shared_ptr<Blob<Dtype> >(new Blob<Dtype>(shape)));
+  }
+}
+
+template <typename Dtype>
+void AdaDeltaSolver<Dtype>::Regularize(int param_id) {
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
```
ronghanghu: The only difference between your `AdaDeltaSolver::Regularize` and the original `SGDSolver::Regularize` seem to be that you use `const vector<shared_ptr<Blob<Dtype> > >& net_params` instead of `const vector<Blob<Dtype>*>& net_params`. The rest are all the same.

So, I believe we don't need a `AdaDeltaSolver::Regularize` here. Let's just use `SGDSolver::Regularize` instead.

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -882,6 +900,91 @@ TYPED_TEST(NesterovSolverTest, TestSnapshotShare) {
...
+  typedef typename TypeParam::Dtype Dtype;
+  const Dtype kLearningRate = 1.0;
+  const Dtype kWeightDecay = 0.1;
+  const Dtype kMomentum = 0.95;
+  const int kNumIters = 4;
+  const int kIterSize = 2;
+  this->CheckAccumulation(kLearningRate, kWeightDecay, kMomentum, kNumIters,
+      kIterSize);
+}
+
```
ronghanghu: #2836 and #2887 modifies solver as well as introduces new share param and snapshot tests. To be consistent, let's also add 4 new test cases:

    TestLeastSquaresUpdateWithEverythingShare
    TestLeastSquaresUpdateWithEverythingAccumShare
    TestSnapshot
    TestSnapshotShare

For TestSnapshot, you can take a look at TestSnapshot in SGDSolverTest as an example. For the other 3 shared cases, you just need to add `this->share_ = true;` to the corresponding test cases.

src/caffe/solver.cpp
```diff
@@ -860,6 +860,214 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
 }
 
 template <typename Dtype>
+void AdaDeltaSolver<Dtype>::PreSolve() {
+  // Add the extra history entries for AdaDelta after those from
+  // SGDSolver::PreSolve
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
```
ronghanghu: You should use `const vector<Blob<Dtype>*>& net_params = this->net_->learnable_params();` instead of `const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();`

src/caffe/solver.cpp
```diff
@@ -860,6 +860,214 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+    break;
+  }
+  default:
+    LOG(FATAL) << "Unknown caffe mode: " << Caffe::mode();
+  }
+}
+
+template <typename Dtype>
+void AdaDeltaSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
```
ronghanghu: You should use `const vector<Blob<Dtype>*>& net_params = this->net_->learnable_params();` instead.

ronghanghu:
@matthiasplappert I just made a few comments above. Let's get the following work done and I think this PR will be ready:

* Rename `AdaDelta::PreSolve` into `AdaDelta::AdaDeltaPreSolve`.
* Remove the `AdaDelta::Regularize()` function entirely.
* Replace `const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();` with `const vector<Blob<Dtype>*>& net_params = this->net_->learnable_params();`
* Add 4 more test cases to be consistent with #2836 and #2887.

After that, squash these 12+ commit into one single commit, and take a further rebase against bvlc/master.

shelhamer:
@matthiasplappert a note about history: instead of squashing to a single commit, please squash the commits by each author into a single commit. This will leave three commits by @mohomran @kevinbache and yourself. In future work please make use of rebase instead of merge, as our policy is to only have merge commits for PRs. Thanks.

> having all solvers in one relatively big file (solver.cpp) proved to be a really big pain while resolving the merge conflicts. [...] I would propose to eventually split out the individual solvers into separate files to avoid this in the future.

Absolutely, and this is noted in #2860.

matthiasplappert:
@ronghanghu Thanks for the thorough review! I'm still very new to caffe, so your feedback is very much appreciated.

I've addressed the remaining feedback and cleaned up the commit history (also: no more merges). All tests pass locally (not sure if Travis will pick this up since the branch was force-pushed to override the history). Let me know if anything else needs to be done before we can land this in master.

ronghanghu:
@matthiasplappert Thanks for the update! I'll take a final review, and I expect to merge it tomorrow. @jeffdonahue could you also take a look?

ronghanghu:
Finished final review. Thanks for the @mohomran, @kevinbache and @matthiasplappert for this excellent AdaDelta solver.

PR #2793, opened 2015-07-20
Added Tensor class.

Russell91:
Sets up the data_ and diff_ members of the Blob class to be Tensors, rather than SyncedMem. This allows a lot of the logic in blob.cpp to be factored out into the Tensor class. Note that this change should be backwards compatible. The primary benefit is that other areas of the framework with simple Tensor operations can call high level methods like axpy rather than using the low-level math utils and #ifdef CPU_ONLY. It may also be possible to wrap this Tensor class using boost python in the future, such that one could perform gpu operations on blobs from python.

### DIFF
--- a/examples/feature_extraction/readme.md
+++ b/examples/feature_extraction/readme.md
@@ -10,7 +10,7 @@ Extracting Features
 ===================
 
 In this tutorial, we will extract features using a pre-trained model with the included C++ utility.
-Note that we recommend using the Python interface for this task, as for example in the [filter visualization example](http://nbviewer.ipython.org/github/BVLC/caffe/blob/master/examples/filter_visualization.ipynb).
+Note that we recommend using the Python interface for this task, as for example in the [filter visualization example](http://nbviewer.ipython.org/github/BVLC/caffe/blob/master/examples/00-classification.ipynb).
 
 Follow instructions for [installing Caffe](../../installation.html) and run `scripts/download_model_binary.py models/bvlc_reference_caffenet` from caffe root directory.
 If you need detailed information about the tools below, please consult their source code, in which additional documentation is usually provided.


BlGene:
Did you miss: ```make lint```?

Russell91:
@BIGene Ah, didn't know you could run that locally :). I was surprised when everything failed on Travis, given that all the tests passed on CPU/GPU.

initialneil:
The lint checking is very strict. You have to make sure lines don't extend 80 or 100 characters and don't end with a white space.

philkr:
There are several versions of this same idea floating around (#2586,#2715). We should probably figure out a way to combine those ideas and convince @longjon to include this in caffe.

Russell91:
@philkr ah, I hadn't seen those, but they look quite nice. I think the most important choice is how to define the semantics of the memory model. Things like multi-gpu support and local cpu computation could be really nice, but they're tricky to get right, esp. given that the interface should feel natural in both python and c++. It would be great to pick a direction to head towards and get something simple that will grow well. I've found that it's just really nice to have a higher level tensor library to use rather than the math utils when appropriate. But I've been pretty hesitant to lean on it too much because I don't know what the final Tensor class will look like. Two sources of difficulty that I've found which are important to consider for back compat in the design are:

1) Weight sharing mechanisms of e.g. diffs in the blob class have to share the SyncedMem, not the Tensor, as different Blobs can have different shapes while sharing the same memory. This is difficult to manage, especially because Reshape() can reset the underlying SyncedMem without notifying other blobs.

2) Caffe works with a global CPU/GPU state. This is ultimately a design choice that works pretty well in my experience. But as soon as you use Tensors rather than math utils, you lose some control and end up wanting some local overriding. I haven't figured out how to do this in a lightweight way, though it seems that both #2586 and #2715 provide some ideas here.

PR #2805, opened 2015-07-22
added weighted loss layer + test

shaibagon:
The PR adds a new loss layer the "weighted loss" layer.
It uses a lookup table (W) to compute the loss between GT label `g` and predicted label `p`:
`W[g,p]`

The code is documented, and tested. passed lint check as well.

Only CPU implementation at the moment.


lukeyeager:
You've got some whitespace issues with `caffe.proto`.

With whitespace differences:
https://github.com/BVLC/caffe/pull/2805/files#diff-e0bd8d028edf0f88febfebe3bf703904

Without:
https://github.com/BVLC/caffe/pull/2805/files?w=#diff-e0bd8d028edf0f88febfebe3bf703904

shaibagon:
@lukeyeager - You are right. got it now. line ending issues (I hate it!). I wonder how come lint checker didn't see it? 
Anyway, I updated my PR accordingly. 
Thanks!

shaibagon:
@lukeyeager how do we progress from here? Are you going to accept this PR? Is there anything I can do on my part to accelerate the process?

bhack:
@shaibagon No. As  you can  see BVLC member status it is unknown (probably on vacation). And nobody knows their activity plans. You need only to wait if you have enough time.

sonnyhu:
@shaibagon I don't have any context here, so I am probably asking dumb questions. But wondering why you closed this PR? Is there a weighted loss in upstream master?

shaibagon:
@sonnyhu no question is dumb.

I closed this loss because it did not work as I expected.

If you want weighted loss you can use `"InfogainLoss"`: see [this SO thread for example](http://stackoverflow.com/a/30497907/1714410). 
I have an open PR for an upgrade infogain loss layer #3855

PR #2812, opened 2015-07-23
Expose blob loss weight to python

philkr:
Makes it easier to identify loss blobs from python.


longjon:
Cool, this looks like a pretty straightforward exposure.

chrischoy:
I guess it wasn't meant to modify the blob_loss_weights from python since the layer forward function doesn't update the modified blob_loss_weights.

Currently, I'm trying to modify a loss weight on the fly. During the backward pass top[i].diff[0] (blob_loss_weights) seems to update correctly. So even though the loss weight were correctly updated. The displayed loss might not reflect the update.

Also

```
net._blob_loss_weights[i] = value  # update correctly
net.blob_loss_weights['layer_name'] = value # does not update the value!
```

PR #2813, opened 2015-07-24
Python net spec cleanups and top-less layers 

longjon:
This PR makes small stylistic improvements to the net spec code and includes an alternative implementation of the functionality provided by #2741.

While #2741 generates top-less layers, it breaks the meaningfulness of `fn.tops` (for `fn` a `Function`); in particular, it need no longer be true that `fn.ntop == len(fn.tops)`, and iterating through `fn.tops` is no longer the same as iterating through `fn`'s tops, which now requires a special case (https://github.com/BVLC/caffe/pull/2741/files#diff-57d5ca5bbea564c189da52386291f6d8R145), which, while mild in the current code, would need to propagate to future code that operates on the (Python-level) graph of `Function`s and `Top`s.

The fake top doesn't really need to be added to `fn.tops`, and could just be generated around https://github.com/BVLC/caffe/pull/2741/files#diff-57d5ca5bbea564c189da52386291f6d8R198, except that this would prohibit the top-less `Function` from discovering a name assigned to this fake top.

We could simply forgo naming of top-less layers, but a more complete solution is to allow naming of `Function`s (layers) separately from `Top`s (the fourth commit here), and have top-less layers return their _`Function`s_ rather than any (fictitious) `Top`s (the fifth). It might be argued that this solution ungraciously overloads the semantics of net spec layer creation, since layers sometimes return `Top`s and sometimes `Function`s, but this is surely no worse than any solution in which top-less layers return something, since something cannot be a top that isn't. It's probably best to view this as an abbreviation of a syntax in which all layers return _both_ their `Function` and all their `Top`s (an option which could actually be added in the future if it becomes desirable to create layers with distinct layer and first-top names).

A simple test for top-less layers is included.

Thanks @philkr for getting the ball rolling on toplessness, and let me know how you feel about this solution.

### DIFF
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -108,7 +108,15 @@ class Function(object):
             del self.params['in_place']
         self.tops = tuple(Top(self, n) for n in range(self.ntop))
 
-    def _get_name(self, top, names, autonames):
+    def _get_name(self, names, autonames):
+        if self not in names and self.ntop > 0:
+            names[self] = self._get_top_name(self.tops[0], names, autonames)
+        elif self not in names:
+            autonames[self.type_name] += 1
+            names[self] = self.type_name + str(autonames[self.type_name])
+        return names[self]
+
+    def _get_top_name(self, top, names, autonames):
         if top not in names:
             autonames[top.fn.type_name] += 1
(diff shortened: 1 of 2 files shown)


jeffdonahue:
Cool! LGTM.

shelhamer:
Sweet. Thanks Jon!

PR #2822, opened 2015-07-27
Add functionality to do per-datum mean / stddev preprocessing

flynnhe:
Can we add the functionality to preprocess each datum/image by subtracting the mean and dividing by its standard deviation?


flynnhe:
Just checking to see what the errors are.

ronghanghu:
I believe this PR is not necessary, since MVNLayer already normalizes the input to have 0-mean and/or unit (1) variance.

https://github.com/BVLC/caffe/blob/master/include/caffe/common_layers.hpp#L330-L362

PR #2831, opened 2015-07-28
Expose tops and bottoms associated with layers in python, fix #2829

### LINKED ISSUE #2829: pycaffe net.forward() assumes end blobs have the same name as their layers
If you call `net.forward(end='somelayer')` from python, where `somelayer` is a layer that has a different name for its output blob, _Net_forward() will attempt to access `net.blobs['somelayer']` (and in fact, returns this as the output).  This happens because [this line](https://github.com/BVLC/caffe/blob/master/python/caffe/pycaffe.py#L80) appends the end layer name to the list of blobs to be output. If the appended name doesn't exist as a blob, the result is an exception.  

Note that there may also be multiple outputs for the end layer, in which case the current behavior doesn't make sense either.

cdoersch:
The main goal is to fix issue #2829, in which `forward()` in the python interface attempts to return the top blobs of the last layer in `forward()`, but instead uses the layer names as if they were blob names. Turns out there's not currently a way for python code to figure out which blobs are connected to which layers, so this PR exposes that functionality and fixes the bug.

### DIFF
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -124,12 +124,18 @@ class Net {
   inline const vector<vector<Blob<Dtype>*> >& bottom_vecs() const {
     return bottom_vecs_;
   }
+  inline const vector<vector<int> >& bottom_id_vecs() const {
+    return bottom_id_vecs_;
+  }
   /**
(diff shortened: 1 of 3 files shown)


cdoersch:
#2975 supersedes this.

PR #2835, opened 2015-07-29
Correct index checking of Blob indices

bearpaw:


shelhamer:
See #792 although note it is PR'd to the obsolete `dev` branch.

Noiredd:
This has been independently implemented in #6021. Turns out that this change is not as trivial as it seems - see conversation under #3420 and let's have the conversation under that issue. For clarity, I will close this PR. Please open a new one if you want to follow up with a more comprehensive solution @bearpaw, thank you.

Issue #2827, opened 2015-07-27
Provide an Adam solver


slavnyk:
Are there any plans to add Adam solver to Caffe in the near future?
[Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
[Python implementation (Theano)](https://github.com/Lasagne/Lasagne/issues/144)
[Lua implementation (Torch)](https://github.com/torch/optim/blob/master/adam.lua)
I have been using the Lua implemenation for training neural networks for low-level image processing (such as [noise reduction and super-resolution](https://github.com/nagadomi/waifu2x)) with great success, but could not reproduce results of similar quality in Caffe with SGD/Adagrad/Nesterov. There are reports of better results (faster, more accurate) when training various types of deep networks using Adam.

kashif:
I would like to try to do this, if no one else has started. I took the time to update the adam implementation for Lasagne and torch to the latest version of the paper.  So next I will try to implement it in caffe.

PatWie:
Just started to write the CPU version:
https://github.com/PatWie/caffe/blob/adam/src/caffe/solver.cpp

Not sure if we really want to use

   this->iter

for computing the effective stepsize.

kashif:
Ok I was thinking perhaps:

```c++
int t = this->iter_ / this->param_.stepsize() + 1;
```

shelhamer:
Thanks for the suggested solver. Closing as addressed by #2856 -- see the PR for status.

PR #2856, opened 2015-08-04
Adam solver

PatWie:
This commit implements the Adam solver by Kingma et. al for CPU and
GPU. All solver parameters are defined in the caffe.proto. This also
adds an example for the MNIST dataset.

see issue #2827 

Before merging, please review the code. I will add changes to this branch (and rebase) if there should be something to change.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -159,6 +159,23 @@ class RMSPropSolver : public SGDSolver<Dtype> {
 };
 
 template <typename Dtype>
+class AdamSolver : public SGDSolver<Dtype> {
+ public:
+  explicit AdamSolver(const SolverParameter& param)
+      : SGDSolver<Dtype>(param) { initAdam();}
+  explicit AdamSolver(const string& param_file)
+      : SGDSolver<Dtype>(param_file) { initAdam(); }
+
+ protected:
+  virtual void ComputeUpdateValue(int param_id, Dtype rate);
+
+  DISABLE_COPY_AND_ASSIGN(AdamSolver);
+
+  void initAdam();
+  vector<shared_ptr<Blob<Dtype> > > m_, v_;
+};
+
+template <typename Dtype>
 Solver<Dtype>* GetSolver(const SolverParameter& param) {
   SolverParameter_SolverType type = param.solver_type();
 
@@ -171,6 +188,8 @@ Solver<Dtype>* GetSolver(const SolverParameter& param) {
       return new AdaGradSolver<Dtype>(param);
   case SolverParameter_SolverType_RMSPROP:
       return new RMSPropSolver<Dtype>(param);
+  case SolverParameter_SolverType_ADAM:
+      return new AdamSolver<Dtype>(param);
   default:
       LOG(FATAL) << "Unknown SolverType: " << type;
   }
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -98,7 +98,7 @@ message NetParameter {
 // NOTE
 // Update the next available ID when you add a new SolverParameter field.
 //
-// SolverParameter next available ID: 39 (last added: rms_decay)
+// SolverParameter next available ID: 40 (last added: momentum2)
 message SolverParameter {
   //////////////////////////////////////////////////////////////////////////////
   // Specifying the train and test networks
@@ -215,10 +215,13 @@ message SolverParameter {
     NESTEROV = 1;
     ADAGRAD = 2;
     RMSPROP = 3;
+    ADAM = 4;
   }
   optional SolverType solver_type = 30 [default = SGD];
-  // numerical stability for AdaGrad
+  // numerical stability for AdaGrad and Adam
   optional float delta = 31 [default = 1e-8];
+  // parameters for the Adam solver
+  optional float momentum2 = 39 [default = 0.999];
 
   // RMSProp decay value
   // MeanSquare(t) = rms_decay*MeanSquare(t-1) + (1-rms_decay)*SquareGradient(t)
--- a/src/caffe/solver.cpp
+++ b/src/caffe/solver.cpp
@@ -934,10 +934,144 @@ void RMSPropSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
   }
 }
 
+
+// template <typename Dtype>
+// void AdamSolver<Dtype>::SnapshotSolverStateToBinaryProto(
+//     const string& model_filename) {
+//   SolverState state;
+//   state.set_iter(this->iter_);
+//   state.set_learned_net(model_filename);
+//   state.set_current_step(this->current_step_);
+//   state.clear_history();
+//   state.clear_history_m();
+//   state.clear_history_v();
+//   for (int i = 0; i < SGDSolver<Dtype>::history_.size(); ++i) {
+//     // Add history
+//     BlobProto* history_blob = state.add_history();
+//     SGDSolver<Dtype>::history_[i]->ToProto(history_blob);
+
+//     BlobProto* history_blob_m = state.add_history_m();
+//     m_[i]->ToProto(history_blob_m);
+
+//     BlobProto* history_blob_v = state.add_history_v();
+//     v_[i]->ToProto(history_blob_v);
+//   }
+//   string snapshot_filename = Solver<Dtype>::SnapshotFilename(".solverstate");
+//   LOG(INFO)
+//     << "Snapshotting solver state to binary proto file" << snapshot_filename;
+//   WriteProtoToBinaryFile(state, snapshot_filename.c_str());
+// }
+
+
+template <typename Dtype>
+void AdamSolver<Dtype>::initAdam() {
+  // Initialize the history
+  const vector<Blob<Dtype>*>& net_params = this->net_->learnable_params();
+  m_.clear();
+  v_.clear();
+  for (int i = 0; i < net_params.size(); ++i) {
+    const vector<int>& shape = net_params[i]->shape();
+    m_.push_back(shared_ptr<Blob<Dtype> >(new Blob<Dtype>(shape)));
+    v_.push_back(shared_ptr<Blob<Dtype> >(new Blob<Dtype>(shape)));
+  }
+}
+
+template <typename Dtype>
+void AdamSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
+  const vector<float>& net_params_lr = this->net_->params_lr();
+  Dtype local_rate = rate * net_params_lr[param_id];
+  const Dtype beta1 = this->param_.momentum();
+  const Dtype beta2 = this->param_.momentum2();
+
+  // we create aliases for convenience
+  // shared_ptr<Blob<Dtype> > val_m = this->history_[param_id];
+  shared_ptr<Blob<Dtype> > val_m = this->m_[param_id];
+  // shared_ptr<Blob<Dtype> > val_v = this->update_[param_id];
+  shared_ptr<Blob<Dtype> > val_v = this->v_[param_id];
+  shared_ptr<Blob<Dtype> > val_t = this->temp_[param_id];
+
+  const int t = this->iter_  + 1;
+  const Dtype correction = std::sqrt(Dtype(1)-pow(beta2, t))/
+      (Dtype(1.)-pow(beta1, t));
+  const int N = net_params[param_id]->count();
+  const Dtype eps_hat = this->param_.delta();
+
+  switch (Caffe::mode()) {
+    case Caffe::CPU: {
+    // update m <- \beta_1 m_{t-1} + (1-\beta_1)g_t
+    caffe_cpu_axpby(N, Dtype(1)-beta1,
+        net_params[param_id]->cpu_diff(), beta1,
+        val_m->mutable_cpu_data());
+
+    // update v <- \beta_2 m_{t-1} + (1-\beta_2)g_t^2
+    caffe_mul(N,
+        net_params[param_id]->cpu_diff(),
+        net_params[param_id]->cpu_diff(),
+    val_t->mutable_cpu_data());
+    caffe_cpu_axpby(N, Dtype(1)-beta2,
+        val_t->cpu_data(), beta2,
+        val_v->mutable_cpu_data());
+
+    // set update
+    caffe_powx(N,
+        val_v->cpu_data(), Dtype(0.5),
+        val_t->mutable_cpu_data());
+    caffe_add_scalar(N, eps_hat, val_t->mutable_cpu_data());
+    caffe_div(N,
+        val_m->cpu_data(),
+        val_t->cpu_data(),
+        val_t->mutable_cpu_data());
+
+    caffe_cpu_scale(N, local_rate*correction,
+        val_t->cpu_data(),
+        net_params[param_id]->mutable_cpu_diff());
+    break;
+  }
+  case Caffe::GPU: {
+#ifndef CPU_ONLY
+    // update m <- \beta_1 m_{t-1} + (1-\beta_1)g_t
+    caffe_gpu_axpby(N, Dtype(1)-beta1,
+        net_params[param_id]->gpu_diff(), beta1,
+        val_m->mutable_gpu_data());
+
+    // update v <- \beta_2 m_{t-1} + (1-\beta_2)g_t^2
+    caffe_gpu_mul(N,
+        net_params[param_id]->gpu_diff(),
+        net_params[param_id]->gpu_diff(),
+        val_t->mutable_gpu_data());
+    caffe_gpu_axpby(N, Dtype(1)-beta2,
+        val_t->gpu_data(), beta2,
+        val_v->mutable_gpu_data());
+
+    // set update
+    caffe_gpu_powx(N,
+        val_v->gpu_data(), Dtype(0.5),
+        val_t->mutable_gpu_data());
+    caffe_gpu_add_scalar(N, eps_hat,
+        val_t->mutable_gpu_data());
+    caffe_gpu_div(N,
+        val_m->gpu_data(),
+        val_t->gpu_data(),
+        val_t->mutable_gpu_data());
+
+    caffe_gpu_scale(N, local_rate*correction,
+        val_t->gpu_data(),
+        net_params[param_id]->mutable_gpu_diff());
+#else
+    NO_GPU;
+#endif
+    break;
+  }
+  default:
+    LOG(FATAL) << "Unknown caffe mode: " << Caffe::mode();
+  }
+}
+
 INSTANTIATE_CLASS(Solver);
 INSTANTIATE_CLASS(SGDSolver);
 INSTANTIATE_CLASS(NesterovSolver);
 INSTANTIATE_CLASS(AdaGradSolver);
 INSTANTIATE_CLASS(RMSPropSolver);
-
+INSTANTIATE_CLASS(AdamSolver);
 }  // namespace caffe
--- a/src/caffe/test/test_gradient_based_solver.cpp
+++ b/src/caffe/test/test_gradient_based_solver.cpp
@@ -236,6 +236,10 @@ class GradientBasedSolverTest : public MultiDeviceTest<TypeParam> {
     Blob<Dtype>& updated_bias = *(*updated_params)[1];
     updated_bias.ReshapeLike(bias);
 
+    const Dtype momentum2 = 0.999;
+    Dtype val_m = 0;
+    Dtype val_v = 0;
+
     for (int i = 0; i <= D; ++i) {
       // Compute the derivative with respect to the ith weight (i.e., the ith
       // element of the gradient).
@@ -289,6 +293,15 @@ class GradientBasedSolverTest : public MultiDeviceTest<TypeParam> {
             + grad * grad * (1 - rms_decay)) + delta_;
         }
         break;
+      case SolverParameter_SolverType_ADAM: {
+        const Dtype m_ = 0.0;
+        const Dtype v_ = 0.0;
+        val_m = (1-momentum)*grad + momentum*m_;
+        val_v = (1-momentum2)*grad*grad + momentum2*v_;
+        Dtype alpha_t = learning_rate*std::sqrt(1-momentum2)/(1-momentum);
+        update_value = alpha_t * val_m / (std::sqrt(val_v) + delta_);
+        break;
+      }
       default:
         LOG(FATAL) << "Unknown solver type: " << solver_type();
       }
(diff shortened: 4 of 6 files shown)


shelhamer:
@philkr could you review this if you have a chance?

src/caffe/proto/caffe.proto
```diff
@@ -191,10 +191,14 @@ message SolverParameter {
...
     NESTEROV = 1;
     ADAGRAD = 2;
+    ADAM = 3;
   }
   optional SolverType solver_type = 30 [default = SGD];
-  // numerical stability for AdaGrad
+  // numerical stability for AdaGrad and Adam
   optional float delta = 31 [default = 1e-8];
+  // parameters for the Adam solver
+  optional float beta1 = 37 [default = 0.9];
```
philkr: Why not use momentum here, to be consistent with other solvers?

src/caffe/proto/caffe.proto
```diff
@@ -191,10 +191,14 @@ message SolverParameter {
...
+  optional float beta2 = 38 [default = 0.999];
```
philkr: If beta1 is renamed, it might make sense to call it momentum2?

src/caffe/solver.cpp
```diff
@@ -409,9 +409,10 @@ Dtype SGDSolver<Dtype>::GetLearningRate() {
   } else if (lr_policy == "step") {
     this->current_step_ = this->iter_ / this->param_.stepsize();
     rate = this->param_.base_lr() *
-        pow(this->param_.gamma(), this->current_step_);
+        caffe_ipow(this->param_.gamma(), this->current_step_);
```
philkr: it seems to me that caffe_ipow is complete overkill. pow is called once per iteration and takes a fraction of a microsecond.

include/caffe/util/math_functions.hpp
```diff
@@ -12,6 +12,18 @@
 
 namespace caffe {
 
+template<typename Dtype>
+Dtype caffe_ipow(Dtype base, int exp) {
```
philkr: I'm not sure that caffe_ipow is needed for this PR. The power function is only used in very few locations, so that the performance difference probably does not matter (a pow call takes 0.1 microseconds on my machine).

shelhamer:
@PatWie thanks for the solver! All SGD solvers need gradient checks. See for instance the AdaGrad tests https://github.com/BVLC/caffe/blob/master/src/caffe/test/test_gradient_based_solver.cpp#L431-L483

src/caffe/solver.cpp
```diff
@@ -775,9 +776,125 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+
+template <typename Dtype>
+void AdamSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
+  const vector<float>& net_params_lr = this->net_->params_lr();
+  Dtype local_rate = rate * net_params_lr[param_id];
+  const Dtype beta1 = this->param_.beta1();
+  const Dtype beta2 = this->param_.beta2();
+
+  const int t = this->iter_ / this->param_.stepsize() +1;
```
philkr: Why divide by stepsize here?

PatWie:
@shelhamer Ah. i didn't realized that there is already a unit-test. I will add one of course.

src/caffe/solver.cpp
PatWie: I think *t* is the epoch rather than the iteration by the definition of Caffe.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,125 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+  const int N = net_params[param_id]->count();
+  const Dtype delta = this->param_.delta();
+
+  switch (Caffe::mode()) {
+  case Caffe::CPU: {
+    // update m <- \beta_1 m_{t-1} + (1-\beta_1)g_t
+    caffe_scal(N, Dtype(beta1), this->val_m_[param_id]->mutable_cpu_data());
+    caffe_cpu_axpby(N, Dtype(1-beta1),
+        net_params[param_id]->cpu_diff(), Dtype(0),
+        this->val_t_[param_id]->mutable_cpu_data());
```
philkr: I don't think this currently does the right thing (val_t_ = (1-beta) net_params + 0*val_t). You can combine the above two lines into one caffe_cpu_axpby using beta1 instead of 0 call.

include/caffe/util/math_functions.hpp
PatWie: I did a small external benchmark. 3sec vs 7sec for caffe_ipow vs std::pow is a difference in speed (multiple iterations). This binary exponenation is common in many (crypto.) programs for several reasons.

src/caffe/proto/caffe.proto
PatWie: Good idea!

src/caffe/solver.cpp
```diff
@@ -775,9 +776,125 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+    caffe_add(N,
+        this->val_t_[param_id]->cpu_data(),
+        this->val_m_[param_id]->cpu_data(),
+        this->val_m_[param_id]->mutable_cpu_data());
```
philkr: The three commands above can be written as a single caffe_cpu_axpby using beta1 instead of 0 and val_m_ instead of val_t_.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,125 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+    // TODO: test whether a simple caffe_mul is faster
+    caffe_powx(N,
+        net_params[param_id]->cpu_diff(), Dtype(2),
+        this->val_t_[param_id]->mutable_cpu_data());
+    caffe_scal(N, Dtype(1.)-beta2,
+      this->val_t_[param_id]->mutable_cpu_data());
+    caffe_add(N,
+        this->val_t_[param_id]->cpu_data(),
+        this->val_v_[param_id]->cpu_data(),
+        this->val_v_[param_id]->mutable_cpu_data());
```
philkr: This can again be combined into a single caffe_cpu_axpby call.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,125 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+
+    // set update
+    caffe_powx(N,
+              this->val_v_[param_id]->cpu_data(), Dtype(0.5),
+              this->val_t_[param_id]->mutable_cpu_data());
```
philkr: If you are concerned about performance, pow is a lot slower than sqrt, so there is some room for improvement here.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,125 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+    caffe_add_scalar(N, delta, this->val_t_[param_id]->mutable_cpu_data());
+    caffe_div(N,
+              this->val_m_[param_id]->cpu_data(),
+              this->val_t_[param_id]->cpu_data(),
+              this->val_t_[param_id]->mutable_cpu_data());
+    caffe_cpu_axpby(N, local_rate*step_correction,
```
philkr: Why use this instead of caffe_cpu_scale?

include/caffe/solver.hpp
```diff
@@ -129,6 +129,28 @@ class AdaGradSolver : public SGDSolver<Dtype> {
...
+ protected:
+  void initAdam();
+  virtual void ComputeUpdateValue(int param_id, Dtype rate);
+  void constructor_sanity_check() {
+    CHECK_EQ(0, this->param_.momentum())
+        << "Momentum cannot be used with Adam.";
+    CHECK_NE(this->param_.stepsize(), 0) << "Adam needs a non-zero stepsize.";
+  }
+
+  vector<shared_ptr<Blob<Dtype> > > val_v_, val_m_, val_t_;
```
philkr: You can reuse the history_ and temp_ if the SGD solver and just add a val_m_ (or history2_) here.

src/caffe/solver.cpp
PatWie: Ah I see. Blas is completely new for me. I will change this tomorrow at all places in my code.

include/caffe/util/math_functions.hpp
philkr: I do understand that in places where pow is called often, this matters. In this solver pow is not call that often though (at most 3 times per iteration and layer with parameters), so optimizing a call that takes 0.1 microseconds with one that takes 0.04 microseconds is complete overkill.

src/caffe/proto/caffe.proto
```diff
@@ -154,7 +154,7 @@ message SolverParameter {
   optional string lr_policy = 8; // The learning rate decay policy.
   optional float gamma = 9; // The parameter to compute the learning rate.
   optional float power = 10; // The parameter to compute the learning rate.
-  optional float momentum = 11; // The momentum value.
+  optional float momentum = 11 [default = 0.9]; // The momentum value.
```
philkr: @shelhamer does a default momentum of 0.9 break other things in caffe? @PatWie it might be easier to have that default be handled in the ADAM layer if desired.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,112 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+
+template <typename Dtype>
+void AdamSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
+  const vector<float>& net_params_lr = this->net_->params_lr();
+  Dtype local_rate = rate * net_params_lr[param_id];
+  const Dtype beta1 = this->param_.momentum();
+  const Dtype beta2 = this->param_.momentum2();
+
+  const int t = this->iter_ / this->param_.stepsize() +1;
```
philkr: I did look the ADAM algorithm again, and I still believe that this value should not be divided by stepsize. The way the weighting works for the normalization (correction) should be (1-beta1^iter_), as the weighting follows a geometric series...

src/caffe/proto/caffe.proto
PatWie: https://github.com/BVLC/caffe/blob/master/src/caffe/test/test_gradient_based_solver.cpp#L334
Default momentum is 0.0.

Nanne:
The example for the MNIST database is very confusing at the moment:

* I'd imagine it's clearer to use the recommended alpha/stepsize value from the paper (0.001) for the base_lr.
* It's not actually using the momentum and momentum2 params, while the defaults in the .proto do define the recommended values it might be clearer to be explicit about this in the prototxt.
* I am not sure why you'd use lr_policy: "step", especially with the given gamma (step updates the rate to be: base_lr * gamma ^ (floor(iter/stepsize)) ). Wouldn't it be clearer (and make more sense) to use fixed for the lr_policy?
* Power isn't used at all, it can be removed.

src/caffe/solver.cpp
Nanne: param_.stepsize() should not be used for this solver at all, it's usage is for the step lr_policy, using it here only makes it unclear what the alpha/stepsize from the paper is.

It does need the +1 to do the (1-beta1^t) correctly, so it should be:
const int t = this->iter_ + 1;

Nanne:
I think there's some confusion going on in the code due to the usage of the stepsize param. Which is made worse by using lr_policy "step" in the MNIST example. 

The alpha/stepsize from the paper should be set via the base_lr, and used with lr_policy: "fixed" as I don't see any recommendations for changing alpha during training. This way you can also get rid of gamma and power in the prototxt (the latter wasn't being used anyway).

The stepsize param should only be used together with lr_policy "step", and if we already set the alpha via base_lr then it is not needed at all. t can just be iter_ +1, as it's afaik not needed to compute the effective stepsize. This also removes the need for a check if stepsize > 0 in the header.

Moreover, it makes sense to change the MNIST example to use the recommended value from the paper for the base_lr (0.001) and explicitly set momentum and momentum2 to 0.9 and 0.999 respectively, rather than relying on the default values.

PatWie:
I applied all changes in memory usage, solver_mnist proto, `t` is now the current iteration instead of epoche. And there is a unit test. 
There are still some trivial things todo:
- [ ] update solver page in the wiki
- [ ] implement learning rate heuristic `1/sqrt(t)` for reproducing results from the paper

src/caffe/solver.cpp
```diff
@@ -775,9 +776,101 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+template <typename Dtype>
+void AdamSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
+  const vector<float>& net_params_lr = this->net_->params_lr();
+  Dtype local_rate = rate * net_params_lr[param_id];
+  const Dtype beta1 = this->param_.momentum();
+  const Dtype beta2 = this->param_.momentum2();
+
+  // we create alias for memory from the SGD for convenience
+  shared_ptr<Blob<Dtype> > &val_m = this->history_[param_id];
```
philkr: A reference to a shared pointer is never good idea, just copy the pointer.

philkr:
The solver looks good to me now. I haven't tested it though. I do think that caffe_ipow should be it's own PR if we really want to add it. Currently it just bloats this PR request and doesn't add any benefit (see https://en.wikipedia.org/wiki/Amdahl%27s_law).

src/caffe/solver.cpp
```diff
@@ -775,9 +776,101 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+  switch (Caffe::mode()) {
+  case Caffe::CPU: {
+    // update m <- \beta_1 m_{t-1} + (1-\beta_1)g_t
+    caffe_cpu_axpby(N, Dtype(1)-beta1,
+        net_params[param_id]->cpu_diff(), beta1,
+        val_m->mutable_cpu_data());
+
+    // update v <- \beta_2 m_{t-1} + (1-\beta_2)g_t^2
+    caffe_mul(N,
+      net_params[param_id]->cpu_diff(),
```
jeffdonahue: please fix indentation -- use 4 space indents when continuing from previous lines (https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Spaces_vs._Tabs)

src/caffe/solver.cpp
```diff
@@ -775,9 +776,101 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+  const Dtype beta2 = this->param_.momentum2();
+
+  // we create alias for memory from the SGD for convenience
+  shared_ptr<Blob<Dtype> > val_m = this->history_[param_id];
+  shared_ptr<Blob<Dtype> > val_v = this->update_[param_id];
+  shared_ptr<Blob<Dtype> > val_t = this->temp_[param_id];
+
+  const int t = this->iter_  + 1;
+  const Dtype step_correction = std::sqrt(Dtype(1)-pow(beta2, t))/
+  (Dtype(1.)-pow(beta1, t));
```
ronghanghu: please fix indent as @jeffdonahue mentioned.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,101 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+    // update m <- \beta_1 m_{t-1} + (1-\beta_1)g_t
+    caffe_gpu_axpby(N, Dtype(1)-beta1,
+        net_params[param_id]->gpu_diff(), beta1,
+        val_m->mutable_gpu_data());
+
+    // update v <- \beta_2 m_{t-1} + (1-\beta_2)g_t^2
+    caffe_gpu_mul(N,
+      net_params[param_id]->gpu_diff(),
+      net_params[param_id]->gpu_diff(),
+      val_t->mutable_gpu_data());
```
ronghanghu: fix indent

src/caffe/solver.cpp
```diff
@@ -775,9 +776,101 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+    caffe_gpu_axpby(N, Dtype(1)-beta2,
+        val_t->gpu_data(), beta2,
+        val_v->mutable_gpu_data());
+
+    // set update
+    caffe_gpu_powx(N,
+              val_v->gpu_data(), Dtype(0.5),
+              val_t->mutable_gpu_data());
+    caffe_gpu_add_scalar(N, eps_hat,
+      val_t->mutable_gpu_data());
```
ronghanghu: fix indent

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -483,6 +495,29 @@ TYPED_TEST(AdaGradSolverTest, TestLeastSquaresUpdateWithEverythingAccum) {
...
+TYPED_TEST_CASE(AdamSolverTest, TestDtypesAndDevices);
+TYPED_TEST(AdamSolverTest, TestAdamLeastSquares) {
+  typedef typename TypeParam::Dtype Dtype;
+  const Dtype kLearningRate = 1.0;
+  const Dtype kMomentum = 0.9;
+  // momentum2 is already set to 0.999
+  const Dtype kWeightDecay = 0.0;
+  this->TestLeastSquaresUpdate(kLearningRate, kWeightDecay, kMomentum);
+}
+
```
ronghanghu: Please add test cases to 4 more test cases: 

* `TestAdamLeastSquaresUpdateLROneTenth` (where you set `kLearningRate = 0.1`).
* `TestAdamLeastSquaresUpdateWithWeightDecay` (where you may set `kWeightDecay = 0.5`).
* `TestAdamLeastSquaresUpdateWithEverything` (where you may set `kWeightDecay = 0.1` and `kNumIters = 4` and loop for `kNumIters` times)
* `TestLeastSquaresUpdateWithEverythingAccum` (where you may set `kNumIters = 4` and `kIterSize = 2` and use `CheckAccumulation`)

You may take a look at those AdaGradSolverTest for details.

ronghanghu:
@PatWie Thank you for this great PR! I just added some comment on the code. Please fix indent to 4 space indents when continuing from previous lines, and add more test cases. After that, squash into one commit and I can merge.

ronghanghu:
#2836 and #2887 introduced new conflicts to be resolved.

src/caffe/solver.cpp
```diff
@@ -775,9 +776,101 @@ void AdaGradSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
   }
 }
 
+
+template <typename Dtype>
+void AdamSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
+  const vector<shared_ptr<Blob<Dtype> > >& net_params = this->net_->params();
```
ronghanghu: Use const vector<Blob<Dtype>*>& net_params = this->net_->learnable_params(); instead.

src/caffe/test/test_gradient_based_solver.cpp
ronghanghu: Note: #2836 and #2887 modifies solver as well as introduces new share param and snapshot tests. To be consistent, let's also add 4 new test cases:

    TestLeastSquaresUpdateWithEverythingShare
    TestLeastSquaresUpdateWithEverythingAccumShare
    TestSnapshot
    TestSnapshotShare

For TestSnapshot, you can take a look at TestSnapshot in SGDSolverTest as an example. For the other 3 shared cases, you just need to add `this->share_ = true;` to the corresponding test cases.

PatWie:
I rebased to the latest master branch commit and fixed up the conflicts with #2836 and #2866.

It seems to be difficult for me to write the other tests (although I am pretty sure, that the implementation is correct) without rewriting fairly large amount of code (mostly duplicating code). In addition, it is not clear what's the favored way and some details like, should the code prevent the usage of weight decay, regularization and other parameters. I think only the following tests make sense (checked the existing tests)

- [x] TestLeastSquaresSingle
- [ ] TestLeastSquaresMultiIter
- [ ] TestSnapshot
- [ ] TestSnapshotShare

One has to deal with the following issues:

- The [snapshot function](https://github.com/BVLC/caffe/blob/master/src/caffe/test/test_gradient_based_solver.cpp#L452) has to write/read 2 additional parameters (`m,v`) from the paper which depends on the actual iteration. Currently, it only supports the history-blob.
- The [test-implementation](https://github.com/BVLC/caffe/blob/master/src/caffe/test/test_gradient_based_solver.cpp#L202) has to save `m, v` for testing the derivative. This depends on the actual iteration.
- I misused the `history_, update_, temp_` members, since Adam does not really rely on them directly. The current commit use again its own members from the Adam-class. How to handle this? Using the initialised members from SDG or use members from the Adam-class?

Possible solutions would be:

- Add two additional fields to the [SolverState](https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto#L236-L240) blob and [add members](https://github.com/BVLC/caffe/blob/master/include/caffe/solver.hpp#L101) to SDGSolver, which would not be used by the other solvers

or 

- Copy and paste most of the SDGSolver class for Adam

This seems to be a serious design issue. Implementing the solver is fairly easy, but writing nearly the same code again and squeezing the code into the current testing class needs hacky solutions.

To refactore the unit test cases one can use the *curiously recurring template pattern* to put everything from the solvers into its derived classes and just refer to the members in the base class. But then again the testing method should use the solver as a template not just as a [member field](https://github.com/BVLC/caffe/blob/master/src/caffe/test/test_gradient_based_solver.cpp#L37)

Since, this needs profound changes in the code it would suggest to let a BVLC maintainer decide the next steps or how to rewrite the solver-interface w.r.t. to these issues.

I am glad to help, but don't want to rewrite mostly everything in reference to #2890.

ronghanghu:
@PatWie Thanks for your update!

I checked the math and am quite confident with your solver implementation. However, at this point the PR is still not working, because snapshot currently won't work for AdamSolver.

>In addition, it is not clear what's the favored way and some details like, should the code prevent the usage of weight decay, regularization and other parameters.

Since right now all the weight decay for AdamSolver is handled in `SGDSolver::Regularize`, these weight decay tests become less important, thought it is still good to have them. It is also good have accumulation tests to test whether the gradient accumulation is working correctly.

>The snapshot function has to write/read 2 additional parameters (m,v) from the paper which depends on the actual iteration. Currently, it only supports the history-blob.

The original design is to use `SGDSolver::history_` to store all history blobs to be saved and recovered during snapshot. The size of history vector does not have to be same as number of learnable parameters . In Adam solver, you can twice as many history blobs as number of learnable parameters, in order to store both m and v. This can be done in a `AdamPreSolve` function. You may take a look at corresponding implementation in AdaDelta solver #2782 (which also stores two history variables) for how this is done.

https://github.com/matthiasplappert/caffe/blob/adadelta/src/caffe/solver.cpp#L937-L947
https://github.com/matthiasplappert/caffe/blob/adadelta/src/caffe/solver.cpp#L975

>The test-implementation has to save m, v for testing the derivative. This depends on the actual iteration.

Test-implementation can have access to solver's history, you may also take a look at AdaDelta #2782 implementation: https://github.com/matthiasplappert/caffe/blob/adadelta/src/caffe/test/test_gradient_based_solver.cpp#L299-L307

>I misused the history_, update_, temp_ members, since Adam does not really rely on them directly. The current commit use again its own members from the Adam-class. How to handle this? Using the initialised members from SDG or use members from the Adam-class?

You should use `SGD::history_` for all history variables that need to be stored and recovered. See above.

>To refactore the unit test cases one can use the curiously recurring template pattern to put everything from the solvers into its derived classes and just refer to the members in the base class. But then again the testing method should use the solver as a template not just as a member field
Since, this needs profound changes in the code it would suggest to let a BVLC maintainer decide the next steps or how to rewrite the solver-interface w.r.t. to these issues. I am glad to help, but don't want to rewrite mostly everything in reference to #2890.

I'll handle #2890 after merging Adam and AdaDelta solvers, and I would like to refactor these solvers and look into curiously recurring template pattern.

For now to get this PR merged, let's simply put val_m and val_v into `SGD::history_`, and read them during test implementation as in #2782, and complete the other 3 tests you think is reasonable. This shouldn't involve too much change in your current code. After that, we can merge this and I can handle the rest in #2890.

jeffdonahue:
Rather than putting both vectors into `history_` (which I think would make the indexing pretty messy and unnatural), I'd recommend keeping the two vectors separate and overriding `SnapshotSolverState` to serialize both of them.

ronghanghu:
@PatWie after a private discussion with @jeffdonahue, I still believe using existing history (and expanding it) is easier.

You can create a reference variable like `val_m_vec = this->history_;` and `val_v_vec = this->history_ + net_params.size();` (either in `AdamSolver` or its member functions as you like), and just use `history_` as a storage buffer. In this way the current snapshot functions still work.

ronghanghu:
Another rebase is needed here after #2909 and #2903

ronghanghu:
carried on in #2918, @PatWie please take a look when you have time.

PR #2859, opened 2015-08-04
ImageData layer default batch size of 1, and check for zero batch size

philkr:
This PR fixes a bug in the ImageData layer when batch size is not specified.

**Current behavior**: Batch size is 0 and the top count is 0 and no data is allocated, which in turn causes an exception in any layer using that data.

**New behavior**: Default batch size is 1 (in prototxt) and a check makes sure that the batch size is always positive.

**Bonus content**: Checking that the first image in a batch exists when computing the top size.


shelhamer:
Thanks for the checks and fix (and whitespace cleanup).

PR #2861, opened 2015-08-04
Allowing netspec to have a phase

philkr:
This is a first attempt to allow netspec to interact with phases. It is admittedly a bit ugly, so feedback would be welcome.
Here is an example how to use this:
```python
from caffe import NetSpec, layers as L, params as P, phase, to_proto

ns = NetSpec()

A = L.ImageData(source="/tmp/test.txt", root_folder="/data/sintel_intrinsic/images/clean/", ntop=2, name='foo')
B = L.ImageData(source="/tmp/train.txt", root_folder="/data/sintel_intrinsic/images/clean/", ntop=2, name='bar')
ns.data, ns.ignore = phase(train=A, test=B)
print(ns.to_proto())

data, ignore = phase(train=A, test=B)
print( to_proto(data, ignore) )
```

and here is the expected output
```
layer {
  name: "foo"
  type: "ImageData"
  top: "data"
  top: "ignore"
  include {
    phase: TRAIN
  }
  image_data_param {
    source: "/tmp/test.txt"
    root_folder: "/data/sintel_intrinsic/images/clean/"
  }
}
layer {
  name: "bar"
  type: "ImageData"
  top: "data"
  top: "ignore"
  include {
    phase: TEST
  }
  image_data_param {
    source: "/tmp/train.txt"
    root_folder: "/data/sintel_intrinsic/images/clean/"
  }
}

layer {
  name: "foo"
  type: "ImageData"
  top: "ImageData1"
  top: "ImageData2"
  include {
    phase: TRAIN
  }
  image_data_param {
    source: "/tmp/test.txt"
    root_folder: "/data/sintel_intrinsic/images/clean/"
  }
}
layer {
  name: "bar"
  type: "ImageData"
  top: "ImageData1"
  top: "ImageData2"
  include {
    phase: TEST
  }
  image_data_param {
    source: "/tmp/train.txt"
    root_folder: "/data/sintel_intrinsic/images/clean/"
  }
}

```

### DIFF
--- a/python/caffe/__init__.py
+++ b/python/caffe/__init__.py
@@ -4,4 +4,4 @@ from .proto.caffe_pb2 import TRAIN, TEST
 from .classifier import Classifier
 from .detector import Detector
 from . import io
-from .net_spec import layers, params, NetSpec, to_proto
+from .net_spec import layers, params, NetSpec, to_proto, phase
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -51,7 +51,7 @@ def to_proto(*tops):
     for top in tops:
         top.fn._to_proto(layers, {}, autonames)
     net = caffe_pb2.NetParameter()
-    net.layer.extend(layers.values())
+    net.layer.extend([l for l in layers.values() if l is not None])
     return net
 
 
@@ -168,7 +168,7 @@ class NetSpec(object):
         for name, top in six.iteritems(self.tops):
             top.fn._to_proto(layers, names, autonames)
         net = caffe_pb2.NetParameter()
-        net.layer.extend(layers.values())
+        net.layer.extend([l for l in layers.values() if l is not None])
         return net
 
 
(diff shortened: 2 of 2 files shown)


longjon:
My general feeling wrt phase (and stage, and level) is that these are ways to inject programmability into protobuf which are not necessary giving net spec, and in fact I wouldn't mind obsoleting these options and simplifying `Net::Init` and layer code accordingly. (Scan issues if you want to see some bugs that have resulted from failing to account for phase in all the right places.)

But I'm happy to hear other opinions and I don't object to merging code that exposes functionality that is already present. It seems like there are two ways to go about this:

1. Implement the straightforward wrapping of protobuf syntax, e.g., `L.Layer(include=dict(phase=?.TRAIN))`
2. Try to come up with some natural syntax for this functionality in the net spec language.

This PR takes the second path (q.v. https://github.com/BVLC/caffe/pull/2959#issuecomment-133947599). The syntax represented here feels pretty natural to me, cf. `switch` or `case` statements from other languages, or (very roughly) `np.where` and related functions.

Given the tone of the PR message, I'm going to tag `sandbox` and not look really carefully into the implementation, but happy to take another look if there's clamoring for merge.

Another thing to consider (though it's a different direction) is reifying this selection into a layer (cf. `FilterLayer`, which is related but not the same).

bchu:
To chime in on an old issue, I'd very much be in favor of the first approach. It's quite simple and fits naturally into the mental model of how NetSpec works. So far I'm using a hacked together version of the first approach when I'm programmatically generating experiments with NetSpec.

PR #2865, opened 2015-08-05
Exposing layer top and bottom names to python

philkr:
This PR allows the python interface to access the top and bottom blob names, which makes the data flow a bit more explicit in python.

### DIFF
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -149,6 +149,18 @@ class Net {
   inline const vector<vector<Blob<Dtype>*> >& top_vecs() const {
     return top_vecs_;
   }
+  /// @brief returns the ids of the top blobs of layer i
+  inline const vector<int> & top_ids(int i) const {
+    CHECK_GE(i, 0) << "Invalid layer id";
+    CHECK_LT(i, top_id_vecs_.size()) << "Invalid layer id";
+    return top_id_vecs_[i];
+  }
+  /// @brief returns the ids of the bottom blobs of layer i
+  inline const vector<int> & bottom_ids(int i) const {
+    CHECK_GE(i, 0) << "Invalid layer id";
+    CHECK_LT(i, bottom_id_vecs_.size()) << "Invalid layer id";
+    return bottom_id_vecs_[i];
+  }
   inline const vector<vector<bool> >& bottom_need_backward() const {
     return bottom_need_backward_;
   }
--- a/python/caffe/pycaffe.py
+++ b/python/caffe/pycaffe.py
@@ -276,6 +276,22 @@ def _Net_batch(self, blobs):
                                                  padding])
         yield padded_batch
 
+
+class _Net_IdNameWrapper:
+    """
+    A simple wrapper that allows the ids propery to be accessed as a dict
+    indexed by names. Used for top and bottom names
+    """
+    def __init__(self, net, func):
+        self.net, self.func = net, func
+
+    def __getitem__(self, name):
+        # Map the layer name to id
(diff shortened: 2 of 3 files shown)


longjon:
Seems like a reasonable thing to expose to pycaffe. A couple questions:

1. Why expose these functions in C++ as well? Currently they're (obviously) not in use; if there's some refactoring to be done using them they might be worth having, but if not it seems we're exposing an unused interface.
2. Why create custom classes instead of using `OrderedDict`s like elsewhere? Right now this seems like an unnecessary (and less functional) mixing of styles.

philkr:
thanks for the feedback. As for the C++ exposition, I didn't see a way of exposing it to python without exposing something to C++, but I'm open to suggestions.
(2) I guess an ordered dict of lists could work, but I felt it might not be as easy as what is there. The main issue is that currently the top and bottom names are computed in the function ```Net<Dtype>::bottom_names```. In order to put them into a OrderedDict one would need to call this function for every layer and store the result somewhere in the python Net, which can get messy quickly. Please let me know if you have any suggestions on how to handle this without creating a mess.

include/caffe/net.hpp
```diff
@@ -149,6 +149,16 @@ class Net {
   inline const vector<vector<Blob<Dtype>*> >& top_vecs() const {
     return top_vecs_;
   }
+  /// @brief returns the ids of the top blobs of layer i
+  inline const vector<int> & top_ids(int i) const {
+    CHECK_LT(i, top_id_vecs_.size()) << "Invalid layer id";
```
longjon: We also need to `CHECK_GE(i, 0)`, yes?

python/caffe/pycaffe.py
```diff
@@ -276,6 +276,22 @@ def _Net_batch(self, blobs):
...
+class _Net_IdNameWrapper:
+    """
+    A simple wrapper that allows the ids propery to be accessed as a dict
+    indexed by names. Used for top and bottom names
+    """
+    def __init__(self, net, func):
+        self.net, self.func = net, func
+
+    def __getitem__(self, name):
+		# Map the layer name to id
```
longjon: No tabs please!

longjon:
The new patch looks pretty reasonable; comments as noted. Also:
* I'm still not sure that constructing an `OrderedDict`, as used by many of the other `Net` methods would be messier than the current approach; the code would be pretty much as it is now (perhaps in comprehension form), cached in `net` if needed per your comment above. Then you could (e.g.) iterate through `top_names` in order, which is a fairly straightforward thing to desire, but won't work right now.
* Please squash cleanups before merge.

python/caffe/pycaffe.py
philkr: tabs are the best!

philkr:
Not sure where to compute the ```OrderedDict``` to store it in net. I have been using this for a while now. I usually iterate through ```_layer_names``` and ```layers``` anyway and rarely directly through top and bottom names. But if you see an easy change to make it work I'm very open to changing it.

longjon:
Right, due to the awkwardness of the current method rewriting scheme, it would need to be done in a similar way to the other `Net` methods, i.e., you would compute the `OrderedDict` during the call. If that creates unacceptable per-call overhead, you could cache the result in the net object (e.g., by testing `hasattr`). Which maybe we should be doing on other calls as well, or maybe we should provide an `__init__` callback, or use some other scheme, but this is the way it is now...

longjon:
Okay, I'll go ahead and merge as is; while it would be nice to have `OrderedDict` functionality, this PR provides a useful interface and that functionality can be added in a future patch.

Thanks for this exposure @philkr!

PR #2867, opened 2015-08-06
RMSprop clean up and rebase

ronghanghu:
Rebased and adapted RMSprop implementation #1890 to the new solver interface #2518 and #1977. The original author is @erogol. Pulled against master instead of dev.

The RMSprop solver is based on G. Hinton's lecture (http://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf). Param gradients are divided by average root mean square of gradients in recent batches. It can be seen as a mini-batch version of using only the sign of gradients.

Update rule:

    MeanSquare(t) = rms_decay * MeanSquare(t-1) + (1 - rms_decay) * gradient(t)^2
    param_update(t) = gradient(t) / (sqrt(MeanSquare(t)) + delta)

Momentum is not supported for RMSprop solver, as in #1890.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -135,6 +135,29 @@ class AdaGradSolver : public SGDSolver<Dtype> {
   DISABLE_COPY_AND_ASSIGN(AdaGradSolver);
 };
 
+
+template <typename Dtype>
+class RMSPropSolver : public SGDSolver<Dtype> {
+ public:
+  explicit RMSPropSolver(const SolverParameter& param)
+      : SGDSolver<Dtype>(param) { constructor_sanity_check(); }
+  explicit RMSPropSolver(const string& param_file)
+      : SGDSolver<Dtype>(param_file) { constructor_sanity_check(); }
+
+ protected:
+  virtual void ComputeUpdateValue(int param_id, Dtype rate);
+  void constructor_sanity_check() {
+    CHECK_EQ(0, this->param_.momentum())
+        << "Momentum cannot be used with RMSProp.";
+    CHECK_GE(this->param_.rms_decay(), 0)
+        << "rms_decay should lie between 0 and 1.";
+    CHECK_LT(this->param_.rms_decay(), 1)
+        << "rms_decay should lie between 0 and 1.";
+  }
+
+  DISABLE_COPY_AND_ASSIGN(RMSPropSolver);
+};
+
 template <typename Dtype>
 Solver<Dtype>* GetSolver(const SolverParameter& param) {
   SolverParameter_SolverType type = param.solver_type();
@@ -146,6 +169,8 @@ Solver<Dtype>* GetSolver(const SolverParameter& param) {
       return new NesterovSolver<Dtype>(param);
   case SolverParameter_SolverType_ADAGRAD:
       return new AdaGradSolver<Dtype>(param);
+  case SolverParameter_SolverType_RMSPROP:
+      return new RMSPropSolver<Dtype>(param);
   default:
       LOG(FATAL) << "Unknown SolverType: " << type;
   }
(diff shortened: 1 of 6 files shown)


erogol:
thanks for handling this :)

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -521,7 +531,7 @@ TYPED_TEST(NesterovSolverTest, TestNesterovLeastSquaresUpdateWithMomentum) {
   const Dtype kMomentum = 0.5;
   const int kNumIters = 1;
   for (int i = 0; i <= kNumIters; ++i) {
-    this->TestLeastSquaresUpdate(kLearningRate, kWeightDecay, kMomentum, i);
+    this->TestLeastSquaresUpdate(kLearningRate, kWeightDecay, kMomentum, 0., i);
```
jeffdonahue: These should be declared as constants (e.g. `const Dtype kRMSDecay = 0`) like the other args to make the meaning clear.

include/caffe/solver.hpp
```diff
@@ -128,6 +128,29 @@ class AdaGradSolver : public SGDSolver<Dtype> {
   DISABLE_COPY_AND_ASSIGN(AdaGradSolver);
 };
 
+
+template <typename Dtype>
+class RMSpropSolver : public SGDSolver<Dtype> {
```
jeffdonahue: The capitalization (everywhere) should be `RMSProp` rather than `RMSprop`.  (I think a global find/replace should do the trick.)

jeffdonahue:
Thanks @erogol for the original work and thanks @ronghanghu for the rebase. This looks good except as noted above.

ronghanghu:
@jeffdonahue OK, I'll handle them. Thanks for the comments!

ronghanghu:
Fixed those issues. I expect this PR to be merged after #2856 and #2782.

jeffdonahue:
Cool, LGTM.  @ronghanghu feel free to merge whenever it's easiest for you, before or after the other two PRs.

ronghanghu:
Took a further rebase on #2866. Authorship preserved for @erogol.

Ready to merge.

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -867,10 +906,124 @@ TYPED_TEST(NesterovSolverTest, TestSnapshotShare) {
...
+  }
+}
+
+template <typename TypeParam>
+class RMSPropSolverTest : public GradientBasedSolverTest<TypeParam> {
+  typedef typename TypeParam::Dtype Dtype;
+
+ protected:
+  virtual void InitSolver(const SolverParameter& param) {
+    this->solver_.reset(new RMSPropSolver<Dtype>(param));
```
shelhamer: Could you set the RMS decay here, instead of introducing the decay argument to least squares and snapshotting tests? Since it is unique to this solver I think it is best handled here.

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -173,6 +174,9 @@ class GradientBasedSolverTest : public MultiDeviceTest<TypeParam> {
     if (momentum != 0) {
       proto << "momentum: " << momentum << " ";
     }
+    if (rms_decay != 0) {
```
shelhamer: I think this could be handled in `RMSPropSolverTest::InitSolver()` since this detail is particular to the RMSProp solver alone.

shelhamer:
@ronghanghu Sorry I didn't catch this earlier, but I have a suggestion for the RMS decay parameter in the tests. Instead of introducing another argument and setting it for every test, this param could be set by the RMSProp test class for encapsulation. Could you send a follow-up PR to make this change?

ronghanghu:
@shelhamer Yes, I can send another PR to do that.

PR #2870, opened 2015-08-06
Multi-GPU

shelhamer:
This is my packaging of #2114 for merge. I figured @cypof @thatguymike and company had made plenty of revisions and that I could help.

This PR is ready to use for data parallel training of networks but

- parallel IO is only coordinated for lmdb / leveldb through a `DataReader`
- it cannot be satisfactorily tested with the current design

which are resolved for merge by #2903.

- [x] **fix multi-GPU solver test failures** see https://github.com/BVLC/caffe/pull/2870#issuecomment-129047932
  - [x] ensure test data determinism #2887 
  - [x] make ParallelDataLayer interface ~~in place of `DataReader`~~ @ronghanghu 
  - [x] test single solver / multi-solver batch equivalence 
- [x] propagate the singleton state to threads without member variables
- [x] adopt persistent check left as TODO
- [x] reconcile with #2866
- [x] replace `CHECK(false)` with `LOG(FATAL)`
- [x] drop timing code
- [x] squash history to clear fixes and such

@cypof @thatguymike @longjon @jeffdonahue please take a look.
@cdoersch could you fire up your parallel training test again?

Reviews and testing by the community are welcome!

### DIFF
--- a/include/caffe/data_reader.hpp
+++ b/include/caffe/data_reader.hpp
@@ -1,0 +1,82 @@
+#ifndef CAFFE_DATA_READER_HPP_
+#define CAFFE_DATA_READER_HPP_
+
+#include <map>
+#include <string>
+#include <vector>
+
+#include "caffe/common.hpp"
+#include "caffe/internal_thread.hpp"
+#include "caffe/util/blocking_queue.hpp"
+#include "caffe/util/db.hpp"
+
+namespace caffe {
+
+/**
+ * @brief Reads data from a source to queues available to data layers.
+ * A single reading thread is created per source, even if multiple solvers
+ * are running in parallel, e.g. for multi-GPU training. This makes sure
+ * databases are read sequentially, and that each solver accesses a different
+ * subset of the database. Data is distributed to solvers in a round-robin
+ * way to keep parallel training deterministic.
+ */
+class DataReader {
+ public:
+  explicit DataReader(const LayerParameter& param);
+  ~DataReader();
+
+  inline BlockingQueue<Datum*>& free() const {
+    return queue_pair_->free_;
+  }
+  inline BlockingQueue<Datum*>& full() const {
+    return queue_pair_->full_;
+  }
+
+ protected:
+  // Queue pairs are shared between a body and its readers
+  class QueuePair {
+   public:
+    explicit QueuePair(int size);
+    ~QueuePair();
+
+    BlockingQueue<Datum*> free_;
+    BlockingQueue<Datum*> full_;
+
+  DISABLE_COPY_AND_ASSIGN(QueuePair);
+  };
+
+  // A single body is created per source
+  class Body : public InternalThread {
+   public:
+    explicit Body(const LayerParameter& param);
+    virtual ~Body();
+
+   protected:
+    void InternalThreadEntry();
+    void read_one(db::Cursor* cursor, QueuePair* qp);
+
+    const LayerParameter param_;
+    BlockingQueue<shared_ptr<QueuePair> > new_queue_pairs_;
+
+    friend class DataReader;
+
+  DISABLE_COPY_AND_ASSIGN(Body);
+  };
+
+  // A source is uniquely identified by its layer name + path, in case
+  // the same database is read from two different locations in the net.
+  static inline string source_key(const LayerParameter& param) {
+    return param.name() + ":" + param.data_param().source();
+  }
+
+  const shared_ptr<QueuePair> queue_pair_;
+  shared_ptr<Body> body_;
+
+  static map<const string, boost::weak_ptr<DataReader::Body> > bodies_;
+
+DISABLE_COPY_AND_ASSIGN(DataReader);
+};
+
+}  // namespace caffe
+
+#endif  // CAFFE_DATA_READER_HPP_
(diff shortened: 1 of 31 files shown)


thatguymike:
All of my quick sanity tests are passing.  Despite knowing that by default this is weak scaling, e.g. the specified batch size in the train_val.prototxt is multiplied by the number of GPUs you choose to run on, I forgot that when validating accuracy graphs.  I still fear that is going to bite users.

cypof:
OK, training works for me. The thread launch code is much better without the fields, that's great.

shelhamer:
Thanks for testing @thatguymike and @cypof. My short test worked so once we hear from @cdoersch about the ec2 test I think this is ready to merge.

src/caffe/solver.cpp
```diff
@@ -188,10 +205,20 @@ void Solver<Dtype>::Step(int iters) {
...
+    timer.Start();
+    timing.str("");
+    timing << "Timing ";
+    if (param().solver_mode() == SolverParameter_SolverMode_GPU) {
+      timing << "(device " << param().device_id() << ") ";
+    }
+    for (int i = 0; i < callbacks_.size(); ++i) {
+      callbacks_[i]->on_start(&timer, &timing);
+    }
     const bool display = param_.display() && iter_ % param_.display() == 0;
```
buaaliyi: must add 'timer.Start();' here to restart timer, or line 266 that timing for grads maybe incorrect.

cdoersch:
Training seems to be working fine on ec2.

src/caffe/solver.cpp
shelhamer: Added to line 224 before forward + backward.

shelhamer:
After discussion with @longjon we decided the timing code is too intrusive to bundle in this change. I have stripped it but archived the branch with timing at `shelhamer/caffe:time-multi_gpu`. It could be re-introduced in a future PR.

src/caffe/solver.cpp
```diff
@@ -211,7 +228,9 @@ void Solver<Dtype>::Step(int iters) {
       losses[idx] = loss;
     }
     if (display) {
-      LOG(INFO) << "Iteration " << iter_ << ", loss = " << smoothed_loss;
+      if (Caffe::root_solver()) {
```
jeffdonahue: probably a bit late to comment on this, and to me not necessary for merge, but these conditional `LOG(INFO)` calls could be made a bit more compact using `LOG_IF`s, e.g. `LOG_IF(INFO, Caffe::root_solver()) << "Iteration..."`

src/caffe/parallel.cpp
```diff
@@ -0,0 +1,439 @@
...
+static size_t total_size(const vector<shared_ptr<Blob<Dtype> > >& params) {
+  size_t size = 0;
+  for (int i = 0; i < params.size(); ++i)
+    size += params[i]->count();
+  return size;
+}
+
+template<typename Dtype>
+Params<Dtype>::Params(shared_ptr<Solver<Dtype> > root_solver)
+    : size_(total_size<Dtype>(root_solver->net()->params())),
```
jeffdonahue: This and the two other calls below should be replaced with `learnable_params` after #2866, I think? (I was debating whether the public `params()` method should just be removed, or if `params()` should just `return learnable_params_`, or...)

shelhamer: Agreed. Making the switch seems to have no effect though and I have the same test failures before and after.

shelhamer:
@cypof @thatguymike after updating this to fit #2866 the multi-GPU solver tests are failing. I tried to combine them without making use of the `Run` object for testing and could have made a mistake. I haven't figured out the issue yet so take a look if you can.

shelhamer:
I'm fairly positive this is a test artifact due to the random Gaussian targets. The multiple solvers can't reproduce random draws equivalent to the single solver sequence:

- The worker solvers don't inherit the root seed in the current test since the seed is set in the singleton, and not the solver param (test_gradient_based_solver.cpp:183). They make their own nondeterministic draws.
- If the test is rewritten to make the workers inherit the seed, they'll all draw the same targets and still be wrong w.r.t. to single solver equivalent.

The solution seems to be making the solver tests take fixed external data, such as the hdf5 data used in the `sample_data.h5` used in the `HDF5DataLayerTest`. I'll follow-up with a separate PR.

shelhamer:
This is now based on #2887 but the multi-GPU solver tests still fail.

cypof:
I was off yesterday, but looking at it now.

shelhamer:
Everyone see #2903 for the rigorously tested and passing multi-GPU branch. @ronghanghu has developed a parallel data layer solution.

ronghanghu:
@shelhamer I'll rebase #2903 shortly, and clean up.

ronghanghu:
Merged in #2903

include/caffe/data_reader.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace caffe {
+
+/**
+ * @brief Reads data from a source to queues available to data layers.
+ * A single reading thread is created per source, even if multiple solvers
+ * are running in parallel, e.g. for multi-GPU training. This makes sure
+ * databases are read sequentially, and that each solver accesses a different
+ * subset of the database. Data is distributed to solvers in a round-robin
+ * way to keep parallel training deterministic.
```
zxiaomzxm: > [``](url)

PR #2871, opened 2015-08-06
PythonLayer takes parameters by string

shelhamer:
Cherry-picking of the PythonLayer string parameter from #2001 by @tnarihi to give `PythonLayer` a `param_str` member for passing parameters through the `python_param` proto message.

This is a simple change in itself that leaves the choice of what and how to parameterize open.
There is no automatic `eval`

### DIFF
--- a/python/caffe/test/test_python_layer_with_param_str.py
+++ b/python/caffe/test/test_python_layer_with_param_str.py
@@ -1,0 +1,59 @@
+import unittest
+import tempfile
+import os
+import six
+
+import caffe
+
+
+class SimpleParamLayer(caffe.Layer):
+    """A layer that just multiplies by the numeric value of its param string"""
+
+    def setup(self, bottom, top):
+        try:
+            self.value = float(self.param_str)
+        except ValueError:
+            raise ValueError("Parameter string must be a legible float")
+
+    def reshape(self, bottom, top):
+        top[0].reshape(*bottom[0].data.shape)
+
+    def forward(self, bottom, top):
+        top[0].data[...] = self.value * bottom[0].data
+
+    def backward(self, top, propagate_down, bottom):
+        bottom[0].diff[...] = self.value * top[0].diff
+
+
+def python_param_net_file():
+    with tempfile.NamedTemporaryFile(mode='w+', delete=False) as f:
+        f.write("""name: 'pythonnet' force_backward: true
+        input: 'data' input_shape { dim: 10 dim: 9 dim: 8 }
+        layer { type: 'Python' name: 'mul10' bottom: 'data' top: 'mul10'
+          python_param { module: 'test_python_layer_with_param_str'
+                layer: 'SimpleParamLayer' param_str: '10' } }
+        layer { type: 'Python' name: 'mul2' bottom: 'mul10' top: 'mul2'
+          python_param { module: 'test_python_layer_with_param_str'
+                layer: 'SimpleParamLayer' param_str: '2' } }""")
+        return f.name
+
+
+class TestLayerWithParam(unittest.TestCase):
+    def setUp(self):
+        net_file = python_param_net_file()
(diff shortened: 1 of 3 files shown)


shelhamer:
@philkr have you adapted this test to Python 3 already?

python/caffe/test/test_python_layer_with_param_str.py
```diff
@@ -0,0 +1,85 @@
...
+        top[0].reshape(bottom[0].num, bottom[0].channels, bottom[0].height,
+                bottom[0].width)
+    def forward(self, bottom, top):
+        self._forward(bottom, top)
+    def backward(self, top, propagate_down, bottom):
+        self._backward(bottom, propagate_down, top)
+
+
+def python_net_file():
+    f = tempfile.NamedTemporaryFile(delete=False)
```
philkr: ```NamedTemporaryFile``` is created with ```mode='w+b'``` by default, which in python3 is a buffer, not a string. To fix this just call ```NamedTemporaryFile(mode='w+', delete=False)```.

philkr:
The python3 error is easy to fix (see above). I have no idea why python2 fails on those warnings...

tnarihi:
@shelhamer @philkr Thank you so much for taking care of my previous PR! Other than this, I still have a bunch of PRs for python that are not finished. If I need to update and polish them, please let me know.

tnarihi:
According to [this post in Stack Overflow](http://stackoverflow.com/questions/10056393/g-with-python-h-how-to-compile), maybe we have to Python.h include first, which means we have to put include of `boost/python.hpp` first in `tools/caffe.cpp`. But I am not sure why this warning suddenly came out in this PR.

philkr:
I think ac6d4b67c2307cae58a98077982d662773492b7c broke it. It seems to be broken for any PR based on the current master. It seems ac6d4b67c2307cae58a98077982d662773492b7c introduced an additional include in caffe.cpp.

shelhamer:
Hmm, the branch CI had passed but this is an issue. I'll try to fix it shortly.

shelhamer:
@jeffdonahue @longjon does anybody have a favorite name or is `param_str_` fine?

jeffdonahue:
I think `param_str` is fine.  It could also just be called `param` or `arg` since the `str` is a little redundant with the protobuf type, but I could also see how it might be a useful hint to have the `str` annotation when parsing in Python.

shelhamer:
I changed the name of the property to `param_str` on the Python side and simplified the test to just check multiplication by the `param_str` value. @longjon take a look and merge if you like it.

python/caffe/test/test_python_layer_with_param_str.py
```diff
@@ -0,0 +1,59 @@
...
+        bottom[0].diff[...] = self.value * top[0].diff
+
+
+def python_param_net_file():
+    with tempfile.NamedTemporaryFile(mode='w+', delete=False) as f:
+        f.write("""name: 'pythonnet' force_backward: true
+        input: 'data' input_shape { dim: 10 dim: 9 dim: 8 }
+        layer { type: 'Python' name: 'mul10' bottom: 'data' top: 'mul10'
+          python_param { module: 'test_python_layer_with_param_str'
+                layer: 'SimpleParamLayer' param_str: "10" } }
```
longjon: Could we be consistent with single-quotes here?

src/caffe/proto/caffe.proto
```diff
@@ -703,6 +703,12 @@ message PowerParameter {
 message PythonParameter {
   optional string module = 1;
   optional string layer = 2;
+  // This value is set to the attribute `param_str_` of your custom
```
longjon: This comment is out-of-date.

longjon:
Looks good except as noted.

shelhamer:
Thanks for the glance @longjon. Fixed and merging.

PR #2873, opened 2015-08-06
Fix gist download model and binary scripts

gut:
Please read the commit text.

### DIFF
--- a/scripts/download_model_from_gist.sh
+++ b/scripts/download_model_from_gist.sh
@@ -18,7 +18,7 @@ fi
 
 echo "Downloading Caffe model info to $MODEL_DIR ..."
 mkdir -p $MODEL_DIR
-wget https://gist.github.com/$GIST/download -O $MODEL_DIR/gist.tar.gz
-tar xzf $MODEL_DIR/gist.tar.gz --directory=$MODEL_DIR --strip-components=1
-rm $MODEL_DIR/gist.tar.gz
+wget https://gist.github.com/$GIST/download -O $MODEL_DIR/gist.zip
+unzip -j $MODEL_DIR/gist.zip -d $MODEL_DIR
+rm $MODEL_DIR/gist.zip
 echo "Done"
--- a/scripts/download_model_binary.py
+++ b/scripts/download_model_binary.py
@@ -32,7 +32,7 @@ def parse_readme_frontmatter(dirname):
     with open(readme_filename) as f:
         lines = [line.strip() for line in f.readlines()]
     top = lines.index('---')
-    bottom = lines[top + 1:].index('---')
+    bottom = lines.index('---', top + 1)
     frontmatter = yaml.load('\n'.join(lines[top + 1:bottom]))
     assert all(key in frontmatter for key in required_keys)
     return dirname, frontmatter


gut:
well, I guess that Travis was failing before my patch as it has nothing to do with the build anyway

scripts/download_model_from_gist.sh
```diff
@@ -18,7 +18,7 @@ fi
 
 echo "Downloading Caffe model info to $MODEL_DIR ..."
 mkdir -p $MODEL_DIR
-wget https://gist.github.com/$GIST/download -O $MODEL_DIR/gist.tar.gz
-tar xzf $MODEL_DIR/gist.tar.gz --directory=$MODEL_DIR --strip-components=1
-rm $MODEL_DIR/gist.tar.gz
+wget https://gist.github.com/$GIST/download -O $MODEL_DIR/gist.zip
```
shelhamer: This isn't needed -- although the interface has a button for "download as ZIP" gists can be downloaded in either format by requesting the desired extension.

gut: The script is intended to avoid the annoyance of going to the website. 
Anyway, if that is the decision then this script should be at least removed. Don't forget to update the wiki pages :-)

shelhamer: Sorry, I don't follow. There's no need to go to the site as the current script downloads the proper tar gz?

gut: when `https://gist.github.com/$GIST/download` is downloaded, you have a zip, not a tarball, hence the need for this patch.
If there is another URL to grab a tarball, then I agree that it'd be better to fix the URL and stick with the old tarball extraction.

shelhamer:
I double-checked that github is actually serving zip files from the download url we're using and apparently tar is just recognizing the zip and unpacking it properly on our platforms. There is a `.tar.gz` url for the full gist hash but not the abbreviation used in the current script.

Although this works as-is (at least where i've tried it) it seems better to accept this patch for clarity.

gut:
Well, it didn't work for me. That's why I did this patch. This is the output without my patch locally:
```
$ ./scripts/download_model_from_gist.sh ksimonyan/fd8800eeb36e276cd6f9  CNN_S
Downloading Caffe model info to CNN_S/ksimonyan-fd8800eeb36e276cd6f9 ...
--2015-08-10 17:36:52--  https://gist.github.com/ksimonyan/fd8800eeb36e276cd6f9/download
Resolving gist.github.com (gist.github.com)... 192.30.252.141
Connecting to gist.github.com (gist.github.com)|192.30.252.141|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://codeload.github.com/ksimonyan/fd8800eeb36e276cd6f9/zip/master [following]
--2015-08-10 17:36:52--  https://codeload.github.com/ksimonyan/fd8800eeb36e276cd6f9/zip/master
Resolving codeload.github.com (codeload.github.com)... 192.30.252.146
Connecting to codeload.github.com (codeload.github.com)|192.30.252.146|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 2022 (2.0K) [application/zip]
Saving to: ‘CNN_S/ksimonyan-fd8800eeb36e276cd6f9/gist.tar.gz’

100%[=========================================================================================================================================================>] 2,022       --.-K/s   in 0s

2015-08-10 17:36:53 (37.4 MB/s) - ‘CNN_S/ksimonyan-fd8800eeb36e276cd6f9/gist.tar.gz’ saved [2022/2022]

gzip: stdin has more than one entry--rest ignored
tar: Child returned status 2
tar: Error is not recoverable: exiting now
Done
```

PR #2877, opened 2015-08-06
[pycaffe] add Blob shape accessor

jeffdonahue:
This adds a direct `shape` accessor for PyCaffe Blobs.  Currently to get the shape one has to do `myblob.data.shape` or `myblob.diff.shape` which can cause a (possibly unnecessary) allocation+copy to host memory.


shelhamer:
Looks good to me.

longjon:
This has actually been on my agenda for a while, so... thanks!

I guess you get a `boost::python` vector wrapper, which is not quite as good as a tuple, but that could be adjusted in a future PR if desired.

PR #2884, opened 2015-08-07
add [] to "delete pixels".

tianzhi0549:
see https://isocpp.org/wiki/faq/freestore-mgmt#delete-array-built-ins.


ronghanghu:
Thanks for the fix!

Issue #2885, opened 2015-08-08
HDF5 snapshot crashes on duplicate layer names


jeffdonahue:
With the new HDF5 snapshotting (#2836), if two or more layers of the net have the same name, Caffe crashes when it tries to snapshot. (This includes the case of two layers with unspecified `name`s, which are interpreted by Caffe via the protobuf default as two layers `name`d `""`.)  Maybe Caffe shouldn't allow layers to have the same name (or unspecified names), but it should probably be checked in `Net::Init` if that's going to be the rule.  We might want to temporarily switch the default serialization back to proto until we figure out what to do about this.

An example of the error message that comes with the crash is below (was training [this model](https://github.com/jeffdonahue/caffe/blob/recurrent-rebase-cleanup/examples/coco_caption/lrcn.prototxt), which has two layers named `data`):

```
I0807 19:34:51.663398 23450 solver.cpp:375] Snapshotting to HDF5 file ./examples/coco_caption/lrcn_iter_5000.caffemodel.h5
HDF5-DIAG: Error detected in HDF5 (1.8.13) thread 139730974058368:
  #000: H5G.c line 312 in H5Gcreate2(): unable to create group
    major: Symbol table
    minor: Unable to initialize object
  #001: H5Gint.c line 194 in H5G__create_named(): unable to create and link to group
    major: Symbol table
    minor: Unable to initialize object
  #002: H5L.c line 1638 in H5L_link_object(): unable to create new link to object
    major: Links
    minor: Unable to initialize object
  #003: H5L.c line 1882 in H5L_create_real(): can't insert link
    major: Symbol table
    minor: Unable to insert object
  #004: H5Gtraverse.c line 861 in H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #005: H5Gtraverse.c line 641 in H5G_traverse_real(): traversal operator failed
    major: Symbol table
    minor: Callback failed
  #006: H5L.c line 1674 in H5L_link_cb(): name already exists
    major: Symbol table
    minor: Object already exists
F0807 19:34:51.663725 23450 net.cpp:871] Check failed: layer_data_hid >= 0 (-1 vs. 0) Error saving weights to ./examples/coco_caption/lrcn_iter_5000.caffemodel.h5.
*** Check failure stack trace: ***
    @     0x7f15a65c3daa  (unknown)
    @     0x7f15a65c3ce4  (unknown)
    @     0x7f15a65c36e6  (unknown)
[...]

PR #2886, opened 2015-08-08
temporarily switch the snapshot_format default back to BINARYPROTO

jeffdonahue:
See #2885 -- Caffe will crash when trying to snapshot nets with duplicate layer names (including nets with multiple unnamed layers).  This is probably going to surprise a lot of users and doesn't have a good error message, so I'm suggesting as a quick patch that we switch the default back until these issues are discussed.

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -179,7 +179,7 @@ message SolverParameter {
     HDF5 = 0;
     BINARYPROTO = 1;
   }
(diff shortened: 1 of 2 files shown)


jeffdonahue:
I'm going to merge after discussion with @shelhamer.  The ["push" Travis build](https://travis-ci.org/BVLC/caffe/builds/74663268) is failing for some reason unrelated to the code (after many retries), but the PR build passes.

PR #2887, opened 2015-08-08
Test solvers on fixed hdf5 data

shelhamer:
For simplicity and determinism the solver tests are given fixed hdf5 input.

The current arrangement of using random dummy data introduces subtleties in the order of execution that add complexity and test artifacts.
In particular, it makes it awkward to test accumulation and multi-GPU solving.


jeffdonahue:
Cool, LGTM.  Feel free to merge as makes sense with #2870.

shelhamer:
Merging as I think this is an improvement in itself and reduces the RNG confusion. The multi-GPU tests will be handled by #2870.

PR #2891, opened 2015-08-09
fix examples/pycaffe/caffenet.py import

fyu:
\_\_future\_\_ imports needs to occur at the beginning of the file in python2


longjon:
Thanks! We should really have Travis running this file (not to mention other examples)...

fyu:
You are welcome. I can confirm that this example can finish successfully with this change.

PR #2892, opened 2015-08-09
Rebase and Clean up Hdf5DataLayer Prefetch

ronghanghu:
Cleaned up #2271 to adapt to #2836. Original authors are @jeffdonahue and @pclove1

@shelhamer please review when you have time.


shelhamer:
It might be best if @erictzeng takes a look at this given his recent hdf5 work.

ronghanghu:
@jeffdonahue If I understand correctly, shuffling is still preserved in `FillHDF5FileData`

https://github.com/ronghanghu/caffe/blob/hdf5-prefetch/src/caffe/layers/hdf5_data_layer.cpp#L61-L64

shelhamer:
This still shuffles the order of hdf5 files but no longer shuffles rows within hdf5. Generally rows >> files so shuffling is limited by this change. This is no different than lmdb / leveldb however.

ronghanghu:
@shelhamer OK, I see. This may be a serious drawback of this PR.

jeffdonahue:
Another disadvantage of the PR (which I didn't realize until I started using HDF5DataLayer myself after writing the initial version of this PR) is the optimization in the case of the entire dataset being a single HDF5 file -- with the current HDF5DataLayer, the entire file is loaded into memory initially and nothing is ever read from disk again.  I'm not sure an HDF5 prefetching PR should be merged until the optimization of this important special case is somehow brought back.

PR #2897, opened 2015-08-10
Update net_spec.py

Russell91:


jeffdonahue:
Thanks for the fix @Russell91!

PR #2899, opened 2015-08-10
fix for layer visualization notebook link

ajkl:


shelhamer:
Thanks for trying to help with the document -- #2951 set the correct link.

PR #2917, opened 2015-08-14
fixing the database param

ajkl:
The example talks about LevelDB as the db backend but has lmdb as the param in the execution.


jeffdonahue:
Thanks @ajkl.  We should probably change all the uses of LevelDB in the docs to LMDB as it's less of a hassle to work with (feel free to send PRs), but this is certainly an improvement over the existing inconsistency in this case.

Issue #2890, opened 2015-08-09
Separate solvers by string type and separate source files


shelhamer:
The `solver_type` field is a proto enum causing conflicts for concurrent development. This should be made a string as #1694 did for layer type.

The solvers are currently defined and implemented in monolithic hpp and cpp that make extension awkward. Each should be split into its own header and source.

After #2860 #2782  #2856

Issue #2860, opened 2015-08-04
Adaptive Solvers: AdaDelta, RMSprop, and ADAM


shelhamer:
There are open PRs for adaptive solvers that need review, revision, and merge:

- [x] RMSprop #1890: finished and ready for review in #2867 
- [x] AdaDelta #2782: need to be aligned with the new solver interface #2518 #1977
- [x] ADAM #2856: finished and ready for review in #2918 

As a last step AdaGrad, AdaDelta, and RMSprop could likely be improved by refactoring out shared code into an `AdaptiveSGDSolver`

@ronghanghu these could be good warm-up projects for development.

@longjon @jeffdonahue thoughts on posting tracking issues of this kind? Does this help navigate or is it noise?

ronghanghu:
OK, I'll go through these PRs.

matthiasplappert:
AdaDelta should already conform to the solver interface. Unless I missed something, I ported it over in https://github.com/BVLC/caffe/pull/2782 when merging in master a couple of weeks ago.

ronghanghu:
Since right now these 3 PRs are blocking each other by enum value in SolverType and SolverParameter next available ID, I expect to merge Adam or AdaDelta first, and RMSprop afterwards. I can take over and clean up #2782 or #2856 when necessary. After that, a solver refactor PR can follow addressing the `solver_type` and split `solver.cpp` into different files.

ronghanghu:
Please add one more test case:

* `TestLeastSquaresUpdateWithEverythingAccum` (where you may set kNumIters = 4 and kIterSize = 2 and use CheckAccumulation)

You may take a look at those AdaGradSolverTest for details.

ronghanghu:
Note: #2836 modifies solver as well as introduces new snapshot tests, so a further rebase is needed here. #2870 also introduces new conflicts.

bhack:
@ronghanghu The merging sequence It is becoming to be a "political" issue because there isn't a clear policy for reviews and merges. Fast track merges on younger PR push adaptation work over the older one (apart of the quality, the in shape status, the debasing effort of the older PR)

ronghanghu:
@bhack Indeed, there isn't a clear policy for PR reviewing process right now. Our merge sequence is mainly based on the status of PRs, that is, whether a PR is ready to merge or needs further modification. Since these 2 out of 3 here (Adam and AdaDelta) are still not ready, I expect to wait to merge therm after #2870.

bhack:
[...]

PR #2918, opened 2015-08-14
Adam solver

ronghanghu:
Carried on Adam solver (originally #2856) for merge.

I completed the tests and rebased it to latest master.

Authorship belongs to @PatWie, and is preserved in git commit.

Original message in #2856 :
>This commit implements the Adam solver by Kingma et. al for CPU and GPU. All solver parameters are defined in the caffe.proto. This also adds an example for the MNIST dataset.

As you may see, now both solver.cpp and test_gradient_based_solver.cpp are growing to 1000+ lines. This problem will be addressed in #2890.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -218,6 +218,21 @@ class AdaDeltaSolver : public SGDSolver<Dtype> {
 };
 
 template <typename Dtype>
+class AdamSolver : public SGDSolver<Dtype> {
+ public:
+  explicit AdamSolver(const SolverParameter& param)
+      : SGDSolver<Dtype>(param) { AdamPreSolve();}
+  explicit AdamSolver(const string& param_file)
+      : SGDSolver<Dtype>(param_file) { AdamPreSolve(); }
+
+ protected:
+  void AdamPreSolve();
+  virtual void ComputeUpdateValue(int param_id, Dtype rate);
+
+  DISABLE_COPY_AND_ASSIGN(AdamSolver);
+};
+
+template <typename Dtype>
 Solver<Dtype>* GetSolver(const SolverParameter& param) {
   SolverParameter_SolverType type = param.solver_type();
 
@@ -232,6 +247,8 @@ Solver<Dtype>* GetSolver(const SolverParameter& param) {
       return new RMSPropSolver<Dtype>(param);
   case SolverParameter_SolverType_ADADELTA:
       return new AdaDeltaSolver<Dtype>(param);
+  case SolverParameter_SolverType_ADAM:
+      return new AdamSolver<Dtype>(param);
   default:
       LOG(FATAL) << "Unknown SolverType: " << type;
   }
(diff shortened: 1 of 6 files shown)


ronghanghu:
Encountered error larger than margin on solver tests. Doesn't seem to be numerical issue. Looking into details.

ronghanghu:
@shelhamer @jeffdonahue @philkr @PatWie Please take a look if you have time. I think this should be ready to merge.

This is last piece of the solver trilogy in #2860. After merging this one, we can address #2890.

include/caffe/solver.hpp
```diff
@@ -218,6 +218,21 @@ class AdaDeltaSolver : public SGDSolver<Dtype> {
 };
```
jeffdonahue: We need to cite the ADAM paper somewhere*.  I suggest putting a reference here, e.g. in a doxygen formatted comment like [this](https://github.com/BVLC/caffe/pull/2033/files#diff-2b4d88968d7da8d11c91d62d40aea3b9R152).  Eventually it would also be good to add sections to the [solver tutorial](http://caffe.berkeleyvision.org/tutorial/solver.html) on these new solvers, where the reference should also then be added.

*We probably also need to go back and add references for some of the other recently merged solvers.

jeffdonahue:
Thanks for the rebase @ronghanghu and thanks @PatWie for the original implementation!  See above comment; otherwise looks good.

ronghanghu:
Citation added for Adam.
>Eventually it would also be good to add sections to the solver tutorial on these new solvers, where the reference should also then be added.
*We probably also need to go back and add references for some of the other recently merged solvers.

Let's address that in #2890 .

src/caffe/solver.cpp
```diff
@@ -1114,11 +1114,116 @@ void AdaDeltaSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
...
+void AdamSolver<Dtype>::ComputeUpdateValue(int param_id, Dtype rate) {
+  const vector<Blob<Dtype>*>& net_params = this->net_->learnable_params();
+  const vector<float>& net_params_lr = this->net_->params_lr();
+  Dtype local_rate = rate * net_params_lr[param_id];
+  const Dtype beta1 = this->param_.momentum();
+  const Dtype beta2 = this->param_.momentum2();
+
+  // we create aliases for convenience
+  size_t update_history_offset = net_params.size();
+  shared_ptr<Blob<Dtype> > val_m = this->history_[param_id];
```
jeffdonahue: No need to create `shared_ptr`s for these `val_*` variables, is there? (I suggest using the raw pointer, e.g. `Blob<Dtype>* val_m = this->history_[param_id].get();`.)

jeffdonahue:
Thanks for adding the citation.  After a final glance I noticed the one other thing I commented above; sorry about not noticing before.  Feel free to merge after addressing that.

PatWie:
Looks good.

src/caffe/solver.cpp
ronghanghu: Yes, you are right. I should use raw pointers.

ronghanghu:
Changed from shared ptrs to raw ptrs in `AdamSolver<Dtype>::ComputeUpdateValue`.

PR #2920, opened 2015-08-14
information about new implemented solvers

PatWie:
This updates the information about implemented solvers of Caffe w.r.t.  #2782, #2918, #2867.


ronghanghu:
@PatWie Thanks for the information!

PR #2921, opened 2015-08-14
Destroy CUDA stream when finished

buaaliyi:
Add missing cudaStreamDestroy in BasePrefetchingDataLayer::InternalThreadEntry. cudaStreamCreate and cudaStreamDestroy function call always come in pairs.


ronghanghu:
@cypof I am not an expert in CUDA myself, but I believe you did forgot to call `cudaStreamDestroy` in #2903 (originally #2114).

ronghanghu:
A simple workaround provided in #2924. Debates can follow in new issues on whether allowing training without learnable parameter.

cypof:
Ah yes, that's correct. Good catch.

PR #2924, opened 2015-08-14
Malloc at least one byte in Parallel

### LINKED ISSUE #2922: Bug in newly added multi-gpu code
Trying out simple networks with multi-gpu segfaults because they don't have learnable parameters:

In parallel.c these lines are critical:

    CUDA_CHECK(cudaMalloc(&data_, size_ * sizeof(Dtype)));
    [...]
    CUDA_CHECK(cudaMalloc(&diff_, size_ * sizeof(Dtype)));
    [...]
    CUDA_CHECK(cudaMalloc(&parent_grads_, size_ * sizeof(Dtype)));

If the net does not have learnable parameters, `size_` will be 0 and `cudaMalloc` will return null pointers.

I currently work around this by adding +1 byte to the allocated size, but there should be a better fix.

ronghanghu:
This seems to be a simple workaround to fix #2922.

I'll merge once Travis CI pass.

### DIFF
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -64,7 +64,9 @@ static size_t total_size(const vector<Blob<Dtype>*>& params) {
   size_t size = 0;
   for (int i = 0; i < params.size(); ++i)
     size += params[i]->count();
-  return size;
+  // Size have at least one byte, otherwise cudaMalloc fails if net has no
+  // learnable parameters.
+  return (size > 0) ? size : 1;
 }
 
 template<typename Dtype>


thatguymike:
Hrmm, probably better to actually look at the root issue.  Seems like on the mallocs we should check layer state.  Although this patch will workaround the issue, we should probably have a more general debate on what the behavior should even be training with no learnable params.

ronghanghu:
>Although this patch will workaround the issue, we should probably have a more general debate on what the behavior should even be training with no learnable params.

I agree. It doesn't make sense to me to train without learnable params (maybe someone just want to see the loss of a network? still doesn't quite make sense to me..). Perhaps train without learnable params should be banned in the future. Since it is a simple 3 line workaround, I'll first merge this, and debates can follow in new issues.

PR #2927, opened 2015-08-14
improve net config and shape mismatch error messages

jeffdonahue:
This PR tries to improve a few unclear error messages that can be triggered from `Net::Init` (many of which were my fault to begin with, sorry about that).  In particular many questions on caffe-users are about the error message `Check failed: ShapeEquals(proto) shape mismatch (reshape not set)` which occurs in the call to `FromProto` in `CopyTrainedLayersFrom`; this new message which displays the layer name and the shape of both blobs (and suggests renaming the layer if you didn't want to finetune) will hopefully help clear up some of that confusion.


shelhamer:
These messages should help spell out issues! Thanks Jeff.

lukeyeager:
:+1:

PR #2928, opened 2015-08-14
New make target to only build the library.

cypof:
The JavaCPP wrapper has a case where it can't compile the examples. It's not easy to fix, so it would be great to be able to skip them. https://github.com/bytedeco/javacpp-presets/pull/77

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -386,11 +386,13 @@ endif
 ##############################
 # Define build targets
 ##############################
-.PHONY: all test clean docs linecount lint lintclean tools examples $(DIST_ALIASES) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -390,7 +390,9 @@ endif
 	py mat py$(PROJECT) mat$(PROJECT) proto runtest \
 	superclean supercleanlist supercleanfiles warn everything
```
jeffdonahue: `lib` needs to be added to list of `PHONY` targets

jeffdonahue:
See above comment, otherwise LGTM

cypof:
@jeffdonahue added.

jeffdonahue:
Great, thanks @cypof.

Issue #2186, opened 2015-03-24
GPU 0 is also used when running on other GPUs (#440 reocurred?)


zym1010:
I just built caffe-rc2 with CUDA 7.0 and Driver 346.47. When running the test on my first GPU (with id 0), everything works fine. However, when running the test on 2nd GPU (with id 1, or `build/test/test_all.testbin 1`), the command `nvidia-smi` shows that both GPUs are being used. This is not the case when I'm running the test on GPU0, nor when I'm running the test using caffe-rc1 (built with CUDA 6.5 a while ago). I tried building caffe-rc2 using CUDA 6.5, and the problem persists.

By setting `export CUDA_VISIBLE_DEVICES=1`, and running `build/test/test_all.testbin 0`, the problem disappeared. So this seems like a problem like that in https://github.com/BVLC/caffe/issues/440?

Update: when I ran `build/test/test_all.testbin 1 --gtest_filter=DataLayerTest*`, with my GPU0's memory filled up using some other software (cuda_memtest in my case), the program failed:

```
build/test/test_all.testbin 1 --gtest_filter=DataLayerTest/3*
Cuda number of devices: 2
Setting to use device 1
Current device id: 1
Note: Google Test filter = DataLayerTest/3*
[==========] Running 12 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 12 tests from DataLayerTest/3, where TypeParam = caffe::DoubleGPU
[ RUN      ] DataLayerTest/3.TestReadLevelDB
F0324 20:34:20.499236 20499 benchmark.cpp:111] Check failed: error == cudaSuccess (2 vs. 0)  out of memory
*** Check failure stack trace: ***
    @     0x7f16777cfdaa  (unknown)
    @     0x7f16777cfce4  (unknown)
    @     0x7f16777cf6e6  (unknown)
    @     0x7f16777d2687  (unknown)
    @     0x7f1675f133b8  caffe::Timer::Init()
    @     0x7f1675f13569  caffe::CPUTimer::CPUTimer()
    @     0x7f1675ebacec  caffe::DataLayer<>::InternalThreadEntry()
    @     0x7f166de8da4a  (unknown)
    @     0x7f16755f5182  start_thread
    @     0x7f167532247d  (unknown)
    @              (nil)  (unknown)
Aborted (core dumped)
```

With other test categories (at least `NeuronLayerTest` and `FlattenLayerTest`), the program works fine.

longjon:
[...]

PR #2929, opened 2015-08-14
Fixes #1399 by preventing CPUTimer from accessing the GPU. This was c…

### LINKED ISSUE #1399: GPU Memory Usage for Multiple GPUS
screen shot 2014-11-04 at 1 10 30 pm
Hi, Guys

I'm wondering if anyone has already posted this issue before. I'm working on a machine with two Tesla K40 gpus. When I trained a network on GPU1, there is a thread in GPU0 which takes a portion of GPU memory. I used to find this problem on early version of Caffe, in which the process in GPU0 took a duplicated memory block but the utilization of GPU is 0. Now it only takes a portion of the total memory usage. 

I guess it is because that some initialization of the code is fixed to use GPU0, but where is it?

xavigibert:
…ausing the allocation of 38MiB of memory on GPU 0 when we instructed caffe to run on another GPU.

### DIFF
--- a/include/caffe/util/benchmark.hpp
+++ b/include/caffe/util/benchmark.hpp
@@ -37,7 +37,7 @@ class Timer {
   float elapsed_microseconds_;
 };
 
-class CPUTimer : public Timer {
+class CPUTimer {
  public:
   explicit CPUTimer();
   virtual ~CPUTimer() {}
@@ -45,6 +45,20 @@ class CPUTimer : public Timer {
   virtual void Stop();
   virtual float MilliSeconds();
   virtual float MicroSeconds();
+  virtual float Seconds();
+
+  inline bool initted() { return initted_; }
+  inline bool running() { return running_; }
+  inline bool has_run_at_least_once() { return has_run_at_least_once_; }
+
+ protected:
+  bool initted_;
+  bool running_;
+  bool has_run_at_least_once_;
+  boost::posix_time::ptime start_cpu_;
+  boost::posix_time::ptime stop_cpu_;
+  float elapsed_milliseconds_;
+  float elapsed_microseconds_;
 };
 
 }  // namespace caffe
--- a/src/caffe/util/benchmark.cpp
+++ b/src/caffe/util/benchmark.cpp
@@ -165,4 +165,8 @@ float CPUTimer::MicroSeconds() {
   return this->elapsed_microseconds_;
 }
 
+float CPUTimer::Seconds() {
+  return MilliSeconds() / 1000.;
+}
+
 }  // namespace caffe


xavigibert:
BTW, the issue fixed by this PR has also been reported as #2186 (GPU 0 is also used when running on other GPUs (#440 reocurred?)). Issue #1399 referenced in the commit is actually closed.

jeffdonahue:
Thanks for the fix. I'm not sure that removing the inheritance and copy-pasting all the code instead is the way to go though.  Maybe we could instead change the `cudaEvent_t`s to `shared_ptr<cudaEvent_t>`s so that they're not automatically initialized on construction of a `Timer`? (Assuming I'm not misunderstanding the issue here.)

xavigibert:
Indeed, `Timer` is not used anywhere in Caffe, except that in `src/caffe/test/test_benchmark.cpp`. On the other hand `CpuTimer` is used in three places, but the tests do not cover it. Also, `cudaEvent_t` needs to be created by an explicit call to `cudaEventCreate` and destroyed by a explicit call to `cudaEventDestroy`, so I don't see how a smart pointer can be helpful.

I believe that a better design would be have `Timer` as an abstract class and move the existing implementation into a new class that we could name `GpuTimer`. Then, both `CpuTimer` and `GpuTimer` could be derived from `Timer` and you could have a timer factory if necessary. However, I would  argue that having two separate timer implementations is too confusing and that a simpler solution would be to deprecate `Timer` and use only `CpuTimer`.

jeffdonahue:
This is reasonable enough and I was thinking we should go ahead and merge, but to try to confirm the bug I ran `caffe time -gpu 1 ...` with current `master`, and I don't see any allocation on GPU 0 (based on `nvidia-smi` -- just the one process gets added and reports running on GPU 1).  Am I missing something?

xavigibert:
I also tested the current `master` and it does not allocate memory on GPU 0. I did more testing and I found that the problem was actually fixed on Aug 9. Commit d94ca3f is the last one that has this issue, while commit e5575cf is the first one that does not have it. I should have tested the latest version myself before submitting this PR.

jeffdonahue:
Cool -- glad the issue was resolved, thanks for investigating!  In light of the fact that this PR isn't a "bugfix" anymore, I'll be a bit pickier on the design and not merge as is, unless there's another strong argument.  If you or someone else wants to fix up the class hierarchy with less duplicated code I can take another look, though.

ronghanghu:
I seem to be observing this issue again. I have a usage GPU 0 memory consumption of 87MiB when I was actually using another GPU.

ronghanghu:
From `nvidia-smi` on a 8 GPU machine.

```
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID  Type  Process name                               Usage      |
|=============================================================================|
|    0     30703    C   /home/ronghang/anaconda/bin/python              87MiB |
|    0     31930    C   /home/ronghang/anaconda/bin/python              87MiB |
|    7     30703    C   /home/ronghang/anaconda/bin/python            1971MiB |
|    7     31930    C   /home/ronghang/anaconda/bin/python            1971MiB |
+-----------------------------------------------------------------------------+
```

wk910930:
This PR works for me!  Much of my work is based on old caffe. This PR saves me a lot of time and GPU memory. Thanks! @xavigibert

bhalajin:
This PR doesn't work for me as well... The code is based on old caffe and it occupies some memory with GPU0 as well.

PR #2930, opened 2015-08-15
Expose LayerFactory::LayerTypeList in pycaffe

lukeyeager:
I want to be able to validate NetParameters with pycaffe without waiting for Caffe to crash on SIGABRT.
```
F0814 16:58:06.833762 31396 layer_factory.hpp:80] Check failed: registry.count(type) == 1 (0 vs. 1) Unknown layer type: not-a-layer-type (known types: AbsVal, Accuracy, ArgMax, BNLL, Concat, ContrastiveLoss, Convolution, Data, Deconvolution, Dropout, DummyData, Eltwise, EuclideanLoss, Exp, Filter, Flatten, HDF5Data, HDF5Output, HingeLoss, Im2col, ImageData, InfogainLoss, InnerProduct, LRN, Log, MVN, MemoryData, MultinomialLogisticLoss, PReLU, Pooling, Power, Python, ReLU, Reduction, Reshape, SPP, Sigmoid, SigmoidCrossEntropyLoss, Silence, Slice, Softmax, SoftmaxWithLoss, Split, TanH, Threshold, WindowData)
Aborted (core dumped)
```
Exposing `LayerTypeList` in pycaffe lets me check the layer types and handle errors in a reasonable way.

Is there any way to catch all CHECK/DCHECK errors with pycaffe and turn them into Python exceptions?

### DIFF
--- a/python/caffe/test/test_layer_type_list.py
+++ b/python/caffe/test/test_layer_type_list.py
@@ -1,0 +1,10 @@
+import unittest
+
+import caffe
+
+class TestLayerTypeList(unittest.TestCase):
+
+    def test_standard_types(self):
+        for type_name in ['Data', 'Convolution', 'InnerProduct']:
+            self.assertIn(type_name, caffe.layer_type_list(),
+                    '%s not in layer_type_list()' % type_name)
--- a/include/caffe/layer_factory.hpp
+++ b/include/caffe/layer_factory.hpp
@@ -41,6 +41,7 @@
 
 #include <map>
 #include <string>
+#include <vector>
 
 #include "caffe/common.hpp"
 #include "caffe/proto/caffe.pb.h"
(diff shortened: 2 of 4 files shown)


ronghanghu:
Look good to me :)

python/caffe/test/test_layer_type_list.py
```diff
@@ -0,0 +1,10 @@
+import unittest
+
+import caffe
+
+class TestLayerTypeList(unittest.TestCase):
+
+    def test_standard_types(self):
+        for type_name in ['Data', 'Convolution', 'InnerProduct']:
+            assert type_name in caffe.layer_type_list(), \
```
jeffdonahue: this should use `self.assertTrue(...)` instead of `assert`

jeffdonahue:
LGTM besides above comment.

lukeyeager:
Thanks for the review. Changed and squashed. Why is `unittest.assertIn()` preferable to `assert`?

Also, can you address this question:

> Is there any way to catch all CHECK/DCHECK errors with pycaffe and turn them into Python exceptions?

That would be a huge help.

jeffdonahue:
Thanks for the update! The `unittest.assert*` functions will make the test pass/fail appropriately according to their result and print a nicely formatted error message when the assertion fails, where `assert` will just raise an Exception if its argument is false. (Maybe the exception would also make the test fail, but presumably with a less nice error message.)

> Is there any way to catch all CHECK/DCHECK errors with pycaffe and turn them into Python exceptions?

I think #1683 tried to do this

Issue #2926, opened 2015-08-14
runtest crash with TEST_GPUID != 0


raingo:
to duplicate this,

TEST_GPUID := 2 in Makefile.config
make runtest

the solver tests will crash. 

no problem with TEST_GPUID := 0 though

ronghanghu:
My experience is that it crashes when TEST_GPUID >= 2 (however, TEST_GPUID = 1 still works). There should be a bug in the test code.

thatguymike:
This was originally discussed in the multi-gpu work.  The problem is how the tests initialize up front before calls to the solver are made.  This causes an alloction/device mismatch that we now catch.

Setting CUDA_VISIBLE_DEVICES to direct testing does work.

raingo:
so maybe just a documentation issue. where can I find info for CUDA_VISIBLE_DEVICES?

thatguymike:
It is an environment variable you can set to control visibility of the GPUs to your app.

ronghanghu:
I can see the problem here. It seems that we shouldn't have initialized the solvers beforehand and feed them to P2PSync afterwards.

raingo:
I see, it's from CUDA SDK. I think we can just remove the TEST_GPUID flag, because CUDA_VISIBLE_DEVICES can do everything.

thatguymike:
We have options here.  Fix the test on initialization order and/or swap the Makefile setup to set/use CUDA_VISIBLE_DEVICES instead of trying to override in the code.  I think we should try to fix the test to delay initialization or actually initialize the right device(s).

thatguymike:
@raingo - We probably want to leave this bug open until we have a proper fix/workaround in place.

xavigibert:
FYI, on my machine it also crashes with TEST_GPUID := 0 (see below). The workaround of doing "setenv CUDA_VISIBLE_DEVICES 0" before doing "make runtest" prevents the crash.

[----------] 9 tests from AdaGradSolverTest/2, where TypeParam = caffe::GPUDevice<float>
[ RUN      ] AdaGradSolverTest/2.TestAdaGradLeastSquaresUpdate
F0814 18:10:39.053434  4412 math_functions.cpp:91] Check failed: error == cudaSuccess (11 vs. 0)  invalid argument
*** Check failure stack trace: ***
    @     0x7f0a0add80dd  google::LogMessage::Fail()
    @     0x7f0a0add9eac  google::LogMessage::SendToLog()
[...]

PR #2931, opened 2015-08-15
Fix MultiGPU Testing

ronghanghu:
This is a patch for multi-gpu testing issue (#2926). The problem is when calling `make runtest` with TEST_GPUID != 0 on a MultiGPU machine, solver tests will crash because gpu ids in multi-gpu tests doesn't match that of single GPU test.

Note: this still doesn't fix the crash if you have 2 widely different device at the same time. https://github.com/BVLC/caffe/issues/2926#issuecomment-131254812


ronghanghu:
@thatguymike This should fix the original issue in #2926, although not for the case in https://github.com/BVLC/caffe/issues/2926#issuecomment-131254812

ronghanghu:
Setting CUDA_VISIBLE_DEVICES may be a better idea. but I can't figure out how to set it up automatically, to ensure Multi-GPU is tested by default.

Issue #2923, opened 2015-08-14
PythonLayer and Multi-GPU random segfaults


philipp-fischer:
Hi, unfortunately I cannot give a very clear bug description, because I get different errors every time.
However:

Using the simplest imaginable PythonLayer and multi-gpu with GPUs 0,1 selected, segfaults in boost Python library pop up. Sometimes I get errors from the garbage collector `Fatal Python error: GC object already tracked`

This seems to be because of parallelizing the Python code in multiple threads.
When I set `share_in_parallel: true` to serialize Python execution, everything works fine.

The Python layer I use:

    import caffe
    import numpy as np

    class PythonLayer(caffe.Layer):

        def setup(self, bottom, top):
            if len(bottom) != 1:
                raise Exception("Need one input.")

        def reshape(self, bottom, top):
            top[0].reshape(bottom[0].num, bottom[0].channels, bottom[0].height, bottom[0].width)

        def forward(self, bottom, top):
            top[0].data[...] = bottom[0].data

        def backward(self, top, propagate_down, bottom):
            if propagate_down[0]:
                bottom[0].diff[...] = top[0].diff

ronghanghu:
seems to be similar to http://stackoverflow.com/questions/8888124/embedded-python-segfaults

ronghanghu:
@philipp-fischer Note
>When I set share_in_parallel: true to serialize Python execution, everything works fine.

This is not true in general. Currently `share_in_parallel: true` (introduced in #2903) only applies to forward, not backward, as I hope one only do so when PythonLayer is used as a data layer and does no backward. However, I didn't expect this bug in #2903. It was my fault.

ronghanghu:
Close as is partially resolved by #3032 as a short-term workaround before we have better solutions. Further discussions on this issue should follow in #2936.

PR #2939, opened 2015-08-17
Serialize python execution in Multi-GPU

### LINKED ISSUE #2936: Reconcile PythonLayer with Multi-GPU
Right now PythonLayer cannot be parallelized in MultiGPU, which is a bug. Python Global Interpreter Lock (GIL) only allows one thread at a time to run in the python interpreter. This issue is reported in #2923.

We should find a solution to to this bug, or only allow shared PythonLayer and serializing PythonLayer forward (currently via `share_in_parallel: true`) it in MultiGPU (then we need a backward lock, too? but it is more complicated than that https://github.com/BVLC/caffe/issues/2923#issuecomment-131504078) Perhaps we may use [multiprocessing](https://docs.python.org/2/library/multiprocessing.html) module to fix this bug? But on second thought it doesn't seem like a good idea either.

ronghanghu:
This PR should fix #2936 (#2923).

PythonLayer cannot be parallelized in MultiGPU, otherwise crash because Python Global Interpreter Lock (GIL) only allows one thread at a time to run in the python interpreter. This PR serialize all calls to any Python code via a ~~global python mutex~~ GIL to allow having only one PythonLayer execution at a time to behave like before, and fix the crash in #2923. The root problem is that Python doesn't support parallelization in a reasonable way.
- This PR shouldn't affect CPU or single GPU training, where all Python calls are naturally serialized.
- Serializing python execution in Multi-GPU shouldn't lead to overhead for CPU Bound PythonLayer since Python interpretation is always serialized and never actually run in parallel (even with multi-thread).
- For IO Bound PythonLayer without backward (using PythonLayer as a data layer without performing backward), one should set `share_in_parallel: true` in Multi-GPU to share it among workers to load data sequentially (avoiding loading all the same data for each solver).
- ~~Even for the rare case of IO Bound PythonLayer with backward, serialization in Multi-GPU is still better than a GIL crash anyway.~~ IO Bound PythonLayer has its chance to release GIL when stalled waiting for IO.

### DIFF
--- a/src/caffe/layers/python_layer.cpp
+++ b/src/caffe/layers/python_layer.cpp
@@ -1,0 +1,71 @@
+#ifdef WITH_PYTHON_LAYER
+
+#include "caffe/python_layer.hpp"
+
+#include <boost/thread.hpp>
+#include <vector>
+
+namespace caffe {
+
+// A class to initialize Python environment, called by init_python_environment()
+class PyInitializer {
+ private:
+  PyInitializer() {
+    Py_Initialize();
+    PyEval_InitThreads();
+    PyEval_ReleaseLock();
+  }
+  friend void init_python_environment();
+};
+
+void init_python_environment() {
+  static PyInitializer py_initializer;
+}
+
+class AcquirePyGIL {
+ public:
+  AcquirePyGIL() {
+    state = PyGILState_Ensure();
+  }
+  ~AcquirePyGIL() {
+    PyGILState_Release(state);
+  }
+ private:
+  PyGILState_STATE state;
+};
+
+template <typename Dtype>
+void PythonLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
+  AcquirePyGIL lock;
+  self_.attr("param_str") = bp::str(
+      this->layer_param_.python_param().param_str());
+  self_.attr("setup")(bottom, top);
+}
+
+template <typename Dtype>
+void PythonLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
+  AcquirePyGIL lock;
+  self_.attr("reshape")(bottom, top);
+}
+
+template <typename Dtype>
+void PythonLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
+  AcquirePyGIL lock;
+  self_.attr("forward")(bottom, top);
+}
+
+template <typename Dtype>
+void PythonLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
+    const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom) {
+  AcquirePyGIL lock;
+  self_.attr("backward")(top, propagate_down, bottom);
+}
+
+INSTANTIATE_CLASS(PythonLayer);
+
+}  // namespace caffe
+
+#endif  // #ifdef WITH_PYTHON_LAYER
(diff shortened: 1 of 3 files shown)


ronghanghu:
An update on this: serialize Python execution via acquiring GIL instead of using boost::mutex. This should give Python code a chance to release GIL and let other Python threads run when it is stalled (e.g. waiting for IO, or entering heavy computation C++ backend).

src/caffe/proto/caffe.proto
```diff
@@ -746,7 +746,9 @@ message PythonParameter {
   optional string param_str = 3 [default = ''];
   // Whether this PythonLayer is shared among worker solvers during data parallelism.
   // If true, each worker solver sequentially run forward from this layer.
-  // This value should be set true if you are using it as a data layer.
+  // This value should be set true if you are using it as a data layer to load
+  // data sequentially (avoiding loading the same data for each solver).
+  // Note: set true only when your layer does NOTHING in backward
```
longjon: Should this comment be in its own commit? (Is this related to the code above?)

shelhamer:
This approach seems reasonable but please double-check the usage of `PyGILState_Ensure()` and `PyGILState_Release()` in particular when executing a net with PythonLayer from pycaffe.

src/caffe/layers/python_layer.cpp
```diff
@@ -0,0 +1,56 @@
+#ifdef WITH_PYTHON_LAYER
+
+#include "caffe/python_layer.hpp"
+
+#include <boost/thread.hpp>
+#include <vector>
+
+namespace caffe {
+
+class aquire_py_GIL {
```
longjon: Should we CamelCase this for style consistency? (Or is this intentionally made to look like a boost-style underscore_case lock or some such thing?)

jeffdonahue: Yeah, I think any classes we add should follow our CamelCase style.  (Also, should be spelled `acquire`.)

ronghanghu:
Right now this PR is working with Caffe's command-line tool, but does not work with PyCaffe, as it releases GIL when creating a PythonLayer (thus if a net has PythonLayer and is run from PyCaffe, PyCaffe will run in NO-GIL after C++ call, and crash).

Ideally I hope to release GIL whenever entering C++ code (from PyCaffe to C++ calls and when a PythonLayer operation is done) and acquire GIL whenever entering Python code (calling a PythonLayer operation or C++ call is done in PyCaffe). However, the current boost::python based PyCaffe interface makes it difficult to do so.

src/caffe/layers/python_layer.cpp
longjon: Wow, I can't believe I noticed the casing but not the misspelling...

124bit:
There is the same situation discussed in this topic. Don't look on the "Python3" in the title, it doesn't matter. 
http://stackoverflow.com/questions/15470367/pyeval-initthreads-in-python-3-how-when-to-call-it-the-saga-continues-ad-naus/15471525#15471525

> Firstly, let's define the scope of the question here: what do I want to do? Well... I want to write a Python extension module in C that will:
> 
> - Spawn worker threads using the pthread API in C
> - Invoke Python callbacks from within these C threads

Find the answer of **thibaut**, he managed with situation similar to ours. His suggestion is to use 
`PyThreadState* st = PyEval_SaveThread();`
instead of 
`PyEval_ReleaseLock();`

Therefore all that have to be changed in this patch is
`PyEval_ReleaseLock(); -> PyThreadState* st = PyEval_SaveThread();`

I want to test this solution, however I can't figure out how to try to run something trough pycaffe on multiple gpus even if I comment out patch in https://github.com/BVLC/caffe/pull/3032. There is no ability in pycaffe API to set multiple GPUs. @ronghanghu, how you tested pycaffe+multiple gpus configuration?
Any thoughts about such solution?

shelhamer:
The NCCL + python parallelism from #4563 make this unnecessary. Now Python layers can execute in their own processes.

PR #2941, opened 2015-08-18
Add some documentation on Multi-GPU support

thatguymike:
Add initial documentation on Multi-GPU support.

### DIFF
--- a/docs/multigpu.md
+++ b/docs/multigpu.md
@@ -1,0 +1,24 @@
+---
+title: Multi-GPU Usage, Hardware Configuration Assumptions, and Performance
+---
+
+# Multi-GPU Usage
+
+Currently Multi-GPU is only supported via the C/C++ paths and only for training.
+
+The GPUs to be used for training can be set with the "-gpu" flag on the command line to the 'caffe' tool.  e.g. "build/tools/caffe train --solver=models/bvlc_alexnet/solver.prototxt --gpu=0,1" will train on GPUs 0 and 1.
+
+**NOTE**: each GPU runs the batchsize specified in your train_val.prototxt.  So if you go from 1 GPU to 2 GPU, your effective batchsize will double.  e.g. if your train_val.prototxt specified a batchsize of 256, if you run 2 GPUs your effective batch size is now 512.  So you need to adjust the batchsize when running multiple GPUs and/or adjust your solver params, specifically learning rate.
+
+# Hardware Configuration Assumptions
+
+The current implementation uses a tree reduction strategy.  e.g. if there are 4 GPUs in the system, 0:1, 2:3 will exchange gradients, then 0:2 (top of the tree) will exchange gradients, 0 will calculate
+updated model, 0\-\>2, and then 0\-\>1, 2\-\>3. 
+
+For best performance, P2P DMA access between devices is needed. Without P2P access, for example crossing PCIe root complex, data is copied through host and effective exchange bandwidth is greatly reduced.
+
+Current implementation has a "soft" assumption that the devices being used are homogeneous.  In practice, any devices of the same general class should work together, but performance and total size is limited by the smallest device being used.  e.g. if you combine a TitanX and a GTX980, peformance will be limited by the 980.  Mixing vastly different levels of boards, e.g. Kepler and Fermi, is not supported.
+
+# Scaling Performance
+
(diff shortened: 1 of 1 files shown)


raingo:
How to check if the "P2P DMA access" is available? Using nvidia-smi?

thatguymike:
'nvidia-smi topo -m' will show you the connectivity matrix.  You should be able to to P2P on most single socket systems.  It's the multi-socket system where you will run into problems as P2P DMAs can't go across QPI on Intel systems.

raingo:
There are four symbols:

```
  X   = Self
  SOC = Path traverses a socket-level link (e.g. QPI)
  PHB = Path traverses a PCIe host bridge
  PXB = Path traverses multiple PCIe internal switches
  PIX = Path traverses a PCIe internal switch
```
which ones are good for P2P?

thatguymike:
Those that start with P should be okay in the case above (352 drivers).  Basically you want things connected via PCIe bridges or switches.  Do note that each bridge can increase latency and reduce BW.  There is no general tool available to show you all the connections in a usable way.  lspci -tv gets pretty close to show the PCIe bridge topology.

For example, from a DIGITS DevBox:

            +-02.0-[07-0a]----00.0-[08-0a]--+-08.0-[0a]--+-00.0  NVIDIA Corporation Device 17c2
             |                               |            \-00.1  NVIDIA Corporation Device 0fb0
             |                               \-10.0-[09]--+-00.0  NVIDIA Corporation Device 17c2
             |                                            \-00.1  NVIDIA Corporation Device 0fb0
             +-03.0-[03-06]----00.0-[04-06]--+-08.0-[06]--+-00.0  NVIDIA Corporation Device 17c2
             |                               |            \-00.1  NVIDIA Corporation Device 0fb0
             |                               \-10.0-[05]--+-00.0  NVIDIA Corporation Device 17c2
             |                                            \-00.1  NVIDIA Corporation Device 0fb0

ID 17c2 are the computing devices in my case (the other is the HDMI audio).  It shows there is a PLX bridge connecting 2 TitanX's to an x16 link to the CPU.

docs/multigpu.md
```diff
@@ -0,0 +1,26 @@
...
+# Hardware Configuration Assumptions
+
+The current implementation uses a tree reduction strategy.  e.g. if there are 4 GPUs in the system, 0:1, 2:3 will exchange gradients, then 0:2 (top of the tree) will exchange gradients, 0 will calculate
+updated model, 0\-\>2, and then 0\-\>1, 2\-\>3. 
+
+For best performance, P2P DMA access between devices is needed. Without P2P access, for example crossing PCIe root complex, data is copied through host and effective exchange bandwidth is greatly reduced.
+
+Current implementation has a "soft" assumption that the devices being used are homogeneous.  In practice, any devices of the same general class should work together, but performance and total size is limited by the smallest device being used.  e.g. if you combine a TitanX and a GTX980, peformance will be limited by the 980.  Mixing vastly different levels of boards, e.g. Kepler and Fermi, is not supported.
+
+"nvidia-smi topo -m" will show you the connectivity matrix.  You can do P2P through PCIe bridges, but not across socket level links at this time, e.g. across CPU sockets on a multi-socket motherboard.
```
cNikolaou: Could you please give an example "nvidia-smi topo -m" output and how can users understand the basics of their topology. Because many Caffe users may be unaware of that and we might see some issues here and there about the topology. I believe, one or two more lines would be enough.

docs/multigpu.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Scaling Performance
+
+Performance is **heavily** dependent on the PCIe topology of the system, the configuration of the neural network you are training, and the speed of each of the layers.  Systems like the DIGITS DevBox have an optimized PCIe topology (X99-E WS chipset).  In general, scaling on 2 GPUs tends to be ~1.8X on average for networks like AlexNet, CaffeNet, VGG, GoogleNet.  4 GPUs begins to have falloff in scaling.  Generally with "weak scaling" where the batchsize increases with the number of GPUs you will see 3.5x scaling or so.  With "strong scaling", the system can become communication bound, especially with layer performance optimizations like those in [cuDNNv3](http://nvidia.com/cudnn), and you will likely see closer to mid 2.x scaling in performance.  Networks that have heavy computation compared to the number of parameters tend to have the best scaling performance.
```
cNikolaou: It might be good to give a matrix with an example topology and the training time that the network needed based on the gpu's that were used. For example, something like that:

| GPUs | Training Time |
| -------- | -------------------- |
| 0             | example_time_0 |
| 0, 1         | example_time_1 |
| 0, 2         | example_time_2 |
| 0, 1, 2, 3 | example_time_3 |

(with the addition of the system's topology) would be a good starting point for Caffe users to understand what they can achieve and why they do not achieve the best possible performance.

This might be combined with my previous comment and it might be possible to add a new header " Multi-GPU Example", where all the information would be provided.

raingo:
Thanks. Very helpful!

PR #2942, opened 2015-08-19
Update io.py

kli-nlpr:
Edit set_transpose function's description.
Because the original description of set_transpose funciton is the same with set_channel_swap.
I think it's confusing.


Noiredd:
The description is indeed wrong, but this change is also confusing, since `set_transpose` allows any kind of transposition, not just HWC to CHW. Thank you for noticing this, I have made the change in 88c9618.

PR #2943, opened 2015-08-19
add generators to MemoryDataLayer

berleon:
With generators it is possible to generate new input data on the fly.
This is especially useful if you have some way to generate infinite
trainings data (e.g from some 3D models).

The three generators types `MatGenerator`, `DatumGenerator`,
and `RawPointerGenerator` correspond to the three existing methods
 `AddMatVector`, `AddDatumVector` and `Reset`.

Please let me know if you are interested in this pull request. If so I would write some tests.


bhack:
The really interesting thing it is to control generation parameters in a generative vs discriminative fashion. Probably something from @automl group could be used like https://github.com/automl/HPOlib

bhack:
For a generative vs discriminative reference see  http://arxiv.org/abs/1506.05751

bhack:
Another scope to select generation samples http://arxiv.org/abs/1506.09016

bhack:
I wrote this comments because we need to think what kind of in formations we need to exchange between the training process and the generator callback to better explore its parameters.

berleon:
It is quite interesting to also explore the parameter space of the generator. Sadly I currently don't have the time to look through all the papers you posted. 
You  add the generators with the c++ caffe interface to the MemoryDataLayer. So the generator might also get a reference to the network. 
E.g:
```
class MyGenerator : public MatGenerator {
     std::shared_ptr<caffe::Net> _net;
     MyGenerator(std::shared_ptr<caffe::Net> net) : _net(net) {
           ...
     }
}
This should be sufficient for all possible data exchange.

PR #2944, opened 2015-08-19
Give the python layer parameter/weight blobs.

philkr:
This PR allows the python layer to have its own parameter/weight blobs.
As any python layer inherits form caffe.Layer it already has a self.blobs variable that gets saved and loaded properly, however it wasn't possible to add a new blob to that BlobVec. This PR adds a function ```add_blob``` to the BlobVec which allows a python layer to extend the vector with a new blob. The arguments of the ```add_blob``` function are the dimensions of the blob to be added.

Here is a simple layer that demonstrates the use.
```python
import caffe

class Param(caffe.Layer):
	def setup(self, bottom, top):
		self.blobs.add_blob(1,2,3)
		self.blobs[0].data[...] = 0

		top[0].reshape(10)

		print(self.blobs[0].data)
		self.blobs[0].data[...] += 1

		pass


def main():
	from caffe import NetSpec, layers as L, params as P, to_proto
	from os import path
	ns = NetSpec()

	A = L.Python(module='test_phase', layer='Param', ntop=1, name='foo')
	open('/tmp/test.prototxt', 'w').write(str(to_proto(A)))
	net = caffe.Net('/tmp/test.prototxt', caffe.TRAIN )
	if path.exists('/tmp/test.caffemodel'):
		net.copy_from('/tmp/test.caffemodel')
	print( "Loaded" )
	net.forward()
	net.forward()
	net.save('/tmp/test.caffemodel')

if __name__ == "__main__":
	main()
```

### DIFF
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -189,18 +189,22 @@ bp::object Blob_Reshape(bp::tuple args, bp::dict kwargs) {
   // We need to explicitly return None to use bp::raw_function.
   return bp::object();
 }
+
 bp::object BlobVec_add_blob(bp::tuple args, bp::dict kwargs) {
-  if (bp::len(kwargs) > 0)
+  if (bp::len(kwargs) > 0) {
     throw std::runtime_error("BlobVec.add_blob takes no kwargs");
+  }
   typedef vector<shared_ptr<Blob<Dtype> > > BlobVec;
   BlobVec* self = bp::extract<BlobVec*>(args[0]);
   vector<int> shape(bp::len(args) - 1);
-  for (int i = 1; i < bp::len(args); ++i)
+  for (int i = 1; i < bp::len(args); ++i) {
     shape[i - 1] = bp::extract<int>(args[i]);
+  }
   self->push_back(shared_ptr<Blob<Dtype> >(new Blob<Dtype>(shape)));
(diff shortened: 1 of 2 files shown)


longjon:
Generally for this purpose I am using #2079, which ought to eventually in some way in accord with #1474 provide a more natural way to deal with parameters, esp. when using net spec.

But this seems like a reasonable exposure of the current state affairs. It would be nice if `Layer.blobs` was treated as a more Pythonic `list`-like thing, but that's understandably difficult with `vector_indexing_suite`, so this interface seems pretty reasonable and the code is unobtrusive.

For merge:
* Could we provide a basic test of this functionality?
* Please do follow the existing style (even where not checked by lint); in particular, one space between functions, and use explicit braces around blocks.

philkr:
Fixed the style and added some tests.

longjon:
Looks good, thanks @philkr!

longjon:
Ah, I just merged this, so this comment is too late, but one more thing for the future: please squash style changes (here they appear in the test commit, which is confusing).

peerajak:
Thank you very much. Exactly what I have been looking for

It terns out that this is perfectly what I want! the solver do update the weights!. I tested it by setting initial weights to zeros and backward with ones. the result diff is with 0-base_lr*1.
the way to access this new blobvec is as follow
solver.net.params['loss'][0].data

PR #2949, opened 2015-08-20
DeconvolutionLayer Backward_gpu fix: don't redo im2col

jeffdonahue:
This tells `forward_gpu_gemm` not to redo im2col when it's already been done by the above `weight_gpu_gemm` call.  I think this was the intended behavior -- this makes it work the same way as `DeconvolutionLayer::Backward_cpu`.  (Currently the `bool skip_im2col` argument to `forward_gpu_gemm` is never used anywhere in the codebase.)


longjon:
Totally.

PR #2950, opened 2015-08-20
Use input_shape instead of input_dim in examples

lukeyeager:
Since #1970, `input_dim` is deprecated.

>  4D input dimensions -- deprecated.  Use "shape" instead.
https://github.com/BVLC/caffe/blob/833b334b53/src/caffe/proto/caffe.proto#L71


longjon:
Great, thanks!

Issue #2017, opened 2015-03-03
MVNLayer across_channels is broken


cdoersch:
In MVNLayer's forward, there's lots of cases where there's a gemv between a num-by-dim matrix and sum_multiplier_.  When across_channels is true, dim will be `channels*width*height`.  Unfortunately for the gemv, sum_multiplier_ is only `width*height`, since it's created with 

```
   sum_multiplier_.Reshape(1, 1,
      bottom[0]->height(), bottom[0]->width());
    Dtype* multiplier_data = sum_multiplier_.mutable_cpu_data();
    caffe_set(sum_multiplier_.count(), Dtype(1), multiplier_data);
```

Hence, the gemv multiplies the matrix by random garbage whenever this flag is active.  

The fix is just to check for across_channels during init and, if it's true, initialize sum_multiplier to be `channels*width*height`  With this fix at least forward() seems to work properly.  Since I'm guessing that the layer hasn't been extensively tested with this flag enabled, somebody may want to test backward() too...

seanbell:
I think #1979 addresses this bug, but you should confirm this.

cdoersch:
#1979 does address this bug, but also does a lot of other things; probably too much for a single PR.  For instance, it has a new file called mvn_blob_helper.cpp which seems like a new, MVN-layer specific abstraction layer; I think there needs to be an very strong reason why such a file needs to exist in caffe, and in five minutes I couldn't figure out why it needs to be there.  This is probably why nobody in BVLC is reviewing it.  

Overall, it's not obvious that this PR will be merged anytime soon, but this is a major bug that can be fixed with a 3-line change.

jyegerlehner:
>  it has a new file called mvn_blob_helper.cpp ...
> and in five minutes I couldn't figure out why it needs to be there.

mvn_blob_helper.cpp is used by InverseMVNLayer can figure out which of its bottom blobs is the mean and which is the variance (and by MVNLayer to do the same for its top blobs). Prior to adding it (and blob_finder.hpp), layers have no way of finding blobs by name, as far as I was able to find. 

I discovered the across_channels bug (and a couple others) in the course of implementing the tests for InverseMVNLayer. These defects had to be fixed in order for InverseMVNLayer tests to be completed and bring them to a passing state. So I fixed them. I didn't mean any harm by it.

> I think there needs to be an very strong reason why such a file needs to exist in caffe, 

[...]

PR #2958, opened 2015-08-21
Bugfix for #2017 broken across_channels computation in MVNLayer

cdoersch:
The current implementation of `MVNLayer` requires multiplying the data matrix by a vector of 1's to perform sums for the mean and variance.  This vector, sum_multiplier, gets created with the incorrect size when across_channels is true (it needs to be `channels*width*height`).  See #2017. This PR checks for the across_channels flag and creates the sum_multiplier with the correct size.  

Note that this is fixed in #1979 but it's looking less and less likely that this PR will be merged. 

@seanbell can you confirm that this fixes all the bugs you're aware of in mvnlayer?

### DIFF
--- a/src/caffe/layers/mvn_layer.cpp
+++ b/src/caffe/layers/mvn_layer.cpp
@@ -18,8 +18,13 @@ void MVNLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
       1, 1);
   temp_.Reshape(bottom[0]->num(), bottom[0]->channels(),
       bottom[0]->height(), bottom[0]->width());
-  sum_multiplier_.Reshape(1, 1,
-      bottom[0]->height(), bottom[0]->width());
+  if (this->layer_param_.mvn_param().across_channels()) {
+    sum_multiplier_.Reshape(1, bottom[0]->channels(),
+        bottom[0]->height(), bottom[0]->width());
+  } else {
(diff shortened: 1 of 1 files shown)


longjon:
Thanks for breaking out this fix, @cdoersch, and to answer @jyegerlehner and @seanbell from #1979, that's the right thing to do in this situation.

@seanbell claimed this was not failing a test because the test was also incorrect; is that still true? Ideally this PR (and most(?) bug fix PRs) should have two commits: one that fixes the test and _fails_ Travis, and one that corrects the bug and passes. That way we have good confidence that the fix won't regress in the future!

jyegerlehner:
@cdoersch  my recollection is there are a couple more. 1) the backward gradient calculation  was wrong in the mean-only case. 2) The tests weren't testing what they purport to,  because they were using Layer::ParseFromString instead of google::protobuf::TextFormat::ParseFromString to parse the mvn_param, so the prototxt was ignored.  3) Something else? I don't remember at the moment.

I created [2964](https://github.com/BVLC/caffe/pull/2964) where I first fixed only the tests per @longjon suggestion, which should show about half the tests failing. I'll cherry-pick the MVNLayer fixes from [1979](https://github.com/BVLC/caffe/pull/1979).

cdoersch:
Looks like #2964 incorporates all these changes.  Closing.

PR #2963, opened 2015-08-23
Remove superfluous code in Net::ToProto

longjon:
Since `Layer::ToProto` calls `layer_param->Clear` (https://github.com/BVLC/caffe/blob/master/include/caffe/layer.hpp#L508), I think filling in fields beforehand has no effect. This appears to be leftover from an ancient protobuf format.


jeffdonahue:
LGTM, thanks Jon.  (And you're right; I should've removed these with the "v0 to v1" proto upgrade ages ago...)

PR #2964, opened 2015-08-24
Fix MVNLayer

jyegerlehner:
This PR will repair some broken MVNLayer behavior. This first commit just fixes the tests, so that they test what they purport to. On my machine I see 12 of 24 tests failing. 

[Edit]
So fixing only the tests did show failing CPU tests on the build server. Looks like the the GPU tests didn't run. I'm guessing this is because the build server doesn't have a GPU to run them. 

So next I'm committing the fixes. 1) shape sum_multiplier_ member blob correctly, depending on whether across_channels is true. 2) Fix the gradient calculation for the mean-only case. The correct calculation is from analysis by @seanbell  in issue [1938](https://github.com/BVLC/caffe/issues/1938).

### DIFF
--- a/src/caffe/layers/mvn_layer.cu
+++ b/src/caffe/layers/mvn_layer.cu
@@ -113,7 +113,12 @@ void MVNLayer<Dtype>::Backward_gpu(const vector<Blob<Dtype>*>& top,
 
     caffe_gpu_div(temp_.count(), bottom_diff, temp_.gpu_data(), bottom_diff);
   } else {
-    caffe_copy(temp_.count(), top_diff, bottom_diff);
+    caffe_gpu_gemv<Dtype>(CblasNoTrans, num, dim, 1. / dim, top_diff,
+            sum_multiplier_.gpu_data(), 0., mean_.mutable_gpu_data());
+    caffe_gpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, num, dim, 1, -1.,
+            mean_.gpu_data(), sum_multiplier_.gpu_data(), 0.,
+            temp_.mutable_gpu_data());
+    caffe_gpu_add(temp_.count(), top_diff, temp_.gpu_data(), bottom_diff);
   }
 }
 
--- a/src/caffe/test/test_mvn_layer.cpp
+++ b/src/caffe/test/test_mvn_layer.cpp
@@ -6,6 +6,7 @@
 #include "caffe/common.hpp"
 #include "caffe/common_layers.hpp"
 #include "caffe/filler.hpp"
+#include "google/protobuf/text_format.h"
 #include "gtest/gtest.h"
 
 #include "caffe/test/test_caffe_main.hpp"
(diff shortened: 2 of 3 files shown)


jyegerlehner:
There was another thing I changed in MVNLayer. I think in the case where the variance is exactly zero (such as image with a solid color, and across_channels=false), the floating point value isn't always exactly zero; I got ever-so-slightly negative values sometimes. When the negative variances then are square-rooted by the subsequent caffe_powx() call, it would produce NaNs. 

I was thinking if we put that change in, perhaps I should write a test to try to reproduce that behavior first. Would you want that in this PR, or a separate one?

src/caffe/layers/mvn_layer.cu
```diff
@@ -113,7 +113,12 @@ void MVNLayer<Dtype>::Backward_gpu(const vector<Blob<Dtype>*>& top,
 
     caffe_gpu_div(temp_.count(), bottom_diff, temp_.gpu_data(), bottom_diff);
   } else {
-    caffe_copy(temp_.count(), top_diff, bottom_diff);
+    caffe_gpu_gemv<Dtype>(CblasNoTrans, num, dim, 1. / dim, top_diff,
+            sum_multiplier_.gpu_data(), 0., mean_.mutable_gpu_diff());
+    caffe_gpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, num, dim, 1, -1.,
+            mean_.gpu_diff(), sum_multiplier_.gpu_data(), 0.,
+            temp_.mutable_gpu_diff());
+    caffe_gpu_add(temp_.count(), top_diff, temp_.gpu_diff(), bottom_diff);
```
jeffdonahue: Is there a reason to use the `mean_` and `temp_` Blobs' `diff`s rather than their `data` here, as in the CPU implementation? If not, `data` should be used to save memory.

src/caffe/test/test_mvn_layer.cpp
```diff
@@ -73,7 +74,9 @@ TYPED_TEST(MVNLayerTest, TestForward) {
 TYPED_TEST(MVNLayerTest, TestForwardMeanOnly) {
   typedef typename TypeParam::Dtype Dtype;
   LayerParameter layer_param;
-  layer_param.ParseFromString("mvn_param{normalize_variance: false}");
+
```
jeffdonahue: nit: rm newline

jeffdonahue:
See above, otherwise LGTM -- thanks for separating out the bug fix @jyegerlehner.

I think the fix for the NaN issue can be a separate PR. Switching to computing the variance by `E(X - E(X))^2` (rather than `E(X^2) - (E(X))^2`) would probably fix that?

jyegerlehner:
> I think the fix for the NaN issue can be a separate PR. 

OK

>Switching to computing the variance by E((X - E(X))^2) (rather than E(X^2) - (E(X))^2) would probably fix that?

Yes but has disadvantage of requiring me to think.

jeffdonahue:
Thanks for the changes @jyegerlehner, LGTM.

jyegerlehner:
You're welcome @jeffdonahue.

PR #2966, opened 2015-08-24
BatchReindexLayer to shuffle, subsample, and replicate examples in a batch

cdoersch:
This layer takes two blobs as input and uses the second blob to index the first blob along the first axis.

This layer is needed for the training procedure described in [this paper](http://arxiv.org/abs/1505.05192), which first samples patches and extracts fc6-like features, before shuffling them around and pairing them up halfway through the network.  @shelhamer agrees that this may be useful to others, hence this PR.  It has documentation in comments and tests; this should be ready for review.

### DIFF
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -71,6 +71,75 @@ class ArgMaxLayer : public Layer<Dtype> {
 };
 
 /**
+ * @brief Index into the input blob along its first axis.
+ *
+ * This layer can be used to select, reorder, and even replicate examples in a
+ * batch.  The second blob is cast to int and treated as an index into the
+ * first axis of the first blob.
+ */
+template <typename Dtype>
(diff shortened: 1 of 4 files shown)


jeffdonahue:
Thanks @cdoersch, this mostly looks good.  For merge the main issues I see are related to the use of legacy 4D blob indexing in this layer (which is not specific to images).  Specifically:

- uses of `num()` should be removed (change to `shape(0)`)
- the shapes in the docs should be updated (e.g., first blob should be `(N \times ...)` and second blob should be 1D `(N)`)
- check should be added that the second blob is 1D (`CHECK_EQ(1, bottom[1]->num_axes())`) and the dimension of its only axis equals the dimension of the first axis of `bottom[0]` (again using `bottom[0]->shape(0)`)

jeffdonahue:
Thanks for the cleanup @cdoersch, LGTM.

PR #2972, opened 2015-08-25
bugfix for ConcatLayer with propagate_down set

jeffdonahue:
Currently, if `propagate_down[i]` is `false`, `offset_concat_axis` is not correctly
updated for any subsequent bottoms (i+1, i+2, ...).

This bug is quite serious for anyone whom it affects -- if you use concat layer and some of its inputs don't need backpropagation (i.e. because they have no learned parameters below them), you should pull and try again with this fix.  Many thanks to @bharath272 for reporting.

This was my bug -- it's been broken since merging #1970 in which I (tried to) generalize blobs and many layers to N dimensions.  Sorry for the trouble.  ~~In the near future I plan to go through all multi-bottom layers and at least glance to make sure they don't have similar `propagate_down` related bugs, since we don't ever really test the `propagate_down` functionality.~~  Edit: After going through the list, I don't think any other (official) Caffe layers are subject to this flavor of bug, but please let me know if I missed anything.

### DIFF
--- a/src/caffe/layers/concat_layer.cpp
+++ b/src/caffe/layers/concat_layer.cpp
@@ -76,13 +76,14 @@ void ConcatLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
   int offset_concat_axis = 0;
   const int top_concat_axis = top[0]->shape(concat_axis_);
   for (int i = 0; i < bottom.size(); ++i) {
-    if (!propagate_down[i]) { continue; }
-    Dtype* bottom_diff = bottom[i]->mutable_cpu_diff();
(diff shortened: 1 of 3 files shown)


jeffdonahue:
The test added in the first commit should cause the build to fail, and even [did so in the build on my public fork](https://travis-ci.org/jeffdonahue/caffe/jobs/77093155), but somehow passes here... Not sure what's up with that.  Regardless, we should probably merge this ASAP.

shelhamer:
Thanks for the fix @jeffdonahue and reporting the issue @bharath272. To echo Jeff, nets with a concat layer that does not backprop to all inputs should be re-examined!

PR #2975, opened 2015-08-25
Fix python net.forward crash when top layer and top blob names mismatch

cdoersch:
The main goal is to fix issue #2829, in which forward() in the python interface attempts to return the top blobs of the last layer in forward(), but instead uses the layer names as if they were blob names. Turns out there's not currently a way for python code to figure out which blobs are connected to which layers, so this PR exposes that functionality and fixes the bug.

This is a redone version of #2831, this time including a test to show the broken functionality.


shelhamer:
Closing in favor of the minimal change in #5519, but leaving #2829 open to note the further enhancement of handling multiple tops. Note that the improved tests are rescued in #5531.

PR #2978, opened 2015-08-26
Allow H5T_INTEGER in HDF5 files

lukeyeager:
Caffe is currently hardcoded to only support the `H5T_FLOAT` datatype class (originally added by @sergeyk in #203). All I had to do was allow the `H5T_INTEGER` class and now I can use HDF5 datasets created with `dtype='uint8'`. That lets me create datasets with much smaller filesizes, comparable to LMDB sizes with uncompressed data (see comparison here - https://github.com/NVIDIA/DIGITS/pull/226#issuecomment-132846885).

Was there a particular reason that `H5T_INTEGER` was disallowed?


src/caffe/util/hdf5.cpp
```diff
@@ -27,7 +27,34 @@ void hdf5_load_nd_dataset_helper(
   status = H5LTget_dataset_info(
       file_id, dataset_name_, dims.data(), &class_, NULL);
   CHECK_GE(status, 0) << "Failed to get dataset info for " << dataset_name_;
-  CHECK_EQ(class_, H5T_FLOAT) << "Expected float or double data";
+  switch (class_) {
+  case H5T_FLOAT:
+    LOG(INFO) << "Datatype class: H5T_FLOAT";
```
jeffdonahue: Could you remove this and the below `LOG(INFO)`s (or change to `VLOG`, or one of the `LOG` variations that only happens the first time it's hit, if you prefer)?  I think this creates too much noise while training nets with an HDF5DataLayer, especially if using relatively small HDF5 files.

jeffdonahue:
This LGTM except as commented above.  Thanks @lukeyeager!

lukeyeager:
Updated `LOG` to `LOG_FIRST_N`. Neat trick!

jeffdonahue:
Great, thanks again @lukeyeager!

src/caffe/test/test_data/generate_sample_data.py
```diff
@@ -36,23 +36,25 @@
...
-        compression='gzip', compression_opts=1
+        compression='gzip', compression_opts=1,
+        dtype='uint8',
     )
 
 with open(script_dir + '/sample_data_list.txt', 'w') as f:
-    f.write(script_dir + '/sample_data.h5\n')
-    f.write(script_dir + '/sample_data_2_gzip.h5\n')
+    f.write('src/caffe/test/test_data/sample_data.h5\n')
+    f.write('src/caffe/test/test_data/sample_uint8_gzip.h5\n')
```
jeffdonahue: Oops -- I missed this in my review: we should have kept the `script_dir`-relative paths here and below.   I'll fix this in the near future (or feel free to send a patch, @lukeyeager or anyone else).

lukeyeager: I feel like I had a good reason for doing this, but I can't remember what it was now...

lukeyeager: Oh right, this lets you run the `generate_sample_data.py` script from the `src/caffe/test/test_data/` directory and still generate the correct paths in the text file. Otherwise, if you run the script again, the paths in the textfiles change from:

    src/caffe/test/test_data/sample_data.h5

to:

    /home/lyeager/caffe/caffe/src/caffe/test/test_data/sample_data.h5

I chose to fix it this way. The other way to fix it would be to remove the `os.path.abspath()` call on line 8. Would you rather me fix it that way?

jeffdonahue: Ah...thanks for the explanation, I didn't realize that was an issue when I merged #2274.  In retrospect I think we should have just stuck with the existing behavior which required this script to be run from Caffe root, unless we were going to move to running this script to generate the data on-the-fly at test time (rather than tracking the test data files as we do now), which I think is probably a better way to do it (but might be rather involved w.r.t. the potential payoff...).  Anyway, I agree with you & retract my previous suggestion, and I think we should probably revert most or all of #2274 since we can't use the `script_dir`-relative paths everywhere.

ih4cku:
Is it ok to use `uint8` in hdf5 file when caffe still only have `H5LTread_dataset_float` and `H5LTread_dataset_double`?

lukeyeager:
Try it out and let me know. It works for me, so I'm assuming `H5LTread_dataset_float` is able to read `uint8` (or really `H5T_INTEGER`) data and automatically convert it to `float32`, which makes sense.

ih4cku:
I had a simple test, no problem. 
More specifically, create the h5 file with `h5py` filled with `uint8` data, read with `C++` into a `float` array and check each value.

PR #2981, opened 2015-08-26
Draw Deconvolution layers like Convolution layers

maaskola:
python/draw_net.py does not show the parameters for Deconvolution layers like it does for Convolution layers. This commit rectifies this by treating Deconvolution layers identically to Convolution layers in the layer drawing python code.


jeffdonahue:
LGTM, thanks @maaskola.  (I wonder if the script should just print the entire `LayerParameter` prototxt representation regardless of layer type, rather than having all these special settings for individual layer types.)

PR #2989, opened 2015-08-27
Fix EmbedLayer compiler warning for unused variable.

jyegerlehner:
Remove the unused variable `count` in `Backward_gpu` method of embed_layer.cu that causes a compiler warning.


jeffdonahue:
Thanks @jyegerlehner

PR #2990, opened 2015-08-27
Add extra OpenBLAS include search path

mattdawkins:
My default package manager on CentOS 5 put OpenBLAS includes in the /usr/include/openblas folder.


jeffdonahue:
Seems reasonable.  Thanks @mattdawkins!

PR #2991, opened 2015-08-27
Add extra check to net forward function

mattdawkins:
Add extra safety check to the forward function of the net class, such that a more descriptive error is written out (as opposed to a segfault) when attempting to input too many blobs for the network, or using a mis-configured net.


mattdawkins:
Accidentally included two commits, closing.

PR #2992, opened 2015-08-27
  Add extra check to forward function

mattdawkins:
Add extra safety check to the forward function of the net class, such that a more descriptive error is written out (as opposed to a segfault) when attempting to input too many blobs for the network, or using a mis-configured net.


longjon:
Yes, thank you. (Note, however, that input blobs are planned for eventual deprecation.)

Perhaps the check should even be `CHECK_EQ`? Otherwise won't input blobs retain stale data? It would also be nice to have a message (just use the check as a stream, `CHECK_???(...) << "Wrong number of input blobs supplied"` or some such thing).

PR #3003, opened 2015-08-31
Added per layer mode selection (CPU/GPU)

ahmadki:
Added a new layer parameter (layer_mode) that can override the global solver mode. This will allow running any layer in CPU/GPU mode irrespective of the solver mode.

### DIFF
--- a/include/caffe/layer.hpp
+++ b/include/caffe/layer.hpp
@@ -39,8 +39,9 @@ class Layer {
    */
   explicit Layer(const LayerParameter& param)
     : layer_param_(param), is_shared_(false) {
-      // Set phase and copy blobs (if there are any).
+      // Set phase, mode and copy blobs (if there are any).
       phase_ = param.phase();
+      layer_mode_ = param.layer_mode();
       if (layer_param_.blobs_size() > 0) {
         blobs_.resize(layer_param_.blobs_size());
         for (int i = 0; i < layer_param_.blobs_size(); ++i) {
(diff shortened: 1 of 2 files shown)


include/caffe/layer.hpp
```diff
@@ -490,7 +503,17 @@ template <typename Dtype>
...
+  if (layer_mode_ == LayerParameter_LayerMode_CPU) {
+    mode = Caffe::CPU;
+  } else if (layer_mode_ == LayerParameter_LayerMode_GPU) {
+    mode = Caffe::GPU;
+  } else if (layer_mode_ == LayerParameter_LayerMode_SOLVER) {
+    mode = Caffe::mode();
+  } else {
+    LOG(FATAL) << "Unknown layer mode.";
+  }
+  switch (mode) {
```
jyegerlehner: This block of code is duplicated. Can we factor it out into a GetMode() method perhaps.

ahmadki:
Done.

jyegerlehner:
I'm thinking this probably won't work in every case.  An example: any layer that uses caffe_copy. If you look at its implementation:
```c++
template <typename Dtype>
void caffe_copy(const int N, const Dtype* X, Dtype* Y) {
  if (X != Y) {
    if (Caffe::mode() == Caffe::GPU) {
#ifndef CPU_ONLY
      // NOLINT_NEXT_LINE(caffe/alt_fn)
      CUDA_CHECK(cudaMemcpy(Y, X, sizeof(Dtype) * N, cudaMemcpyDefault));
#else
      NO_GPU;
#endif
    } else {
      memcpy(Y, X, sizeof(Dtype) * N);  // NOLINT(caffe/alt_fn)
    }
  }
}
```
You can see that it's assuming it's being given either a host or device pointer based only on the `Caffe::mode()`. So I think this is going to crash in the case where, say, you've set a layer's mode to CPU while the Caffe::mode() is GPU. `caffe_copy()` will call `cudaMemcpy()` with the CPU pointers. 

I haven't looked if there are other cases other than `caffe_copy()` that choose their behavior based on `Caffe::mode()`. 

I suppose a possible way forward is to find all the math functions that call `Caffe::mode()` directly, and change them to take a mode argument, and pass in the result of `GetMode()` wherever they are called.

PR #3007, opened 2015-09-01
Compute backward for negative lr_mult

philkr:
Caffe currently does not update parameters for layers with negative lr_mult. This PR changes this behavior to update parameters for all layers with non-zero lr_mult.
I belief this is a bug after a conversation with @jeffdonahue .


jeffdonahue:
Thanks @philkr.  This is indeed a bug since there is no `CHECK_GE(lr_mult, 0)` or the like, so `Net` will proceed thinking negative `lr_mult` layers don't "need backward", and the parameters aren't updated unless earlier layers do need backward.

PR #3008, opened 2015-09-01
Cleanup: Fixup capitalisation of Caffe_POSTFIX.

darrengarvey:
Replaces CAffe_POSTFIX -> Caffe_POSTFIX.


jeffdonahue:
Thanks for the fix @darrengarvey.  (Current output of `git grep CAffe | wc -l` is 6 so this should be all of them.)

PR #3009, opened 2015-09-01
Fix up documentation errors.

darrengarvey:
Fix some doxygen warnings about an undocumented argument in Blob and
incorrect documentation in a constructor of SoftmaxWithLossLayer.

### DIFF
--- a/include/caffe/loss_layers.hpp
+++ b/include/caffe/loss_layers.hpp
@@ -712,7 +712,6 @@ class SoftmaxWithLossLayer : public LossLayer<Dtype> {
   virtual inline int MaxTopBlobs() const { return 2; }
 
  protected:
-  /// @copydoc SoftmaxWithLossLayer
   virtual void Forward_cpu(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top);
   virtual void Forward_gpu(const vector<Blob<Dtype>*>& bottom,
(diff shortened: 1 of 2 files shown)


include/caffe/loss_layers.hpp
```diff
@@ -691,14 +691,7 @@ template <typename Dtype> class SoftmaxLayer;
...
 class SoftmaxWithLossLayer : public LossLayer<Dtype> {
  public:
-   /**
-    * @param param provides LossParameter loss_param, with options:
-    *  - ignore_label (optional)
-    *    Specify a label value that should be ignored when computing the loss.
-    *  - normalize (optional, default true)
-    *    If true, the loss is normalized by the number of (nonignored) labels
-    *    present; otherwise the loss is simply summed over spatial locations.
-    */
```
jeffdonahue: I think this is correct? (See https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto#L414)

darrengarvey: You're right, that documentation is correct. I was misreading it, sorry. It's actually the ``@copydoc SoftmaxWithLossLayer`` in Forward_cpu that triggers the doxygen warning.

I can update the PR with the doc for Forward_cpu removed as I don't think the same docs make sense in Forward_cpu.

jeffdonahue: Sure, makes sense to remove the `copydoc`.  Thanks!

jeffdonahue:
Thanks @darrengarvey!

PR #3010, opened 2015-09-01
Fix memory leak in convert_mnist_siamese_data.

darrengarvey:
This fixes a memory leak by using delete[] rather than plain delete.


jeffdonahue:
Like #3011, I'm not sure why Travis isn't running here... given that this is a one-line change and looks right, I'll merge. Thanks for the fix @darrengarvey.

darrengarvey:
Thanks @jeffdonahue. I'm not sure either why Travis isn't running, as it ran for one or two of the PRs today (status looks OK: https://www.traviscistatus.com/).

PR #3014, opened 2015-09-02
Allow trivial slice/concat

jeffdonahue:
I'm pulling this out from #2033 for separate review.  This allows trivial (identity) operation of the slice & concat layers by changing their `MinTopBlobs` or `MinBottomBlobs` (respectively) from 2 to 1.  For performance, it is special cased to use `ShareData` and `ShareDiff` in this case, as done for other operations that correspond to the identity when the input & output Blobs are viewed as a flat data array (e.g. Reshape and Flatten).

This is used in #2033 to allow single-timestep operation of recurrent layers without an annoying special case.

@ronghanghu you recently commented that this would make your work easier as well -- maybe you could review this?

### DIFF
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -85,7 +85,7 @@ class ConcatLayer : public Layer<Dtype> {
       const vector<Blob<Dtype>*>& top);
 
   virtual inline const char* type() const { return "Concat"; }
-  virtual inline int MinBottomBlobs() const { return 2; }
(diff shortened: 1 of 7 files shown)


ronghanghu:
LGTM :) I'll take a closer look tomorrow.

longjon:
Looks good to me. This is the correct behavior IMO.

ronghanghu:
Took another pass today and looks ready to merge.

Issue #2969, opened 2015-08-24
`caffe --version` should print the version/git hash


berleon:
caffe does not print its version. 
Current behaviour:
```
$ caffe --version
caffe
```
The expected behaviour would be to print the git hash.

```
$ caffe --version
caffe version rc2-470-g6a5425f
```

This issue can be resolved by setting gflags::SetVersionString.
The git hash can be obtained by `git describe --tags`. See this stackoverflow question for greater
detail: http://stackoverflow.com/questions/1435953/how-can-i-pass-git-sha1-to-compiler-as-definition-using-cmake

lukeyeager:
:+1: 

FYI, the CMake build already grabs this information, so it should be relatively easy to add it to the build:

https://github.com/BVLC/caffe/blob/0dfc5dac3d/cmake/Summary.cmake#L58-L76

shelhamer:
Mostly done in #3311 although the full `git describe` output is not yet included.

Issue #486, opened 2014-06-10
Use the semantic versioning and add a version header file


kloudkl:
Recently released versions are called 0.9, 0.99 and 0.999. Another popular method to set the number is the [semantic versioning](http://semver.org/). Even HBase, one of the Hadoop ecosystem projects, that has been approaching version 1.0 for a long time names its recent versions as 0.9x.x.

OpenCV updates its version in a [header file](https://github.com/Itseez/opencv/blob/master/modules/core/include/opencv2/core/version.hpp) for the users to know the current version they are using. It also enables CMake to [automatically version some other files](https://github.com/Itseez/opencv/blob/95006af17076e4787e38eb56e5d716d892779daf/cmake/OpenCVVersion.cmake).

boaz001:
:+1:

shelhamer:
Keeping open to remind us of the version header.

shelhamer:
Caffe now declares a version, but we haven't followed semantic versioning practices. Closing as close enough for now.

PR #3015, opened 2015-09-03
Cmake add version and soversion

CDLuminate:
Changed CMake files in order to add SONAME for libcaffe.so.
  
Adding SONAME would be convenient for Distribution maintainers to maintain a package.
  
This patch is derived from debian packaging repo of caffe.
  
Please have a look at it, thanks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,10 @@ cmake_minimum_required(VERSION 2.8.7)
 # ---[ Caffe project
 project(Caffe C CXX)
 
+# ---[ Caffe version
+set(CAFFE_TARGET_VERSION "0")
+set(CAFFE_TARGET_SOVERSION "0")
+
 # ---[ Using cmake scripts and modules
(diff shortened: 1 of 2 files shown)


lukeyeager:
I would like to see all of these things together into one patch:

1. Mark v1.0 (#486)
1. **CMake** builds the lib with an SONAME (this)
1. **Make** builds the lib with an SONAME (example: https://github.com/NVIDIA/caffe/commit/18a3ff4e12c3431d5124c8e997944b75407ff442)
1. `build/tools/caffe --version` returns the version (#2969)
1. `pycaffe.__version__` returns the version
1. *Something similar for matcaffe*

I started working on this but had to abandon it for something else. Feel free to steal some of this if anything is helpful:
https://github.com/BVLC/caffe/compare/master...lukeyeager:cmake-versioning

lukeyeager:
Also, you could update the summary to print the version:
https://github.com/BVLC/caffe/blob/9a244f9bcf/cmake/Summary.cmake#L77

shelhamer:
Thanks for submitting this @CDLuminate but let's focus on the complete patch by @lukeyeager in #3311.

PR #3021, opened 2015-09-03
Add ScalarLayer to multiply two Blobs with broadcasting

jeffdonahue:
This adds `ScalarLayer`, which takes two Blobs and (in effect) multiplies them elementwise, after broadcasting the axes of the second Blob to match the first as necessary.

For example, if bottom[0] has shape `(2, 3, 4, 5)` and bottom[1] has shape `(3, 4)` and `axis == 1`, then the computation of this layer is equivalent to reshaping bottom[1]  to `(1, 3, 4, 1)`, then tiling it to `(2, 3, 4, 5)`, then multiplying the result elementwise with bottom[0].

In the most general case, `Backward` to bottom[1] is accomplished with elementwise multiplication followed by 2 `gemv`s.  For special cases (when bottom[1]'s shape corresponds to the beginning or end of bottom[0]'s shape, e.g. if it were instead shape `(2, 3)` and `axis == 0` or shape `(4, 5)` with `axis == 2`) one or both of the `gemv`s is skipped (or replaced with a dot product).

My use case for this comes from #2033 -- I am replacing the hacky `coeff_blob` I added to `Eltwise` to perform the binary multiplications with this layer.  It could also replace the channel-wise scalar in `PReLU` (I think this backward implementation is faster), or be used to learn a channel-wise scalar after batch normalization.

Thanks to @longjon for the name for this layer and initial implementation of a previous version.

### DIFF
--- a/include/caffe/layers/scalar_layer.hpp
+++ b/include/caffe/layers/scalar_layer.hpp
@@ -1,0 +1,70 @@
+#ifndef CAFFE_INNER_PRODUCT_LAYER_HPP_
+#define CAFFE_INNER_PRODUCT_LAYER_HPP_
+
+#include <vector>
+
+#include "caffe/blob.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+namespace caffe {
+
+/**
+ * @brief Computes a product of two input Blobs, with the shape of the
+ *        latter Blob "broadcast" to match the shape of the former.
+ *        Equivalent to tiling the latter Blob, then computing the elementwise
+ *        product.
+ */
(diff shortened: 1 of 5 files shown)


jeffdonahue:
I've updated this with some cleanup in the original commit, and two extra commits.  The first (0437545) makes the layer learn the scalar as a parameter if only one bottom is specified. The second (4199be6) makes it support in-place computation, which necessitates(?) adding a temp storage blob of the input size and copying the input in forward for use in backward. Happy to squash if desired (though the commits are cleanly separated).

jeffdonahue:
Replaced by #3591

PR #3022, opened 2015-09-03
Net: expose param_display_names_

jeffdonahue:
This exposes `param_display_names_` from Net.  Pulled out of #2033, where I use this to log the param names of the internal unrolled net.


jeffdonahue:
Merging this minor dependency of #2033.  (No functional change here, just harmlessly exposes this piece of `Net`'s private state.)

PR #3023, opened 2015-09-03
Net debug_info fix

jeffdonahue:
After #2866 created `learnable_params_`,  this code which computes summary statistics over all net parameters should have been modified to loop over `learnable_params_`, rather than looping over `params_` and checking `param_owners_`.


shelhamer:
Thanks for the fix Jeff.

PR #2935, opened 2015-08-15
Output accuracies per class.

rmanor:

PR #3024, opened 2015-09-03
Enables the alternate solvers to be accessed by the python interface.

danielgordon10:


shelhamer:
Thanks for filling out the list of solvers @danielgordon10.

shelhamer:
Thanks for including more solvers @danielgordon10, but please revise to add `AdaDeltaSolver` too.

jeffdonahue:
I'll note that the other solvers can already be accessed by `get_solver`.  But given that we already directly expose `SGDSolver`, it would be more consistent to expose the others as well.  (An alternate route to consistency would be to remove `SGDSolver` and always use `get_solver`; I don't have a strong opinion either way.)

shelhamer:
@jeffdonahue right -- I'm fine with either exposing them all like this or documenting the use of `get_solver()` even though `SGDSolver` is a special case by its popularity.

danielgordon10:
@shelhamer Whoops, totally missed that one.
Yeah, I saw get_solver at the tail end of my debugging, after I had written the other exposers. I appreciate the explicitness of <Solver>Solver though, especially because SGDSolver exists. I think if you want to encourage get_solver, you shouldn't even have SGDSolver as its own method.

danielgordon10:
Either way, I'm pretty sure the NesterovSolver and AdaGradSolver lines in _caffe.cpp were uncallable without a change. So something needs to be done.

danielgordon10:
Anyone know what the build fail is about?

shelhamer:
From the travis log:

```
[  FAILED  ] 1 test, listed below:
[  FAILED  ] AccuracyLayerTest/1.TestForwardCPUPerClass, where TypeParam = double
```

which blame points to 374fb8c7 in #2935. @ronghanghu could you check what this is about?

@danielgordon10 the test failure has nothing to do with your PR, so don't worry about that. However do squash your fixup commit for AdaDelta for a clear history. Thanks.

ronghanghu:
@shelhamer I'll checkout shortly.

ronghanghu:
@shelhamer There is a bug with AccuracyLayer test in #2935 which I didn't observe. It was my fault.

danielgordon10:
@shelhamer Is there an easy way to squash the commits into an already pushed branch? I'm having issues pushing the squashed version saying I'm behind the pushed one. And if I pull first, then I have a bunch of commits, which is exactly what squash is trying to avoid.

shelhamer:
@danielgordon10 once you squash, force push by `git push -f` to your PR branch to update the PR.

danielgordon10:
@shelhamer naughty naughty, using -f. I'll do it, but I won't enjoy it.

shelhamer:
Not at all -- "published history" in the sense of the project is the canonical BVLC/caffe master branch. Force push is not only fine but right to wrap up PRs.

PR #3025, opened 2015-09-04
FlattenLayer fix(?) -- top should always Share with bottom

jeffdonahue:
I pulled this commit from #2033, where it is no longer necessary (the current version doesn't use `FlattenLayer`), but in some previous version of my RNN implementation, this was causing incorrect behavior.  The current implementation does `bottom[0]->ShareDiff(*top[0])`, which means that if `bottom[0]` has its own diff `SyncedMemory`,  it gets deleted, which must have interacted with my sharing of input diffs in weird ways.

This changes the behavior to work the same way as `ReshapeLayer`.  It also seems somewhat more natural as the `bottom` blobs really "belong" to some other layer (whichever one outputs them as `top`s), whereas a layer's `top`s sort of belong to it, so it seems less dangerous for a layer to delete the underlying `SyncedMemory` of its own `top`s.  However, I don't know of any actual cases where this causes problems in current Caffe, so I'll leave it to others to decide whether this should be merged.

### DIFF
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -313,7 +313,9 @@ class FlattenLayer : public Layer<Dtype> {
    *      the outputs -- i.e., the (virtually) copied, flattened inputs
    */
   virtual void Forward_cpu(const vector<Blob<Dtype>*>& bottom,
-      const vector<Blob<Dtype>*>& top);
+      const vector<Blob<Dtype>*>& top) {}
+  virtual void Forward_gpu(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {}
 
   /**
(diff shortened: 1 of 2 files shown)


longjon:
I haven't thought carefully (yet?) about whether the current behavior is okay, but I agree that this patch changes the code to what I would expect.

ronghanghu:
Except for the "ownership" issue, in my opinion a layer should never change its bottom during forward and reshape (and perhaps never change its top during backward). Ideally we should have `const Blob<Dtype>*` instead of `Blob<Dtype>*`. So I advocate for putting sharing in Reshape instead of Forward and Backward;

longjon:
@ronghanghu yes; for details on why we don't (currently?) have `const Blob*` bottoms in forward, see #945.

PR #3027, opened 2015-09-04
Fix AccuracyLayerTest for per-class accuracy.

ronghanghu:
Fix AccuracyLayerTest for per-class accuracy. Previously in #2935, it crashes since the test accuracy is nan (0/0) when instances of a class never appear.


shelhamer:
Thanks for finding and fixing this Ronghang!

PR #3032, opened 2015-09-04
Disallow PythonLayer in Multi-GPU training

ronghanghu:
This PR disallow PythonLayer during Multi-GPU training phase. Otherwise it crash due to GIL issues #2936.

@jeffdonahue @longjon @shelhamer We should merge either this PR or #2939, to fix PythonLayer crash in Multi-GPU.


ronghanghu:
This is meant to be a partial workaround before we have better solutions. #2939 seems better than this PR (to me) but needs more testing.

PR #3034, opened 2015-09-05
fixed a problem of wasted memory.

tianzhi0549:
Actually, we don't need a so large buffer. I think we only need a  buffer of `inner_num_*sizeof(Dtype) ` bytes to save intermediate results.


ronghanghu:
This buffer size is designed for GPU implementation. Take a look at 
https://github.com/tianzhi0549/caffe/blob/master/src/caffe/layers/softmax_layer.cu#L21

tianzhi0549:
You're right. Thank you:-).

PR #3044, opened 2015-09-08
NoiseLayer

jyegerlehner:
This PR adds a new layer type `NoiseLayer`. It can be used in a strategy that requires adding noise to a blob, e.g. de-noising autoencoders, [target prop (with corruption)](http://arxiv.org/abs/1407.7906), ["dither" as it is called here](http://arxiv.org/abs/1508.04826).

Example usage is a layer like this used to add gaussian noise to the blob `data`. Top and bottom blobs can be the same for in-place noising, or they can be different.

```
layer {
  name: "noise"
  type: "Noise"
  bottom: "data"
  top: "data"
  include {
    phase: TRAIN
  }
  noise_param {
    filler_param {
      type: "gaussian"
      mean: 0
      std: 0.5
    }
  }
}
```

I also considered having the noise layer generate a top blob, and EltWise layer to add it in, but that ends up using more memory because you end up with 3 blobs whereas 2 this way (or 1.5 with in-place noise).

### DIFF
--- a/include/caffe/neuron_layers.hpp
+++ b/include/caffe/neuron_layers.hpp
@@ -475,6 +475,85 @@ class ReLULayer : public NeuronLayer<Dtype> {
       const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
 };
 
+/**
+ * @brief NoiseLayer for adding noise to a blob.
+ */
+template <typename Dtype>
+class NoiseLayer : public NeuronLayer<Dtype> {
+ public:
+  /**
+   * @param param provides NoiseParameter noise_param.
+   *
+   */
+  explicit NoiseLayer(const LayerParameter& param)
+      : NeuronLayer<Dtype>(param), inplace_(false) {}
+  virtual inline const char* type() const { return "Noise"; }
+  virtual void LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+  virtual void Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+
+ protected:
+  /**
+   * @param bottom input Blob vector (length 1)
+   *   -# @f$ (N \times C \times H \times W) @f$
+   *      the inputs @f$ x @f$
+   * @param top output Blob vector (length 1)
+   *   -# @f$ (N \times C \times H \times W) @f$
+   *      the noised output.
+   *        y = \max(0, x)
+   */
+  virtual void Forward_cpu(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+  virtual void Forward_gpu(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+
+  /**
+   * @brief Computes the error gradient w.r.t. the inputs.
+   *
+   * @param top output Blob vector (length 1), providing the error gradient with
+   *      respect to the outputs
+   *   -# @f$ (N \times C \times H \times W) @f$
+   *      containing error gradients @f$ \frac{\partial E}{\partial y} @f$
+   *      with respect to computed outputs @f$ y @f$
+   * @param propagate_down see Layer::Backward.
+   * @param bottom input Blob vector (length 1)
+   *   -# @f$ (N \times C \times H \times W) @f$
+   *      the inputs @f$ x @f$; Backward fills their diff with the top diff.
+   */
+  virtual void Backward_cpu(const vector<Blob<Dtype>*>& top,
+      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
+  virtual void Backward_gpu(const vector<Blob<Dtype>*>& top,
+      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
+
+  const NoiseParameter& NoiseParam() const {
+    return this->layer_param_.noise_param();
+  }
+
+  const FillerParameter& FillerParam() const {
+    return NoiseParam().filler_param();
+  }
+
+  std::string NoiseType() const {
+    const FillerParameter& filler_param = this->NoiseParam().filler_param();
+    return filler_param.type();
+  }
+
+  // Returns true if we are noising in-place. I.e. the top and bottom blob are
+  // the same.
+  bool Inplace() const {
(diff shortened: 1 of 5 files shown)


ducha-aiki:
@jyegerlehner nice job!
Have you actually tried dither? I could not manage it to work (not as in paper, but instead of dropout between layers)

jyegerlehner:
I tried it just to test the code. I put exactly the prototxt above in the stock Caffe mnist example. Running for 10000 iters (what's in the solver prototxt) it didn't seem to make any difference. I then ran with and without "dither" for 50K iterations hoping that might give the model a chance to overfit, and the one without dither got to test accuracy 99.12% and the one with dither got a bit better 99.3% test accuracy. I don't know if that was just run-to-run variance or a real effect. Also, the mnist example in Caffe might be a small enough model that perhaps it doesn't overfit, and regularization isn't needed or helpful. Don't know. One more difference is I recall the dither paper said he used uniform noise distribution, whereas I ran gaussian. And my choice of std_dev=0.5 was arbitrary.

jeffdonahue:
I haven't looked closely at this, but is there any reason for this to be a new layer rather than a new capability of `DummyDataLayer`?

jyegerlehner:
I wasn't aware of `DummyDataLayer`. So are you suggesting producing the noise blob using `DummyDataLayer`, and  `EltwiseLayer` to add in the noise?

Reasons why NoiseLayer might be preferable:
1) `NoiseLayer` allows in-place noising, so that the diffs of the blob to be noised don't have to be duplicated in memory, and copied in the `EltwiseLayer` Backward method.
2) The `DummyDataLayer` always uses fillers to produce the top blob. Fillers only ever run on the CPU. NoiseLayer didn't use Fillers so it could generate noise on the GPU calling `caffe_gpu_rng_gaussian` or `caffe_gpu_rng_uniform`. If we use Fillers as in `DummyDataLayer`, we'd have to run code on the CPU to fill a potentially large blob, and ship it across to the GPU in the middle of our Forward propagation.

jeffdonahue:
> 1) NoiseLayer allows in-place noising, so that the diffs of the blob to be noised don't have to be duplicated in memory, and copied in the EltwiseLayer Backward method.

`EltwiseLayer` could be modified to work in-place for the first bottom to the (only) top, in which case the `DummyDataLayer` + `EltwiseLayer` combination shouldn't use any more memory than this unless I'm missing something.

> 2) The DummyDataLayer always uses fillers to produce the top blob. Fillers only ever run on the CPU. NoiseLayer didn't use Fillers so it could generate noise on the GPU calling caffe_gpu_rng_gaussian or caffe_gpu_rng_uniform. If we use Fillers as in DummyDataLayer, we'd have to run code on the CPU to fill a potentially large blob, and ship it across to the GPU in the middle of our Forward propagation.

For this it seems like it would be more natural to modify the `FIller`s to work on GPU as well. I've tried to do this, but I recall there are problems with the GPU RNG methods -- e.g. the CUDA Gaussian RNG can only fill an even number of Gaussian values...  Probably there would need to be a `FillerParameter` switch to manually set the device for this to work.

jyegerlehner:
> `EltwiseLayer` could be modified to work in-place for the first bottom to the (only) top,

In-place calculation isn't always what someone wants with `EltwiseLayer`, nor is suppressing backward prop to a subset of its bottom blobs. So it adds complexity to put some mechanism to do those things, making the code harder to understand. I find this kind of complexity to be a bigger liability than adding a new layer type. Whereas factoring differing behavior into distinct layer types allows each to be simpler and easier to understand, less error-prone to maintain.

If modified `EltwiseLayer` and `DummyDataLayer` is your preferred approach I'll close this PR.

jeffdonahue:
> In-place calculation isn't always what someone wants with EltwiseLayer, 

Agreed -- it should work the same way all in-place computation in Caffe currently does -- it only happens when bottom[0] has the same name as top[0].

> nor is suppressing backward prop to a subset of its bottom blobs.

This will be correctly determined automatically -- `propagate_down` will be set to false for the noise blob if it's the direct output of a noise layer (because no parameters were used to compute it).

I do prefer the modified EltwiseLayer + DummyDataLayer combination if this has no technical advantage over that.  Another convenience here is that your NoiseLayer takes an input Blob for shape, thereby not requiring the user to manually specify the output shape if there is another similarly shaped Blob in the net -- I think that would be nice to have in DummyDataLayer as well. (I've done this modification in my own work and could PR it at some point.)

jyegerlehner:
> takes an input Blob for shape, thereby not requiring the user to manually specify the output shape if there is another similarly shaped Blob in the net -- I think that would be nice to have in DummyDataLayer as well.

To me that sounds like a violation of the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).  DataLayers produce a top blob from some source; it never takes bottom blob. So people looking at the code or trying to figure out how to use are going to wonder what's going on.

To try to make my previous point another way: if I have the choice between more layers that are each simpler and do a single thing, or fewer, more complicated layers that have had more features shoe-horned into them, I'd prefer the many simple layers. Seems like better separation of concerns.

I'll ponder if I'm up to putting this in a form that's just takes a little different design approach, but isn't obviously superior to me.

jeffdonahue:
> To try to make my previous point another way: if I have the choice between more layers that are each simpler and do a single thing, or fewer, more complicated layers that have had more features shoe-horned into them, I'd prefer the many simple layers. Seems like better separation of concerns.

I agree with the preference for many simpler layers... to me this layer falls into the former category of more "complicated layers" in that it performs not one basic operation but two (random data generation + elementwise addition), whereas the DummyDataLayer+EltwiseLayer combination is two simpler layers that each do just one simple thing and are naturally composable.  Having DummyDataLayer accept a bottom blob to specify shape might be a little odd in the sense that no other "data layers" accept bottom blob inputs, but it's a very small amount of additional logic and doesn't seem too confusing to me, if documented.  

I am open to hearing others' opinions on any of this, though.

ducha-aiki:
@jyegerlehner Originally I liked your solution more, but then realized that "ShapeLike" for DummyData is very useful feature (i.e. when we have ImageDataLayer in single-no-resize-image mode with fully convolutional net).

jyegerlehner:
> DummyDataLayer+EltwiseLayer combination is two simpler layers that each do just one simple thing and are naturally composable.

Yes you're right. I think I see your point about how Eltwise can be made as memory efficient for in-place computation case.

PR #3049, opened 2015-09-09
Check that the snapshot directory is writeable before starting training

seanbell:
When training, if a snapshot cannot be written (directory does not exist, or insufficient permissions, or invalid snapshot prefix), a lot of time can potentially be lost waiting for the error.  This tries to open an empty test file (and delete it) before training starts, so that failure happens as soon as possible.

This does not check for disk space, but usually that's a problem for later in training.

### DIFF
--- a/src/caffe/solver.cpp
+++ b/src/caffe/solver.cpp
@@ -55,6 +55,7 @@ void Solver<Dtype>::Init(const SolverParameter& param) {
     << std::endl << param.DebugString();
   param_ = param;
   CHECK_GE(param_.average_loss(), 1) << "average_loss should be non-negative.";
+  CheckSnapshotWritePermissions();
   if (Caffe::root_solver() && param_.random_seed() >= 0) {
     Caffe::set_random_seed(param_.random_seed());
   }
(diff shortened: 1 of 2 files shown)


ronghanghu:
This should be helpful. Thanks @seanbell !

src/caffe/solver.cpp
```diff
@@ -435,6 +435,26 @@ void Solver<Dtype>::Snapshot() {
 }
 
 template <typename Dtype>
+void Solver<Dtype>::CheckSnapshotWritePermissions() {
+  if (param_.snapshot()) {
+    if (!param_.has_snapshot_prefix()) {
+      LOG(FATAL) << "In solver params, snapshot is specified "
+        << "but snapshot_prefix is not";
+    }
```
ronghanghu: A minor issue: better to use `CHECK` marcos here and below to make code more succinct.

tools/caffe.cpp
```diff
@@ -195,6 +195,7 @@ int train() {
   shared_ptr<caffe::Solver<float> >
     solver(caffe::GetSolver<float>(solver_param));
 
+  solver->CheckSnapshotWritePermissions();
```
ronghanghu: Maybe put this check into `Solver::Init` instead of here, to also help pycaffe and matcaffe?

seanbell:
Thanks for the review!  I addressed your comments and squashed into a single commit.

src/caffe/solver.cpp
```diff
@@ -435,6 +436,24 @@ void Solver<Dtype>::Snapshot() {
 }
 
 template <typename Dtype>
+void Solver<Dtype>::CheckSnapshotWritePermissions() {
+  if (Caffe::root_solver() && param_.snapshot()) {
+    CHECK(param_.has_snapshot_prefix())
+      << "In solver params, snapshot is specified but snapshot_prefix is not";
```
ronghanghu: Better to indent 4 spaces in continuing lines here and below.

seanbell: OK -- updated to 4 spaces (and below) and rebased again.

hjc1028:
the snapshot directory cannot write when i training the lenent ,what should i do?

PR #3052, opened 2015-09-09
Confusion Matrix

borisgin:
New layer, which produces Confusion Matrix (https://en.wikipedia.org/wiki/Confusion_matrix ):
- input: the same as for accuracy layer
- output: 2D matrix ( Blob) [CxC], where C is number of classes

I also modified Solver::Test()  to print 2D outputs of test net into log file.

The example of using new layer:  
- examples/cifar10/cifar_cm_train_test.prototxt

The example of Visualization of confusion matrix:
- plot_conf_matrrix.py

### DIFF
--- a/plot_conf_matrix.py
+++ b/plot_conf_matrix.py
@@ -1,0 +1,65 @@
+#!/home/bginsburg/anaconda/bin/python
+
+"""Extract confusion matrix from log file
+   usage: file.log
+   
+"""
+
+import sys
+import re
+import os
+import math
+import matplotlib.pyplot as plt
+import matplotlib.animation as animation
+import numpy as np
+
+#--------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


ducha-aiki:
Nice!

plot_conf_matrix.py
```diff
@@ -0,0 +1,65 @@
+#!/home/bginsburg/anaconda/bin/python
```
ozabluda: Specific to your environment. Has to be fixed.

TimZaman:
@borisgin any updates on this?

journey-wang:
not merged ?

shelhamer:
Closing as better addressed by a Python layer. There are many metrics out there. Thanks @borisgin for the metric effort all the same.

PR #3038, opened 2015-09-06
Fix opencv 3.0 linking issue

z24:
There is a opencv 3.0 linking issue when building `cpp_classification` example, as shown below

```
ld: .build_release/examples/cpp_classification/classification.o: undefined reference to symbol '_ZN2cv6imreadERKNS_6StringEi'
/usr/lib64/libopencv_imgcodecs.so.3.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make: *** [.build_release/examples/cpp_classification/classification.bin] Error 1
```

Please check http://answers.opencv.org/question/46755/first-example-code-error/?answer=49952#post-id-49952 for details.

But it may not work for opencv 2.x.

PR #3054, opened 2015-09-10
Add opencv_imgcodecs to Makefile

zoharby:
opencv_imgcodecs needed for compilation (at least with OpenCV 3.0)


zoharby:
Hmm, I guess it fails on OpenCV 2, and anyway I now realize it's a duplicate of #3038.

Issue #1318, opened 2014-10-18
Should pooling regions be identical to convolution regions? 


longjon:
Pooling and convolution both aggregate inputs over rectangular regions defined by kernel size, stride, and padding parameters. (In fact, average pooling is convolution with a fixed filter, and max pooling is a special case of "tropical convolution".)

However, Caffe currently uses different sets of regions (and consequently, produces different output sizes) for pooling and convolution. Convolution regions are never allowed outside of the padded input, while pooling is performed when a _strided_ pooling region extends off the ends of the input.

This inconsistency causes some annoyance when the exact sizes of things need to be computed and can vary (due to to #594). (The issue doesn't normally come up when using networks like AlexNet that ensure that their pooling regions don't encounter this edge case.)

Should the behavior be made consistent, or is there a good reason for its current state?

(See also #988, but note that the expression presented there is different from both the current behaviors.)

dujianyi:
I met with the same problem, just wonder if there is already a solution to this? 
It might be good to have a switch that I can choosing between taking ceiling and floor?

wenwei202:
Thanks for details. Wired dimension of outputs of pooling is confusing.

drcege:
Hey guys, it has been a very long time since the issue was opened. 
When can we solve this problem？

PR #3057, opened 2015-09-10
Add parameter for pooling layer to specify "ceil" or "floor"

dujianyi:
When I am trying to convert [`torch.nn`](https://github.com/torch/nn) network to caffe, I met the problem that I can't specify `floor` mode for MaxPooling to calculate the output size. In `nn`, you can specify it freely between `:ceil()` and `:floor()`, so it could be better to include this parameter for users to choose.

### DIFF
--- a/include/caffe/vision_layers.hpp
+++ b/include/caffe/vision_layers.hpp
@@ -416,7 +416,7 @@ class PoolingLayer : public Layer<Dtype> {
   int height_, width_;
   int pooled_height_, pooled_width_;
   bool global_pooling_;
-  bool ceil_mode;
+  bool ceil_mode_;
   Blob<Dtype> rand_idx_;
   Blob<int> max_idx_;
 };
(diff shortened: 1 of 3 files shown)


seanbell:
It might be more intuitive/readable to have this be an enum with choices FLOOR and CEIL.

longjon:
See also #1318...

dujianyi:
@longjon Has that issue been solved?

include/caffe/vision_layers.hpp
```diff
@@ -416,6 +416,7 @@ class PoolingLayer : public Layer<Dtype> {
   int height_, width_;
   int pooled_height_, pooled_width_;
   bool global_pooling_;
+  bool ceil_mode;
```
Franciszzj: bool ceil_mode_;

mazakaro:
working C++ code for floor and ceil can be found at https://www.tutorialcup.com/array/floor-ceil.htm

kli-nlpr:
how to modify the pooling_layer.cu ?

askerlee:
Thank you. Can you also update pooling_layer.hpp? Currently the new variable ceil_mode_ is only added to vision_layers.hpp (not sure where this file will be used; apparently pooling_layer.hpp doesn't include it).

shaibagon:
@Noiredd in light of PR#6282, can we close this PR?

PR #3058, opened 2015-09-11
Add a caffe.io.write_mean function to the MATLAB interface

zoharby:
Useful for exporting models from MATLAB (e.g. MatConvNet) to Caffe

### DIFF
--- a/matlab/+caffe/private/caffe_.cpp
+++ b/matlab/+caffe/private/caffe_.cpp
@@ -478,6 +478,29 @@ static void read_mean(MEX_ARGS) {
   mxFree(mean_proto_file);
 }
 
+// Usage: caffe_('write_mean', mean_data, mean_proto_file)
+static void write_mean(MEX_ARGS) {
+  mxCHECK(nrhs == 2 && mxIsSingle(prhs[0]) && mxIsChar(prhs[1]),
+      "Usage: caffe_('write_mean', mean_data, mean_proto_file)");
+  char* mean_proto_file = mxArrayToString(prhs[1]);
+  int ndims = mxGetNumberOfDimensions(prhs[0]);
+  mxCHECK(ndims >= 2 && ndims <= 3, "mean_data must have at 2 or 3 dimensions");
+  const mwSize *dims = mxGetDimensions(prhs[0]);
+  int width = dims[0];
+  int height = dims[1];
+  int channels;
+  if (ndims == 3)
+    channels = dims[2];
+  else
+    channels = 1;
+  Blob<float> data_mean(1, channels, height, width);
+  mx_mat_to_blob(prhs[0], &data_mean, DATA);
+  BlobProto blob_proto;
+  data_mean.ToProto(&blob_proto, false);
+  WriteProtoToBinaryFile(blob_proto, mean_proto_file);
+  mxFree(mean_proto_file);
+}
+
 /** -----------------------------------------------------------------
  ** Available commands.
  **/
@@ -515,6 +538,7 @@ static handler_registry handlers[] = {
   { "get_init_key",       get_init_key    },
   { "reset",              reset           },
   { "read_mean",          read_mean       },
+  { "write_mean",         write_mean      },
   // The end.
   { "END",                NULL            },
 };
--- a/matlab/+caffe/io.m
+++ b/matlab/+caffe/io.m
@@ -29,5 +29,13 @@ classdef io
       CHECK_FILE_EXIST(mean_proto_file);
       mean_data = caffe_('read_mean', mean_proto_file);
     end
+    function write_mean(mean_data, mean_proto_file)
+      % write_mean(mean_data, mean_proto_file)
+      %   write image mean data to binaryproto file
+      %   mean_data should be W x H x C with BGR channels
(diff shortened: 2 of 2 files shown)


matlab/+caffe/private/caffe_.cpp
```diff
@@ -478,6 +478,29 @@ static void read_mean(MEX_ARGS) {
...
+// Usage: caffe_('write_mean', mean_mat, mean_proto_file)
+static void write_mean(MEX_ARGS) {
+  mxCHECK(nrhs == 2 && mxIsSingle(prhs[0]) && mxIsChar(prhs[1]),
+      "Usage: caffe_('write_mean', mean_mat, mean_proto_file)");
+  char* mean_proto_file = mxArrayToString(prhs[1]);
+  int ndims = mxGetNumberOfDimensions(prhs[0]);
+  mxCHECK(ndims >= 2 && ndims <= 3, "mean_mat must have at 2 or 3 dimensions");
+  const mwSize *dims = mxGetDimensions(prhs[0]);
+  int height = dims[0];
+  int width = dims[1];
```
ronghanghu: I think you've messed up with dimensions, it should be `int width = dims[0]; int height = dims[1];`

zoharby:
Actually, MATLAB has a Column-major order, so the order of the dimensions in memory is:
(height, width, channels, num) 
and not:
(width, height, channels, num) as mentioned in line 9. So I think actually all the rest of the conventions in the interface are incorrect and should be reversed. But I agree that currently my commit is inconsistent with the rest of the code, so we should align all the functions to the same convention.

Images in MATLAB are definitely stored as HxWxC, so the following is correct:
```
const mwSize *dims = mxGetDimensions(prhs[0]);
int height = dims[0];
int width = dims[1];
int channels = dims[2];
```

Did you have any specific reason why you used (width, height, channels, num) for the MATLAB memory layout?

ronghanghu:
>Actually, MATLAB has a Column-major order, so the order of the dimensions in memory is:
(height, width, channels, num) 
and not:
(width, height, channels, num) as mentioned in line 9. So I think actually all the rest of the conventions in the interface are incorrect and should be reversed. But I agree that currently my commit is inconsistent with the rest of the code, so we should align all the functions to the same convention.

Yes, you are right. There has been a lot of confusion on Caffe's dimensions, so let me try to clarify here.

If you load an image using Matlab's `imread`, it is going to be (H x W x C, column-major), *where H is the fastest (memory continuous) dimension*.

However, this characteristic of Matlab's image toolbox (height being the fastest dimension) is inconsistent with many other image tools, such as OpenCV. In OpenCV (both C/C++ and Python), the default memory layout will be (H x W x C, row-major) in C-contiguous array, that is, channel being the fastest dimension, followed by width, and height being the slowest dimension*. Since C/C++ is row-major and Matlab is column major, OpenCV's memory layout (H x W x C, row-major) is exactly the opposite from Matlab (H x W x C, column-major).

When designing Caffe, to be consistent with C/C++ and numpy's default, and to make it easier to do convolution with existing BLAS libraries, we adopted (N x C x H x W, row-major) memory layout. This is neither Matlab convention nor OpenCV convention. Also, OpenCV uses BGR channel order by default while Matlab uses RGB channel order, and we follow OpenCV to use BGR in Caffe.

So if you load an image using Matlab's imread, you will need to:
1. Permute height and width to make width the fastest dimension
2. Permute channel order from RGB to BGR
3. Convert from uint8 to (single precision) float
as is illustrated in `caffe.io.load_image`.

In your code above, to be consistent with the rest of Caffe, you will need to set

```
int width = dims[0];
int height = dims[1];
int channels = dims[2];
```

And remember convert channel order to BGR before running your `caffe.io.write_mean`

zoharby:
Right, my mistake, thanks for the clear & detailed explanation. 
I now realize the conversion to WxHxC BGR is even demonstrated in the classification_demo.m file and in load_image as you mentioned.

Fixed & Pushed, I think it's all good now.

ronghanghu:
@zoharby Please squash into one single commit, and I'll merge.

matlab/+caffe/private/caffe_.cpp
```diff
@@ -478,6 +478,29 @@ static void read_mean(MEX_ARGS) {
   mxFree(mean_proto_file);
 }
 
+// Usage: caffe_('write_mean', mean_mat, mean_proto_file)
+static void write_mean(MEX_ARGS) {
+  mxCHECK(nrhs == 2 && mxIsSingle(prhs[0]) && mxIsChar(prhs[1]),
+      "Usage: caffe_('write_mean', mean_mat, mean_proto_file)");
```
ronghanghu: Better to be consistent with matlab code in nomination here and use 'mean_data' instead of 'mean_mat'.

zoharby:
Fixed he naming you suggested and squashed, but mistakenly squashed before I pushed the last change so it resulted in an empty merge commit. Hope it's OK, not sure how to get rid of it now...

zoharby:
Also, I still see the squashed commits here in GitHub although they're gone from my local repo. Kinda new with all this advanced git stuff, sorry for the trouble...

ronghanghu:
Everything looks good now :) @zoharby Thank you for this PR!

PR #3059, opened 2015-09-11
device_query: Interrogate all GPUs when "-gpu" is not passed.

darrengarvey:
This makes the following two commands equivalent:

    caffe device_query
    caffe device_query -gpu all

The former is what a new user is likely to use and it would be helpful if
it showed all of the GPUs available on the device.


lukeyeager:
:+1:

seanbell:
Won't this have the effect of using all GPUs for training as well? (https://github.com/darrengarvey/caffe/blob/improve-device_query-usability/tools/caffe.cpp#L175)

darrengarvey:
Good point @seanbell, that wasn't the intention. I've update the PR to fix this.

PR #3060, opened 2015-09-13
Always Use cudaMallocHost to allocate and pin CPU memory

### LINKED ISSUE #3053: Error with cudaFreeHost(ptr) in syncedmem.hpp:30
In my case, when caffe model finishs predicting an image. There is the below error:

F0910 15:09:52.590445 21913 syncedmem.hpp:30] Check failed: error == cudaSuccess (11 vs. 0)  invalid argument
Aborted

It comes from CUDA_CHECK(cudaFreeHost(ptr));  in syncedmem.hpp.
Everything else works well. Any ideas for fixing this? I use latest code in the master.

I'm using Ubuntu 14.04, NVIDIA driver 352.41, Cuda 7.5 and CuDNN v2.

ronghanghu:
Unless in CPU_ONLY build, always use cudaMallocHost and cudaFreeHost to allocate (and pin) and free cpu memory, so that it does not depend on Caffe::mode(), which is changable during runtime and can result in inconsistency between host memory allocation and free. This should fix #3053.

Eventually it is better to give mode and device to net and layers, in the spirit of #1500. But it seems like a short-term fix to always use `cudaMallocHost` to allocate cpu memory regardless of the mode of the net to avoid this crash without significant overhead.

However, although mode/device is right now not a member of Net and Layer and thus changable in runtime, it is better to set them in advance and not change them during a lifecycle of a net.

### DIFF
--- a/include/caffe/syncedmem.hpp
+++ b/include/caffe/syncedmem.hpp
@@ -8,30 +8,28 @@
 
 namespace caffe {
 
-// If CUDA is available and in GPU mode, host memory will be allocated pinned,
+// If CUDA is available, host memory will be allocated pinned,
 // using cudaMallocHost. It avoids dynamic pinning for transfers (DMA).
 // The improvement in performance seems negligible in the single GPU case,
 // but might be more significant for parallel training. Most importantly,
 // it improved stability for large models on many GPUs.
 inline void CaffeMallocHost(void** ptr, size_t size) {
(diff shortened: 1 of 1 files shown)


ronghanghu:
Ah, seem to be running into the same issue as in https://github.com/BVLC/caffe/commit/46a431ac7423e08a2db1894ef8cea0803eba96b2

PR #3069, opened 2015-09-15
Add argmax_param "axis" to maximise output along the specified axis

timmeinhardt:
This PR adds an option to the ArgMaxLayer which lets the user specify an axis along which the layer should maximise the output. This now works similar to e.g. [numpy.argmax](http://docs.scipy.org/doc/numpy/reference/generated/numpy.argmax.html) and might be useful for deploying segmentation task networks. 
The default option is axis = 0, which computes the argmax of the flattened bottom blob per image. So existing deployments of the layer work as expected. Negative indexing is also possible, but eventually the axis value must be between 0 and 3.
If axis != 0 and out_max_val is set to true the layer outputs max_val instead of max_ind. Specifying an axis and outputting max_val and max_ind at the same time was not possible due to the general architecture of a blob.

### DIFF
--- a/src/caffe/layers/argmax_layer.cpp
+++ b/src/caffe/layers/argmax_layer.cpp
@@ -11,11 +11,23 @@ namespace caffe {
 template <typename Dtype>
 void ArgMaxLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
-  out_max_val_ = this->layer_param_.argmax_param().out_max_val();
-  top_k_ = this->layer_param_.argmax_param().top_k();
-  CHECK_GE(top_k_, 1) << " top k must not be less than 1.";
-  CHECK_LE(top_k_, bottom[0]->count() / bottom[0]->num())
-      << "top_k must be less than or equal to the number of classes.";
+  const ArgMaxParameter& argmax_param = this->layer_param_.argmax_param();
+  out_max_val_ = argmax_param.out_max_val();
+  top_k_ = argmax_param.top_k();
+  has_axis_ = argmax_param.has_axis();
+  CHECK_GE(top_k_, 1) << "top k must not be less than 1.";
+  if (has_axis_) {
+    axis_ = bottom[0]->CanonicalAxisIndex(argmax_param.axis());
+    CHECK_GE(axis_, 0) << "axis must not be less than 0.";
+    CHECK_LE(axis_, bottom[0]->num_axes()) <<
+      "axis must be less than or equal to the number of axis.";
+    CHECK_LE(top_k_, bottom[0]->shape(axis_))
+      << "top_k must be less than or equal to the dimension of the axis.";
+  } else {
+    CHECK_LE(top_k_, bottom[0]->count(1))
+      << "top_k must be less than or equal to"
+        " the dimension of the flattened bottom blob per instance.";
+  }
 }
 
 template <typename Dtype>
--- a/src/caffe/layers/argmax_layer.cpp
+++ b/src/caffe/layers/argmax_layer.cpp
@@ -33,13 +33,19 @@ void ArgMaxLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
 template <typename Dtype>
 void ArgMaxLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
-  if (out_max_val_) {
+  std::vector<int> shape(4, 1);
+  shape[0] = bottom[0]->shape(0);
+  // Produces max_ind
+  shape[2] = top_k_;
+  if (has_axis_) {
+    // Produces max_ind or max_val per axis
+    shape = bottom[0]->shape();
+    shape[axis_] = top_k_;
+  } else if (out_max_val_) {
     // Produces max_ind and max_val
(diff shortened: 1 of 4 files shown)


seanbell:
Wouldn't axis = 1 make more sense as a default?  By convention in caffe, axis = 0 is the dimension that indexes across images in a batch.  Typically, axis = 1 is the "channel" that separates classes.

timmeinhardt:
In order to argmax a classification task output as (10,1000,1,1) with 10 images and 1000 classes you could either set axis = 0 or axis = 1. The result would be the same, but in the first case the bottom blob is treated as a flattened array and in the second the width and height of the blob are considered.
I went for axis = 0 as default option because that's how the layer worked before and processing the flattened input is usually the default option for argmax implementations.

seanbell:
OK I see what you've done, and I don't think it is intuitive.

If axis > 0, then you are computing what you expect (argmax along that axis).  If axis == 0, then you are reverting to the old layer behavior (which is not actually computing argmax along axis 0).  I think it's unintuitive and surprising to do it this way.  For example, I would expect that for a (N, C) input, computing argmax along axis 0 would give an output shape of (top_k, C).  I would be surprised to discover that it has actually given me an output of shape (N, top_k) and this would seem like a bug to me.

The current ArgmaxLayer didn't get updated during the 4D --> ND conversion in #1970, and has extra unnecessary dimensions of size 1 in the output.  Also, we have easy efficient reshaping as of #2217.  So the automatic reshaping that you get with this layer isn't really worth trying to preserve.

I think a better way to preserve backwards compatibility would be: if `axis` is specified at all, perform the old behavior.  Otherwise, if it is set, compute the argmax along that actual axis, rather than making `axis: 0` do something special that's not actually argmax along axis 0.

Or, you could have a flag (e.g. `reshape` with default `true`) that enables automatic reshaping of the blob, and make the default `axis: 1`.  This would also preserve the legacy behavior of this layer.  But it would be a lot more intuitive.

timmeinhardt:
I got your point. Basically I was already using axis = 0 as if no axis was specified so I changed that to use has_axis instead. Now one can even argmax across multiple images using axis = 0. Perhaps this might be useful for someone but it is definitely more intuitiv. Thank you! Sometimes you do not see the wood for the trees when you are in the middle of something...
So now if no axis is specified the maximisation is computed along the flattened bottom blob per image and if axis is set it maximises along the specified axis.

src/caffe/layers/argmax_layer.cpp
```diff
@@ -11,23 +11,41 @@ namespace caffe {
...
+  out_max_val_ = argmax_param.out_max_val();
+  top_k_ = argmax_param.top_k();
+  has_axis_ = argmax_param.has_axis();
+  CHECK_GE(top_k_, 1) << "top k must not be less than 1.";
+  if (has_axis_) {
+    axis_ = bottom[0]->CanonicalAxisIndex(argmax_param.axis());
+    CHECK_LE(top_k_, bottom[0]->shape(axis_))
+      << "top_k must be less than or equal to the dimension of the axis.";
+    CHECK_GE(axis_, 0) << "axis must not be less than 0.";
+    CHECK_LE(axis_, 3) <<
```
shelhamer: This should check `bottom[0]->num_axes()` for generality now that blobs are N-D.

src/caffe/layers/argmax_layer.cpp
```diff
@@ -11,23 +11,41 @@ namespace caffe {
...
+      << "top_k must be less than or equal to"
+        " the dimension of the flattened bottom blob per image.";
+  }
 }
 
 template <typename Dtype>
 void ArgMaxLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
-  if (out_max_val_) {
+  std::vector<int> shape(4, 1);
```
shelhamer: Instead of fixing 4 dimensions try for N-D.

src/caffe/proto/caffe.proto
```diff
@@ -443,6 +443,10 @@ message ArgMaxParameter {
   // If true produce pairs (argmax, maxval)
   optional bool out_max_val = 1 [default = false];
   optional uint32 top_k = 2 [default = 1];
+  // The axis along which to maximise -- may be negative to index from the
+  // end (e.g., -1 for the last axis).
+  // By default ArgMaxLayer maximises blobs as flattened array per image.
```
shelhamer: Instead of focusing on images, "per instance" or "ArgMaxLayer maximizes over the flattened trailing dimensions for each index of the first / num dimension` might be better since all kinds of predictions could be argmaxed.

shelhamer:
Thanks for contributing this @timmeinhardt and thanks for guiding the development @seanbell! This looks good in general but I'd like to see it done in N-D instead of the fixed 4 dimensions since it's nearly there. You could look at `ConcatLayer` or other layers with an `axis` arg.

timmeinhardt:
I adjusted my commits and now it should work for N-D blobs as well. I updated the documentation as well. I did not know that you made this update to the blob architecture :+1:

shelhamer:
Thanks @timmeinhardt !

AndKo1201:
@timmeinhardt could you please check the ArgMaxLayer<Dtype>::Reshape.

I'm getting a assertion when i work with debug lvl == 2 while executing the caffe unit tests.
It crashes in `NetTest, TestSkipPropagateDown`.

In this test the num_axes of bottom[0] is 2 and the value of has_axis_ is false.
But within the else you are trying to change the value of shape[2], which is an OutOfBounds access.

PR #3073, opened 2015-09-16
Ensure consistency between memory alloc and free

### LINKED ISSUE #3053: Error with cudaFreeHost(ptr) in syncedmem.hpp:30

ronghanghu:
Add a bool flag to record whether a host memory is allocated using malloc or cudaMallocHost, and free correspondingly using this flag, instead of depending on Caffe::mode(), which is mutable during runtime.

This should fix #3053 .

### DIFF
--- a/src/caffe/syncedmem.cpp
+++ b/src/caffe/syncedmem.cpp
@@ -8,7 +8,7 @@ namespace caffe {
 
 SyncedMemory::~SyncedMemory() {
   if (cpu_ptr_ && own_cpu_data_) {
-    CaffeFreeHost(cpu_ptr_);
+    CaffeFreeHost(cpu_ptr_, cpu_malloc_use_cuda_);
   }
 
 #ifndef CPU_ONLY
@@ -27,7 +27,7 @@ SyncedMemory::~SyncedMemory() {
 inline void SyncedMemory::to_cpu() {
   switch (head_) {
   case UNINITIALIZED:
-    CaffeMallocHost(&cpu_ptr_, size_);
+    CaffeMallocHost(&cpu_ptr_, size_, &cpu_malloc_use_cuda_);
     caffe_memset(size_, 0, cpu_ptr_);
     head_ = HEAD_AT_CPU;
     own_cpu_data_ = true;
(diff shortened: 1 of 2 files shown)


ronghanghu:
I hope to merge this PR, if no one opposes. This ensures that memories are allocated and freed consistently.

shelhamer:
Let's hold off a bit. Although this addresses the immediate issue the root cause is that Nets do not know their own state / mode and adding more state to SyncedMem isn't necessarily how we want to work around that. However I might be convinced this is the way to go for now since it works.

ronghanghu:
>Although this addresses the immediate issue the root cause is that Nets do not know their own state / mode and adding more state to SyncedMem isn't necessarily how we want to work around that.

I agree. In the spirit of #1500 I think it is better to have mode and device in net (or in layers if we plan to have individual mode/device for each layer).

ronghanghu:
However, I still believe even the net knows their state/mode/device, it is still not enough, because in order to make SyncMemory and Blob self-contained data structures (not depending on the net holding it), SyncMemory class needs to know things like how memory is allocated, on which device etc.

So personally I still think a bool flag in this PR could be sort of necessary for SyncMemory. But I am open to opinions and suggestions on this issue.

src/caffe/syncedmem.cpp
```diff
@@ -154,4 +154,3 @@ void SyncedMemory::async_gpu_push(const cudaStream_t& stream) {
 #endif
 
 }  // namespace caffe
-
```
longjon: Can we remove this whitespace change from the commit?

shelhamer:
Alright, this is good to merge for solving the immediate issue and we will further plan #1500 for diffusing mode + device through Net, Layer, and Blob.

src/caffe/syncedmem.cpp
ronghanghu: Resolved.

ronghanghu:
This should be good now :)

PR #3081, opened 2015-09-17
include IO dependencies by default

shelhamer:
Follow-up to #2523: keep old behavior by including leveldb, lmdb, and opencv by
default.


jeffdonahue:
Thanks @shelhamer, LGTM

eelstork:
@shelhamer nice.

Issue #3077, opened 2015-09-17
Why is `snapshot` not exposed through the Python interface?


gustavla:
I have written a Python script to manage my training and I'm adding a signal interrupt handler that upon receiving `SIGTERM` will finish its current iteration and then snapshot and safely quit.

However, I noticed that `Solver::Snapshot` is `protected` and thus not exposed to Python. I don't understand why this is the case, since my use case seems perfectly reasonable. Changing this locally and adding `snapshot` to `python/_caffe.cpp` has worked out well for me so far.

A way around it might be to somehow change the snapshot parameter through Python and that way force a snapshot to be made at the next iteration. I don't even know if this is possible, but even it is it's quite an ugly hack compared to just calling `snapshot()`. It would also delay the snapshot by one iteration.

seanbell:
See https://github.com/rbgirshick/fast-rcnn/blob/master/lib/fast_rcnn/train.py#L48 as an example of performing a snapshot in Python.

gustavla:
Calling `save` only produces a `.caffemodel` file with the weights. To correctly resume SGD with momentum, I also need the latest gradients in a `.solverstate` file.

ronghanghu:
Personally I think it is better to expose Snapshot function to pycaffe. I can write a PR for that when less busy.

gustavla:
I agree. I can also arrange a PR, since I have already made these changes locally.

longjon:
Closing as addressed by #3082.

PR #3082, opened 2015-09-18
Expose `Solver::Snapshot` to pycaffe

gustavla:
This addresses #3077, making it possible to manually save a snapshot (caffemodel+solverstate) from Python.

On a related note, does it also make sense to expose `Solver::TestAll`, so that it's possible to manually instigate a test run through Python?

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -60,6 +60,11 @@ class Solver {
   // RestoreSolverStateFrom___ protected methods. You should implement these
   // methods to restore the state from the appropriate snapshot type.
   void Restore(const char* resume_file);
+  // The Solver::Snapshot function implements the basic snapshotting utility
+  // that stores the learned net. You should implement the SnapshotSolverState()
+  // function that produces a SolverState protocol buffer that needs to be
+  // written to disk together with the learned net.
+  void Snapshot();
   virtual ~Solver() {}
   inline const SolverParameter& param() const { return param_; }
   inline shared_ptr<Net<Dtype> > net() { return net_; }
@@ -87,11 +92,6 @@ class Solver {
  protected:
   // Make and apply the update value for the current iteration.
   virtual void ApplyUpdate() = 0;
-  // The Solver::Snapshot function implements the basic snapshotting utility
-  // that stores the learned net. You should implement the SnapshotSolverState()
-  // function that produces a SolverState protocol buffer that needs to be
-  // written to disk together with the learned net.
-  void Snapshot();
   string SnapshotFilename(const string extension);
   string SnapshotToBinaryProto();
   string SnapshotToHDF5();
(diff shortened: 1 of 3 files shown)


longjon:
Looks good -- I wonder if we could have a basic test though?

Exposing `TestAll` in a different PR sounds good to me.

gustavla:
Sorry for delay on this. I added a test that runs `solver.snapshot()` and checks if the expected files were generated.

shelhamer:
Thanks for the pycaffe extension @gustavla!

mpkuse:
Is there also a way to load a snapshot into the solver with pycaffe?

rogertrullo:
Hi @gustavla , could you please provide an example on how to use this feature?
in #3077 you mentioned that you wrote a `SIGTERM` interrupt handler?
Thanks!

rayryeng:
@mpkuse Just use the restore method that accompanies the solver.  The input is the path to your .solverstate file that gets saved during snapshots --> solver.restore('your_solverstate_file.solverstate').  It's also mentioned in the Python section under Interfaces in the Caffe documentation: http://caffe.berkeleyvision.org/tutorial/interfaces.html

VasLem:
Is there a way to set snapshot file name from inside a Python program, or is it necessary to have the name inside the solver prototxt from the beginning?  Sorry if I ask something obvious, but there is no documentation, apart from this page and http://caffe.berkeleyvision.org/tutorial/interfaces.html.

naibaf7:
@VasLem 
This example demonstrates the Python interface to it's full extent. However do note some options may only be available on the opencl branch and not (yet) mainline. But it should give you a good idea of it:
https://github.com/naibaf7/opencl_caffe_examples/blob/master/mnist_lenet/mnist_lenet.ipynb

Coderx7:
The current snapshot seems to have a problem, or at least I cant find a way around it.
When someone tries to train using Pycaffe and save the best model, one way to do so is have a main loop like this: 
```
for it in range (max_iter):
    solver.step(1)

    #check if its the best so far, if so , save it  
    solver.snapshot()
```

However, at this point ,`solver.iter` has increased by one already and when a `solver.snapshot()` is issued, it saves the "**Next iterations model**". 
Is there a way around this? or is it me doing something wrong here? 
Trying to get the accuracy according to the examples in [IPython notebooks](https://github.com/BVLC/caffe/blob/master/examples/01-learning-lenet.ipynb) also doesn't work. Simply doing : 
```
def run_test(solver, test_iter):
    acc = 0
    batch_size_test =  solver.test_nets[0].blobs['data'].data.shape[0]
    
    for i in range(test_iter):
        #testing the network on all test set and calculate the test accuracy
        solver.test_nets[0].forward()
        acc += solver.test_nets[0].blobs['accuracy'].data
    
    final_result= '{0:.4f}'.format( acc / test_iter ) 
  return float(final_result)
```
Won't capture the best model, it captures the next immediate one instead. 
So Simply put, it if you have set your test_interval to 500, and you get a model with x accuracy. 
you wont be able to save that specific model, in that specific iteration. 
All you can do is to save the next immediate one. 
Any idea how to fix this?

PR #3083, opened 2015-09-18
refine format of switch case statement in solver

yanchen036:
refine format of switch case statement in solver


longjon:
This does appear to more consistent with our existing `switch` statements, thanks!

PR #3095, opened 2015-09-20
Fix broken DeconvolutionLayer GPU backward caused by ND conv typo

longjon:
#2049 broke deconvolution GPU backward with a typo (https://github.com/BVLC/caffe/pull/2049/files#diff-5911f1d28796ada5f80449e8bd05d270R54 omits the `n *` coefficient). This PR fixes.

This should really really be covered by the tests, but this PR is just the fix for master.

### DIFF
--- a/src/caffe/layers/deconv_layer.cu
+++ b/src/caffe/layers/deconv_layer.cu
@@ -51,7 +51,7 @@ void DeconvolutionLayer<Dtype>::Backward_gpu(const vector<Blob<Dtype>*>& top,
         }
         // gradient w.r.t. bottom data, if necessary.
         if (propagate_down[i]) {
-          this->forward_gpu_gemm(top_diff + this->top_dim_, weight,
+          this->forward_gpu_gemm(top_diff + n * this->top_dim_, weight,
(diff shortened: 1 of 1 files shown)


jeffdonahue:
Yikes! Thanks for the fix. Looking quickly (since I want to merge ASAP) I don't understand why the buggy version passes the current gradient checks...it seems like they shouldn't pass unless we were only testing the `n == 1` case, which we're not -- the tests have `n == 2`.

shelhamer:
@jeffdonahue @longjon It is clearly good that the indexing of the top diff is fixed but oddly enough the code was correct as it was. The correctness hinges on passing `this->param_propagate_down_[0]` as the `skip_im2col` arg of `forward_gpu_gemm`: in this way the weight gradient call first `im2col`s the right part of the top diff according to `top_diff + n * this->top_dim_` and then the bottom gradient call ignores its own input arg in favor of the cached `im2col` result. As the gradient checker backprops to weights and bottom(s) in `CheckGradientExhaustive()` this condition always holds.

All-in-all this was subtle and weird but right.

PR #3096, opened 2015-09-20
Fix broken conv/deconv reshaping

longjon:
#2049 moved reading and storing of `bottom` shapes to `LayerSetUp`, which (silently) breaks reshaping to different input shapes.

This PR fixes that issue. I'm also not sure why another blob was introduced to store the input shape; this PR uses an inline function that wraps `bottom[0]->shape` instead. (Note that that's now somewhat awkwardly stored as `bottom_shape_` in order to get around the fact that `bottom` is neither a member variable nor argument to functions to such as `compute_output_shape`.)

Again, this should definitely be tested, but this PR is just the fix...

### DIFF
--- a/include/caffe/vision_layers.hpp
+++ b/include/caffe/vision_layers.hpp
@@ -58,6 +58,10 @@ class BaseConvolutionLayer : public Layer<Dtype> {
   void backward_gpu_bias(Dtype* bias, const Dtype* input);
 #endif
 
+  /// @brief The spatial dimensions of the input.
+  inline int input_shape(int i) {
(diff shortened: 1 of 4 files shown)


shelhamer:
This looks right to me but @jeffdonahue should check as the original author.

jeffdonahue:
Looks reasonable to me, but hard to know without any tests.  Since this is a bugfix though, I think this can be merged given the existing tests pass and that you've successfully used this for reshaping in your work @longjon.

longjon:
I'll go ahead and merge then since this passes the tests and a cursory check which the old code failed, but we should still fix up the tests so that this is checked more carefully.

PR #3099, opened 2015-09-21
Separate LIB addition to Makefile.config

nrailgun:
I guess avoiding directly changing `Makefile` might be a good idea. I added variable `ADDITIONAL_LIBS` so those who want to add layers for themselves only don't have to modify `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ ifneq ($(CPU_ONLY), 1)
 	LIBRARIES := cudart cublas curand
 endif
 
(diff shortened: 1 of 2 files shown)


nrailgun:
Some test cases failed without my modifications. I don't know reason.

shelhamer:
Sorry, I don't see the need for this as-is since, and were further libraries needed I'd rather adjust the Makefile to grow by `LIBRARIES += ` so that it could be defined in the Makefile.config too.

Issue #2334, opened 2015-04-18
MemoryDataLayer Problem


TJKlein:
Hi,

when choosing MemoryDataLayer as input in Python I get segmentation faults and Check failed: status == CUBLAS_STATUS_SUCCESS (14 vs. 0)  CUBLAS_STATUS_INTERNAL_ERROR.
For small data sets it runs for a while until the Python kernel crashes with the above error. For large data sets this error occurs immediately. Tested in on both OSX and Linux, with the latest master branch. Same issue

However, when using HDF5 layer input for the exactly same data / network, it works perfectly. I used the networks as defined in the tutorial - with the adaption of MemoryDataLayer: 

http://nbviewer.ipython.org/github/BVLC/caffe/blob/tutorial/examples/01-learning-lenet.ipynb

I am also training / testing on MNIST data set

```
layer {
  name: "data"
  type: "MemoryData"
  top: "data"
  top: "label"
  memory_data_param {
    batch_size: 50
    channels: 1
    height: 28
    width: 28
  }
}
layer {
  name: "conv1"
  type: "Convolution"
  bottom: "data"
  top: "conv1"
  convolution_param {
    num_output: 20
    kernel_size: 5
    weight_filler {
      type: "xavier"
    }
  }
}
layer {
  name: "pool1"
  type: "Pooling"
  bottom: "conv1"
  top: "pool1"
  pooling_param {
    pool: MAX
    kernel_size: 2
    stride: 2
  }
}
layer {
  name: "conv2"
  type: "Convolution"
  bottom: "pool1"
  top: "conv2"
  convolution_param {
    num_output: 50
    kernel_size: 5
    weight_filler {
      type: "xavier"
    }
  }
}
layer {
  name: "pool2"
  type: "Pooling"
  bottom: "conv2"
  top: "pool2"
  pooling_param {
    pool: MAX
    kernel_size: 2
    stride: 2
  }
}
layer {
  name: "ip1"
  type: "InnerProduct"
  bottom: "pool2"
  top: "ip1"
  inner_product_param {
    num_output: 500
    weight_filler {
      type: "xavier"
    }
  }
}
layer {
  name: "relu1"
  type: "ReLU"
  bottom: "ip1"
  top: "ip1"
}
layer {
  name: "ip2"
  type: "InnerProduct"
  bottom: "ip1"
  top: "ip2"
  inner_product_param {
    num_output: 10
    weight_filler {
      type: "xavier"
    }
  }
}
layer {
[...]

PR #3100, opened 2015-09-21
add transformation to memory data layer

bwilbertz:
This PR adds transformations according to TransformParam to the MemoryDataLayer for the case that data is directly set by a pointer via reset. The transformations (on the blobs) are done on-the-fly during Forward. Including test cases.

### DIFF
--- a/include/caffe/data_layers.hpp
+++ b/include/caffe/data_layers.hpp
@@ -301,6 +301,8 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
   Blob<Dtype> added_data_;
   Blob<Dtype> added_label_;
   bool has_new_data_;
+  Blob<Dtype> transform_data_;
+  bool is_transformation_applied_;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


dfagnan:
This is exactly what I was looking for and I'm not sure why it wasn't supported. Can you elaborate what you mean by set by a pointer via reset? Would this also work using python.set_input_arrays? (I got a cublas error attempting to use this through python and I'm trying to diagnose).

dfagnan:
@bwilbertz It looks like this code assumes height and width are the same for the transformed data, which is usually not the case, am I missing something?

bwilbertz:
Hi dfagnan.

Yes, python.set_input_arrays uses this mechanism. 

What do you exactly mean by "height/width are preserved during transformation"?

This PR uses (like in the DataLayer) DataTransformer<Dtype>::Transform(Blob<Dtype>* input_blob, Blob<Dtype>* transformed_blob) for the transformation.

That means, if you have input pictures in 1024x768, you would specify this height/width in the proto file and when setting a crop of 256, all the output blobs would have size 256x256.

dfagnan:
Sorry for the delay in following up but this code worked exactly as I hoped. My error was an unrelated bug with memory data layer which has been addressed but not merged here: #2334. Thanks for the great work. I still don't understand the details of the code, but will leave that for another day!

bwilbertz:
Great :)

shelhamer:
Closing according to #5528, but thanks @bwilbertz for proposing an improvement to the layer (in good form too with a test).

PR #3101, opened 2015-09-21
Updated relighting augmentation by yanii

dfagnan:
This is an attempt at an updated pull request on the relighting augmentation work done by @yanii in #1865. I'm happy to perform any remaining steps in order to have this reviewed. I think this an important feature that really needs to make it into master. Testing for my own examples has reduced over-fitting significantly.

### DIFF
--- a/include/caffe/data_transformer.hpp
+++ b/include/caffe/data_transformer.hpp
@@ -5,6 +5,7 @@
 
 #include "caffe/blob.hpp"
 #include "caffe/common.hpp"
+#include "caffe/filler.hpp"
 #include "caffe/proto/caffe.pb.h"
 
 namespace caffe {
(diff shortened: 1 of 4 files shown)


mtamburrano:
I've not tried this augmentation yet, but I'm interested on this PR

hyojinie:
Has anyone tested this code? Can I directly use it in current version of Caffe or the version right before?

Also, why is "repeated float eigen_value = 8;"  instead of 3?

yxchng:
@hyojinie have u tried? Does this work? Also are you able to calculate PCA for the entire imagenet dataset? My computer is not sufficient to calculate PCA for such a large dataset. If you have, can you share the result? Thanks.

hyojinie:
It does work. I subsampled images in my training set to compute PCA.

PR #3103, opened 2015-09-22
CNN triplet loss version with flexible triplet sets

Wangyida:
* Ready for a review
* Part of GSoC 2015

### DIFF
--- a/include/caffe/syncedmem.hpp
+++ b/include/caffe/syncedmem.hpp
@@ -13,20 +13,22 @@ namespace caffe {
 // The improvement in performance seems negligible in the single GPU case,
 // but might be more significant for parallel training. Most importantly,
 // it improved stability for large models on many GPUs.
-inline void CaffeMallocHost(void** ptr, size_t size) {
+inline void CaffeMallocHost(void** ptr, size_t size, bool* use_cuda) {
 #ifndef CPU_ONLY
   if (Caffe::mode() == Caffe::GPU) {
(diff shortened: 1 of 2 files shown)


bhack:
Hi probably the history it is still too corrupted and full of "work in progress" to be reviewed. I don't know if it his better to squash the history in a single commit or you can re-edit  you history to let review new introduced functionalities by commit. Every isolable feature you want to introduce in caffe could be reviewable as a commit. It hard to review work in progress commit, commit where you introduce something that then you remove etc. So it is better to have a single squashed commit or to reorganize the history with squashing to have very few commits strictly related to the new features you propose to introduced in caffe.

Wangyida:
Codes are collected, the commits are the same as before, when they are squashed, it will be easier to be reviewed. @bhack

Wangyida:
@bhack  @mtamburrano The commits are clear now, the main inappropriate act is that I used to using my master branch for development on codes and some commits are just pushed on my master branch, so when I have my development on cnn_triplet branch and have each rebase on the updated master branch, the commits on my master branch will also be restored on the new branch. I clean the commits on master branch and have a rebase again, they are tidy at last. Codes differs a lot from the beginning, So you could just have a look on whole.

bhack:
Seems to me that you have introduced spurious changes in https://github.com/BVLC/caffe/commit/e4db1311d0d7da2da1666fc0b5cd77e06a07b076. See the diff of each file.

PR #3112, opened 2015-09-23
Test Reshape More Rigorously

shelhamer:
Follow-up to #3096.

- Check that the output shape actually varies with the input shape while making sure the input and output num match.
- Check small then large to catch potential issues with realloc.
- `EXPECT*` instead of `CHECK*` to see full results.

master @ 58e4ea4 fails the new test while master @ head passes.

### DIFF
--- a/src/caffe/test/test_net.cpp
+++ b/src/caffe/test/test_net.cpp
@@ -2262,8 +2262,8 @@ TEST_F(FilterNetTest, TestFilterInOutByExcludeMultiRule) {
 TYPED_TEST(NetTest, TestReshape) {
   typedef typename TypeParam::Dtype Dtype;
   // We set up bottom blobs of two different sizes, switch between
-  // them, and check that forward and backward both run and the results
-  // are the same.
+  // them, check that forward and backward both run and the results
+  // are the same, and check that the output shapes change.
   Caffe::set_random_seed(this->seed_);
   Caffe::set_mode(Caffe::CPU);
   FillerParameter filler_param;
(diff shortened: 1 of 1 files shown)


src/caffe/test/test_net.cpp
```diff
@@ -2262,15 +2262,16 @@ TEST_F(FilterNetTest, TestFilterInOutByExcludeMultiRule) {
...
+  // them, check that forward and backward both run and the results
+  // are the same, and check that the output shapes change.
   Caffe::set_random_seed(this->seed_);
   Caffe::set_mode(Caffe::CPU);
   FillerParameter filler_param;
   filler_param.set_std(1);
   GaussianFiller<Dtype> filler(filler_param);
-  Blob<Dtype> blob1(4, 3, 9, 11);
-  Blob<Dtype> blob2(2, 3, 12, 10);
+  // Check smaller shape first as larger first could hide realloc failures.
```
jeffdonahue: Could add an `ASSERT_GT` for the resulting output counts.  (It's not immediately clear to me which one should give the bigger result.)

src/caffe/test/test_net.cpp
```diff
@@ -2313,8 +2314,20 @@ TYPED_TEST(NetTest, TestReshape) {
...
   this->net_->Backward();
   for (int i = 0; i < output2.count(); ++i) {
-    CHECK_EQ(*(output2.cpu_data() + i), *(output_blob->cpu_data() + i));
+    EXPECT_FLOAT_EQ(*(output2.cpu_data() + i), *(output_blob->cpu_data() + i));
   }
+
+  EXPECT_EQ(output1.num(), blob1.num());
+  EXPECT_EQ(output2.num(), blob2.num());
+  bool same_spatial_shape = true;
+  int kFirstSpatialAxis = 2;
```
jeffdonahue: `const int`

jeffdonahue:
Thanks for the test @shelhamer; LGTM besides above nits.

shelhamer:
@jeffdonahue thanks for reviewing -- updated accordingly for merge when CI passes.

shelhamer:
CI failed for an irrelevant timeout issue but I've checked this independently so it's merge time.

PR #3115, opened 2015-09-24
Fix generate_sample_data.py

lukeyeager:
*Bug from #2978*

Somewhere in development, I had considered changing the filename. I forgot to change the `generate_sample_data.py` script back to use the original filename before committing.


jeffdonahue:
Thanks for catching and fixing this @lukeyeager!

PR #3118, opened 2015-09-25
Fix parse_log.sh against "prefetch queue empty" messages

ducha-aiki:
With new data layer sh plot parser sometimes parses incorrectly because of "prefetch queue empty" and "Waiting for data" lines. This PR fixes by taking lines between starting testing and continue training lines, then clean up them.


shelhamer:
Thanks for catching this.

PR #3127, opened 2015-09-28
Install libs as non-executable files

lukeyeager:
According to the Debian policy manual:

>Shared libraries should not be installed executable, since the dynamic linker does not require this and trying to execute a shared library usually results in a core dump."

https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#s-sharedlibs-runtime

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -652,7 +652,7 @@ $(DISTRIBUTE_DIR): all py | $(DISTRIBUTE_SUBDIRS)
 	cp $(EXAMPLE_BINS) $(DISTRIBUTE_DIR)/bin
 	# add libraries
 	cp $(STATIC_NAME) $(DISTRIBUTE_DIR)/lib
-	cp $(DYNAMIC_NAME) $(DISTRIBUTE_DIR)/lib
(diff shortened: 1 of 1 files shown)


lukeyeager:
Travis failed with:

    The command "git clone --depth=50 https://github.com/BVLC/caffe.git BVLC/caffe" failed and exited with 128 during .

Can someone restart it please?

lukeyeager:
It looks like @CDLuminate is using the CMake build for debian packaging. 
http://anonscm.debian.org/cgit/debian-science/packages/caffe.git/tree/debian/rules

CMake already does this correctly.

CDLuminate:
@lukeyeager Yes, using `CMake` for Debian packaging.  
  
By my side, the mode of shared lib doesn't disturb since
* CMake handles shared lib mode.
* dh_fixperms will fix it if anything wrong according to `Policy`.
(the permission checker in debian packaging suite)
  
FYI: Generally `.so` permission is `Access: (0644/-rw-r--r--)`.

shelhamer:
Thanks for the Debian conformance @lukeyeager.

PR #3132, opened 2015-09-30
Fix loss of last iteration when average_loss > 1

bwilbertz:
Hi,

The loss which is computed for the last iteration in Solver::Solve(...) does not contain any smoothing as it specified by average_loss and as it is applied in all the previous iterations. 
As a consequence, one currently gets a nice outlier (because it's based on only one forward pass) for the final loss...

This PR also applies smoothing for the very last iteration.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -107,6 +107,7 @@ class Solver {
   virtual void RestoreSolverStateFromHDF5(const string& state_file) = 0;
   virtual void RestoreSolverStateFromBinaryProto(const string& state_file) = 0;
   void DisplayOutputBlobs(const int net_id);
+  void UpdateSmoothedLoss(Dtype loss, int start_iter, int average_loss);
 
   SolverParameter param_;
   int iter_;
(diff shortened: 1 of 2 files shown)


jeffdonahue:
Thanks @bwilbertz, this has bothered me as well.  Please factor the 8 lines of newly-duplicated loss smoothing code into  a separate method.

bwilbertz:
@jeffdonahue: I have re-factored the duplicate code into a separate function and also merged in the current master, since master and this branch had diverged.

shelhamer:
@bwilbertz please rebase onto the latest master instead of merging for clear history once your PR is merged.

bwilbertz:
@shelhamer I did a rebase onto master and squashed my changes into one single commit

include/caffe/solver.hpp
```diff
@@ -107,6 +107,7 @@ class Solver {
   virtual void RestoreSolverStateFromHDF5(const string& state_file) = 0;
   virtual void RestoreSolverStateFromBinaryProto(const string& state_file) = 0;
   void DisplayOutputBlobs(const int net_id);
+  void updateSmoothedLoss(Dtype loss, int start_iter, int average_loss);
```
seanbell: This should be `UpdateSmoothedLoss` to be consistent with other method names.

(I personally don't like the caffe capitalization style, but consistency is important)

bwilbertz:
@seanbell: Thanks for the hint. I have update the commit accordingly.

jeffdonahue:
Thanks for this fix @bwilbertz! LGTM.

PR #3138, opened 2015-10-02
add a way to construct a Blob with externally allocated GPU memory

vakopian:
When caffe is used with other libraries that manage their own GPU memory (e.g. ArrayFire), it would be useful to be able to create a `Blob` by passing the GPU memory pointer that was allocated in the 3rd party library directly to it. Otherwise we would have to do an extra copy to the host then back to the device.
This pull request adds the special constructors for `SyncedMemory` and `Blob` to handle this scenario.
An example of the usage would be
```
af::array my_array = af::randu(3, 4, 5);
std::vector<int> shape = {{3, 4, 5}};
Blob<float> blob = new Blob<float>(shape, my_array.device(), 0);
```

### DIFF
--- a/include/caffe/blob.hpp
+++ b/include/caffe/blob.hpp
@@ -32,6 +32,8 @@ class Blob {
       const int width);
   explicit Blob(const vector<int>& shape);
 
+  Blob(const vector<int>& shape, void * pre_allocated_gpu_ptr, int gpu_device);
+
   /// @brief Deprecated; use <code>Reshape(const vector<int>& shape)</code>.
   void Reshape(const int num, const int channels, const int height,
       const int width);
--- a/include/caffe/syncedmem.hpp
+++ b/include/caffe/syncedmem.hpp
@@ -53,6 +53,14 @@ class SyncedMemory {
       : cpu_ptr_(NULL), gpu_ptr_(NULL), size_(size), head_(UNINITIALIZED),
         own_cpu_data_(false), cpu_malloc_use_cuda_(false), own_gpu_data_(false),
         gpu_device_(-1) {}
+  SyncedMemory(void * pre_allocated_gpu_ptr, size_t size, int gpu_device)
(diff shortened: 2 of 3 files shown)


willyd:
Hi think it would probably be best to use the `SyncedMem::set_gpu_data`. You could add the method to the `Blob` class for convenience. Otherwise, you can use:
```C++
Blob<Dtype>* blob = ...
blob->data()->set_gpu_data(externally_allocated_gpu_memory)
```

bhack:
It could be interesting to extend also memory data layer to accept [opencv UMAT](http://docs.opencv.org/master/d7/d45/classcv_1_1UMat.html#gsc.tab=0)

bhack:
Probably here  we are talking of cuda side of the GPU moon. @naibaf7 memory layer support of OpenCV UMAT could be interesting in your opencl branch.

vakopian:
@willyd, I'm curious as to why you think using the `set_gpu_data()` is better? I would think initializing the objects the way we want them right away would be preferred to initialize+set approach.
Is it just to avoid adding a new constructor to `SyncedMem`?

willyd:
@vakopian IMHO I think `set_gpu_data` makes the intent more explicit. Also, it keeps things *symmetric* between cpu and gpu. With the constructor approach you have to choose between a cpu or gpu version or have a constructor that accepts two pointers and then you need to look up the arguments name to know where to pass cpu pointer or gpu pointer.

naibaf7:
@willyd 
I agree, CPU and GPU should be completely symmetric. But I also do not like how memory location is handled in Caffe currently. I.e. it is becoming a mess with multi-GPU approaches.
Definitely an urgent need for a device abstraction.

vakopian:
I don't really have strong feelings as to which way is better. I guess if one of the committers tells us that the PR would be accepted if we change it to the `set_gpu_data` method, I have no problem changing it.
However, notice that we would need a `set_gpu_data` overload - one that also takes `gpu_device` as argument.

PR #3153, opened 2015-10-05
NetSpec: type-check Function inputs (they must be Top instances)

jeffdonahue:
This adds simple type-checking for a common (at least for me) class of NetSpec use errors: the use of non-`Top` instances as inputs to `Function`s.  The exception is raised when the `Function` is `__init__`ed rather than lazily when `to_proto` is called, making it easier to debug (IMO).

### DIFF
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -103,6 +103,10 @@ class Function(object):
 
     def __init__(self, type_name, inputs, params):
         self.type_name = type_name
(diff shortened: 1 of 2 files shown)


jeffdonahue:
(Fails Python3 Travis checks due to apparently different behavior of either regexps or `unittest.assertRaisesRegexp`; will look into debugging that pending @longjon's general approval of this.)

shelhamer:
Thanks for the net spec check Jeff!

PR #3160, opened 2015-10-06
Basic cuDNN v3 support (update)

shelhamer:
This is the same as #2737 except for

- Restoring the pooling layer fallback for max + argmax output when the layer is configured to have two tops. The padding is dropped since it is now supported by cuDNN pooling.
- Clearing the warnings for unused variables in `cuDNNConvolutionLayer::Forward_gpu()` now that algo and workspace are determined in `Reshape()`.

Thanks @slayton58 for the integration.

### DIFF
--- a/src/caffe/layers/cudnn_conv_layer.cu
+++ b/src/caffe/layers/cudnn_conv_layer.cu
@@ -14,11 +14,6 @@ __global__ void sync_conv_groups() { }
 template <typename Dtype>
 void CuDNNConvolutionLayer<Dtype>::Forward_gpu(
     const vector<Blob<Dtype>*>& bottom, const vector<Blob<Dtype>*>& top) {
-  const int* kernel_shape_data = this->kernel_shape_.cpu_data();
-  const int kernel_h = kernel_shape_data[0];
-  const int kernel_w = kernel_shape_data[1];
-  const size_t workspace_limit_bytes =
-      kernel_h * kernel_w * this->channels_ * sizeof(int) + 1;
   const Dtype* weight = this->blobs_[0]->gpu_data();
   for (int i = 0; i < bottom.size(); ++i) {
     const Dtype* bottom_data = bottom[i]->gpu_data();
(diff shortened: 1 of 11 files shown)


slayton58:
Regarding the test failure for Groups -- it seems like this->weight_offset_ for the cuDNN routines is getting set incorrectly / not set (either way, it's wrong!) This seems to have been introduced in https://github.com/BVLC/caffe/commit/9d8206e0f906069e7c04f08dfddefa1357f3915c

Setting it back explicitly to:
this->weight_offset_ = (this->num_output_ / this->group_) * (this->channels_ / this->group_) * kernel_h * kernel_w;

in CuDNNConvolutionLayer::Setup seems to fix the issue, not sure if there's a better way - let me know and I'll update the PR

ronghanghu:
:+1:

shelhamer:
cuDNN v3 is not itself backward compatible with v2, so adopting v3 in this PR does deprecate v2. We plan to follow the latest cuDNN version in master but keep compatability as the cuDNN interface itself allows.

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -101,10 +57,12 @@ void CuDNNConvolutionLayer<Dtype>::Backward_gpu(const vector<Blob<Dtype>*>& top,
   if (this->param_propagate_down_[0]) {
     weight = this->blobs_[0]->gpu_data();
     weight_diff = this->blobs_[0]->mutable_gpu_diff();
+    caffe_gpu_set(this->blobs_[0]->count(), Dtype(0), weight_diff);
   }
   Dtype* bias_diff = NULL;
   if (this->bias_term_ && this->param_propagate_down_[1]) {
     bias_diff = this->blobs_[1]->mutable_gpu_diff();
+    caffe_gpu_set(this->blobs_[1]->count(), Dtype(0), bias_diff);
```
ronghanghu: @shelhamer @slayton58 I am a bit confused here... Why do we need to zero out the diff? It confuses me as parameter gradients should be accumulated.

shelhamer: This is definitely a bug -- thanks for the fix in #3254.

PR #3162, opened 2015-10-07
Improve numerical stability of variance computation in MVNLayer

kkhoot:
This PR is to avoid numerical precision error in MVNLayer when the input image has large intensity values and relatively small variance, which possibly causes negative values in calculating var(X) = E(X^2) - (EX)^2.

I found this error when dealing with some real data having almost white color intensities resulting in NaN loss or loss divergence. 

Solution to this problem is to compute the mean intensity and subtract it first, and compute variance using var(X) = E((X-EX)^2) - E(X-EX)^2, which simply is E((X-EX)^2).

Updated: removed redundant code (subtracting mean) in the forward functions.

### DIFF
--- a/src/caffe/layers/mvn_layer.cpp
+++ b/src/caffe/layers/mvn_layer.cpp
@@ -42,29 +42,21 @@ void MVNLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
 
   int dim = bottom[0]->count() / num;
 
-  if (this->layer_param_.mvn_param().normalize_variance()) {
-    // put the squares of bottom into temp_
-    caffe_powx(bottom[0]->count(), bottom_data, Dtype(2),
-        temp_.mutable_cpu_data());
+  // subtract mean
+  caffe_cpu_gemv<Dtype>(CblasNoTrans, num, dim, 1. / dim, bottom_data,
+      sum_multiplier_.cpu_data(), 0., mean_.mutable_cpu_data());  // EX
+  caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, num, dim, 1, -1.,
+      mean_.cpu_data(), sum_multiplier_.cpu_data(), 0.,
+      temp_.mutable_cpu_data());
+  caffe_add(temp_.count(), bottom_data, temp_.cpu_data(), top_data);  // X-EX
 
-    // computes variance using var(X) = E(X^2) - (EX)^2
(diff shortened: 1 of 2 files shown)


src/caffe/layers/mvn_layer.cpp
```diff
@@ -42,29 +42,21 @@ void MVNLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
 
   int dim = bottom[0]->count() / num;
 
-  if (this->layer_param_.mvn_param().normalize_variance()) {
-    // put the squares of bottom into temp_
-    caffe_powx(bottom[0]->count(), bottom_data, Dtype(2),
-        temp_.mutable_cpu_data());
+  // subtract mean
+  caffe_cpu_gemv<Dtype>(CblasNoTrans, num, dim, 1. / dim, bottom_data,
+  sum_multiplier_.cpu_data(), 0., mean_.mutable_cpu_data());  // EX
```
ronghanghu: Please indent 4 spaces in continuing line, to be consistent with our code style.

src/caffe/layers/mvn_layer.cu
```diff
@@ -20,29 +20,22 @@ void MVNLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
...
-  if (this->layer_param_.mvn_param().normalize_variance()) {
-    // put the squares of bottom into temp_
-    caffe_gpu_powx(bottom[0]->count(), bottom_data, Dtype(2),
-        temp_.mutable_gpu_data());
+  // subtract mean
+  caffe_gpu_gemv<Dtype>(CblasNoTrans, num, dim, 1. / dim, bottom_data,
+      sum_multiplier_.gpu_data(), 0., mean_.mutable_gpu_data());  // EX
+  caffe_gpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, num, dim, 1, -1.,
+          mean_.gpu_data(), sum_multiplier_.gpu_data(), 0.,
+          temp_.mutable_gpu_data());
```
ronghanghu: Please indent 4 spaces.

ronghanghu:
@kkhoot Thanks for this PR!

Please fix the style issue commented above. Everything else looks good to me :)

kkhoot:
@ronghanghu  Thanks for reviewing this PR. :) The style issue has been fixed. 
However many lines in the rest of the code in MVNLayer have irregular space indents (from 2 to 8). If you think it is better to fix all of them, I will fix them and push it again.

ronghanghu:
Let's just leave the rest code as-is, to provide a more readable diff in this PR :)

Thanks @kkhoot !

PR #3170, opened 2015-10-08
Remove the 4D constraint of blobproto IO in python

sh1r0:
As the 4D dimensions (num, channels, height, width) are deprecated for BlobProto, the IO should not have a condition that the shape is 4D only.


jeffdonahue:
LGTM. Thanks @sh1r0!

PR #3180, opened 2015-10-12
Fix predict(..oversample=False) to use integer slices.

davidbau:
net.predict(... oversample=False) fails to work due to the fact that when it does a single center crop, the numbers crop[0] etc are not ints on the affected lines.  This change converts these numbers to ints before using them as slice indices.


davidbau:
Note - integer subscripts are needed on the latest numpy versions - see https://github.com/numpy/numpy/issues/2810

PR #3182, opened 2015-10-13
Fix class label index error in tutorial 00.

bchu:
`print("Predicted class is #{}.".format(out['prob'].argmax()))`
should be:
`print("Predicted class is #{}.".format(out['prob'][0].argmax()))`

in https://github.com/BVLC/caffe/blob/master/examples/00-classification.ipynb

### DIFF
--- a/examples/00-classification.ipynb
+++ b/examples/00-classification.ipynb
@@ -119,7 +119,7 @@
    "source": [
     "net.blobs['data'].data[...] = transformer.preprocess('data', caffe.io.load_image(caffe_root + 'examples/images/cat.jpg'))\n",
     "out = net.forward()\n",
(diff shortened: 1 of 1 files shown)


antran89:
Hi,
If ```out['prob']``` is always a list just having only one element, these original version is still okay. Right?

bchu:
`out['prob'].shape == (50, 1000)` (the test batch size is 50, with 1000 categories).

This was a weird one to track down. argmax() by default gives the index into the flattened array. This normally wouldn't be a problem in the tutorial (all 50 images are the same, so you would expect each row of `out['prob']` to be the same), but when compiled with OpenBLAS, not each row of `out['prob']` is the same.

shelhamer:
Thanks for this fix to clarify the example.

PR #3184, opened 2015-10-13
Update store2hdf5.m

lalelale:
Fixed a bug in two assertions (the condition input argument must be a scalar logical)


ronghanghu:
Thanks @lalelale !

PR #3185, opened 2015-10-13
PReLU speed up

mausoomsarkar:
Moved the loop inside PReLUParamBackward to do the reduction inside the kernel

Now PReLU backward is taking the same time as forward

### DIFF
--- a/src/caffe/layers/prelu_layer.cu
+++ b/src/caffe/layers/prelu_layer.cu
@@ -31,10 +31,15 @@ __global__ void PReLUBackward(const int n, const int channels, const int dim,
 
 // CUDA kernel for element-wise parameter backward
 template <typename Dtype>
-__global__ void PReLUParamBackward(const int n, const Dtype* in_diff,
+__global__ void PReLUParamBackward(const int n,
+    const int rows, const int rowPitch, const Dtype* in_diff,
     const Dtype* in_data, Dtype* out_diff) {
   CUDA_KERNEL_LOOP(index, n) {
     out_diff[index] = in_diff[index] * in_data[index] * (in_data[index] <= 0);
+    for ( int k = 1; k < rows; k++ ) {
+        out_diff[index] += in_diff[index + k*rowPitch]
+           * in_data[index + k*rowPitch] * (in_data[index + k*rowPitch] <= 0);
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/caffe/layers/prelu_layer.cu
```diff
@@ -83,13 +88,13 @@ void PReLULayer<Dtype>::Backward_gpu(const vector<Blob<Dtype>*>& top,
     Dtype* slope_diff = this->blobs_[0]->mutable_gpu_diff();
     int cdim = channels * dim;
     Dtype dsum = 0.;
-    for (int n = 0; n < bottom[0]->num(); ++n) {
+
```
ronghanghu: I don't think you can simply remove the loop here. The `caffe_gpu_dot` and `caffe_gpu_gemv` needs to be in this loop.

mausoomsarkar: Yes, that had been true had the multiplier been variable for every N. But because it is 1 the caffe_gpu_dot is essentially doing a sum across the CxHxW and the loop is adding that sum for all the inputs in the batch.
Also for caffe_gpu_gemv the sum of HxW for every channel C is computed which is then added up for all inputs. 
So what I am doing is that for all the N (inputs) in NxCxHxW blob, I am calculating sum across N for each of the CxHxW elements. After this just a single call to caffe_gpu_dot or caffe_gpu_gemv should work.

ronghanghu: You are right. I see what you are doing here. Thanks for this! I'll take a look today.

src/caffe/layers/prelu_layer.cu
```diff
@@ -102,7 +107,7 @@ void PReLULayer<Dtype>::Backward_gpu(const vector<Blob<Dtype>*>& top,
             backward_buff_.gpu_diff(), multiplier_.gpu_data(), 1.,
             slope_diff);
       }
-    }
+
     if (channel_shared_) {
       caffe_gpu_add_scalar(this->blobs_[0]->count(), Dtype(dsum), slope_diff);
```
ronghanghu: After removing the loop, let's put merge this `if (channel_shared_)` into the `if (channel_shared_)` above.

src/caffe/layers/prelu_layer.cu
ronghanghu: After removing the loop, let's delete the fix the indent here. No block indent needed here after removing the loop.

ronghanghu:
@mausoomsarkar Thanks for this PR!

Let's squash the two commits above into one single commit, and I'll merge.

ronghanghu:
Thanks @mausoomsarkar !

PR #3187, opened 2015-10-13
Add pyyaml as a requirement

akash1810:
In getting the [web demo](http://caffe.berkeleyvision.org/gathered/examples/web_demo.html) started I get an `ImportError: No module named yaml` error when running `./scripts/download_model_binary.py models/bvlc_reference_caffenet`.


shelhamer:
Thanks for adding the missing demo dependency!

PR #3189, opened 2015-10-13
Set CaffeNet train_val test mirroring to false

### LINKED ISSUE #3183: CaffeNet solver.prototxt test mirroring should not be true?
In https://github.com/BVLC/caffe/blob/master/models/bvlc_reference_caffenet/train_val.prototxt, in the test phase data layer, shouldn't `#    mirror: true` be `#    mirror: false`?

Would be happy to submit a quick PR is this is a bug.

bchu:
Resolves #3183


shelhamer:
Thanks for the cleanup.

PR #3190, opened 2015-10-14
Fixed drawing problems with repeated convolution

DrSleep:


shelhamer:
Thanks for the fix!

nnop:
Is there any documentation for the protobuf `_value` variable?

PR #3200, opened 2015-10-15
Allow old-style shape in blobproto_to_array

### LINKED ISSUE #3199: error in reading binaryproto in Python
Hi,
Today I met the odd. 
I use the following snippet of code to reading binaryproto into npy array. The code is very standard that I found in #290, #808.
```
#test caffe
import caffe

MEAN_FILE = '/home/tranlaman/BLVC-caffe/data/ilsvrc12/imagenet_mean.binaryproto'
blob = caffe.proto.caffe_pb2.BlobProto()
data = open(MEAN_FILE, 'rb').read()
blob.ParseFromString(data)
arr = np.array(caffe.io.blobproto_to_array(blob))
out = arr[0]
print out.shape
```
The error is as follows
```
  File "/home/tranlaman/Desktop/python-workspace/test-python/test_python.py", line 34, in <module>
    arr = np.array(caffe.io.blobproto_to_array(blob))
  File "/home/tranlaman/BLVC-caffe/python/caffe/io.py", line 26, in blobproto_to_array
    return np.array(blob.data).reshape(*blob.shape.dim)
TypeError: function takes exactly 1 argument (0 given)
```

My caffe was installed properly. I run this code just fine with Caffe version of 1 month ago.
I am just asking is there any major changes in Caffe recent days.
Thanks,

lukeyeager:
Fixes #3199
Bug introduced in #3170

### DIFF
--- a/python/caffe/io.py
+++ b/python/caffe/io.py
@@ -20,11 +20,18 @@ def blobproto_to_array(blob, return_diff=False):
     Convert a blob proto to an array. In default, we will just return the data,
     unless return_diff is True, in which case we will return the diff.
     """
+    # Read the data into an array
     if return_diff:
-        return np.array(blob.diff).reshape(*blob.shape.dim)
+        data = np.array(blob.diff)
     else:
-        return np.array(blob.data).reshape(*blob.shape.dim)
+        data = np.array(blob.data)
 
+    # Reshape the array
+    if blob.HasField('num') or blob.HasField('channels') or blob.HasField('height') or blob.HasField('width'):
+        # Use legacy 4D shape
+        return data.reshape(blob.num, blob.channels, blob.height, blob.width)
+    else:
+        return data.reshape(blob.shape.dim)
 
 def array_to_blobproto(arr, diff=None):
     """Converts a N-dimensional array to blob proto. If diff is given, also
--- a/python/caffe/test/test_io.py
+++ b/python/caffe/test/test_io.py
@@ -1,0 +1,41 @@
+import numpy as np
+import unittest
+
+import caffe
+
+class TestBlobProtoToArray(unittest.TestCase):
+
+    def test_old_format(self):
+        data = np.zeros((10,10))
+        blob = caffe.proto.caffe_pb2.BlobProto()
+        blob.data.extend(list(data.flatten()))
+        shape = (1,1,10,10)
+        blob.num, blob.channels, blob.height, blob.width = shape
+
+        arr = caffe.io.blobproto_to_array(blob)
+        self.assertEqual(arr.shape, shape)
+
+    def test_new_format(self):
+        data = np.zeros((10,10))
+        blob = caffe.proto.caffe_pb2.BlobProto()
+        blob.data.extend(list(data.flatten()))
+        blob.shape.dim.extend(list(data.shape))
+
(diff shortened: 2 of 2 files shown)


jeffdonahue:
Thanks @lukeyeager, sorry for neglecting to think about this in #3170. I think the logic should be reversed though -- i.e., should check if any of the legacy fields are present and use them if so, othereise use the shape. This way, a BlobProto with neither the legacy fields nor a shape will be interpreted as having an empty shape (a scalar). This is consistent with the behavior of Blob::FromProto.

lukeyeager:
I've updated the logic and added a third test.

python/caffe/io.py
```diff
@@ -20,11 +20,20 @@ def blobproto_to_array(blob, return_diff=False):
...
+        data = np.array(blob.data)
 
+    if blob.HasField('shape'):
+        return data.reshape(blob.shape.dim)
+    elif blob.HasField('num') or blob.HasField('channels') or blob.HasField('height') or blob.HasField('width'):
+        # Use legacy 4D shape
+        return data.reshape(blob.num, blob.channels, blob.height, blob.width)
+    else:
+        # Return scalar
+        return data
```
jeffdonahue: This wasn't quite what I meant -- this will give a 1D array, not a 0D array (scalar).  You'd still want to `return data.reshape(blob.shape.dim)`, or equivalently `return data.reshape(())`.  The invariant we want is `len(data.shape) == len(blob.shape.dim)` (unless any of the legacy fields are present).  It will only work in the case where `len(data) == 1`, but that's okay since a `BlobProto` with no `shape` field (and no legacy dims) is kind of an odd corner case.

python/caffe/test/test_io.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_old_format(self):
+        data = np.zeros((10,10))
+        blob = caffe.proto.caffe_pb2.BlobProto()
+        blob.data.extend(list(data.flatten()))
+        shape = (1,1,10,10)
+        blob.num, blob.channels, blob.height, blob.width = shape
+
+        arr = caffe.io.blobproto_to_array(blob)
+        self.assertEqual(
```
jeffdonahue: Can you put these asserts on a single line?

lukeyeager: Sure.

python/caffe/io.py
lukeyeager: Well I'm not letting it fall back to this worthless error:
```
  File "/home/tranlaman/BLVC-caffe/python/caffe/io.py", line 26, in blobproto_to_array
    return np.array(blob.data).reshape(*blob.shape.dim)
TypeError: function takes exactly 1 argument (0 given)
```
How about raising a `ValueError`?
```python
          if len(data) != 1:
              raise ValueError('blob has no shape')
          # Return scalar
          return data[0]
```

jeffdonahue: If you remove the `*` (`reshape(blob.shape.dim)`) it should work?

lukeyeager: Oh yeah, that works. Good call.

lukeyeager:
Updated logic again and added a fourth test.

jeffdonahue:
Thanks for the fixes and tests @lukeyeager! LGTM.

PR #3205, opened 2015-10-17
Test Layer Type Upgrade

shelhamer:
The V1 layer type test should always be run regardless of the IO dependencies.

@eelstork can you explain your thinking in f3a933a6? Let me know if this is somehow needed.


eelstork:
Yes, it is bogus. The nested OpenCV guard supersedes the outer, didn't see this, sorry about that.

shelhamer:
No worries. Thanks for confirming.

Issue #3042, opened 2015-09-08
PyCaffe MemoryError loading model


SimoV8:
Hi, I'm trying to load the VGG 16-layer model (downloaded from here http://www.robots.ox.ac.uk/~vgg/research/very_deep/) with pycaffe but I always get a MemoryError message when I try to execute `net = caffe.Net(prototxt, model, caffe.TEST)`. I'm running it on a Jetson TK1 board that has about 2GB of ram and more then 1GB of memory is free to use. I changed the prototxt file reducing the first input dimension from 10 to 1 and I removed the last 3 layers , cutting the network at the relu7 layer, to reduce memory usage. 

When I run the script debug messages say "Memory required for data: 114583552" and then "The total number of bytes read was 553432081". 
Summing up these numbers, it seems that the memory allocation should be around 668 MB. Considering an expansion of the model in memory and other variables, it shouldn't exceed 1GB, that fits with my hw requirements. The problem is that probably I'm missing something because when I run the script it takes much more then 1GB and when it finish all the available memory it crashes. Can someone explain me why? How can I solve that?
I also tried to use the modified version of caffe provided by the #2009 pull request but without success.

PS: I tried both with gpu and cpu mode. The cuda version installed is 6.5 with cuDNN v2.

seanbell:
That reported number is just for blob storage _between_ the layers.  It does not include all of the internal buffers _inside_ each layer (e.g. inside each convolution layer).  As far as I know there's currently no code for counting up all of the internal buffers, as they're hidden inside of the layer classes.

longjon:
Closing as a usage issue; discuss on caffe-users as desired.

From https://github.com/BVLC/caffe/blob/master/CONTRIBUTING.md:
> _Please do not post usage, installation, or modeling questions, or other requests for help to Issues._
Use the [caffe-users list](https://groups.google.com/forum/#!forum/caffe-users) instead. This helps developers maintain a clear, uncluttered, and efficient view of the state of Caffe.

Issue #2387, opened 2015-04-29
What does "Memory required for data:" measure?


mingtop:
Here is the output of debug   minist_letnet  :

I1203 18:47:00.397809  4610 net.cpp:170] relu1 needs backward computation.
I1203 18:47:00.397816  4610 net.cpp:170] ip1 needs backward computation.
I1203 18:47:00.397825  4610 net.cpp:170] pool2 needs backward computation.
I1203 18:47:00.397832  4610 net.cpp:170] conv2 needs backward computation.
I1203 18:47:00.397843  4610 net.cpp:170] pool1 needs backward computation.
I1203 18:47:00.397851  4610 net.cpp:170] conv1 needs backward computation.
I1203 18:47:00.397860  4610 net.cpp:172] label_mnist_1_split does not need backward computation.
I1203 18:47:00.397867  4610 net.cpp:172] mnist does not need backward computation.
I1203 18:47:00.397874  4610 net.cpp:208] This network produces output accuracy
I1203 18:47:00.397884  4610 net.cpp:208] This network produces output loss
I1203 18:47:00.397905  4610 net.cpp:467] Collecting Learning Rate and Weight Decay.
I1203 18:47:00.397915  4610 net.cpp:219] Network initialization done.
I1203 18:47:00.397923  4610 net.cpp:220] Memory required for data: 8086808

memory required for data 8086808   is  the total caffe needed for gpu , or only for input data(mnist)?
the 8006808 is Byte?or bit ?

lukeyeager:
Yeah, the memory requirement reported by caffe seems to have no correlation with either the actual memory used on the GPU or the file size of the .caffemodel.

So, what **does** it mean?

Is it the amount of memory set aside for prefetching the data?

shelhamer:
It's the memory required for all the top blobs or "data" in the sense of layer outputs. It excludes the diffs, the parameters, and any intermediate blobs within layers. @jeffdonahue do you remember the motivation for this number?

lukeyeager:
Ah, so it's `sizeof(data layer input)` + `sizeof(label layer input)`, in most cases? That doesn't seem particularly useful information.

[...]

PR #3206, opened 2015-10-17
Clean Up Logging for Net Init

shelhamer:
- condense conditions by `LOG_IF`
- only log memory usage once after all tops are made

Comment to vote for the current memory usage logging if there's a use case out there.

### DIFF
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -46,10 +46,9 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
   // the current NetState.
   NetParameter filtered_param;
   FilterNet(in_param, &filtered_param);
-  if (Caffe::root_solver()) {
-    LOG(INFO) << "Initializing net from parameters: " << std::endl
-              << filtered_param.DebugString();
-  }
+  LOG_IF(INFO, Caffe::root_solver())
+      << "Initializing net from parameters: " << std::endl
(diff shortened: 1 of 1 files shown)


src/caffe/net.cpp
```diff
@@ -684,25 +660,22 @@ void Net<Dtype>::ForwardDebugInfo(const int layer_id) {
...
     const string& blob_name = blob_names_[top_id_vecs_[layer_id][top_id]];
     const Dtype data_abs_val_mean = blob.asum_data() / blob.count();
-    if (Caffe::root_solver()) {
-      LOG(INFO) << "    [Forward] "
-                << "Layer " << layer_names_[layer_id]
-                << ", top blob " << blob_name
-                << " data: " << data_abs_val_mean;
-    }
-  }
+    LOG_IF(INFO, Caffe::root_solver())
```
ronghanghu: You seem to have removed extra } here.

shelhamer: @ronghanghu fixed, thanks.

seanbell:
I think it makes sense to not print memory usage, since it's not the total memory usage and thus leads to confusion such as #2387 and #3042.

shelhamer:
Dropping the memory usage log line is fine by me. It does cause confusion
and I don't know that anyone is making use of it.

Evan Shelhamer

PR #3217, opened 2015-10-19
Qualify messages issued by CMake when CUDA is unavailable

eelstork:
Unqualified messages incorrectly stall build workers; conversely, errors may go unnoticed if messages are not correctly qualified.


shelhamer:
Thanks -- build polishing is always welcome.

PR #3221, opened 2015-10-20
Endorse CMP0046, CMP0054

eelstork:
Set policies to silence warnings in CMake 3.02 and later.
Later versions of CMake require the user to specify whether they wish to reject or endorse the referred policies.
CMP0046 requires that we explicitly disable adding pycaffe as a dependency when it is not present.
https://cmake.org/cmake/help/v3.0/policy/CMP0046.html
CMP0054 does not require changes.
https://cmake.org/cmake/help/v3.1/policy/CMP0054.html
This improves forward compatibility with newer versions of CMake.


shelhamer:
Thanks for the build help.

Issue #1646, opened 2014-12-29
Split headers to improve build granularity


longjon:
#1472 brings proper incremental building to the Makefile (which presumably the CMake build has as well). However, interactive building can be still be pretty slow, largely due to the use of monolithic header files (such as `vision_layers.hpp`).

Unless there's some good reason to have things the way they are, it seems better to give each layer its own header file. (Of course, we can still provide `vision_layers.hpp` etc. to include the others if that's actually useful somewhere.)

futurely:
I tried to do that but gave up the idea several months ago. The Makefile became too cumbersome without the automatic dependency generation mechanism.

It's also a good time to split the whole project into modules which is a common practice for large and quickly growing projects such as OpenCV.

bhack:
:+1:

shelhamer:
Solved by #3315 and #3226.

PR #3226, opened 2015-10-20
Clean includes

kkhoot:
This PR reduces header includes to speed up (partial) builds.

1. Give appropriate subcategory layer headers to each layer/test file.
Most of the layers/tests include the monolithic header vision_layers.hpp which causes recompiling almost all layer/test cpps when changes in other subcategory layer headers (such as common_layers.hpp, etc.) are made.

2. Remove unnecessary headers 
Many of the files include unneeded headers (such as cstring). Most of these includes were erased.

-During this work, I found that HDF5 defines are defined in a different layer header, so moved it to the right place.

### DIFF
--- a/include/caffe/blob.hpp
+++ b/include/caffe/blob.hpp
@@ -8,7 +8,6 @@
 #include "caffe/common.hpp"
 #include "caffe/proto/caffe.pb.h"
 #include "caffe/syncedmem.hpp"
-#include "caffe/util/math_functions.hpp"
 
 const int kMaxBlobAxes = 32;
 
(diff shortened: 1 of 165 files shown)


ronghanghu:
@kkhoot Thanks for the header cleanup. I'll try to review tonight or tomorrow.

ronghanghu:
The headers seem much cleaner after this PR. Thanks again @kkhoot

shelhamer:
Thanks for this @kkhoot. This is a nice reform and step in the direction of #1646.

kkhoot:
@ronghanghu  @shelhamer  Thanks for quick review :)

Issue #2623, opened 2015-06-19
`FilterLayer` may produce empty empty blob and lead to crash


prclibo:
Hi,
When a empty selection mask is ported to `FilterLayer`, it will produces empty tops. I found this empty tops may cause several issues.

1. As mentioned in #2569, https://github.com/BVLC/caffe/blob/master/src/caffe/layers/filter_layer.cu#L38 this line will be divided by 0.
2. Many loss layers (e.g. `EuclideanLoss`) will produces NaN since they divide the summation loss by `N`, the input blob size.
3. GPU operations (e.g. `add_kernel` for `caffe_gpu_add`) will fail because they use `CAFFE_GET_BLOCKS(N)` and when `N == 0` this produces 0 blocks, which makes CUDA unhappy.

So maybe there should be some check added to handle empty blob case generated by `FilterLayer`

bhack:
This was a know problem. But maintainers ignored pur advice to review and merge https://github.com/BVLC/caffe/pull/2053 with filterlayer.
/cc @mtamburrano

bhack:
@BVLC ping

PR #3238, opened 2015-10-23
Detect division by zero

dblN:
There happens division by zero if the setting is inappropriate.

- When creating the layer such as convolution or pooling layer, the shape can be 0.
- When `stepsize` is not specified, it can be 0.

I added checks for these cases.

### DIFF
--- a/src/caffe/blob.cpp
+++ b/src/caffe/blob.cpp
@@ -29,7 +29,7 @@ void Blob<Dtype>::Reshape(const vector<int>& shape) {
   }
   int* shape_data = static_cast<int*>(shape_data_->mutable_cpu_data());
   for (int i = 0; i < shape.size(); ++i) {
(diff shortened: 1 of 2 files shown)


ronghanghu:
I think we should have this, or at least the `stepsize` check. Note this is against #2053.

bhack:
@ronghanghu We could always improve that if somebody give us any review feedbacks by months. This one got a reply only after 5 days. See also BAIDU researcher catch almost ignored at https://github.com/BVLC/caffe/issues/2623 and related PR.

mtamburrano:
hi @ronghanghu,
I'm not sure how this PR is against #2053, where that PR let the computation to continue even when some batch becomes 0-sized (probably due to the use of FilterLayer).
This PR only checks if the blob has size 0 and stop computation in that case, so it doesn't fix the use case when a blob is completely filtered on purpose

Issue #3036, opened 2015-09-05
Gradient accumulation is untested


seanbell:
I often run into the same problem as #2532, where it is necessary to zero-initialize the bottom diff, but _not_ the parameter diffs.  This is unintuitive, but acceptable as long as it's both documented and tested.  The problem is that there are no tests to check each layer for this behavior.

If you are opposed to fixing the test as per https://github.com/tnarihi/caffe/commit/7d455263dc66b7d663340c12458795dc86f7466d (the "+1, -1" trick), then an alternative would be to add a function to every layer that says whether or not it supports gradient accumulation (default false).  Then, the gradient checker would apply the "+1, -1" trick only to those that claim to support it.  In the case that someone uses iter_size > 1, the net would check all its layers and raise an exception if any layer has parameters but doesn't support gradient accumulation.

I'm opening this issue because I think it needs to be addressed in one way or another.

PR #3254, opened 2015-10-28
CuDNNConvolutionLayer accumulate gradients

ronghanghu:
After #1977, all layers should accumulate gradients for parameters. These two lines introduced in #3160 seems to be a bug.

@shelhamer @slayton58 can you take a look?

### DIFF
--- a/src/caffe/layers/cudnn_conv_layer.cu
+++ b/src/caffe/layers/cudnn_conv_layer.cu
@@ -53,12 +53,10 @@ void CuDNNConvolutionLayer<Dtype>::Backward_gpu(const vector<Blob<Dtype>*>& top,
   if (this->param_propagate_down_[0]) {
     weight = this->blobs_[0]->gpu_data();
     weight_diff = this->blobs_[0]->mutable_gpu_diff();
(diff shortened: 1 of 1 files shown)


longjon:
Good catch @ronghanghu -- this definitely looks wrong. I'll leave open for the moment since I haven't read this code before, but we should take care of this immediately.

We should really have better testing for this situation... if this is bugged as it appears, this breaks weight sharing and gradient accumulation.

slayton58:
Yep, this does look like a bug -- when I wrote the initial support on our fork and merged it across, it looks like this got dragged in (it's still present in our code)

shelhamer:
Good catch @ronghanghu and sorry for missing this is my review. While there
are cuDNN layer tests and weight sharing + gradient accumulation tests
there are not weight sharing + gradient accumulation tests for cuDNN.

ronghanghu:
We need to adapt the gradient checker to gradient accumulation, in order to prevent future mistakes like this one and #2532.

seanbell:
:+1: I agree that the gradient checker needs to be updated.  A simple fixed was proposed earlier:
https://github.com/tnarihi/caffe/commit/7d455263dc66b7d663340c12458795dc86f7466d

See also #3036.

PR #3257, opened 2015-10-28
Support to specify layer name to extract from the network.

sjuvekar:
You can pass an argument like '--layer_name fc7' to classify.py
to extract the required layer instead of output probabilities -- a
useful feature to have for feature extraction.

Change-Id: I468504b156b0c0d8fad54ad5569229f592216979


shelhamer:

PR #3261, opened 2015-10-29
Update plot_training_log.py.example

kli-nlpr:
I find there is no plot_log.sh file.
I think this is a small mistake.


shelhamer:
Thanks for the typo fix.

PR #3272, opened 2015-11-03
OSX 10.10 (and more) use Accelerate Framework instead of veclib

ixartz:
OSX Yosemite 10.10 (and more) use Accelerate Framework instead of veclib.
In Accelerate Framework, cblas.h does not exist and we need to include Accelerate/Accelerate.h

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -102,6 +102,12 @@ elseif(APPLE)
   find_package(vecLib REQUIRED)
   include_directories(SYSTEM ${vecLib_INCLUDE_DIR})
   list(APPEND Caffe_LINKER_LIBS ${vecLib_LINKER_LIBS})
+
+  if(VECLIB_FOUND)
+    if(NOT vecLib_INCLUDE_DIR MATCHES "^/System/Library/Frameworks/vecLib.framework.*")
+      add_definitions(-DUSE_ACCELERATE)
+    endif()
+  endif()
 endif()
 
 # ---[ Python
--- a/include/caffe/util/mkl_alternate.hpp
+++ b/include/caffe/util/mkl_alternate.hpp
@@ -7,9 +7,14 @@
 
 #else  // If use MKL, simply include the MKL header
 
+#ifdef USE_ACCELERATE
+#include <Accelerate/Accelerate.h>
+#else
 extern "C" {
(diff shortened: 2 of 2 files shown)


scripts/travis/travis_install.sh
```diff
@@ -23,7 +23,7 @@ apt-get install \
 # Caffe requires a minimum CMake version of 2.8.8.
 if $WITH_CMAKE; then
   # cmake 3 will make sure that the python interpreter and libraries match
-  wget http://www.cmake.org/files/v3.2/cmake-3.2.3-Linux-x86_64.sh -O cmake3.sh
+  wget --no-check-certificate http://www.cmake.org/files/v3.2/cmake-3.2.3-Linux-x86_64.sh -O cmake3.sh
```
lukeyeager: This belongs in a separate commit, but I can verify that it's needed.

https://travis-ci.org/BVLC/caffe/jobs/88939834
https://travis-ci.org/BVLC/caffe/jobs/89066109
https://travis-ci.org/NVIDIA/caffe/jobs/89073960

lukeyeager: Merged in #3276.

ixartz:
Update the pull request after merging wget cmake with --no-check-certificate

Duffycola:
This is not working for me (OS X El Capitan 10.11.5 (15F34)), because after my recent upgrade the system still contains those old library paths. For instance:

> $ cat /System/Library/Frameworks/vecLib.framework/Headers/vecLib.h 
> #error "The vecLib framework is deprecated. Please include <Accelerate/Accelerate.h> instead of <vecLib/vecLib.h> and link to the Accelerate framework instead of the vecLib framework."

Duffycola:
Can you have a look at cmake/Modules/FindvecLib.cmake  ? I think any fix should be placed there

shelhamer:
Thanks for the improved OS X build logic @ixartz. I agree with @Duffycola that this could still be an issue due to the deprecated `vecLib.h`. A further patch along the lines of https://github.com/BVLC/caffe/pull/4247#issuecomment-225416151 might be needed to make this work correctly.

If anybody on OS X wants to follow-up with a patch for that (potentially simplifying the just-merged #4247), please do.

p.s. FWIW the Makefile build seems to be compatible with El Cap.

PR #3275, opened 2015-11-03
fix detect.py (invalid model path)

darxriggs:


longjon:
Thanks for the correction!

(The Travis issue is unrelated and awaiting a fix.)

PR #3276, opened 2015-11-03
TravisCI: wget cmake with --no-check-certificate

lukeyeager:
Pulling fix from https://github.com/BVLC/caffe/pull/3272.

```
--2015-11-03 22:31:11--  http://www.cmake.org/files/v3.2/cmake-3.2.3-Linux-x86_64.sh
Resolving www.cmake.org (www.cmake.org)... 66.194.253.19
Connecting to www.cmake.org (www.cmake.org)|66.194.253.19|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: http://cmake.org/files/v3.2/cmake-3.2.3-Linux-x86_64.sh [following]
--2015-11-03 22:31:11--  http://cmake.org/files/v3.2/cmake-3.2.3-Linux-x86_64.sh
Resolving cmake.org (cmake.org)... 66.194.253.19
Connecting to cmake.org (cmake.org)|66.194.253.19|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://cmake.org/files/v3.2/cmake-3.2.3-Linux-x86_64.sh [following]
--2015-11-03 22:31:11--  https://cmake.org/files/v3.2/cmake-3.2.3-Linux-x86_64.sh
Connecting to cmake.org (cmake.org)|66.194.253.19|:443... connected.
ERROR: no certificate subject alternative name matches
        requested host name `cmake.org'.
To connect to cmake.org insecurely, use `--no-check-certificate'.
```

PR #3277, opened 2015-11-03
Revert "TravisCI: wget cmake with --no-check-certificate"

shelhamer:
This causes its own trouble:
https://github.com/BVLC/caffe/pull/3276#issuecomment-153516349

This can be re-included in a future PR that also fixes Python 3.


lukeyeager:
I'm not sure I caused the PYTHON3 issue. I don't see how downloading the CMake install script insecurely could cause a boost vs. python conflict.

shelhamer:
> I don't see how downloading the CMake install script insecurely could cause a boost vs. python conflict.

True. Closing this and deferring to a CI fix for Python 3. It seems like Travis boost and / or Python versions have changed. Sorry for the noise.

PR #3286, opened 2015-11-05
[travis] fix boost/python3 conda conflict

longjon:
This is an attempt to fix the Python 3 issue that's arisen on Travis; see #3277.

Don't merge before it's passed; I don't know if this works!


longjon:
After a few attempts, this is now passing.

It appears that the Python 3 boost version we're using here depends on Python 3.4, but miniconda is now installing Python 3.5 by default. I fixed by explicitly installing 3.4 when in Python 3 mode.

I'm going to go ahead and merge this simple fix to Travis, but feel free to comment or send a new PR if you have a more enlightened solution.

shelhamer:
Thanks for fixing this Jon!

Evan Shelhamer

PR #3287, opened 2015-11-05
Fix whitespace in travis_install.sh

longjon:
Caffe uses spaces, never tabs, and definitely never ever mixed whitespace. (Hi @philkr!)


longjon:
Merging this simple, passing style fix.

PR #3294, opened 2015-11-06
Switched order of two layers for simpler diff with untuned file

panmari:
Untuned file is in models/bvlc_reference_caffenet/train_val.prototxt.

The diff between the two files is shown in the example 03-fine-tuning.ipynb.


panmari:
Rebased commit on master.

shelhamer:
Thanks for the example tweak.

PR #3295, opened 2015-11-06
Fix issue #3274

### LINKED ISSUE #3274: ArgMaxLayer Reshape missing checks for legacy path
- build caffe_test with debug_lvl==2
- run test NetTest, TestSkipPropagateDown
- you will get an OutOfBound assertion in ArgMaxLayer<Dtype>::Reshape for __shape[2] = top_k_;__
 - num_axes of bottom[0] is 2 and the value of has_axis_ is false
 - shape will be initialized with a size of two
 - the access to index 2 in shape causes the assert

I don't know if this is really a bug or if just the net configuration for this unittest is incorrect.
The change in the ArgMaxLayer was introduced with PR #3069.

timmeinhardt:
This PR resolves #3274 by fixing the OutOfBound assertion error for any shape of bottom blob in the ArgMaxLayer::Reshape function. Before the reshape function always used the bottom blob shape to shape the top blob even when the num of axes of the bottom blob is smaller than 3. But 3 axes are necessary to output the different max_ind or max_val options.

### DIFF
--- a/include/caffe/common_layers.hpp
+++ b/include/caffe/common_layers.hpp
@@ -53,8 +53,8 @@ class ArgMaxLayer : public Layer<Dtype> {
    *   -# @f$ (N \times C \times H \times W) @f$
    *      the inputs @f$ x @f$
    * @param top output Blob vector (length 1)
-   *   -# @f$ (N \times 1 \times K \times 1) @f$ or, if out_max_val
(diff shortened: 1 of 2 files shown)


shelhamer:
Thanks for the fix @timmeinhardt ! Sorry I didn't catch this at merge time.

PR #3297, opened 2015-11-07
Add a -c to wget so that it continues interrupted downloads

sjbrown:
This would've saved me an overnight download (slow connection here)

I tested it, and it worked for me.


jeffdonahue:
Seems reasonable, thanks @sjbrown.

PR #1063, opened 2014-09-10
Create temporary path with boost::filesystem for cross platform compatibility

kloudkl:

PR #3300, opened 2015-11-07
Replace unistd functions with cross platform counterparts

eelstork:
Via unistd.h, caffe uses a handful of functions not readily available on non-Unix systems; boost is cross platform and affords more concise, more readable implementations (notably, with reference to MakeTempFilename and MakeTempDir).
To support this, CMake and Make builds are updated to link against boost_filesystem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ ifneq ($(CPU_ONLY), 1)
 	LIBRARIES := cudart cublas curand
 endif
 
-LIBRARIES += glog gflags protobuf boost_system m hdf5_hl hdf5
+LIBRARIES += glog gflags protobuf boost_system boost_filesystem m hdf5_hl hdf5
 
 # handle IO dependencies
 USE_LEVELDB ?= 1
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -2,7 +2,7 @@
 set(Caffe_LINKER_LIBS "")
 
 # ---[ Boost
-find_package(Boost 1.46 REQUIRED COMPONENTS system thread)
+find_package(Boost 1.46 REQUIRED COMPONENTS system thread filesystem)
 include_directories(SYSTEM ${Boost_INCLUDE_DIR})
 list(APPEND Caffe_LINKER_LIBS ${Boost_LIBRARIES})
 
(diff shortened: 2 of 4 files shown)


eelstork:
Thanks to @flx42 for suggesting a truly cross platform temp dir

ronghanghu:
LGTM :) Replacing unistd.h with boost should make it easier to build caffe across platforms for community windows ports.

flx42:
You have another place where you are using "/tmp", I guess you could also remove it and use boost too.

eelstork:
@ronghanghu thanks for looking at this. 
@flx42 fixing.

ronghanghu:
I plan to merge this PR if other people don't object.

shelhamer:
:+1:

bhack:
Are we not ready for [tmpnam](http://en.cppreference.com/w/cpp/io/c/tmpnam) et al.?

eelstork:
@bhack I appreciate [difficulties running Boost on iOS and Android](http://stackoverflow.com/questions/14036311/official-boost-library-support-for-android-and-ios). In the meantime I'm not sure that C++11 is a silver bullet for cross compatibility. With a little more refactoring we'll be able to handle this.

futurely:
Did this solve the problem of potential race condition https://github.com/BVLC/caffe/pull/1063?

futurely:
It seems not. https://github.com/BVLC/caffe/issues/3324

ronghanghu:
Oh, I didn't notice the race condition in #1063. My bad. This is probably causing directory creation to fail in #3324?

futurely:
The easiest solution is to add a macro USE_BOOST_FILESYSTEM and keep both versions.

ronghanghu:
It seems that both versions (old and new) have potential race conditions, so it is a bit unclear to me why #3324 is appearing

eelstork:
@futurely I think we are going to do this the hard way.
@ronghanghu if I understand correctly #3324 isn't a regression caused by this PR, but we're seeing significantly more occurrences since merging this?

ronghanghu:
I'm not sure if there is more occurrence of #3324 after this PR. Seems tricky to do a perfect fix to the race here. At this point the function is only used in test cases so I guess the race condition shouldn't affect too much.

eelstork:
@ronghanghu I've seen a couple more. I think it's worth fixing.

PR #3303, opened 2015-11-08
GetDB must return a value

eelstork:
As noted by @danst18, when USE_LEVELDB and USE_LMDB are disabled, a compiler error is issued since GetDB no longer returns a value.
At runtime a fatal error would be issued anyways. However to help users who don't need a DB backend, NULL should be returned here.


ronghanghu:
LGTM. Thank you @eelstork !

Issue #3302, opened 2015-11-08
Could caffe display message "Ignore source layer xxx" by default when fine-tuning or testing a net?


tianzhi0549:
https://github.com/BVLC/caffe/blob/master/src/caffe/net.cpp#L748

I suggest that caffe should change DLOG() to LOG() and displays "Ignore xxx" when debug_info=true or always displays that. Now this message only appears when building caffe with DEBUG:=1.

If do so, when someone types a error mismatch layer name, caffe could hint him even if he doesn't turn DEBUG on when building.

Thank you:-)!

ronghanghu:
Yes, I think we should have this message displayed not only in debug mode. I will create a PR on that.

ronghanghu:
Note that to correctly display messages during fine-tuning or testing a net, one also needs to modify `Net::CopyTrainedLayersFrom`.

tianzhi0549:
Thank you for your PR:-).

PR #3305, opened 2015-11-08
display 'ignore source layer' when initializing from existing parameters

ronghanghu:
This PR helps in the case to see which layer is initialized from existing parameters, and which layer is ignored. This helps identify the cases where the user types a error mismatch layer name, as mentioned in #3302.


seanbell:
:+1:

Issue #842, opened 2014-08-03
Random seeding


netheril96:
The current random seeding `cluster_seedgen()` use a combination of pid and current time, with a maximum output of 104729 (17 bits), which is quite small. But since the program is already linked to boost.random, one can simply generate the seed from `boost::random_device`, or read from `/dev/urandom`. Both approach uses entropy collected by the OS from hardware, have more bits available, and do not require many lines of code.

PR #2619, opened 2015-06-18
Make HDF5 components optional for configurable build

sh1r0:
This PR simply intends to make HDF5 as an optional dependency.

PR #3306, opened 2015-11-09
Provide cross-platform seed generation

eelstork:
Previously (#842), the use of an entropy source was added so that caffe's RNG uses a random seed by default. Although @netheril96 suggested that we avoid `boost::random_device`, I contend that incurring a dependency on boost random is a small price to pay for cross-platform compatibility.

Additionally, is there a practical reason to preserve the fallback solution via `getpid` ?


eelstork:
We shouldn't use `std::random_device`.

In C++11 [std::random_device][1] *may be implemented in terms of an implementation-defined pseudo-random number engine if a non-deterministic source (e.g. a hardware device) is not available to the implementation. In this case each std::random_device object may generate the same number sequence*.

In constrast, Boost specifies that  *for those environments where a non-deterministic random number generator is not available, class [random_device][2] must not be implemented*.

[1]: http://en.cppreference.com/w/cpp/numeric/random/random_device
[2]: http://www.boost.org/doc/libs/1_59_0/doc/html/boost/random/random_device.html

bhack:
@eelstork If you are interested I've removed all boost dependencies other than for python some months ago at https://github.com/BVLC/caffe/pull/2537

eelstork:
@bhack although I am interested, I feel that this (removing boost) may be premature. Additionally, the problem at hand (seed generation) inspires caution: C++ 11 isn't always an ideal replacement for Boost.
What is the main reason to remove boost? Smaller footprint on handhelds?

bhack:
The effort started to minimize the dependencies for mobile and embedded (specially for forward pass only of trained nets). I.e. in android boost it is not so much a first grade citizen. See https://github.com/BVLC/caffe/pull/2619

eelstork:
@ronghanghu thoughts on the above?

bhack:
I repeat again. Are we sure that we are not  embracing a [rearguard position](https://github.com/tensorflow/blob/master/tensorflow/core/lib/random/random.cc)?

eelstork:
@bhack currently caffe builds default to C++98. As mentioned before I'm looking into a workaround so that we can be more inclusive.
@shelhamer can you have a look at this?

bhack:
@eelstork I know. But after 6 months  only Nvidia guys commented to the introduction of c++11 and Boost removal in c++. Also the big https://github.com/BVLC/caffe/pull/2610 rely on c++11. As pointed also Tensorflow already use c++11 features. The problem is ignoring instances and not what kind of direction maintainers want to take.

eelstork:
Ready for review/merge.

eelstork:
@ronghanghu can you guys have a look at this?

PR #3310, opened 2015-11-09
DOC: Fix consistent typo in contrastive loss

gustavla:
If a pair is similar, the contrastive loss should take the squared distance and not the distance. This is clearly what the code is doing, but the documentation gets it consistently wrong.

This is straight from the code with an arrow added:

```cpp
if (static_cast<int>(bottom[2]->cpu_data()[i])) {  // similar pairs
  loss += dist_sq_.cpu_data()[i];  // <---- clearly d^2 and not d
} else {  // dissimilar pairs
  if (legacy_version) {
    loss += std::max(margin - dist_sq_.cpu_data()[i], Dtype(0.0));
  } else {
    Dtype dist = std::max(margin - sqrt(dist_sq_.cpu_data()[i]), 0.0);
    loss += dist*dist;
  }
}
```

### DIFF
--- a/include/caffe/loss_layers.hpp
+++ b/include/caffe/loss_layers.hpp
@@ -132,7 +132,7 @@ class LossLayer : public Layer<Dtype> {
 
 /**
  * @brief Computes the contrastive loss @f$
- *          E = \frac{1}{2N} \sum\limits_{n=1}^N \left(y\right) d +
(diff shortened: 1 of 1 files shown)


seanbell:
LGTM

shelhamer:
Thanks for the doc fix @gustavla!

PR #3312, opened 2015-11-10
Constrain Prelu negative slope

jyegerlehner:
This PR allows the negative slope on a PReLU unit to be constrained between min and max values which default to zero and -1.0. Which means it constrains the PReLU to be between ReLU and a linear unit. As the PR stands, this is now the default behavior, but could be changed to leave the negative slope unconstrained by default. 

[Edited for clarity, after seanbell replied.]

### DIFF
--- a/include/caffe/layer.hpp
+++ b/include/caffe/layer.hpp
@@ -316,10 +316,21 @@ class Layer {
     param_propagate_down_[param_id] = value;
   }
   /**
-   * @brief Called after net's parameters have been updated. May be overridden
-   *        by a layer to apply constraints to parameters.
+   * @brief Called on each layer after net's parameters have been updated
+   *        using the solver.
    */
-  virtual void PostUpdateProcessing() {}
+  void PostUpdateProcessing() {
+    switch (Caffe::mode()) {
+    case Caffe::CPU:
+      PostUpdateProcessing_cpu();
+      break;
+    case Caffe::GPU:
+      PostUpdateProcessing_gpu();
+      break;
+    default:
+      LOG(FATAL) << "Unknown caffe mode.";
+    }
+  }
 
  protected:
   /** The protobuf that stores the layer parameters */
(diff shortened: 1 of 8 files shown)


include/caffe/layer.hpp
```diff
@@ -513,6 +517,9 @@ void Layer<Dtype>::ToProto(LayerParameter* param, bool write_diff) {
   }
 }
 
+
+
+
```
jyegerlehner: Oops

src/caffe/proto/caffe.proto
```diff
@@ -1231,4 +1231,11 @@ message PReLUParameter {
   optional FillerParameter filler = 1;
   // Whether or not slope paramters are shared across channels.
   optional bool channel_shared = 2 [default = false];
+  // Whether or not to constrain the negative slope between min_neg_slope and
+  // max_neg_slope.
+  optional bool constrain_neg_slope = 3 [default = true];
```
seanbell: Since there is no migration procedure for those who have not specified `constrain_neg_slope` yet, I think that the default should be `false`.  Otherwise this silently changes the behavior of any existing nets, and this can be really hard to debug.

jyegerlehner: Right. But it has no effect once a model is trained. If a trained model is just deployed, its behavior  won't change. So to the extent that it changes behavior, it won't be until after they've done some more training. At which point they would be evaluating before deploying it anyway. 

I don't mind changing the default to false. But I doubt most people will be aware that the option is there, and so will continue suffering along with crazy Prelu values (I was getting NaNs, and after instrumenting found it was due to an inf coming out of a Prelu in forward, going in to a softmax that generated NaN when dividing inf by inf. So if you're worried about hard-to-debug problems, I think a wacky negative slope is a bigger threat than a Prelu that's constrained to be somewhere in-between a ReLU (neg-slope == 0.0) and a linear unit (neg slope == 1.0).

seanbell: Hmm, yeah.  I agree that this is a reasonable updated default.  It does change experiments that people may want to re-run, but caffe has already changed training defaults in the past (https://github.com/BVLC/caffe/pull/2321).

PR #3315, opened 2015-11-11
Dismantle Layer Headers

shelhamer:
Monolithic layer includes trigger grossly unnecessary recompilation whenever a change is made to even a single layer definition. Dismantling the agglomerated `vision_layer.hpp`, `common_layers.hpp`, and so forth makes the dependencies and build more modular. This should likewise reduce conflicts in layer development and PRs.
                
Regarding dependencies, I do not promise that this PR is free of recursive includes. Occasionally a `make clean` might be needed (if this is in fact a problem at all; I need to review dependency generation...) but this is no different than the current arrangement.

I admit there is essentially never a good time for this PR since it will impact any layers in progress but at some point the headers need breaking for mending.

To reconcile your own branches with this you should rebase and transplant changes to the monolithic headers to the relevant layer header (which you need to create for your own layers).

Now the only obstacle to decoupled layer development is the common caffe.proto.


ducha-aiki:
>I admit there is essentially never a good time for this PR since it will impact any layers in progress but at some point the headers need breaking for mending.
Even with this, it is greatly welcomed PR :)

shelhamer:
@ronghanghu can I ask for your help to check if this is compatible with matcaffe?

ronghanghu:
@shelhamer Of course! I am checking.

ronghanghu:
All matcaffe tests passed with this PR on a BVLC machine.

PR #3320, opened 2015-11-12
Cast std::max args to Dtype

eelstork:
This is to avoid ambiguous literals. Consistent with similar idioms in Caffe source. Fixes errors with MSVC compiler.

### DIFF
--- a/include/caffe/test/test_gradient_check_util.hpp
+++ b/include/caffe/test/test_gradient_check_util.hpp
@@ -169,8 +169,9 @@ void GradientChecker<Dtype>::CheckGradientSingle(Layer<Dtype>* layer,
           || fabs(feature) > kink_ + kink_range_) {
         // We check relative accuracy, but for too small values, we threshold
         // the scale factor by 1.
-        Dtype scale = std::max(
-            std::max(fabs(computed_gradient), fabs(estimated_gradient)), 1.);
+        Dtype scale = std::max<Dtype>(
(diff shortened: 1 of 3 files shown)


seanbell:
You can achieve the same effect more cleanly with `std::max<Dtype>(a, b)` instead of `std::max(Dtype(a), Dtype(a))`.

eelstork:
@seanbell I am aware of this (@willyd and others fixed it this way). Is it cleaner, or only more concise?

seanbell:
I personally think it's more clear/clean, since `max<T>` is saying "use the `max` function for type `T`."  I started using it after I ran into the problem that `abs` can sometimes lead to unintentional casting to integers (I needed `fabs`), but `abs<T>` would either do the right thing or generate an error (depending on which `abs` definition is included).  We're talking about a different function (`max`), but that's why I like explicitly specifying which template is used via `function<T>`.

Anyway, that's just my opinion.  Both versions work.

longjon:
I agree with @seanbell here... in addition, it's best to avoid C-style casting (I can see an argument being made for literals, but we are not only casting literals here).

shelhamer:
Agreed with @seanbell and @longjon. Please revise for merge. Thanks for the the compatibility checks.

eelstork:
Updated to use suggested form.

ronghanghu:
Thanks @eelstork !

PR #3321, opened 2015-11-12
In syncedmem.cpp, functions shall return a value

eelstork:
with CPU_ONLY builds gpu_data() and mutable_gpu_data() must return a value.


shelhamer:
`NO_GPU` triggers a fatal error so the return will never execute; see https://github.com/BVLC/caffe/blob/master/include/caffe/util/device_alternate.hpp#L10. Is this to make a compiler happy?

eelstork:
Yes, this is to make a compiler happy.

shelhamer:
Alright, thanks for the compiler therapy!

PR #3323, opened 2015-11-12
Remove un-necessary includes

eelstork:
Removing unnecessary includes which cause compatibility issues


shelhamer:
Thanks for the header hygiene.

PR #3326, opened 2015-11-12
MultiClass label support

mtamburrano:
This is an uncomplete PR to support multiclass loss in caffe.
I doubt this will be ever merged but I wanted to share some modifications I made to memoryDataLayer to accept multiple labels and to softmax to handle mutliclass problems.
A good discussion about multilabel, multiclass and multitask problems in caffe is [here](https://groups.google.com/forum/#!topic/caffe-users/RuT1TgwiRCo)

The main news this PR introduce are:
- a new method for `MemoryDataLayer` called `addMatVectorMultilabel` that accepts a `vector < vector < int > >` as label for each data Mat. So a Mat can be easily associated with multiple classes. The resulting label blob will have the label values on the channel dimension (or more exactly in the dimension `1`since the N-dimension support).
- extended `SoftmaxLossLayer` to let it to compute the loss on different labels and on slices of `bottom[0]`. The number of slices are specified with the prototxt parameter `slice` and must match `bottom[1]->shape(softmax_axis_)`
- extended `SoftmaxLayer` to let it to compute the softmax independently on slices of `bottom[0]`, so the `SoftmaxLossLayer` can compute the loss for each slice comparing it with each corresponding label.

The idea is to feed the network with `MemoryDataLayer` and `addMatVectorMultilabel`, then attaching a `SoftmaxLossLayer` specifying the parameter `slice` to be equal to the size of the vector of the labels added. It is important to set the output of `bottom[0]` of the `SoftmaxLossLayer` to be a multiple of the number of classes and of the slices.
For example let's say we want to add a vector of 3 labels, and each labels can have 4 different classes, the input for `addMatVectorMultilabel` could be a vector like '{{0,0,0,1}, {1,0,0,0}, {0,1,0,0}}.
Now after some conv layer we attach an `Inner_product` layer with `num_output: 12` (3*4 = 12). Then we attach to the `InnerProductLayer` layer a `SoftmaxLossLayer` specifying `slice: 3`. So the softmax will be applied independently to each slice of 4 elements of the output of the `InnerProductLayer` and the loss calculated for each softmax.

Actually the cpu version of `SoftmaxLayer` and `SoftmaxLossLayer` is working, but the gpu is not, I wrote some modification on the gpu part of the `SoftmaxLayer` but some index is wrong ( the test doesn't pass when `batch_size>2`.
I don't know when I will have time to fix that, if someone wants to contribute, any help is appreciated :)


xdshang:
@mtamburrano I also found a definition of multiclass, multilabel and multi-task classification [here](http://scikit-learn.org/stable/modules/multiclass.html). According to this reference, softmax is inherently a multiclass classifier. 

I think your modification is more related to multi-task classification, which can be implemented by setting <code>inner_num_</code> in caffe. In the above example, you can use <code>ReshapeLayer</code> to reshape <code>(12,)</code> into <code>(4, 3)</code> before forwarding into <code>SoftmaxWithLossLayer</code>, and set <code>softmax_axis_</code> to the 1st axis. Then <code>inner_num_</code> will be automatically set to 3 and multi-task classification is executed.

Basically, multilabel classification can also be done in this way, but one needs to duplicate the <code>inner_product</code> output (4-D in your example) <code>number_of_labels (number_of_tasks)</code> times. This needs write a new <code>DuplicateLayer</code> as well as the new layer consumes more memory space. So in PR #3268, I did some modification in <code>SoftmaxWithLossLayer</code> to support multilabel classification without more memory requirement.

PR #3332, opened 2015-11-14
Fix outs and diffs being overwritten in forward_backward_all.

alexlee-gk:
The nparrays out_blobs and out_diffs need to be copied. The same already happens in _Net_forward_all.


shelhamer:
Thanks for catching this. The `_Net_forward_backward_all()` logic should indeed match `_Net_forward_all()`.

PR #3352, opened 2015-11-18
Skip python layer tests if WITH_PYTHON_LAYER unset

### LINKED ISSUE #2093: make pytest fails unless WITH_PYTHON_LAYER is set
I just noticed to day that there is a pytest path in the Makefile. Looks like it was added here:  #1473. When I run `make pytest`, I get this error:

    layer_factory.hpp:77] Check failed: registry.count(type) == 1 (0 vs. 1) Unknown layer type: Python
    
    ...
    
    Aborted (core dumped)
    make: *** [pytest] Error 134

I've got a pretty standard installation here. I assume this is just not something most people check, so it doesn't get updated. Is this really tested and passing in travis?

P.S. If caffe would just exit with an error code instead of coredumping, then the rest of the tests could run and I could get an actual test report. This is an issue across the board with the python interface.

lukeyeager:
*Close #2093*

With `caffe_layer_list()` from #2930, this is trivial.

### DIFF
--- a/python/caffe/test/test_python_layer.py
+++ b/python/caffe/test/test_python_layer.py
@@ -77,6 +77,8 @@ def parameter_net_file():
         return f.name
 
 
+@unittest.skipIf('Python' not in caffe.layer_type_list(),
+    'Caffe built without Python layer support')
 class TestPythonLayer(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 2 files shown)


ronghanghu:
Thanks @lukeyeager !

PR #3357, opened 2015-11-19
add lines to move the heads of the syncedmems of the learnable params…

lim0606:
The current implementation of the multigpu support does not move the heads (of the syncedmems) of the learnable params in non-root_solvers to HEAD_AT_GPU.

Because of that, the data in the pointer not synced when you need to call cpu_data().

To solve this issue, added lines to call mutable_gpu_data() in on_start() function in parallel.cpp

### DIFF
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -318,6 +318,15 @@ void P2PSync<Dtype>::on_start() {
     CUDA_CHECK(cudaStreamSynchronize(cudaStreamDefault));
     children_[i]->queue_.push(this);
   }
+
+  // Move syncedmem head to HEAD_AT_GPU
+  const vector<Blob<Dtype>*>& params = solver_->net()->learnable_params();
+  Dtype* mutable_gpu_data;
+  for (int i = 0; i < params.size(); ++i) {
+    mutable_gpu_data = params[i]->mutable_gpu_data();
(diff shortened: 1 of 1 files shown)


src/caffe/parallel.cpp
```diff
@@ -318,6 +318,15 @@ void P2PSync<Dtype>::on_start() {
     CUDA_CHECK(cudaStreamSynchronize(cudaStreamDefault));
     children_[i]->queue_.push(this);
   }
+
+  // Move syncedmem head to HEAD_AT_GPU
+  const vector<Blob<Dtype>*>& params = solver_->net()->learnable_params();
+  Dtype* mutable_gpu_data;
+  for (int i = 0; i < params.size(); ++i) {
+    mutable_gpu_data = params[i]->mutable_gpu_data();
+    mutable_gpu_data++; // added to prevent the warning message of compilers
```
seanbell: Why do you need to hold onto the return value?
You could just call the `mutable_gpu_data` function:
```
  for (int i = 0; i < params.size(); ++i) {
    params[i]->mutable_gpu_data();
  }
````

lim0606: That sounds great :+1:  Thanks :)

seanbell:
Hmm, is a similar fix needed here too?  https://github.com/lim0606/caffe/blob/master/src/caffe/parallel.cpp#L376

PR #3358, opened 2015-11-19
Openmp for convolution, deconvolution and relu

jczaja:
Hi,

I would like to suggest changes introducing parallelization of CPU layers of caffe. Changes
presented here are a rewritten and modified (tiny optimization) code that comes from this PR:

https://github.com/BVLC/caffe/pull/439

This PR focus only only on convolution, deconvolution and relu layers modified (more to come if this PR is approved). 

## Desc:
Changes are introducing Openmp based parallelization for convolutional and relu layers. By default OpenMP path is not enabled.
Motivation behind changes is performance gain. **Also after introducing OpenMP enabling changes, other layers (pooling, LRN) will also be optimized in a next step.**

## Performance Results :
Time measures are: average Forward-Backward as stated by *caffe time*. *speedup* is (master branch measure / openmp-conv-relu branch measure)

#### Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz (MKL 11.3, GCC 4.8.3):
Branch | googlenet(speedup: 1.5) | caffenet (speedup: 2.7) | alexnet(speedup: 2.7)
----------|-----------------------|-------------------------------|-----------------------------
openmp-conv-relu (using OMP_NUM_THREADS=36)| 2834ms                   |2997ms                   |3523ms 
master                                                              |4438ms                    |8164ms                   |8644ms 

#### Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz (OpenBLAS 0.2.14, GCC 4.8.3):
Branch | googlenet(speedup: 1.4) | caffenet (speedup: 2.7) | alexnet(speedup: 1.2)
----------|-----------------------|-------------------------------|-----------------------------
openmp-conv-relu (using OMP_NUM_THREADS=36)| 11711ms       |            9417ms                   | 49557ms
master |                                                              16848ms        |           26130ms     |   62091ms  

#### Intel(R) Xeon(R) CPU E3-1285 v3 @ 3.60GHz (MKL 11.3, GCC 4.8.3):
Branch | googlenet(speedup: 1.2) | caffenet (speedup: 1.2) |  alexnet(speedup: 1.3)
----------|-----------------------|-------------------------------|-----------------------------
openmp-conv-relu (using OMP_NUM_THREADS=4) | 3239ms    |  6881ms    |       6946ms  
master |      3902ms   |      8519ms     |              9360ms  

#### Intel(R) Xeon(R) CPU E3-1285 v3 @ 3.60GHz (OpenBLAS 0.2.14, GCC 4.8.3):
Branch |  googlenet(speedup: 1.3)  |  caffenet (speedup: 2.2)  | alexnet(speedup: 1.1)
----------|-----------------------|-------------------------------|-----------------------------
openmp-conv-relu (using OMP_NUM_THREADS=4) | 11256ms  | 11308ms  |  55081ms
master |  14827ms  |         25432ms    |              64692ms   

####  Intel(R) Core(TM) i7-4770R CPU @ 3.20GHz(MKL 11.3, GCC 4.9.2):
Branch |  googlenet(speedup: 1.2)  |  caffenet (speedup: 1.5)  | alexnet(speedup: 1.4)
----------|-----------------------|-------------------------------|-----------------------------
openmp-conv-relu (using OMP_NUM_THREADS=4) | 3209ms | 6344ms    |  7024ms
master |  4025ms  |         9568ms    |              10127ms   

####  Intel(R) Core(TM) i7-4770R CPU @ 3.20GHz(OpenBLAS 0.2.14, GCC 4.9.2):
Branch |  googlenet(speedup: 1.2)  |  caffenet (speedup: 2.1)  | alexnet(speedup: 1.1)
----------|-----------------------|-------------------------------|-----------------------------
openmp-conv-relu (using OMP_NUM_THREADS=4) | 10088ms | 12342ms    |  42884ms
master |  12633ms  |  26547ms    |              49261ms   

So there is significant speedup, that depends on how many CPU cores the platform does have. Tests were made using MKL(it is available free of charge now) and OpenBLAS.

- For disabled openmp build (Default setting) performance remain comparable to the current master branch. 
- **It is best to have HT disabled in BIOS as using OMP_NUM_THREADS may not always prevent
OS kernel for running two OpenMP threads on the same phyisical core (eg. using HT).** 
- Without using OMP_NUM_THREADS and with HT enabled performance is still better than single threaded version (data not included here)

### Building:
a) Modify Makefile.config so it has enabled USE_OPENMP := 1 and CPU_ONLY := 1  then build as usual (choose your BLAS)
b) *MKL_ROOT=$MKLROOT cmake <path to CmakeLists.txt> -DBLAS=mkl -DUSE_OPENMP=ON -DCPU_ONLY=ON    # for mkl*
   *cmake <path to CmakeLists.txt> -DBLAS=open -DUSE_OPENMP=ON -DCPU_ONLY=ON                     # for openblas*

### Running:
It is best NOT to use Hyperthreading . So either disable it in BIOS or limit OpenMP threads number by using OMP_NUM_THREADS env variable. If not sure how to set OMP_NUM_THREADS  and you cannot disable HT in BIOS, then do not do it, you should still observe performance gain , but not that
significant as when not relying on HT.

###### Xeon E5 , two sockets, 18 cpu cores in each socket
*OMP_NUM_THREADS=36 ./build/tools/caffe time -iterations 20  --model=models/bvlc_googlenet/train_val.prototxt*  
###### Xeon E3 , one socket, 4 cpu cores in each socket
*OMP_NUM_THREADS=4 ./build/tools/caffe time -iterations 20  --model=models/bvlc_googlenet/train_val.prototxt*   
###### Brix i7-4770R (Haswell), one socket, 4 cpu cores in each socket
*OMP_NUM_THREADS=4 ./build/tools/caffe time -iterations 20  --model=models/bvlc_googlenet/train_val.prototxt*   

##### Notes:
To check if you have HT enabled:
*sudo dmidecode -t processor | grep HTT*
or 
*cat /proc/cpuinfo | grep ht*

With HT (usually) You have twice as much processors online that physical cpu cores. So get this number of processors online , divide by 2 and make this result a value to which
 OMP_NUM_THREADS will be set to.

If you have any questions, please do not hesitate to ask.

Regards,


talda:
Note that mkl by itself uses multiple threads. I think in this case you will get a better result by setting the number of threads used by mkl to 1. 

You can do this by setting MKL_NUM_THREADS environment variable to 1.

ronghanghu:
IMO, if you use Intel MKL or OpenBLAS as CPU backend (not sure about Atlas), then it is already utilizing multi-core, so OpenMP doesn't seem necessary for Conv and Deconv. OpenMP can still help for ReLU, but may introduce more difficulties for debugging.

jczaja:
@talda Thanks for suggesstion. I tried that but did not get better performance than when not restricting mkl threads. MKL libs are providing multithreaded version: mkl_rt.so and singlethreaded: mkl_sequential.so. Caffe is using multithreaded, but I made modification (just to observere if I can get better performance) to run against single threaded MKL. But results I got were weaker than when using multithreaded MKL.

jczaja:
@ronghanghu Thanks for opinion. All three mentioned BLAS libs comes in singlethreaded as well as multithreaded editions. Here In performance result I used multithreaded MKL (mkl_rt.so), and singlethreaded OpenBLAS. As those configurations perform best with OpenMP changes being reviewed here. 

Few words about OpenBLAS tested. Result comes from single threaded, but I was able to make multithreaded build from source , for target architecture I was testing on (HASWELL). And I did not get better results that what I presented here.

Now, considering question if there is a point in optimizing convolution code while having multithreaded BLAS lib used, I checked if convolution itself gives performance boost (and it does) , and if relu gives performance boost (it does too). They both contribute to speedup of layer execution.
Why there is performance boost in convolution using openmp? Well, conv layer is iterating over batches eg. each loop iteration is processing one batch of data and then MKL/OpenBLAS/Atlas is to perform multiplication using multithreading. Perhaps for BLAS libs is not that easy to utilise full platform resources when working on single batch of data. On the other hand caffe OpenMP changes are implementing utilisation of threads so each of thread is running separate BLAS operation, also done in a way to avoid false sharing.

jczaja:
@ronghanghu I realised that I did not answered fully on your questions. As for concerns about difficulty of debugging of openmp code, there are few points worth mentioning:
1) OpenMP is optional path, disabled by default . So contributors not focused on CPU performance should not be concerned with OpenMP path.
2) range of modifications is kept to minimum , so in case mentioned Relu, this is only one #pragma .. line that was added. OpenMP and sequential path share most of code path.
3) As for debugging OpenMP solution. With current approach of keeping sequential path of CPU alive and adding minimum openmp changes, I do not think there will be much problems with debugging OpenMP code, other than making sure that normal, sequential CPU path works.
**Moreover, Myself and my colleagues are more than happy to help if such a need of debugging OpenMP path , arise.**
4) Performance results (presented in this PR) and performance boost that will be a result of next OpenMP based PR (will get up to ~8 times better performance comparing to sequential path, when pooling and LRN are modified to use OpenMP) are making this PR worth consideration to be merged:)

If something in code is not clear of you have some suggestions to code , please let me know.

Looking forward to your feedback,

jczaja:
I have updated PR :
- OpenMP is now used if available , if not avaialble then non-OpenMP build is made. 
- MKL usage from cmake was fixed/improved

bhack:
There is a specific openblas branch at https://github.com/xianyi/OpenBLAS/tree/optimized_for_deeplearning?files=1

jczaja:
@bhack Thanks for letting me know about effort of tuning openblas for deeplearning needs.

PR #3359, opened 2015-11-19
Fix MaxTopBlobs in Accuracy Layer

ronghanghu:
Fix the typo "MaxTopBlos" to "MaxTopBlobs". This typo causes maximum top number to be incorrect.


shelhamer:
Thanks for the fix @ronghanghu and thanks for the report @dbcam.

PR #3361, opened 2015-11-20
replace snprintf with a C++98 equivalent

eelstork:
`snprintf` is not consistently available on Windows.
Although patchy support is [possible][1] (search snprintf in suggested reference), I find that `snprintf` may be advantageously replaced by a solution which doesn't involve defining arbitrarily sized buffers and converting between C and C++ strings.
As I understand that C++ streams are not very popular and the main use of `snprintf` in Caffe is producing numbered batches, I suggest the introduction of a utility function (`format_int`).
[1]: https://github.com/BVLC/caffe/blob/6eae122a8eb84f8371dde815986cd7524fc4cbaa/src/gtest/gtest-all.cpp

### DIFF
--- a/examples/cifar10/convert_cifar_data.cpp
+++ b/examples/cifar10/convert_cifar_data.cpp
@@ -16,6 +16,7 @@
 
 #include "caffe/proto/caffe.pb.h"
 #include "caffe/util/db.hpp"
+#include "caffe/util/format.hpp"
 
 using caffe::Datum;
 using boost::scoped_ptr;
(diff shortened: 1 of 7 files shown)


ronghanghu:
IMO `snprintf` is a standard io function in C/C++, available in `<stdio.h>` / `<cstdio>` and is (or at least should be) cross-platform. http://www.cplusplus.com/reference/cstdio/snprintf/

Note: Microsoft has finally implemented `snprintf` in Visual Studio 2015.

eelstork:
@ronghanghu specifically, `snprintf` is available in C99 and C++11, not C++98 (for proof, replacing `<stdio.h>` with `<cstdio>` will break C.I. builds).
VS2015 is a bit too new for most Windows build workers.
This is a compatibility issue.

eelstork:
Opinions on this PR?

ronghanghu:
This all looks good to me. I'll try to review during the Thanksgiving (got a bit overwhelmed myself these days since it is near the end of the semester).

eelstork:
Thanks! Don't forget to relax during holidays.

ronghanghu:
Thanks @eelstork !

PR #3362, opened 2015-11-20
GetDB must return a value.

eelstork:
Currently compilation will fail with some compilers when LevelDB and LMDB are disabled.
Similar to #3303.


ronghanghu:
Thanks @eelstork

PR #3363, opened 2015-11-20
Exclude core.hpp when building without OpenCV

eelstork:
When building without OpenCV, of course we do not want any OpenCV header to be included.


ronghanghu:
Thanks @eelstork for the patch!

PR #3365, opened 2015-11-20
Fix syntax inside switch

eelstork:
glog macros inside `switch` cause the MSVC compiler to emit [error C2360](https://msdn.microsoft.com/en-us/library/61af7cx3.aspx). 
I find that GCC may be lenient in this case; standards compliance should we require that we manually define scopes to declare and initialise variables within `case` statements.
Additionally, `LOG(FATAL)` doesn't exonerate us from the requirement to `break` after `case`.

### DIFF
--- a/src/caffe/util/hdf5.cpp
+++ b/src/caffe/util/hdf5.cpp
@@ -29,10 +29,10 @@ void hdf5_load_nd_dataset_helper(
   CHECK_GE(status, 0) << "Failed to get dataset info for " << dataset_name_;
   switch (class_) {
   case H5T_FLOAT:
-    LOG_FIRST_N(INFO, 1) << "Datatype class: H5T_FLOAT";
+    { LOG_FIRST_N(INFO, 1) << "Datatype class: H5T_FLOAT"; }
     break;
   case H5T_INTEGER:
-    LOG_FIRST_N(INFO, 1) << "Datatype class: H5T_INTEGER";
+    { LOG_FIRST_N(INFO, 1) << "Datatype class: H5T_INTEGER"; }
     break;
   case H5T_TIME:
     LOG(FATAL) << "Unsupported datatype class: H5T_TIME";


lukeyeager:
See #3130 as well.

longjon:
Note that, by my quick read of https://github.com/google/glog/blob/master/src/glog/logging.h.in, it looks like this issue only affects `LOG_FIRST_N`.

This feels like an upstream issue to me; in general, macros should behave like function calls. We could consider a workaround (e.g., a wrapper around `LOG_FIRST_N`) pending an upstream fix (i.e., patch this in glog), but I'm not a big fan of mysteriously augmenting lots of calls that look perfectly fine.

> Additionally, `LOG(FATAL)` doesn't exonerate us from the requirement to `break` after `case`.

Sure it does, just as `return` or `exit(0)` does.

eelstork:
[Raising an issue and suggesting a fix](https://github.com/google/glog/issues/72) to the glog team.
In the meantime, I have updated the proposed fix to reflect suggested improvements.

eelstork:
(closed by mistake; I wouldn't close this since I don't know whether it will be fixed in glog)

eelstork:
With reference to google/glog#71 and google/glog#72, it appears that this can't be fixed in glog. @longjon, would you consider merging this?

shelhamer:
@willyd is this patch still helpful for windows, and if so is #3130 even better for compatibility?

eelstork:
Summary of what I know: patch is helpful; not fixable as an upstream issue (we tried); @longjon already debunked superfluous changes in #3130

willyd:
@shelhamer I agree with @eelstork the patch is needed for windows compatibility. The current state of the windows branch looks more like #3130 but this PR has only the required changes so if it was my call I would merge this one and close #3130 without merging it. Current windows branch also as a comment explaining the new scope creation see this [comment](https://github.com/BVLC/caffe/blob/windows/src/caffe/util/hdf5.cpp#L32-L34).

shelhamer:
Thanks @eelstork for the fix, @longjon for the investigation, and @willyd for the current branch perspective.

PR #2561, opened 2015-06-05
Added the unpooling layer, that does unpooling operation

nyf-nyf:
I've added the unpooling layer. It's described in articles of using Deconvolutional Neural Networks.

http://www.matthewzeiler.com/pubs/iccv2011/iccv2011.pdf

But i'm not expert in c++, would appreciate any fixes.
The version for cuda is missing, but works fine for me with cpu.

It helps to debug a convolutional network from bottom to top using unpooling, relu and deconvolution.

PR #3376, opened 2015-11-24
Add the Unpooling Layer.

mariolew:
Inspired by #2561 , I added the unpooling layer which is more general than #2561 (work with flexible pool size)with test and example use. Since this layer will never participate in the training stage, I didn't write the gpu version. Also, I only write unpooling for max pooling like #2561 , I will make it for average pooling later.


bhack:
Seems that this not merge and travis is failing.

mariolew:
@bhack  Actually, it works. If you are interested, you can give it a try and I have example to guide you through the usage.

chuong:
Hi All. Is there still any interest to merge this unpooling feature into caffe? It is useful to my project.

mariolew:
@chuong Actually, you can use the code although it is not merged.

chuong:
@mariolew I actually merged part of your code to a branch of Caffe. Then I realised that I need 3D-unpooling layer. As I cannot find elsewhere, I probably have to implement this by myself.

mariolew:
@chuong You can implement 3D unpooling layer yourself, I might try to implement one when I got some time. However, I don't think it can be much more difficult than to implement the 2D one.

Issue #3324, opened 2015-11-12
Directory creation fails intermittently


eelstork:
Referring:
- https://travis-ci.org/BVLC/caffe/jobs/90421120
- https://travis-ci.org/BVLC/caffe/jobs/90689090

I'll be looking into this. Suggestions welcome.

futurely:
Probably due to race condition. https://github.com/BVLC/caffe/pull/3300#issuecomment-158844531

eelstork:
@ronghanghu looking at #1063 I will try to find a fix for this (for now, checking the python source)

futurely:
http://stackoverflow.com/questions/15324947/boostfilesystemexists-crashs
http://stackoverflow.com/questions/11213940/boost-interprocess-race-condition-prevention
Create Temporary Files Correctly:
https://developer.apple.com/library/mac/documentation/Security/Conceptual/SecureCodingGuide/Articles/RaceConditions.html

eelstork:
Referring Python implementation of `mkdtemp`, if at first you don't succeed... ([tempfile.py, line 322](https://github.com/python-git/python/blob/715a6e5035bb21ac49382772076ec4c630d6e960/Lib/tempfile.py#L296))

ajtulloch:
FWIW,  bumping up the entropy in the random directory name (`caffe_test.%%%%%%` => `/caffe_test.%%%%%%%%%%%%%`) seemed to fix the issue on our CI.

ronghanghu:
> FWIW, bumping up the entropy in the random directory name (caffe_test.%%%%%% => /caffe_test.%%%%%%%%%%%%%) seemed to fix the issue on our CI.

Sounds like a good approach :+1:

eelstork:
We can fix this issue. For directories the approach I link is practically safe. I'm porting the fix and will submit a patch.
I suspect the way temporary files (vs directories) are created is more dangerous. It looks to me that this can fail silently. I want to raise a separate issue so I can submit a fix for directory creation today.

ronghanghu:
@eelstork Thank you for that!

PR #3378, opened 2015-11-24
Safely create temporary files and directories

eelstork:
With reference to #3324 I propose a modified implementation of `MakeTempDir`.
@futurely and @netheril96 suggested that the designated issue may be caused by a race condition. 

Upon closer inspection, I find that:

* Boost `unique_path` is specified to return a randomly generated path, which may or may not be unique. Boost implementations typically system level entropy sources to generate the path.
* Typical `mkdtemp` implementations rely on `mkdir` to generate temporary directories. Race conditions are avoided by checking the return value of `mkdir` instead of checking whether a generated path already exists.
* Paths generated using Boost use 4 bits of randomness per character. By comparison typical `mkdtemp` implementations may use 5 or 6 bits of randomness.

With this in mind the proposed solution is similar to how Python implements `mkdtemp` ([tempfile.py, line 322] (https://github.com/python-git/python/blob/715a6e5035bb21ac49382772076ec4c630d6e960/Lib/tempfile.py#L296)).

### DIFF
--- a/include/caffe/util/io.hpp
+++ b/include/caffe/util/io.hpp
@@ -2,12 +2,15 @@
 #define CAFFE_UTIL_IO_H_
 
 #include <boost/filesystem.hpp>
+#include <iomanip>
+#include <iostream>  // NOLINT(readability/streams)
 #include <string>
 
 #include "google/protobuf/message.h"
 
 #include "caffe/common.hpp"
 #include "caffe/proto/caffe.pb.h"
+#include "caffe/util/format.hpp"
 
 #ifndef CAFFE_TMP_DIR_RETRIES
 #define CAFFE_TMP_DIR_RETRIES 100
@@ -17,13 +20,6 @@ namespace caffe {
 
 using ::google::protobuf::Message;
 using ::boost::filesystem::path;
-
-inline void MakeTempFilename(string* temp_filename) {
-  temp_filename->clear();
-  const path& model = boost::filesystem::temp_directory_path()
-    /"caffe_test.%%%%%%";
-  *temp_filename = boost::filesystem::unique_path(model).string();
-}
 
 inline void MakeTempDir(string* temp_dirname) {
   temp_dirname->clear();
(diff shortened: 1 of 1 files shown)


ronghanghu:
Thanks @eelstork ! This looks good

eelstork:
Adding a commit targeting safe creation of temporary files. In this case the approach used to create temporary directories cannot be reused:

* `mkstemp` and `open` are not cross platform.
* `fopen` provides an `x` flag similar to `O_EXCL` however this requires C11.
*  I do not know of an equivalent using standard C++ or Boost.

The proposed solution consists in (safely) creating a temporary directory. This temporary directory is then used as a container for all temporary files.
A similar approach is described in [Apple's secure coding guide](https://developer.apple.com/library/mac/documentation/Security/Conceptual/SecureCodingGuide/Articles/RaceConditions.html#//apple_ref/doc/uid/TP40002585-SW10). In our case, however, temporary files are labelled using a counter, which eliminates the need to test for homonyms.

Optionally, this approach may be extended to the creation of temporary directories (meaning: safely create a temp directory, then use it as a container for all temp files and directories).

eelstork:
@ronghanghu this needs a little more work on my end, not compiling in VS.

ronghanghu:
@eelstork Thank you and don't worry. I'll take a look when it's ready :)

eelstork:
@ronghanghu ready for review.

ronghanghu:
@eelstork Thanks! I'll try to take a look tomorrow

include/caffe/util/io.hpp
```diff
@@ -9,26 +11,43 @@
...
 using ::boost::filesystem::path;
 
-inline void MakeTempFilename(string* temp_filename) {
-  temp_filename->clear();
-  const path& model = boost::filesystem::temp_directory_path()
-    /"caffe_test.%%%%%%";
-  *temp_filename = boost::filesystem::unique_path(model).string();
-}
+static path temp_files_subpath;
+static uint64_t next_temp_file = 0;
```
ronghanghu: Better to move this two static variables into function `MakeTempFilename` so that they are only accessible in that function.

include/caffe/util/io.hpp
```diff
@@ -9,26 +11,43 @@
...
+inline void MakeTempFilename(string* temp_filename) {
+  temp_filename->clear();
+  if ( temp_files_subpath.empty() ) {
+    string path_string="";
+    MakeTempDir(&path_string);
+    temp_files_subpath = path_string;
+  }
+  std::ostringstream s;
+  s << std::setw(9) << std::setfill('0') << next_temp_file++;
+  *temp_filename = (temp_files_subpath/s.str()).string();
```
ronghanghu: @eelstork So I see this implementation follows the Apple guide "The best way to handle this (Create Temporary Files Correctly) is to create a safe temporary directory that only you can access, then write the files into that directory."

Looks good to me.

ronghanghu:
@eelstork This should be ready. Let's fix the minor issue mentioned above and I'll merge.

eelstork:
Updated; looks good to me.

ronghanghu:
Thanks @eelstork !

PR #3379, opened 2015-11-24
Support OpenCV 2 and 3 in Preprocess function.

milq:
Support both versions of OpenCV in Preprocess function. This fix allow users to install Caffe using the last two major versions of OpenCV (branch 2 and branch 3).


flx42:
Duplicate of #2693

milq:
An updated and more simple pull request will be created due to ```cv::COLOR_BGR2GRAY``` and so on are compatible in both OpenCV (2 and 3).

PR #3390, opened 2015-11-27
Makefile: add -pg argument for debug

acgtyrant:
pg is use for generate extra code to write profile information suitable for the
analysis program gprof.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -246,6 +246,11 @@ endif
 ifeq ($(LINUX), 1)
 	CXX ?= /usr/bin/g++
 	GCCVERSION := $(shell $(CXX) -dumpversion | cut -f1,2 -d.)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -297,7 +302,7 @@ endif
 
 # Debugging
 ifeq ($(DEBUG), 1)
-	COMMON_FLAGS += -DDEBUG -g -O0
+	COMMON_FLAGS += -DDEBUG -g -O0
```
seanbell: Unnecessary trailing space

acgtyrant: @seanbell Sorry for my stupid typo, fix it done.

acgtyrant:
I should not put `-pg` with `-O2` together, and gprof is not a good profiler actually too. So I give up to maintain this branch.

PR #3392, opened 2015-11-27
Support OpenCV 2 and 3 in Preprocess function

milq:
Support both versions of OpenCV in Preprocess function. This fix allow users to install Caffe using the last two major versions of OpenCV (branch 2 and branch 3).


ronghanghu:
Thanks for the update.

@milq Note: you could update your existing PR #3379 instead of opening this new one.

ronghanghu:
Resolved in #2693.

Also thanks @milq for the effort!

PR #3395, opened 2015-11-29
Fix compatibility issues with extract_features

eelstork:
In some environments, `boost::shared_ptr` may conflict with `std` (with C++ TR1 or C++11). Suggesting we avoid ambiguous notations in this case.
`uint` is nonstandard.
This PR addresses compatibility issues.

### DIFF
--- a/tools/extract_features.cpp
+++ b/tools/extract_features.cpp
@@ -16,7 +16,6 @@ using caffe::Blob;
 using caffe::Caffe;
 using caffe::Datum;
 using caffe::Net;
-using boost::shared_ptr;
 using std::string;
 namespace db = caffe::db;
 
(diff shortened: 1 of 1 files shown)


tools/extract_features.cpp
```diff
@@ -51,7 +50,7 @@ int feature_extraction_pipeline(int argc, char** argv) {
   arg_pos = num_required_args;
   if (argc > arg_pos && strcmp(argv[arg_pos], "GPU") == 0) {
     LOG(ERROR)<< "Using GPU";
-    uint device_id = 0;
+    unsigned int device_id = 0;
```
longjon: This probably should have just been `int` all along, yes? `atoi` returns `int`, the check below is currently bogus, and `SetDevice` expects `int`. And we generally don't use `unsigned` per Google style guide.

This is the only appearance of `uint` in Caffe; I wonder if it's a typo!

eelstork: Maybe a typo.

eelstork:
Updated to reflect @longjon's sensible remark.

longjon:
Looks good, thanks @eelstork!

PR #3402, opened 2015-12-02
Make popcount cross-platform

eelstork:
`__builtin__popcount` is a [built-in GCC function] (1). There is no standard C/C++ equivalent.

For portability I provide an explicit popcount implementation; for performance on Windows, MSVC built-ins are suggested.

Compiler specific code may be avoided; should we make an exception in this case?

**Implementation note**

`_GNUC_` is also defined with llvm/clang; testing for `__llvm__` and/or `__clang__` will not behave as expected. [Less naive compiler macros][2] might be used (at the cost of adding Boost headers).

Currently pop count is only used in `hamming_distance`; It would appear that Caffe [does not use this function][3].

[1]: https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
[2]: http://www.boost.org/doc/libs/1_59_0/libs/predef/doc/html/predef/reference/boost_comp_compiler_macros.html
[3]: https://github.com/BVLC/caffe/search?utf8=✓&q=hamming

### DIFF
--- a/src/caffe/util/math_functions.cpp
+++ b/src/caffe/util/math_functions.cpp
@@ -7,6 +7,26 @@
 #include "caffe/util/math_functions.hpp"
 #include "caffe/util/rng.hpp"
 
+#if defined(__GNUC__)
+  #define popcount __builtin_popcount
+  #define popcountl __builtin_popcountl
+#elif defined(_MSC_VER)
+  #define popcount __popcnt
+  #define popcountl __popcnt64
+#else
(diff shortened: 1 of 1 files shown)


ronghanghu:
I guess you don't need this one. You can just remove `caffe_cpu_hamming_distance` IMO.

Actually I don't think you need `caffe_cpu_hamming_distance` at all, and it is perhaps a mistake to merge it one year ago...

longjon:
Obsoleted by #3404.

Issue #3403, opened 2015-12-02
Hamming distance not used in Caffe, hinders portability. Remove?


eelstork:
Caffe [does not use][1] `hamming_distance`; indirectly, this function limits portability via `__builtin__popcount`.

In #3402 I offer a solution to the compatibility issue.

@ronghanghu suggested that `hamming_distance` may be removed. @longjon , @bhack would you agree that this might be removed?
@kloudkl clearly some work went into this; would you favour removing this function from Caffe?

[1]: https://github.com/BVLC/caffe/search?utf8=✓&q=hamming

shelhamer:
I'm ok with dropping this function since it was never put to use by a completed PR and the disabled tests for it have led to confusion in the past. I'm open to other opinions on this though.

longjon:
Dead code should be removed.

It looks like this was added for use in an example that was never merged. However, even that example does not appear to require this as a feature of Caffe. Unless I have missed something, @ronghanghu is right in #3402; this code does not belong in Caffe.

eelstork:
Submitted a PR.

longjon:
Merged, thanks!

futurely:
It's inevitable to say goodbye to the [Done+Dusted](http://www.doneanddusted.com/). C'est la vie.

PR #3404, opened 2015-12-02
Remove hamming_distance and popcount

eelstork:
With reference to #3403 and #3402, this PR implements the removal of `hamming_distance()` GPU, CPU implementations, along with related unit tests.


longjon:
Looks good, thanks @eelstork. There's no code like no code.

As this does look like dead code with no greater justification for its existence, and as @ronghanghu and @shelhamer seem to agree, I'm merging.

Issue #778, opened 2014-07-24
Group parameter


ardila:
I can't find anywhere what the meaning of the "group" parameter is. It seems to halve the number of locations a filter gets applied when set to 2?

```
layers {
  name: "conv2"
  type: CONVOLUTION
  bottom: "norm1"
  top: "conv2"
  blobs_lr: 1
  blobs_lr: 2
  weight_decay: 1
  weight_decay: 0
  convolution_param {
    num_output: 256
    pad: 2
    kernel_size: 5
    group: 2
  }
}
```

lxduan:
Any follow-ups regarding the "group" parameter? Right now, the `bvlc_reference_caffenet/train_val.prototxt` still has `group` set, e.g. [here](https://github.com/BVLC/caffe/blob/master/models/bvlc_reference_caffenet/train_val.prototxt#L128).

Based on the above discussion, grouping will eventually be removed from Caffe?

CRIPAC:
I print the bvlc_alexnet model's parameters, and I found that there are only half parameters in CONV2(48), CONV4(128) and CONV5(192). I think the parameters is should be twice. Is any problem here ?

dangweili:
I print the bvlc_alexnet model's parameters, and I found that there are only half parameters in CONV2(48), CONV4(128) and CONV5(192). I think the parameters is should be twice. Is any problem here ?

kli-nlpr:
Oh God, I just notice the group parameter today in AlexNet.
Do other papers who use AlexNet also use the group parameter?

wk910930:
Group is helpful in my case where I need to do the channel-wise convolution.

myyan92:
Group is helpful in channel-wise deconvolution too. I think it just adds flexibility to layer setup and should not do harm. Just delete it in the proto if you don't need it.

Jongchan:
@wk910930 @myyan92 

I am trying to use the 'group' option to perform channel-wise convolution. Intermediate channels perform 3x3x64x64 convolution, where I want to use 'group: 64' option.

However, the gradient seems to explode, and the memory requirement goes up, and the training becomes slower. Is there any extra tips for using this option? I can't find any good example using the group option.

```
layer {
  name: "conv4"
  type: "Convolution"
  bottom: "conv3_1"
  top: "conv4"
  param {
    lr_mult: 1
  }
  param {
    lr_mult: 0.1
  }
  convolution_param {
    num_output: 64
    kernel_size: 3
    stride: 1
[...]

PR #3413, opened 2015-12-04
Remove group parameter from Alexnet model

gheinrich:
Grouping is non standard in Alexnet and causes trained models
to feature fewer filters than standard models do.

As highlighted on https://github.com/BVLC/caffe/issues/778, grouping
should be removed.

### DIFF
--- a/models/bvlc_alexnet/deploy.prototxt
+++ b/models/bvlc_alexnet/deploy.prototxt
@@ -70,7 +70,6 @@ layer {
     num_output: 256
     pad: 2
     kernel_size: 5
-    group: 2
   }
 }
 layer {
@@ -143,7 +142,6 @@ layer {
     num_output: 384
     pad: 1
     kernel_size: 3
-    group: 2
   }
 }
 layer {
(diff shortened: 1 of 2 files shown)


seanbell:
1) I don't think it manes sense to talk about "non standard" when talking about a reference implementation of AlexNet, the original model defined in Krizhevsky et al 2012 (http://www.cs.toronto.edu/~fritz/absps/imagenet.pdf).

2) Editing this in place breaks existing BVLC models.  If you want to remove the group parameter, you should define a new model with a different name.  I don't think we should overwrite the original reference implementation.  That's the whole point of a reference implementation.

3) I think empirical studies are needed to justify this kind of change (i.e. training with and without grouping).  I don't think you can predict a priori which will do better without a proper comparison.  There are theoretical reasons why you might think either would do better (removing grouping adds parameters, but leaving grouping in adds sparsity and a division of information).

If it turns out that removing the grouping leads to higher performance, I think the right approach is to define a new model with a new name, and encourage people to that.  Not overwrite the original reference implementation.

jeffdonahue:
@seanbell is right on all points.

nmoezzi:
Is there a trained model of AlexNet on ImageNet **without grouping**?

PR #3433, opened 2015-12-09
Fix ImageDataLayer's silent failure on file paths with spaces 

### LINKED ISSUE #1935: ImageData requires file paths with no spaces
I'm not really sure if this is a bug or a feature request, but it would be nice if the file paths passed into the ImageData layer could have spaces in them and respected the shell '\ ' escape sequences.

Here's the relevant line:
https://github.com/BVLC/caffe/blob/f998127ef2224b15dbcee9d392b080b978788bd0/src/caffe/layers/image_data_layer.cpp#L40

If I have some free time I'll make a PR with the changes.

bchu:
by splitting on last space.

Also fixes its test to contain newlines. Fixes #1935 and #1951.

Note: I don't work with C++ very much, so please let me know if there's a better way of doing this.
I've also changed window_data_layer and hdf5_data_layer to read in lines. Please check that this is correct (I have never used hdf5 or window_data).

### DIFF
--- a/src/caffe/layers/hdf5_data_layer.cpp
+++ b/src/caffe/layers/hdf5_data_layer.cpp
@@ -81,7 +81,7 @@ void HDF5DataLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
   std::ifstream source_file(source.c_str());
   if (source_file.is_open()) {
     std::string line;
-    while (source_file >> line) {
+    while (std::getline(source_file, line)) {
       hdf_filenames_.push_back(line);
     }
   } else {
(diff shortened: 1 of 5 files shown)


bchu:
I realized that the issue is already fixed by this PR: https://github.com/BVLC/caffe/pull/1971, so I closed this PR.

Re-opening in case people think it's a good idea. This method is less forgiving than the regex method, with the sole benefit of avoiding a dependency on boost regex.

vimalthilak:
I would be happy to see this fix merged since we ran into this problem while using caffe in our group at work. With that said, it'd be nice to use a comma or something that's more robust than just a space as a separator.

bchu:
Resolved by https://github.com/BVLC/caffe/pull/4059

PR #3439, opened 2015-12-10
Fix CuDNNConvolutionLayer for cuDNN v4

flx42:
Add a macro to check the current cuDNN version

### DIFF
--- a/include/caffe/util/cudnn.hpp
+++ b/include/caffe/util/cudnn.hpp
@@ -7,6 +7,9 @@
 #include "caffe/common.hpp"
 #include "caffe/proto/caffe.pb.h"
 
+#define CUDNN_VERSION_MIN(major, minor, patch) \
+    (CUDNN_VERSION >= (major * 1000 + minor * 100 + patch))
+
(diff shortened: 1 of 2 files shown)


flx42:
The CMake website is under maintenance, causing the automated cmake builds to fail.
ping @shelhamer for review, it should be a quick one.

shelhamer:
@flx42 will there be further ifdef guards for v4? If not this is simple and done. I'm happy to keep v3 compatibility and add v4 if it doesn't obscure the code. So far we have followed the current cuDNN version and dropped old versions (because of API changes).

Thanks for the PR.

thatguymike:
This is the only change until we get to more feature support like batch normalization.  We should generally support the macro to better deal with API change-overs.

flx42:
For v3 and v4 only, I believe you could simply replace `cudnnAddTensor` by `cudnnAddTensor_v3`. But this function is also marked as deprecated so it seems less future-proof.

shelhamer:
Ok, agreed on the macro and this is fine as-is.

PR #3444, opened 2015-12-11
Replace uses of POSIX open/close

eelstork:
Protobuf related i/o in Caffe relies on POSIX `open`/`close`. These functions are not cross-platform.
If performance concerns are not critical, the proposed solution (suggested by @willyd and @keenbrowne) may be adequate.

Previously, `ReadXXX`/`ReadXXXOrDie`, `WriteXXX`/`WriteXXXOrDie` variants were available; Fail-slow variants have no uses in higher level APIs and possibly unsafe uses in tests and examples.

### DIFF
--- a/examples/cpp_classification/classification.cpp
+++ b/examples/cpp_classification/classification.cpp
@@ -119,7 +119,7 @@ std::vector<Prediction> Classifier::Classify(const cv::Mat& img, int N) {
 /* Load the mean file in binaryproto format. */
 void Classifier::SetMean(const string& mean_file) {
   BlobProto blob_proto;
-  ReadProtoFromBinaryFileOrDie(mean_file.c_str(), &blob_proto);
+  ReadProtoFromBinaryFile(mean_file.c_str(), &blob_proto);
 
   /* Convert from BlobProto to Blob<float> */
   Blob<float> mean_blob;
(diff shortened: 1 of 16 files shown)


ajtulloch:
- What's the perf change for loading a nontrivial model (say, a hundred MB) with this diff?

- Having the non-CHECK'ing APIs are useful for cases where you're embedding Caffe in a larger process, and a failure to read a file doesn't need (and shouldn't) take down the whole process.  I'd be in favor of not removing them.

eelstork:
@ajtulloch a performance dip is expected according to `IstreamInputStream` [docs] [1]; I don't mind running a couple of tests if interested parties perceive that this is important - might you contribute data?

Although I agree that error handling should be customisable, the current implementation falls short. For sake, have a look at `ReadProtoFromTextFile`. An assertive check is carried out regardless, with the returned flag only covering parsing correctness. Is this expected behaviour?

Raising standard or custom i/o exceptions might be appropriate.

[1]: https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.io.zero_copy_stream_impl#IstreamInputStream

willyd:
+1 for this as it will help many of us maintaining Windows forks of Caffe.

PR #3451, opened 2015-12-14
Fix to cmake build for clang

jczaja:
Failure scenario that is fixed:

1) mkdir mybuild; cd mybuild
2) CXX=/usr/bin/clang++ cmake \<path to CMakeLists.txt\> -DBLAS=open
3) make all; make test
4) make runtest

Root cause of problem:
cmake/Targets.cmake  is having a check for BUILD_SHARED_LIBS , but at the moment 
of inclusion of mentioned file, such variable does not exist yet. 

Solution:
Condition checking for BUILD_SHARED_LIBS was wraped into macro , and invoked
lateron eg. after BUILD_SHARED_LIBS is initialized

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,8 @@ if(UNIX OR APPLE)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall")
 endif()
 
+caffe_set_caffe_link()
+
 if(USE_libstdcpp)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libstdc++")
   message("-- Warning: forcing libstdc++ (controlled by USE_libstdcpp option in cmake)")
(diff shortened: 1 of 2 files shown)


shelhamer:
@Nerei can you check this?

Nerei:
Hi, @jczaja This doesn't reproduce for me. What is an effect of these repro-steps? Non-registered layers? Shouldn't be. Nonetheless the change looks reasonable. Looks good for me.

jczaja:
Hi, @Nerei ,

I missed (not intentionally) important detail here. Reproduction is to be achieved only on *Linux*. 

 Effect of reproduction is that linking won't be successful. Build won't be successful. This is because if BUILD_SHARED_LIBS is not present at the time of checking then MacOS's Clang specific line will be executed eg.

set(Caffe_LINK -Wl,-force_load caffe)

instead of:

 set(Caffe_LINK caffe)

So my Bad that I did not explicitly mentioned that this is Linux specific. It is very likely that it works fine on MacOS.

Nerei:
Thanks. Anyway good change

shelhamer:
Thanks for the fix @jczaja and check @Nerei.

PR #3460, opened 2015-12-17
Document exception, add details to exception message in _Net_forward

achalddave:


achalddave:
@shelhamer Sorry to ping you directly but you merged my other PR quickly so I figured I'd bug you :) This one's also fairly minor and has been left around for a while, could you take a look?

lukeyeager:
I like it, but I think a `TypeError` is more appropriate than a raw `Exception`.
https://docs.python.org/2/library/exceptions.html#exceptions.TypeError

python/caffe/pycaffe.py
```diff
@@ -109,8 +116,12 @@ def _Net_forward(self, blobs=None, start=None, end=None, **kwargs):
         outputs = set(self.outputs + blobs)
 
     if kwargs:
-        if set(kwargs.keys()) != set(self.inputs):
-            raise Exception('Input blob arguments do not match net inputs.')
+        expected_kwargs = set(kwargs.keys())
+        received_kwargs = set(self.inputs)
```
shelhamer: Expected and received are swapped, or that's my interpretation. Inputs are the expected args (if any).

achalddave: Ah, I fixed this on one machine and forgot to commit it, sorry. Should be fixed now.

achalddave:
@lukeyeager I'm not entirely sure if ValueError or TypeError is the right exception, but I'll take your suggestion and I've made it a TypeError for now unless anyone disagrees.

PR #3468, opened 2015-12-20
Replace blobs_lr with lr_mult in readme.md.

mnogu:
models/finetune_flickr_style/deploy.prototxt uses `lr_mult` now.


longjon:
Thanks!

PR #3471, opened 2015-12-21
New tutorial on python datalayers and multilabel classification

beijbom:
The goal of this tutorial is twofold. First, it illustrates how to write python datalayers. Both syncronous and asyncronous. Second, it clarifies a common misconception that caffe "doesn't support multi-label classification". It does, and the correct loss layer is SigmoidCrossEntropyLoss.


wangg12:
How about the evaluation metrics of multilabel classification like the accuracy of the multi-label one?

beijbom:
hey Wangg12, I use the hamming distance during test-time directly in the notebook, but did not put it in a python layer. Is that what you refer to?

examples/pycaffe/layers/pascal_multilabel_datalayers.py
```diff
@@ -0,0 +1,258 @@
...
+            # Store in a result list.
+            self.result['data'].append(self.transformer.preprocess(im))
+            self.result['label'].append(multilabel)
+            self._cur += 1
+
+
+def load_pascal_annotation(index, pascal_root):
+    """
+    This code is borrowed from Ross Girshick's FAST-RCNN code. It parses the PASCAL .xml metadata files.
+    Thanks Ross!
```
seanbell: You should cite the URL (https://github.com/rbgirshick/fast-rcnn) and tech report (http://arxiv.org/abs/1504.08083) that accompanies Fast R-CNN.  The current citation isn't quite enough.

beijbom: citation updated. thanks for the pointer!

wangg12:
Hi, @beijbom 

According to [Multi-label_classification](https://en.wikipedia.org/wiki/Multi-label_classification),the hamming loss and multi-label version accuracy are equivalent to some extent. 

In my view, hamming distance cares more about the difference between prediction and groundtruth, while the multi-label version accuracy focuses more on how much groundtruth and top-k prediction are the same.(Correct me if I'm wrong.) 

So if I want to use the multi-label version accuracy during test, could you give me some instructions on how to write this with python layer? 

Thanks!

kli-nlpr:
Great, I have been looking for python data layer for a long time.
Thank you very much.   @beijbom

lukeyeager:
This tutorial may also be useful. It may be more beginner friendly.

https://github.com/NVIDIA/DIGITS/blob/digits-3.0/examples/python-layer/README.md

kli-nlpr:
@lukeyeager  Thank you very much, I will check it out. o(^▽^)o

examples/pascal_multilabel_with_datalayer_tools.py
```diff
@@ -0,0 +1,111 @@
+import numpy as np
```
shelhamer: I think this would be at home as `caffe/examples/pycaffe/tools.py` to make it more clearly related to pycaffe examples.

beijbom: sounds good. I just moved it.

examples/pycaffe/layers/pascal_multilabel_datalayers.py
```diff
@@ -0,0 +1,262 @@
...
+
+import numpy as np
+import os.path as osp
+
+from xml.dom import minidom
+from random import shuffle
+from threading import Thread
+from PIL import Image
+
+from pascal_multilabel_with_datalayer_tools import SimpleTransformer
```
elezar: Should this not now be "from tools import SimpleTransformer"?

beijbom: You are right, I'll fix that. Thanks!

elezar: @beijbom I have submitted a pull request to your personal branch that also addresses this. Please have a look.

By the way, the tutorial is great! I was able to reproduce it with my multi-label data layer that I am working on for #3653 as well (still need to push the lastest changes).

beijbom: Thanks a ton Evan, very nicely done. I'm not sure the new layer code is ideal from a tutorial perspective though. While certainly more professional, I think it obscures things a bit, and the reader have to jump back and forth to follow the line of execution. I tend to favour showing people the easiest, dumbest, and most straight-forward way of doing things, and let them take it from there. Does that makes sense? What do other folks think? @shelhamer?

elezar: Now that you mention it, I might have gotten a bit carried away.

I was thinking about it this morning again, and I don't think a base class is necessarily required, but I would still think that using the `BatchLoader` in both the sync and async cases is a good idea as it clarifies what the steps in the process are, and users can "drill down" further if they need more info as to what actually happens when loading an image.

beijbom: Ok, sure, that makes sense. That's a good compromize. :). Do you want to make a new pull requests, or should I edit yours?

elezar: I don't think that I'm going to get much of a chance to have a look this weekend. If it can wait until Monday, then I don't mind doing it myself. Otherwise, feel free to edit it.

shelhamer:
Thanks for the multi-label tutorial Oscar!

Potential follow-ups for the willing:

- train the defined multi-label CaffeNet classifier to convergence on PASCAL to have a more exciting conclusion and yet another model for the zoo
- switch to proto solver definition like in the fine-tuning notebook
- add more explanation of classification/multi-class classification/multi-label classification with pointers to the relevant losses
- include a pre-fetching Python data layer as an IO optimization (this should be another data layer to make this easier on beginners)

beijbom:
tnx Evan!

I'll revisit some of the follow-ups after ECCV.

The tutorial doesn't appear right on http://caffe.berkeleyvision.org/. Do
we need to take further action to make that happen? Perhaps I missed
setting one of the ipython notebook metadata fields?

I did set:
"include_in_docs": true,

but are there others?

cheers

shelhamer:
I had to push the new site from the latest master. It's there now!

kli-nlpr:
Hi, @beijbom . Why did you remove the  asyncronous  layer?

In my opinion, It's benifical to prefetch a batch for next iteration when caffe is running.

For the syncronous layer, caffe have to waite for batch loadiing, right?

Thanks!

beijbom:
hi @kli-nlpr! We decided to remove it because we didn't see any significant time savings (I agree that it SHOULD make sense, it just didn't :). It was unclear whether it was a implementation or hardware issue. Did you notice any time saving?

kli-nlpr:
Hi, @beijbom   I didn't test it, I just wondering why you delete it.
Does this mean that when writing python data layer, one don't have to consider prefetching?

elezar:
@kli-nlpr If you are really interested in performance, you can have a look at #3653, which adds multi-label support to the standard image data layer. This already implements batch prefetching and is quite a bit faster than the python layers here. Some testing and feedback would be welcome, and if there is more interest from the community, then maybe the PR could get merged sooner. I tried to reproduce @beijbom's pascal tutorial there too, so it should be simple enough to get started.

Note: I am trying to knock @beijbom's layers. They are great examples of some of the more advanced uses of Caffe's python interface!

kli-nlpr:
@elezar Thank you very much, I will check it out.

In fact, I am much more comfortable with python than C++.
I think it’s much easier to do some complex data preparation using python layer. 
But the speed is much slower than C++ layers.
I am looking for methods which can speed up python data layers. ^_^

kli-nlpr:
Is there a **pre-fetching Python data layer** right now? Thanks

beijbom:
We have experimented with a pre-fetching Python data layers, but were unable to see a consistent speed-up so far. Let me know if you get it to work.

PR #3476, opened 2015-12-22
Links fixed

noshaba:
Just some broken links updated/fixed.


longjon:
Closing as this does not appear correct -- these links are designed to work from the generated pages (served on caffe.berkeleyvision.org), not from the markdown viewed on GitHub. (Ideally they would work in both situations -- you're welcome to look into that, if it's possible.)

PR #3483, opened 2015-12-25
Improve the AdaDelta performance

buaaliyi:
I found that the AdaDelta computing is quite slow, due to the bad performance of pow(x, 2) and pow(x, 0.5). So it's better to use x*x and sqrt(x) to instead of pow, especially the sqrt may takes advantage of Intel hardware optimization in CPU mode.

For more information, here is the reference: http://stackoverflow.com/questions/17417490/difference-between-sqrtx-and-powx-0-5


seanbell:
There's no need for a separate `sqr` function -- you can use `mul` and provide the same argument twice.

e.g. `caffe_gpu_mul(n, x, x)` instead of `caffe_gpu_sqr(n, x)`

PR #3492, opened 2015-12-29
ImageDataLayer minor refactor and optimization

jeffdonahue:
This PR has two commits.

The first (e8d770d) shouldn't change anything functionally -- it just cleans up ImageDataLayer a bit, adding a `ReadCurrentImageToCVMat` method to remove a bit of duplicate code and declares variables for `this->layer_param_.image_data_param()` to reduce line noise.

The second (c8b94bc) changes `load_batch` to avoid reading the first image of each batch twice by moving the `top_shape` inference logic to the inside of the loop's first iteration.  This is a 50% reduction in disk reads (and resizing if applicable) for fully convolutional training done using this layer (or more generally, any training with batch size 1).


vimalthilak:
@jeffdonahue Would it make sense to change ImageDataLayer to accommodate spaces in file names. A simple change would be to split string based on the last space but a better way might be to use comma-separated value instead of space-delimited values for image name and ID values?

PR #3493, opened 2015-12-29
minor cleanup/refactor of ReadImageToCVMat methods

jeffdonahue:
This is a minor refactoring of the `ReadImageToCVMat` methods that I think makes the code a bit cleaner.  Rather than all the methods calling the most general case (which does both loading and resizing), the load+resize version now calls the simpler load-only version and then does the resizing.

It also uses default parameters for `is_color` to get rid of two extra versions of the method.  (I realize these go against the Google style guide recommendations, but they're used enough elsewhere in Caffe that it seems we've decided they're acceptable for this project.  Aside from merely being acceptable though, it's a matter of opinion and taste whether this code is an improvement over the existing version and I'm happy to switch back to the existing "explicit defaults" if that's the consensus.)


src/caffe/util/io.cpp
```diff
@@ -72,34 +72,21 @@ void WriteProtoToBinaryFile(const Message& proto, const char* filename) {
...
-    CV_LOAD_IMAGE_GRAYSCALE);
-  cv::Mat cv_img_origin = cv::imread(filename, cv_read_flag);
-  if (!cv_img_origin.data) {
-    LOG(ERROR) << "Could not open or find file " << filename;
-    return cv_img_origin;
-  }
-  if (height > 0 && width > 0) {
-    cv::resize(cv_img_origin, cv_img, cv::Size(width, height));
-  } else {
-    cv_img = cv_img_origin;
```
jeffdonahue: Aside from style cleanup, this PR also eliminates this unnecessary copy in the no resize case by returning immediately on line 76.  (Not sure whether it's significant, but the performance should be strictly better.)

PR #3496, opened 2015-12-30
TestDataTransformer: fix some memory leaks

jeffdonahue:
This fixes some memory leaks in `TestDataTransformer` caused by the use of `new` (with no corresponding `delete` or other cleanup).


jeffdonahue:
I'll go ahead and merge this minor unit test fix.

PR #3500, opened 2015-12-31
Handle signals with boost Asio

eelstork:
Referring #2253 and #2012, signal handling was implemented in Caffe using `sigaction`. I suggest a cross platform implementation.
@willyd suggested using Boost Asio to make signal handling more cross platform.
@ronghanghu as you were involved in testing the original implementation, would you please have a look at this?


longjon:
I'm not ready to review in detail, but this looks like like a good direction; we already have a strong boost dependency, and this seems to simplify the code.

PR #2824, opened 2015-07-27
Nd conv pool

WGW101:
Hi !

Following my issue ticket #2671, here is my pull request for nD convolution and pooling using CuDNN primitives.

The nD convolution by itself seems to work, but the biases addition using cudnnAddTensor() returns NOT_SUPPORTED status.

The nD pooling doesn't work, and returns NOT_SUPPORTED.
Apparently the nD pooling descriptor might only be a place-holder in this version, so this might work with future version of cuDNN...

I inherits my layers directly from the Layer class and not the BaseConvolutionLayer and BasePoolingLayer to avoid modifying any existing (and working..) features.
The major drawback of this approach is that it won't be able to rely on other engines if CuDNN is not supported by the user configuration. But as I declared a LayerFactory for NdConvolution and NdPooling, it might be relatively easy to solve this behaviour.

Don't hesitate to give me feedbacks on this two new layers,
and share any new insight about why it doesn't work.

I'm already aware of the #2049 PR for nD convolution, but I'm still missing nD pooling (actually I only need 3D pooling in my application).

Cheers,

PR #3515, opened 2016-01-05
nd convolution and pooling revision

lfz:
This is a revision of #2824 , since it is already deprecated.

I fixed some bugs mentioned in the discussion, now it supports cudnn v3

an example of using:
```

layer {
  name: "conv1"
  type: "NdConvolution"
  bottom: "3d_data" # actually 5 dimension here, you may need reshape
  top: "conv1"
  convolution_param {
    num_output: 10
    stride:1
    # specifies the index of the "channels" axis --
    # may be omitted as 1 is the default
    kernel_shape { dim: 4 dim: 4 dim: 3}
    weight_filler {type: "xavier"}
  }
}



# layer {
#   name: "pool1"
#   type: "NdPooling"
#   bottom: "conv1" 
#   top: "pool1" 

#   pooling_param {
#     pool: MAX
#     kernel_shape {dim: 3,dim: 3,dim: 2}
#     stride_shape  {dim: 3,dim: 3, dim: 2}
#   }
# }
```


lfz:
I am sorry that this only works with a gpu, 

and because of that the checks didn't pass without cuda. 

And I don't know how to bypass this

include/caffe/blob.hpp
```diff
@@ -137,10 +137,10 @@ class Blob {
   /// @brief Deprecated legacy shape accessor width: use shape(3) instead.
   inline int width() const { return LegacyShape(3); }
   inline int LegacyShape(int index) const {
-    CHECK_LE(num_axes(), 4)
-        << "Cannot use legacy accessors on Blobs with > 4 axes.";
-    CHECK_LT(index, 4);
-    CHECK_GE(index, -4);
+    CHECK_LE(num_axes(), 5)
```
seanbell: Is this a bug?  The legacy shape accessors should not allow for 5D blobs.

src/caffe/proto/caffe.proto
```diff
@@ -804,21 +808,24 @@ message PoolingParameter {
...
   optional uint32 stride_h = 7; // The stride height
   optional uint32 stride_w = 8; // The stride width
+  optional BlobShape stride_shape = 12;
   enum Engine {
     DEFAULT = 0;
     CAFFE = 1;
     CUDNN = 2;
   }
-  optional Engine engine = 11 [default = DEFAULT];
+  optional Engine engine = 14 [default = DEFAULT];
```
seanbell: It breaks existing models to change the id of existing parameters.

include/caffe/blob.hpp
lfz: it's because I want to use the xavier initializer, I'll change it

ddetone:
caffe/src/caffe/layers/cudnn_ndpooling_layer.cpp line 13 has a typo, const is written twice

lfz:
Thanks @ddetone , it has been fixed

YOUNGING:
Is it possible to implement a ND deconvolutional layer using cudnn?

YOUNGING:
Is there anyone can make sure that this PR works?I tried this Ndpooling and the master branch 3D convolutionLayer,but I have to reduce my learningrate to 1e-18 to avoid loss rising to Nan.Is this normal?

lfz:
@YOUNGING really? Please supply me with the sample proto and result

YOUNGING:
@lfz  When compiled with  "USE_CUDNN" option , caffe will use CUDNN engine,right?
But most of cudnn layers don't have a ND implementation.
It will give an error like this:


neural layer like ReLU can use CAFFE engine by setting engine to CAFFE,but SoftmaxWithLoss layer don't have an engine option,so what should I do to make things right?

BTW,reshaping the input of loss to 2D may solve the problem,but I want to find a more efficient way.Any ideas?

Thanks.

include/caffe/util/cudnn.hpp
```diff
@@ -105,7 +139,26 @@ inline void setConvolutionDesc(cudnnConvolutionDescriptor_t* conv,
...
+  std::vector<int> shape(pad.size()+2);
+  cudnnDataType_t cudnn_type;
+  cudnnGetFilterNdDescriptor(filter, shape.size(), &cudnn_type, &nbDims,
+      shape.data());
+  CHECK_EQ(nbDims, pad.size()+2) <<
+      "Dimensions of filters and pad don't match !";
+  CHECK_EQ(nbDims, stride.size()+2) <<
+      "Dimensions of filters and stride don't match !";
+  std::vector<int> upscale(pad.size(), 1);
+  CUDNN_CHECK(cudnnSetConvolutionNdDescriptor(*conv,
```
chuckcho: This function takes an additional input arg in CUDNN v4 (or possibly earlier version too)
Something like following will take care of this:

#if CUDNN_VERSION >= 4007
  CUDNN_CHECK(cudnnSetConvolutionNdDescriptor(*conv,
              pad.size(), pad.data(), stride.data(), upscale.data(),
              CUDNN_CROSS_CORRELATION, cudnn_type));
#else
   CUDNN_CHECK(cudnnSetConvolutionNdDescriptor(*conv,
               pad.size(), pad.data(), stride.data(), upscale.data(),
               CUDNN_CROSS_CORRELATION));
#endif

src/caffe/layers/cudnn_ndconv_layer.cu
```diff
@@ -0,0 +1,166 @@
...
+                    cudnn::dataType<Dtype>::one,
+                    top_descs_[i],  top_diff + top_offset_ * g,
+                    cudnn::dataType<Dtype>::one,
+                    bias_desc_, bias_diff + bias_offset_ * g));
+      }
+
+      // Gradient w.r.t. weights.
+      if (this->param_propagate_down_[0]) {
+        const Dtype* bottom_data = bottom[i]->gpu_data();
+        CUDNN_CHECK(cudnnConvolutionBackwardFilter(handle_[1*this->group_ +
```
chuckcho: Consider the following mod:
```
#if CUDNN_VERSION >= 4007
        CUDNN_CHECK(cudnnConvolutionBackwardFilter_v2(handle_[1*this->group_ +
                    g],
                    cudnn::dataType<Dtype>::one,
                    bottom_descs_[i], bottom_data + bottom_offset_ * g,
                    top_descs_[i],    top_diff + top_offset_ * g,
                    conv_descs_[i],
                    cudnn::dataType<Dtype>::one,
                    filter_desc_, weight_diff + weight_offset_ * g));
#else
        CUDNN_CHECK(cudnnConvolutionBackwardFilter(handle_[1*this->group_ +
                    g],
                    cudnn::dataType<Dtype>::one,
                    bottom_descs_[i], bottom_data + bottom_offset_ * g,
                    top_descs_[i],    top_diff + top_offset_ * g,
                    conv_descs_[i],
                    cudnn::dataType<Dtype>::one,
                    filter_desc_, weight_diff + weight_offset_ * g));
#endif
```

src/caffe/layers/cudnn_ndconv_layer.cu
```diff
@@ -0,0 +1,166 @@
...
+                    filter_desc_, weight_diff + weight_offset_ * g));
+      }
+
+      // Gradient w.r.t. bottom data.
+      if (propagate_down[i]) {
+        if (weight == NULL) {
+          weight = this->blobs_[0]->gpu_data();
+        }
+        Dtype* bottom_diff = bottom[i]->mutable_gpu_diff();
+        CUDNN_CHECK(cudnnConvolutionBackwardData(handle_[2*this->group_ + g],
```
chuckcho: Consider this:
```
#if CUDNN_VERSION >= 4007
        CUDNN_CHECK(cudnnConvolutionBackwardData_v2(handle_[2*this->group_ + g],
                    cudnn::dataType<Dtype>::one,
                    filter_desc_, weight + weight_offset_ * g,
                    top_descs_[i], top_diff + top_offset_ * g,
                    conv_descs_[i],
                    cudnn::dataType<Dtype>::zero,
                    bottom_descs_[i], bottom_diff + bottom_offset_ * g));
#else
        CUDNN_CHECK(cudnnConvolutionBackwardData(handle_[2*this->group_ + g],
                    cudnn::dataType<Dtype>::one,
                    filter_desc_, weight + weight_offset_ * g,
                    top_descs_[i], top_diff + top_offset_ * g,
                    conv_descs_[i],
                    cudnn::dataType<Dtype>::zero,
                    bottom_descs_[i], bottom_diff + bottom_offset_ * g));
#endif
```

src/caffe/layers/cudnn_ndpooling_layer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  cudnn::setTensorNdDesc<Dtype>(&bottom_desc_, input_shape_);
+  cudnn::setTensorNdDesc<Dtype>(&top_desc_, pooled_shape_);
+}
+
+template <typename Dtype>
+void CudnnNdPoolingLayer<Dtype>::compute_output_shape() {
+  pooled_shape_ = std::vector<int>(input_shape_.begin(),
+                  input_shape_.begin()+2);
+  for (int i = 2; i < input_shape_.size(); ++i) {
+    int dim = (input_shape_[i] + 2 * pad_shape_[i-2] - kernel_shape_[i
```
chuckcho: Shouldn't this be rounded up (ceil) rather than down (floor) to be consistent with non-Nd-version pooling?

futurely:
Any updates? FYI, https://github.com/BVLC/caffe/pull/3983 supports cuDNN v5.

Issue #2848, opened 2015-08-03
 How to load two pre-trained different models into a new model to realize finetune operation?


takecareofbigboss:
![qq 20150803151457](https://cloud.githubusercontent.com/assets/11883059/9032556/77821bf0-39f2-11e5-99ab-6095a0ec319e.jpg)

Hi, I have two pretrained models, which have been pre-trained on different datasets. Now I want to load the parameters w & b into the new model, the parameters of CNN1 is loaded into the left part of the new model, and the parameters of CNN2 is loaded into the right part of the new model. Also, the input in the new model is different.
What i wanna ask is that how can i realize the mentioned function by python or .sh?

THX...

takecareofbigboss:
![qq 20150803151457](https://cloud.githubusercontent.com/assets/11883059/9032627/0765576e-39f3-11e5-968c-d7adc4b96a98.jpg)

takecareofbigboss:
![qq 20150803151457](https://cloud.githubusercontent.com/assets/11883059/9032645/277c4710-39f3-11e5-86f1-7506ad630c5b.jpg)

takecareofbigboss:
wanna get ur reply...Best Regards...lol

takecareofbigboss:
@Yangqing @rbgirshick

BlGene:
I would look at the python net surgery script for a start. My tip would be to create a third model and load the parameters of both sub networks into the third, then this can be run.

This is a usage question and belongs on the mailing list. Please close.

BR, Max

ChenSiqin:
@BlGene I wonder that whether i can load this two sub networks by using two '--weights=___.caffemodel'?

rorschachhb:
Here's an example of loading two pretrained models into a new model, the author used a pretrained CNN and a pretrained LSTM to build a lrcn model.
https://github.com/jeffdonahue/caffe/blob/recurrent-rebase-cleanup/examples/coco_caption/train_lrcn.from_lm.sh

BlGene:
@ChenSiqin As far as I know when using the weights command line option you are limited to loading only from one network and the load copies parameters to layers that share the same name.

RogerShaw:
@rorschachhb do you mean 
WEIGHTS=\
./models/bvlc_reference_caffenet/bvlc_reference_caffenet.caffemodel\
can have two and more mdels

dongfeng951:
Have you found the solution to the problem? I also face the same problem. Thanks! @takecareofbigboss

[...]

PR #3516, opened 2016-01-05
load multi weights when test

chenjx1005:
we can load multi weights when socre a model by:
caffe test -model model.prototxt -weights weight1 weight2 ...
useful when we want to test a model combained with some submodel like https://github.com/BVLC/caffe/issues/2848


seanbell:
It would be good to update your PR with a description.

PR #3521, opened 2016-01-06
Add CUDA_POST_KERNEL_CHECK to math_functions.cu

xkszltl:
I get CUDA error (Error code 11, invalid augument) in BatchNorm layer when I increase the batch size to >=145, and the error is caught in the next PReLU layer since there is no CUDA_POST_KERNEL_CHECK in this math functions


xkszltl:
The 65536 issue can be fixed by compiling with compute_35,sm_35, which allows 0x7FFFFFFF blocks.

PR #3525, opened 2016-01-06
CMake python version fix

philkr:
The new python3 releases for ubuntu decided to add a plus sign to the end of the version number (e.g. '3.4.3+'). This break the cmake script that finds a matching boost_python version.
This PR fixes this.


longjon:
This string replacement seems pretty hacky, but at least the patch is not (much?) more hacky. Merging, thanks!

PR #3527, opened 2016-01-07
Allow the access of same LMDB for parallel caffe instances.

xdever:
Running multiple instances of caffe (with different networks) on the same LMDB was impossible because the mdb_env_open() locked the database. I removed locking of the database when opening it in read only mode.


lukeyeager:
Try compiling with `ALLOW_MDB_NOLOCK` (#3088).

longjon:
Closing as this patch apparently ignores the following code and surrounding discussion as linked by @lukeyeager; feel free to raise an issue if one remains.

PR #3529, opened 2016-01-07
Feat key value data layer

cseward:
The KeyValueDataLayer allows fetching of information from an LMDB or level_db in arbitrary order. The order is specified by creating a CSV file with the keys for the values of data.

Many things are made possible by the KeyValueDataLayer. For example, if there are 100000 images in a database, and it is desired to train a network on specific pairs of these images, then it is easy put pairs of keys into the CSV file and have two KeyValueDataLayers, with each reading out one image of the pair. In this way, it is possible to create millions of training examples from a database with far fewer entries.

Another useful application is new test / training splits can easily be created by shuffling a CSV file instead of creating a whole new set of databases.

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -306,7 +306,7 @@ message ParamSpec {
 // NOTE
 // Update the next available ID when you add a new LayerParameter field.
 //
-// LayerParameter next available layer-specific ID: 140 (last added: batch_norm_param)
+// LayerParameter next available layer-specific ID: 141 (last added: key_value_data_param)
 message LayerParameter {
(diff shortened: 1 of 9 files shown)


src/caffe/proto/caffe.proto
```diff
@@ -745,6 +746,14 @@ message InnerProductParameter {
   optional int32 axis = 5 [default = 1];
 }
 
+message KeyValueDataParameter {
+  // CSV file with ';' delimiter that specifies which keys to read from the
```
elezar: Could the delimiter not also be part of the protocol definition, with ";" the default?

PR #3536, opened 2016-01-11
Performance related update of im2col() and col2im() functions

intelmm:
Improvement of im2col() and col2im() functions on CPU.

### DIFF
--- a/src/caffe/util/im2col.cpp
+++ b/src/caffe/util/im2col.cpp
@@ -5,6 +5,16 @@
 
 namespace caffe {
 
+// Function uses casting from int to unsigned to compare if value of
+// parameter a is greater or equal to zero and lower than value of
+// parameter b. The b parameter is of type signed and is always positive,
+// therefore its value is always lower than 0x800... where casting
+// negative value of a parameter converts it to value higher than 0x800...
+// The casting allows to use one condition instead of two.
+inline bool is_a_ge_zero_and_a_lt_b(int a, int b) {
+  return static_cast<unsigned>(a) < static_cast<unsigned>(b);
+}
+
 template <typename Dtype>
 void im2col_cpu(const Dtype* data_im, const int channels,
     const int height, const int width, const int kernel_h, const int kernel_w,
@@ -12,22 +22,33 @@ void im2col_cpu(const Dtype* data_im, const int channels,
     const int stride_h, const int stride_w,
     const int dilation_h, const int dilation_w,
     Dtype* data_col) {
-  const int height_col = (height + 2 * pad_h -
-      (dilation_h * (kernel_h - 1) + 1)) / stride_h + 1;
-  const int width_col = (width + 2 * pad_w -
-      (dilation_w * (kernel_w - 1) + 1)) / stride_w + 1;
-  const int channels_col = channels * kernel_h * kernel_w;
-  for (int c_col = 0; c_col < channels_col; ++c_col) {
-    int w_offset = c_col % kernel_w;
-    int h_offset = (c_col / kernel_w) % kernel_h;
-    int c_im = c_col / kernel_h / kernel_w;
-    for (int h_col = 0; h_col < height_col; ++h_col) {
-      for (int w_col = 0; w_col < width_col; ++w_col) {
-        int h_im = h_col * stride_h - pad_h + h_offset * dilation_h;
-        int w_im = w_col * stride_w - pad_w + w_offset * dilation_w;
-        data_col[(c_col * height_col + h_col) * width_col + w_col] =
-            (h_im >= 0 && w_im >= 0 && h_im < height && w_im < width) ?
-            data_im[(c_im * height + h_im) * width + w_im] : 0;
+  const int output_h = (height + 2 * pad_h -
+    (dilation_h * (kernel_h - 1) + 1)) / stride_h + 1;
+  const int output_w = (width + 2 * pad_w -
+    (dilation_w * (kernel_w - 1) + 1)) / stride_w + 1;
+  const int channel_size = height * width;
+  for (int channel = channels; channel--; data_im += channel_size) {
+    for (int kernel_row = 0; kernel_row < kernel_h; kernel_row++) {
+      for (int kernel_col = 0; kernel_col < kernel_w; kernel_col++) {
+        int input_row = -pad_h + kernel_row * dilation_h;
+        for (int output_rows = output_h; output_rows; output_rows--) {
+          if (!is_a_ge_zero_and_a_lt_b(input_row, height)) {
+            for (int output_cols = output_w; output_cols; output_cols--) {
+              *(data_col++) = 0;
+            }
+          } else {
+            int input_col = -pad_w + kernel_col * dilation_w;
+            for (int output_col = output_w; output_col; output_col--) {
+              if (is_a_ge_zero_and_a_lt_b(input_col, width)) {
+                *(data_col++) = data_im[input_row * width + input_col];
+              } else {
+                *(data_col++) = 0;
+              }
+              input_col += stride_w;
+            }
+          }
+          input_row += stride_h;
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


ajtulloch:
What is the performance impact of this on a model like AlexNet or VGG?

intelmm:
Hi, I made only some general improvements of both mentioned functions on master. The changes affected Average Forward-Backward time: using CPU only, on platform with Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz, compiled with MKL, I got x1.189 speedup for AlexNet, x1.077 for GoogleNet and x1.217 for CaffeNet; while compiled with OpenBLAS (single threaded): x1.095 for AlexNet and x1.056 for GoogleNet.

intelmm:
Here are some results from this layerwise benchmark with the new PR: https://github.com/soumith/convnet-benchmarks#layer-wise-benchmarking-last-updated-april-2015

Configuration is as previous: CPU_ONLY set to 1, with MKL and platform Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz (36 threads each on separate core).

Results captured before and after the modification are attached: 
[Results.zip](https://github.com/BVLC/caffe/files/90409/Results.zip)

Below are listed Average Forward-Backward times [ms] before and after CPU optimizations and improvement ratio:

run_imagenet.sh
 alexnet.prototxt:  3215.80,  2737.10,  x1.175
 overfeat.prototxt:  5225.80,  4580.60,  x1.141
 vgg_a.prototxt:  12640.20,  10909.70,  x1.159
 googlenet.prototxt:  14614.90,  13508.50,  x1.082

run_forcegradinput.sh
 conv1.prototxt:  2781.70,  2031.80,  x1.369
 conv2.prototxt:  8927.90,  6924.90,  x1.289
 conv3.prototxt:  3325.40,  2798.00,  x1.188
 conv4.prototxt:  479.60,  380.60,  x1.260
 conv5.prototxt:  397.00,  349.50,  x1.136

run_nogradinput.sh
 conv1.prototxt:  1976.00,  1400.00,  x1.411
 conv2.prototxt:  6115.80,  4354.50,  x1.404
 conv3.prototxt:  2297.70,  1855.10,  x1.239
 conv4.prototxt:  344.60,  258.30,  x1.334
 conv5.prototxt:  282.50,  248.00,  x1.139

src/caffe/util/im2col.cpp
```diff
@@ -6,44 +6,92 @@
...
-      for (int w_col = 0; w_col < width_col; ++w_col) {
-        int h_im = h_col * stride_h - pad_h + h_offset * dilation_h;
-        int w_im = w_col * stride_w - pad_w + w_offset * dilation_w;
-        data_col[(c_col * height_col + h_col) * width_col + w_col] =
-            (h_im >= 0 && w_im >= 0 && h_im < height && w_im < width) ?
-            data_im[(c_im * height + h_im) * width + w_im] : 0;
+void im2col_cpu(const Dtype *data_im, int channels, int input_h, int input_w,
+    int kernel_h, int kernel_w, int pad_h, int pad_w,
+    int stride_h, int stride_w, int dilation_h, int dilation_w,
+    Dtype *data_col) {
```
longjon: Most of these arguments can remain `const`, right?

src/caffe/util/im2col.cpp
```diff
@@ -6,44 +6,92 @@
...
+  const int output_w = (input_w + 2 * pad_w -
+    (dilation_w * (kernel_w - 1) + 1)) / stride_w + 1;
+  const int channel_size = input_h * input_w;
+  for (; channels--; data_im += channel_size) {
+    for (int kernel_row = 0; kernel_row < kernel_h; kernel_row++) {
+      for (int kernel_col = 0; kernel_col < kernel_w; kernel_col++) {
+        int input_row = -pad_h + kernel_row * dilation_h;
+        for (int output_rows = output_h; output_rows; output_rows--) {
+          if (((unsigned) input_row) >= ((unsigned) input_h)) {
+            for (int output_cols = output_w; output_cols; output_cols--)
```
longjon: Generally Caffe style is to always include curly braces; I don't see a reason to deviate here.

src/caffe/util/im2col.cpp
```diff
@@ -6,44 +6,92 @@
...
+    (dilation_h * (kernel_h - 1) + 1)) / stride_h + 1;
+  const int output_w = (input_w + 2 * pad_w -
+    (dilation_w * (kernel_w - 1) + 1)) / stride_w + 1;
+  const int channel_size = input_h * input_w;
+  for (; channels--; data_im += channel_size) {
+    for (int kernel_row = 0; kernel_row < kernel_h; kernel_row++) {
+      for (int kernel_col = 0; kernel_col < kernel_w; kernel_col++) {
+        int input_row = -pad_h + kernel_row * dilation_h;
+        for (int output_rows = output_h; output_rows; output_rows--) {
+          if (((unsigned) input_row) >= ((unsigned) input_h)) {
```
longjon: Why are we casting here? I can't readily imagine a case where this cast both has an effect and is correct. (In any case, we shouldn't be C-style casting.)

src/caffe/util/im2col.cpp
```diff
@@ -6,44 +6,92 @@
...
+    int pad_h, int pad_w, int stride_h, int stride_w,
+    int dilation_h, int dilation_w, double* data_col);
+template void col2im_cpu<float>(const float* data_col, int channels,
+    int input_h, int input_w, int kernel_h, int kernel_w,
+    int pad_h, int pad_w, int stride_h, int stride_w,
+    int dilation_h, int dilation_w, float* data_im);
+template void col2im_cpu<double>(const double* data_col, int channels,
+    int input_h, int input_w, int kernel_h, int kernel_w,
+    int pad_h, int pad_w, int stride_h, int stride_w,
+    int dilation_h, int dilation_w, double* data_im);
```
longjon: (This patch would be easier to read if we didn't reorder the instantiations. It's okay as is, but in general let's factor out irrelevant changes.)

src/caffe/util/im2col.cpp
```diff
@@ -201,5 +206,4 @@ template void col2im_nd_cpu<double>(const double* data_col,
     const int* kernel_shape, const int* pad, const int* stride,
     const int* dilation, double* data_im);
 
-
```
longjon: Please don't include whitespace changes in a commit that changes code somewhere else. (It's okay to make whitespace fixes, but those should go in their own commits unless they're adjacent to modified code.)

longjon:
Thanks @intelmm, this looks like a nice boost. I've made a few comments as noted. Two points I'd like to clear up before merge:

1. Let's be confident there are no serious regressions from this change. The provided benchmarks seem like pretty good evidence of this, but others may want to comment.
2. Can you explain the change? Why is this code faster?

intelmm:
Thank you @longjon for review and comments! I'll commit updated version very soon.

Regarding your questions, here are some hints in general:
- We have made some profiling using callgrind and it showed that the im2col function takes some serious part of total execution time.
- Therefore even small performance changes in this function slightly affects the total execution time.
- In our cases, function improvement changed the performance factor to x1.5 - x5 depending on conv layer type.
- On the other hand, the col2im function does not consume as much resources, however its construction is almost the same as im2col, therefore can be easily replaced.
- Moreover the replacement of col2im is recommended to have both pieces of code looking similarly for easier maintenance.

Why is this code faster?
- In fact, the code has changed, but functionally it still works the same.
- First of all, output elements are stored one by one, so instead of calculating index of the output element in array, it has been replaced by *(data_col++) = x;
- Current condition used for padding checking is quite costly, therefore I made two changes:
  - For vertical coordinates the condition is moved into higher loop, and fills whole lines with zeros
  - The tricky part is to change condition from "if(0 <= a && a < N)" into "if((unsigned) a < (unsigned) N)"
    - This reduces two conditions into single by changing negative values to 0xFFFFF... like,
    - which is certain to be always higher than N (N is of type int i.e. is lower than 0x80000...).
- All divide and modulo operations has been replaced by consecutive "for" loops (i.e. replaced by increment operations).

What could we do next?
- The col2im function however could be still improved, but it should be redesigned from scratch.
- Current implementation clears the output matrix by assigning zeros to each output cell, and then accumulate the value with multiple read-write operations in fashion which completely misuse CPU cache.
- It should accumulate the value for each single output cell in one variable and then store it.

longjon:
Thanks for the explanation, that makes this clearer (and might be a useful reference for future optimization).

The `unsigned` casting is pretty tricky (although it does seem well-defined)... could we wrap it in an inline function that describes its purpose? (And use `static_cast` per note.)

An improved `col2im` implementation would be welcome in a future PR.

intelmm:
Dear @longjon, all the changes have been committed according to your comments.

src/caffe/util/im2col.cpp
```diff
@@ -5,6 +5,16 @@
...
 namespace caffe {
 
+// Function uses casting from int to unsigned to compare if value of
+// parameter a is greater or equal to zero and lower than value of
+// parameter b. The b parameter is of type signed and is always positive,
+// therefore its value is always lower than 0x800... where casting
+// negative value of a parameter converts it to value higher than 0x800...
+// The casting allows to use one condition instead of two.
+inline bool is_a_ge_zero_and_a_lt_b(int a, int b) {
+    return static_cast<unsigned>(a) < static_cast<unsigned>(b);
```
longjon: This should be a two-space indent per Google style guide (not sure why lint doesn't catch it).

longjon:
The latest version seems understandable and the diff is much cleaner, thanks!

I've commented on one space error, and then please squash into one commit before merge.

intelmm:
Dear @longjon thank you for review. The commit is ready.

longjon:
Clean diff, clean history, clear code... merging! Thanks, and looking forward to future CPU optimizations!

PR #3545, opened 2016-01-13
use relative paths on making build/tools/ links

gdh1995:
The old uses `abspath`, which I think is so harmful:
* If I `cp -a` the whole project, `build/tools/caffe` still refer to
    the old file, until `make clean`, making debugging very hard
* For `tar` and `scp`, the soft links can not work
    unless the target project folder has the same path

The solution is to use `notdir`. I've re-compiled the master branch with it
and `build/tools/caffe` works well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -601,7 +601,7 @@ $(TEST_CXX_BINS): $(TEST_BIN_DIR)/%.testbin: $(TEST_CXX_BUILD_DIR)/%.o \
 # Target for extension-less symlinks to tool binaries with extension '*.bin'.
 $(TOOL_BUILD_DIR)/%: $(TOOL_BUILD_DIR)/%.bin | $(TOOL_BUILD_DIR)
 	@ $(RM) $@
-	@ ln -s $(abspath $<) $@
+	@ ln -s $(notdir $<) $@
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -589,8 +589,8 @@ $(TEST_CXX_BINS): $(TEST_BIN_DIR)/%.testbin: $(TEST_CXX_BUILD_DIR)/%.o \
 
 # Target for extension-less symlinks to tool binaries with extension '*.bin'.
 $(TOOL_BUILD_DIR)/%: $(TOOL_BUILD_DIR)/%.bin | $(TOOL_BUILD_DIR)
-	@ $(RM) $@
-	@ ln -s $(abspath $<) $@
+	@ $(RM) $@
```
jeffdonahue: remove added EOL whitespace

jeffdonahue:
Thanks @gdh1995! This makes sense.  Please amend your commit as I commented above and I'll merge this.

gdh1995:
Okay, I've rebased it and remove the space char.

jeffdonahue:
Thanks again @gdh1995!

lukeyeager:
:+1:

PR #3548, opened 2016-01-13
Python implementation of the orthonormal and LSUV initializations

ducha-aiki:
LSUV initialization is described in ICLR2016 submission http://arxiv.org/abs/1511.06422
First, it fills weight by orthonormal matrices (as suggested by Saxe), 2nd - scale weights using data, until activations vars become 1.0. 
It gives state-of-art or better results for all tested activations, i.e. improving BVLC-googlenet result by 0.8% top-1 accuracy.

Orthonormal init is described in
Andrew M. Saxe, James L. McClelland, Surya Ganguli (2013). Exact solutions to the nonlinear dynamics of learning in deep linear neural networks.  http://arxiv.org/abs/1312.6120 and is used by Kaggle winners in NDSB (Plankton) and Retinopathy challenges.

### DIFF
--- a/tools/extra/LSUV.py
+++ b/tools/extra/LSUV.py
@@ -1,0 +1,92 @@
+import os
+import sys
+from pylab import *
+import random
+import numpy as np
+
+caffe_root_dir=os.path.dirname(os.path.realpath(__file__))
+caffe_root_dir+='/../../python'
+sys.path.insert(0, caffe_root_dir)
+if len (sys.argv) < 4:
+    raise RuntimeError('Usage: python ' + sys.argv[0] + ' path_to_solver path_to_save_model mode')
+import caffe
+
+solver_path = str(sys.argv[1])
+init_path = str(sys.argv[2])
+init_mode =  str(sys.argv[3])
+margin = 0.02;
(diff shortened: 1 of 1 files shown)


ducha-aiki:
Ready to review.

seanbell:
Generally, caffe promises to support both Python 2 and 3 (https://github.com/BVLC/caffe/blob/master/docs/installation.md), so it would be good to change your `print` statements to be `python3`-compatible.

tools/extra/LSUV.py
```diff
@@ -0,0 +1,92 @@
+import os
```
seanbell: This is an executable script, but missing the hashbang at the top.

ducha-aiki:
@seanbell added py3 print and hashbang. 
But I got used to run python scripts via python script.py, never used them as executables :)

ducha-aiki:
@shelhamer, @ronghanghu PR is ready for review long time ago. 
If you want, you could merge instead https://github.com/philkr/magic_init from BVLC, it does almost the same. Data-driven initialization works good and it would be good to some implementation in caffe.

PR #3550, opened 2016-01-13
Add BiasLayer to add two Blobs with broadcasting

jeffdonahue:
This adds BiasLayer, designed analogously to ScalarLayer (#3021), to add blobs with arbitrary axes broadcasted.  This could be used together with ScalarLayer to learn the batch norm scale and shift parameters.  It could also be used independently anywhere in a network to learn a bias without a corresponding multiplication.  And even more generally, it could be used to efficiently add two blobs with any number of corresponding axes, which can currently only be accomplished (in the most general case) rather inefficiently: with a pair of `Reshape`s and `Tile`s (to broadcast leading and trailing axes) followed by the `Eltwise` `SUM` operation.

This is currently based on ScalarLayer (for caffe.proto ID sequencing), with the last two commits being the relevant ones -- I'm happy to rebase this without ScalarLayer if we want to merge this before or without that.

Both this and ScalarLayer can either take two bottoms, specifying both inputs to the function, or take a single bottom and learn the second as a parameter.

A different approach for learning the BN scale/shift parameters that I haven't looked at yet is in #2996 (by @ducha-aiki), which learns both sets of parameters together.  @cdoersch and I and anyone else interested (possibly @longjon and @shelhamer) should take a look at both and evaluate the benefits, with merge priority for any shared functionality given to @ducha-aiki's #2996 as the earlier PR. 

Personally I do like the approach of having layers do as little as possible, which is why for my own work I've taken the approach of using two independent layers.

### DIFF
--- a/include/caffe/layers/bias_layer.hpp
+++ b/include/caffe/layers/bias_layer.hpp
@@ -1,0 +1,54 @@
+#ifndef CAFFE_INNER_PRODUCT_LAYER_HPP_
+#define CAFFE_INNER_PRODUCT_LAYER_HPP_
+
+#include <vector>
+
+#include "caffe/blob.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+namespace caffe {
+
+/**
+ * @brief Computes a sum of two input Blobs, with the shape of the
+ *        latter Blob "broadcast" to match the shape of the former.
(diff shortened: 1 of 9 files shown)


cdoersch:
@longjon @shelhamer @jeffdonahue the discussion of the various options for channelwise affine operations has been happening in #3229...may be good to have all the discussion in one place.

jeffdonahue:
Replaced by #3591

Issue #3345, opened 2015-11-17
Boost python 3 library not found with cmake


pixelou:
cmake/Dependencies.cmake assumes boost's python3 libraries have a name libboost-py{concatenatedVerNum} such as:

> libboost_python-py35.so.1.58.0

Which is the convention used by ubuntu for example ([see here](http://packages.ubuntu.com/xenial/amd64/libboost-python1.58.0/filelist)). However, other distributions ([archlinux](https://www.archlinux.org/packages/extra/x86_64/boost/), [fedora](https://apps.fedoraproject.org/packages/boost-python3/)) name the library like so:

> libboost_python3.so.1.59.0

The patch below should fix the compilation process, but I haven't checked wether it causes troubles or not in practice since I am a complete newbie with caffe.

```patch
diff --git a/cmake/Dependencies.cmake b/cmake/Dependencies.cmake
index 51a803c..9501156 100644
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -118,6 +118,11 @@ if(BUILD_python)
     find_package(Boost 1.46 COMPONENTS "python-py${boost_py_version}")
     set(Boost_PYTHON_FOUND ${Boost_PYTHON-PY${boost_py_version}_FOUND})
     
+    if(NOT Boost_PYTHON_FOUND)
+      find_package(Boost 1.46 COMPONENTS "python${boost_py_version}")
+      set(Boost_PYTHON_FOUND ${Boost_PYTHON${boost_py_version}_FOUND})
+    endif()
+
     while(NOT "${version}" STREQUAL "" AND NOT Boost_PYTHON_FOUND)
       STRING( REGEX REPLACE "([0-9.]+).[0-9]+" "\\1" version ${version} )
       
@@ -125,6 +130,11 @@ if(BUILD_python)
       find_package(Boost 1.46 COMPONENTS "python-py${boost_py_version}")
       set(Boost_PYTHON_FOUND ${Boost_PYTHON-PY${boost_py_version}_FOUND})
       
+      if(NOT Boost_PYTHON_FOUND)
+        find_package(Boost 1.46 COMPONENTS "python${boost_py_version}")
+        set(Boost_PYTHON_FOUND ${Boost_PYTHON${boost_py_version}_FOUND})
+      endif()
+
       STRING( REGEX MATCHALL "([0-9.]+).[0-9]+" has_more_version ${version} )
       if("${has_more_version}" STREQUAL "")
         break()
```

[...]

PR #3553, opened 2016-01-14
Fix boost_python discovery for distros with different naming scheme

rrika:
The current method probes

    boost_python_py351
    boost_python_py35
    boost_python_py3
    boost_python

On Arch Linux however the name of the library is

    boost_python3

So I added that to the cmakefile.


pixelou:
As far as I can tell, this is the same bug as #3345.

rrika:
@pixelou you're right, your patch additionally checks for

    boost_python351
    boost_python35

though I don't know any distribution that names its libraries like that

rarensu:
this merge would be pretty great since it's still an issue for some folks like me.

PR #3555, opened 2016-01-15
Update io.py

n3011:
Fixes unnecessary error reporting. Mean shape incompatible with input shape

### DIFF
--- a/python/caffe/io.py
+++ b/python/caffe/io.py
@@ -256,10 +256,10 @@ class Transformer:
                 raise ValueError('Mean shape invalid')
             if ms != self.inputs[in_][1:]:
                 print(self.inputs[in_])
-    		 in_shape = self.inputs[in_][1:]
-    		 m_min, m_max = mean.min(), mean.max()
-    		 normal_mean = (mean - m_min) / (m_max - m_min)
-    		 mean = resize_image(normal_mean.transpose((1,2,0)),
+    		    in_shape = self.inputs[in_][1:]
+    		    m_min, m_max = mean.min(), mean.max()
+    		    normal_mean = (mean - m_min) / (m_max - m_min)
+    		    mean = resize_image(normal_mean.transpose((1,2,0)),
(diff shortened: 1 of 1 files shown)


seanbell:
(1) Why is this unnecessary error reporting?  The mean image is spatially dependent, so it doesn't make sense to always blindly resize it. I think a safer option to get your feature is to add a parameter (default `False`) that allows for resizing if the shapes don't match.

(2) If your mean image a constant value (which is a reasonable input) then this will introduce a NaN since `m_min = m_max`.

python/caffe/io.py
```diff
@@ -255,7 +255,14 @@ def set_mean(self, in_, mean):
...
             if ms != self.inputs[in_][1:]:
-                raise ValueError('Mean shape incompatible with input shape.')
+                print(self.inputs[in_])
+    		    in_shape = self.inputs[in_][1:]
+    		    m_min, m_max = mean.min(), mean.max()
+    		    normal_mean = (mean - m_min) / (m_max - m_min)
+    		    mean = resize_image(normal_mean.transpose((1,2,0)),
+                        in_shape[1:]).transpose((2,0,1)) * \
+                        (m_max - m_min) + m_min
+                #raise ValueError('Mean shape incompatible with input shape.')
```
seanbell: The added lines 258-265 are indented by 2 spaces too many.

python/caffe/io.py
```diff
@@ -255,7 +255,14 @@ def set_mean(self, in_, mean):
             if len(ms) != 3:
                 raise ValueError('Mean shape invalid')
             if ms != self.inputs[in_][1:]:
-                raise ValueError('Mean shape incompatible with input shape.')
+                print(self.inputs[in_])
```
seanbell: This message isn't very good debugging information.  A better message would explain that the sizes were incompatible and the mean shape was auto-resized for you, listing the sizes.

Noiredd:
Fixed in #5719

PR #3565, opened 2016-01-18
Simplified CropLayer

BlGene:
This is for cropping one blob to another using coordinates from parameters.

continues from PR #1976.

### DIFF
--- a/include/caffe/layers/crop_layer.hpp
+++ b/include/caffe/layers/crop_layer.hpp
@@ -1,0 +1,49 @@
+#ifndef CAFFE_CROP_LAYER_HPP_
+#define CAFFE_CROP_LAYER_HPP_
+
+#include <utility>
+#include <vector>
+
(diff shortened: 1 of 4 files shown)


longjon:
Thanks for posting this, this is basically what I had in mind in #1976. However, the more substantial task (which I haven't attempted yet) is to build the ND version. I think it would need an interface similar to the current convolution one, i.e., an `axis` param and a repeated `crop` param. The former would indicate that dimensions following the indicated one were spatial dimensions, while the latter would either be specified once, indicating that all spatial dimensions should be cropped the same amount, or specified a number of times equal to the number of spatial dimensions. Feel free to take a stab at that if you like!

I do already have a private version of the Python code mentioned in #1976. I'll try to open a PR for that shortly, whether completely finished or not.

BlGene:
Closing in favor of #3570.

PR #3568, opened 2016-01-19
Add a Restricted Boltzmann Machine layer

elezar:
With this pull request, a Restricted Boltzmann Machine layer is made available. The aim is to use this as a basis for more complex unsupervised learning tasks including Deep Belief networks.

In its current state, an example is provided that allows for an example similar to that presented at: http://deeplearning.net/tutorial/rbm.html and as presented in [1](although the example does not consider the labels or classification error).

We (@cseward and myself) are still cleaning up the code, but would appreciate feedback with regards to naming, etc.

An example of the filters learned by the RBM are shown below:
rbm_filters

[1] H. Larochelle, M. Mandel, R. Pascanu, and Y. Bengio, “Learning algorithms for the classification restricted boltzmann machine,” J. Mach. Learn. Res., vol. 13, no. 1, pp. 643–669, 2012.


cbalint13:
@elezar ,

I am also working on RBM for caffe. I am looking at your implementation and have few Q:

* Can't see where we could specify number of CD steps. Perhaps we could do PCD too ?
* I believe naming simply "RBM" instead of "RBMInnerProduct" would be more fine.
* RBM use two bias layers one for hidden one for visible, its not clear if you do.
* Can document a bit the steps, e.g where do you do Gibbs ?!

elezar:
@cbalint13 Thanks for the questions. Maybe @cseward  would be better suited to answer your questions (I will also have a look as soon as I have a free moment).

We have been working on making things a little clearer, but have not had a chance to push the changes yet.

cbalint13:
@elezar , @cseward,

  I've looked more deeper into RBM implementation. I am impressed by its very good design ! Ignoring few minor typos (e.g. s/visable/visible/) It is very readable and clean, many of my initial Q are answered !

 May I propose to you patches to this in private ? (to not grow this PR text in length). I am thinking:

* Few addons like: GRBM (gaussian visible), adjustable amount of CD steps (CD-x), P-CD ?

* Also, activation function (a.t.m. logistic) could be specified by user (e.g ReLU in case of GRBM).

* DBN, multimodal & joint layers with adjustable sampling phases (e.g. in case of PCD).

---------

We can compare results with Nitish Shrivastava's excelent (and the only true one) RBM/DBN suite: https://github.com/nitishsrivastava/deepnet

elezar:
@cbalint13  Thank for the info. We will have a look at the references (and address some of the points that you mention). You are also welcome to contact us offline if you don't want to add more comments to the pull request.

elezar:
@cbalint13 We have finally had a chance to rework our RBM implementation somewhat. This includes making it more flexible in terms of specifying the activation and sampling functions. We would appreciate it if you could have a look and let us know what you think.

cbalint13:
@elezar ,

  * Looks *really* nice ! The layout is now very flexible and can be extended with ease to Gaussian-RBM instead of (Bernoulli) & P-CD (instead of standard CD), and also the activation could be ReLU (e.g. for G-RBM). Some sparsity too can be added, in time. Even beyond of these "know to work" one can experiment various setup. Thus, i believe all RBM related functionalities can be covered with ease with actual flexibility.

  * I looked at actual implementation, unfortunate is that operation with batch_size=100 does not converge, i am looking at it right now. I try help out, also if all fine a CIFAR10 & (P)CD example can be added with some nice python notebooks for booth actual MNIST & potential CIFAR10 ( expecting to see such very nice filters on cifar10: http://openrisc.rdsor.ro/grbm-color-patches.png 16x16px cifar10 sub-samples).

Will propose some +patches, but first need to figure out the batch_size issue while i try understand the full proposed work flow within this patch (may take some time if i reach to lowest printf() way of doing it, i hope not to reach that).

elezar:
@cbalint13 thanks for the quick response. We did test with a larger batch size, but it could be that we missed something and I will re-run some tests. If you could provide an example for CIFAR, that would be great. We will work at setting up a notebook for training the MNIST example from our side.

You are welcome to create a pull request against https://github.com/zalando/caffe/tree/feat_rbm_layer with your changes (if you discover the source of the bug).

elezar:
@cbalint13 I have just checked the MNIST example with a batch size of 100, and it still seems to work, and the filters that are generated look similar to the single image case.

cbalint13:
@elezar ,

- I believe once batch issue is solved and also FREE_ENERGY metric works this PR is done.
- Regarding CIFAR10 you can account on me, i wish to do a nice GRBM sample notebook .
- Will PR against your mentioned branch, I am also looking at our batch=100 issue.


My time is quite limited during normal week, i am expecting to work in weekends.

elezar:
@cbalint13 that sounds great! I have just tested the larger batch sizes here, and the results do not look quite right. It seems as if the filters for a number of digits are overlaid. We will investigate further on this side. 

An example and an iPython notebook for the CIFAR10 dataset would be a really valuable addition. We look forward to your PR.

cseward:
@cbalint13 we took a look at the issue and there was indeed a small indexing error. We're now getting better results, when we run the RBM as in the example we get filters that look like this:

rbm_batch_1_iter_65000

If we increase the batch size to 10, run for 7000 iterations (and leave everything else unchanged) the results look similar

rbm_batch_10_iter_7000

The filters for larger batch sizes start to look different (at least for < 10^5 iterations) since larger batch sizes means each image was seen many times and more was learned. With a batch size of 100, learning rate of 0.01 and 40000 iterations more point like filters are created:

rbm_batch_100_iter_40000

The reconstruction error is also a lot lower with the point filters, so I think the point filters are a good thing. Do you now get better results with your example? Thanks.

cbalint13:
@elezar ,

 - I can confirm that I am getting same [filters](http://openrisc.rdsor.ro/rbm_minst_0.01_40000.png).
 - I used a very simple cudamat rbm [implementation](https://github.com/cudamat/cudamat/blob/master/examples/rbm_cudamat.py) for the test, visualised by a small [script](http://openrisc.rdsor.ro/vis-cudamat.py) .

The convergence of euclidian loss is strange a bit (sometimes grows back a little), in contrast with the evolution in the cudamat sample. Also the value is high >20 unit instead of 0.0xxx fractional things, will try check the source of issue.

encore0117:
Hi,

Thank you so much for creating this layer!
I'm trying to use this layer to build up a deep belief network consisting several rbm_inner_product layers. I'm wondering if the ordering of top blobs in each such layer matters as I assume there are two blobs, states of hidden units and reconstruction, in each rbm_inner_product layer. 

Also, I've found out in the "reshape" function, it assumes loss_measure blobs are proceeding other blobs.

Thanks~

encore0117:
Hi @elezar ,

Thank you so much for creating this layer!

I'm trying to build up a deep belief network by stacking rbm_inner_product layers. However, for each rbm layer, I have two top blobs, the state of hidden units and reconstruction, and when I ran the program, it complains about the sizes of bottom blobs not being equal in 'reconstruction error' layer.

I read the comment in the header file about the ordering of top blobs in rbm layer. Is it possible to have only the last rbm layer having two top blobs? 

Thanks~

encore0117:
Hi @elezar ,

Thank you so much for writing up this layer!

I am trying to build up a deep belief network by stacking several rbm layers.

Is it possible to achieve this goal? Or should I just write another layer from ground up?

Please advise.

Thanks~

encore0117:
Hi!

Thank you so much for writing up this layer!

I'm trying to build up a deep belief network by stacking rbm layers. However, I've found out that the way it calculates the reconstruction error does not allow change of dimensions. 

Is there a way to solve this issue? 

Please advice.

Thanks~

elezar:
I think @cseward was working on DBNs, so it should be possible to use these layers as a building block. I don't know if the PR in its current state is able to deal with this use case, though.

encore0117:
Hi @elezar ,

Thanks for your prompt response!

Will look into this.

robosmith:
I am confused about the current state of this commit. The only RBM files I could find in Caffe/master appear to contain empty method definitions, and even they do not seem to be in the current master repo. I am working on a DBN implementation using a container layer of multiple RBM layers, but am finding some inconsistencies between the current implementation here and other descriptions. For instance, I've seen a reference which says the RBM hidden layer is binary, but this implementation appears to be float. Can someone familiar with these issues clarify what's going on with this effort?

robosmith:
@cseward, @elezar
I am confused about the current state of this commit. The only RBM files I could find in Caffe/master appear to contain empty method definitions, and even they do not seem to be in the current master repo. I am working on a DBN implementation using a container layer of multiple RBM layers, but am finding some inconsistencies between the current implementation here and other descriptions. For instance, I've seen a reference which says the RBM hidden layer is binary, but this implementation appears to be float. Can someone familiar with these issues clarify what's going on with this effort?

robosmith:
@cseward, @elezar
I've been testing my Deep Belief Network implementation based on your RBM layer, and there seems to be a problem with the diffs being reset by the net.cpp at the start of every iteration. Since the bottom mutable_cpu_diff is being used in the Forward_cpu function, resetting the diffs at the start of each iter_, changes the Forward_cpu results dramatically. Is this really correct behavior? It seems to me that only the Backward_cpu function should use diffs that get reset, but I am not that experienced with how this is supposed to work.

elezar:
@robosmith Thanks for the message. Unfortunately I haven't really touched this in a while and am unable to help you further. 

I don't known if @cseward can field this question as he is a little more involved in the space at the moment.

If I don't hear back from him, I'm going to close this PR so as to prevent further confusion. You are, of course, welcome to use this as a basis for further work of your own and submit your own PR to the upstream repo.

Sorry for the inconvenience.

robosmith:
@elezar Thanks for your response.

I have since figured out that once an RBM layer in a DBN is trained, the forward_is_update_ variable must be set to false in order to not be affected by subsequent solver iterations which reset the diffs.

robosmith:
@cseward  @elezar

I am training my DBN (container layer) with 4 of your RBM layers in sequence, first training layer 1, then 2, etc. I am seeing behavior which I don't understand and wonder if either of you can comment to clarify what is happening. 

The first layer has ~100K inputs and 1K outputs. During training when both sample_h_given_v() and gibbs_hvh() are being called in sequence, the output of layer 1 appears to have reasonable values, mostly between 0 and 1 and all less than 10. However, when I disable training on layer 1 (set forward_is_update_ = false & disable Backward_cpu()), so that sample_h_given_v(), but NOT gibbs_hvh(), is called, the  layer 1 output values are much larger, ranging between ~200K & 10^11. 

This seems very odd to me and seems to lead to saturated outputs in later layers, but I don't know enough about the RBM layer to know if this is reasonable and expected. Thank you for any explanation or clarification you may provide.

cseward:
@robosmith 

So this is a pretty old project that we're not using any more (tensorflow makes a lot of this stuff much easier). But to answer your question when a forward pass is called with forward_is_update_ = false then all that happens is sample_h_given_v(). I'm not sure what's going wrong, but there's a few possible places where errors could creep in, like the weights of the filter matrix not being loaded properly or the size of the top member variable being incorrectly set (since lots of parts of the update process are saved in top) or the sampling running too long / not long enough.

Hope these few pointers could help you move in the right direction

elezar:
I'm closing this PR.

PR #3569, opened 2016-01-19
solver trace feature to trace weights, loss, diffs and much more

cseward:
This pull request allows a running history to be created of any solver.  Only a very small user selected subset of values are saved, but these are enough to visualize the progress of training.

With this pull request, plots tracing the progress of learning can be created that look like this:

test_trace

weight_traces

The changes made were the following:
 * messages for saving solver history are added to caffe.proto
 * options for how solver history is saved (how often, how much) is added to caffe.proto
 * a SolverTrace class is created to manage solver history
 * Solver class is modified to save its pass its state to SolverTrace at appropriate times
 * caffe executable is updated to allow resumption of training with a solver trace file
 * mnist example that creates a solver trace added
 * python script to visualize a solver trace file added

### DIFF
--- a/examples/mnist/lenet_solver_with_trace.prototxt
+++ b/examples/mnist/lenet_solver_with_trace.prototxt
@@ -1,0 +1,32 @@
+# The train/test net protocol buffer definition
+net: "examples/mnist/lenet_train_test.prototxt"
+# test_iter specifies how many forward passes the test should carry out.
+# In the case of MNIST, we have test batch size 100 and 100 test iterations,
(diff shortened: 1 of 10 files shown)


elezar:
This PR supersedes #2327. @cseward, could you close the previous one.

elezar:
Looking at #2836, it may be useful to switch to the new (hdf5) format. @erictzeng could you (or someone else) comment on whether the changes that @cseward  proposes here are better suited to the new format?

PR #3570, opened 2016-01-19
ND Crop layer

BlGene:
Something along the lines of what a ND crop layer would look like.
Based on: PR #1976 -> PR #3565 -> this

### Todo
- [x] Test for D > 4  
- [x] Test backwards
- [x] GPU kernel copy instead of memcopy (?)
- [x] const indices and offsets
- [x] GPU backwards/forwards
- [x] CPU backwards
- [x] parameter size checks
- [x] lint

Please see PR #3613 for how to set the crop parameters.

### DIFF
--- a/include/caffe/layers/crop_layer.hpp
+++ b/include/caffe/layers/crop_layer.hpp
@@ -1,0 +1,49 @@
+#ifndef CAFFE_CROP_LAYER_HPP_
+#define CAFFE_CROP_LAYER_HPP_
+
+#include <utility>
+#include <vector>
+
+#include "caffe/blob.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+namespace caffe {
+
+/**
+ * @brief Takes a Blob and crop it along either the width or height dimension,
+ *        outputting a cropped Blob.
+ *
+ * TODO(dox): thorough documentation for Forward, Backward, and proto params.
+ */
+
+template <typename Dtype>
+class CropLayer : public Layer<Dtype> {
+ public:
+  explicit CropLayer(const LayerParameter& param)
+      : Layer<Dtype>(param) {}
+  virtual void LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+  virtual void Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+
+  virtual inline const char* type() const { return "Crop"; }
+  virtual inline int ExactNumBottomBlobs() const { return 2; }
+  virtual inline int ExactNumTopBlobs() const { return 1; }
+
+ protected:
+  virtual void Forward_cpu(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+  virtual void Backward_cpu(const vector<Blob<Dtype>*>& top,
+      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
+  virtual void Forward_gpu(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+  virtual void Backward_gpu(const vector<Blob<Dtype>*>& top,
+      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
+
+  int crop_h_, crop_w_;
+};
+
+}  // namespace caffe
+
+#endif  // CAFFE_CROP_LAYER_HPP_
--- a/include/caffe/layers/crop_layer.hpp
+++ b/include/caffe/layers/crop_layer.hpp
@@ -11,8 +11,8 @@
 namespace caffe {
 
 /**
- * @brief Takes a Blob and crop it along either the width or height dimension,
- *        outputting a cropped Blob.
+ * @brief Takes a Blob and crop it, to the shape specified by the second input
+ *  Blob, across all dimensions after the specified axis.
  *
  * TODO(dox): thorough documentation for Forward, Backward, and proto params.
  */
--- a/include/caffe/layers/crop_layer.hpp
+++ b/include/caffe/layers/crop_layer.hpp
@@ -41,9 +41,27 @@ class CropLayer : public Layer<Dtype> {
   virtual void Backward_gpu(const vector<Blob<Dtype>*>& top,
       const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
 
-  int crop_h_, crop_w_;
-};
+  vector<int> offsets;
 
+ private:
+  void crop_copy(const vector<Blob<Dtype>*>& bottom,
+               const vector<Blob<Dtype>*>& top,
+               const vector<int>& offsets,
+               vector<int> indices,
+               int cur_dim,
+               const Dtype* src_data,
+               Dtype* dest_data,
+               bool is_forward);
+
+  void crop_copy_gpu(const vector<Blob<Dtype>*>& bottom,
+                const vector<Blob<Dtype>*>& top,
+                const vector<int>& offsets,
+                vector<int> indices,
+                int cur_dim,
+                const Dtype* src_data,
+                Dtype* dest_data,
+                bool is_forward);
+};
 }  // namespace caffe
 
 #endif  // CAFFE_CROP_LAYER_HPP_
--- a/src/caffe/layers/crop_layer.cpp
+++ b/src/caffe/layers/crop_layer.cpp
@@ -1,0 +1,78 @@
+#include <algorithm>
+#include <map>
+#include <set>
+#include <vector>
+
+#include "caffe/layer.hpp"
+#include "caffe/layers/crop_layer.hpp"
+#include "caffe/net.hpp"
+
+
+namespace caffe {
+
+template <typename Dtype>
+void CropLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
+  const CropParameter& param = this->layer_param_.crop_param();
+  CHECK_EQ(bottom.size(), 2) << "Wrong number of bottom blobs.";
+  CHECK_EQ(bottom[0]->num_axes(), 4) << "Only works with 4D blobs.";
+  CHECK_EQ(bottom[1]->num_axes(), 4) << "Only works with 4D blobs.";
+  crop_h_ = param.offset_height();
+  crop_w_ = param.offset_width();
+}
+
+template <typename Dtype>
+void CropLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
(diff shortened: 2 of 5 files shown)


BlGene:
@longjon So, my attempt at a ND crop layer, if this is approximately what you are looking for I try to fix the GPU part.

longjon:
Actually I'm confused about the parameter interface here... what I had in mind is something mirroring the conv layer interface, with a _scalar_ (i.e., `optional`) param `axis` alongside a repeated param of crop values (as you have currently).

The way it would work would be the same way conv layer works: _if_ only one crop value is specified, all the dimensions after the specified `axis` would be cropped (by the same amount); otherwise the number of crop values specified must be equal to the number of dimensions following `axis`, and the trailing dimensions would be cropped accordingly.

The reason for doing it this way is because the crop values need to be determined from the net specification, and the net specification doesn't know how many dimensions the data will have! If the layers we are "cropping through" include conv/pool layers with non-singleton kernel sizes, then we know exactly how many dimensions to crop; but if not, all we can do is propagate the `axis` param and determine how many dimensions to crop at run time. (At least, I think that should work.)

It would be nice to have an ND GPU kernel, but it's fine to punt that to get the basic functionality in.

Thanks!

BlGene:
@longjon The parameters should do what you want now. 

I substituted the copy kernel with repeated calls to caffe_copy, which if I understood correctly will call cudaMemcpy, and thus perform the copy operation on the GPU. My intuition is that calling this thousands of times is probably less efficient than writing a custom kernel.

longjon:
The parameters look right now.

`caffe_copy` will perform the copy on the GPU, but (as you suggest) has overhead -- even in the 2D case I found the overhead unacceptable, which is why the kernel exists. A fine solution for now would be to keep the existing kernel around, and use it only for the innermost two dimensions, looping around the outer ones. That should keep the performance the same in the 2D case while supporting ND.

BlGene:
@longjon The `crop_copy_gpu` function now calls the kernel for the innermost two dimensions. This should hopefully fix the speed problem.

bjosv79:
It seems to a few problems with the backward pass of `crop_layer.cu`:
- A typo `src_outer_stride` instead of `src_inner_stride` in line 94.
- Incorrect number of elements in `caffe_gpu_set` in line 89.
- Is the `caffe_gpu_set` really necessary? It is already done in line 119?

BlGene:
@bjosv79 Thanks. All those things seemed correct, I changed them ( but haven't tested yet ).

YOUNGING:
@BlGene How to determine the offset parameter?Could you please show an example?

BlGene:
@YOUNGING Please see PR #3613.

YOUNGING:
I have merged this PR with the latest caffe branch(I download it today),and I tried to train the fcn32s model.When using CPU mode ,it works;but when I use GPU  mode, caffe just break.The log file is like this

I0222 19:41:53.105304 21292 net.cpp:283] Network initialization done.
I0222 19:41:53.105304 21292 solver.cpp:60] Solver scaffolding done.
I0222 19:41:53.105304 21292 caffe.cpp:219] Starting Optimization
I0222 19:41:53.105304 21292 solver.cpp:280] Solving FCN
I0222 19:41:53.105304 21292 solver.cpp:281] Learning Rate Policy: step  

PR #1976 works in both CPU and GPU mode.
Do you have any idea about this problem?

YOUNGING:
@BlGene This is the prototxt file I used. 
[fcn32s_net.txt](https://github.com/BVLC/caffe/files/140525/fcn32s_net.txt)

BlGene:
@YOUNGING. Its not possible to deduce the source of the error from the log that you posted. Have you tried using PR #3565? I'm a bit busy with other things, I'll try to have a look at it later this week, but feel free to try to debug it yourself if you want to.

YOUNGING:
Thanks for the quick response.PR#3565 has the same problem.I will try to debug it.

YOUNGING:
@BlGene I put an output at the entrance of forward_gpu function,like this:
std::cout << "forward_gpu entry" << std::endl;
but it cannot print anything.

BlGene:
Try a ' LOG(FATAL) << "This code is being run." '
In case the program does not quit at this line the code is not being called. In case it is, you can use LOG(INFO) << "forward_gpu entry."

src/caffe/proto/caffe.proto
```diff
@@ -306,7 +306,7 @@ message ParamSpec {
 // NOTE
 // Update the next available ID when you add a new LayerParameter field.
 //
-// LayerParameter next available layer-specific ID: 143 (last added: scale_param)
+// LayerParameter next available layer-specific ID: 144 (last added: crop_param)
```
shelhamer: Needs ID++ after #3211 -- sorry for the rebase.

include/caffe/layers/crop_layer.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <vector>
+
+#include "caffe/blob.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+namespace caffe {
+
+/**
+ * @brief Takes a Blob and crop it along either the width or height dimension,
```
shelhamer: This doc comment is out-of-date now that the layer is N-D and cropping is not one dimension or another, but across all dimensions after the axis.

src/caffe/layers/crop_layer.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+template <typename Dtype>
+void CropLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
+  const CropParameter& param = this->layer_param_.crop_param();
+  // bottom[0] supplies the data
+  // bottom[1] supplies the size
+  int input_dim = bottom[0]->num_axes();
+  CHECK_LT(param.axis(), input_dim) << "crop axis bigger than input dim";
```
shelhamer: It seems safe to move any logic that depends only on the number of dimensions to `LayerSetUp()`; switching the number of dims. would break many layers.

src/caffe/layers/crop_layer.cpp
```diff
@@ -0,0 +1,142 @@
...
+void CropLayer<Dtype>::crop_copy(const vector<Blob<Dtype>*>& bottom,
+             const vector<Blob<Dtype>*>& top,
+             const vector<int>& offsets,
+             vector<int> indices,
+             int cur_dim,
+             const Dtype* src_data,
+             Dtype* dest_data,
+             bool is_forward) {
+  if (cur_dim + 1 < top[0]->num_axes()) {
+    // We are not yet at the final dimension, call copy recursivley
```
shelhamer: recursivley -> recursively

src/caffe/layers/crop_layer.cu
```diff
@@ -0,0 +1,127 @@
...
+          top_diff, bottom_diff);
+    }
+  }
+}
+
+template <typename Dtype>
+void CropLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
+  std::vector<int> indices(top[0]->num_axes(), 0);
+  // This works because crop_copy uses caffe_copy which calls cudaMemcpy.
```
shelhamer: Drop out-of-date comment as the implementation now uses a copy kernel.

shelhamer:
@BlGene I pushed a rebase on the latest master to my fork: https://github.com/shelhamer/caffe/tree/crop-nd

src/caffe/proto/caffe.proto
```diff
@@ -597,6 +598,16 @@ message ConvolutionParameter {
   optional bool force_nd_im2col = 17 [default = false];
 }
 
+message CropParameter {  
+  // If only one crop_offset is specified, all the dimensions after the specified
+  // axis would be cropped (by the same amount); otherwise the number of crop
+  // values specified must be equal to the number of dimensions following axis, and
+  // the trailing dimensions would be cropped accordingly.
+  // Protip: standard dimensions are ( N,C,H,W )
+  optional uint32 axis = 1 [default = 2];
```
shelhamer: Let's make `axis` mirror the `axis` arg of other layers to allow for negative indexing through an `int32` field and the `CanonicalAxisIndex()` helper method? See for instance https://github.com/BVLC/caffe/blob/master/src/caffe/layers/flatten_layer.cpp#L12-L13

BlGene:
@shelhamer this last commit should address all the points you mentioned.

src/caffe/layers/crop_layer.cpp
```diff
@@ -0,0 +1,148 @@
...
+  const CropParameter& param = this->layer_param_.crop_param();
+  CHECK_EQ(bottom.size(), 2) << "Wrong number of bottom blobs.";
+  int input_dim = bottom[0]->num_axes();
+  const int start_axis = bottom[0]->CanonicalAxisIndex(param.axis());
+  CHECK_LT(start_axis, input_dim) << "crop axis bigger than input dim";
+  if (param.offset_size() > 1) {
+    // the number of crop values specified must be equal to the number
+    // of dimensions following axis
+    CHECK_EQ(start_axis + param.offset_size(), input_dim)
+      << "number of crop values specified must be equal to the number of "
```
shelhamer: `crop values` -> `offset values`

shelhamer:
Other then the last trivial comment all that's needed now are tests. @BlGene do you have these planned, or should I take a stab at it?

BlGene:
@shelhamer 
Fixed `crop values` -> `offset values`
Added tests.

shelhamer:
@BlGene I revised the tests in my [latest commits to shelhamer/caffe:nd-crop](https://github.com/shelhamer/caffe/commits/crop-nd) to

- align with test conventions
- add a 5D test
- switch to the standard gradient checker

so that all the TODOs are covered. Can you incorporate my changes into this branch for merge? Please squash commits to simplify history (including mine, I don't mind) into something like

- a40439d original crop layer (longjon)
- 1e9d279 extend to N-D (BlGene) with 409fa12, 9a4c15c, and 6437271 squashed
- tests (BlGene)
- test grooming: style, 5D, standard grad check (shelhamer)

I've added tests to #3613 so we can merge both once this is ready.

BlGene:
@shelhamer Squashed!

shelhamer:
@BlGene thanks for the quick reply! Sorry to be so obsessed with history but can you reword the messages to make the subject line (first line of the message) more complete? That is, b6ee629 could be "Crop: fixes, tests, and negative axis indexing" and 631ff09 could be "Crop: more tests and test tuning" with the rest of the details in a list.

With that this will be a model PR and finally merged! Thanks for all your work on this and the attention to detail. I'm excited to have all the FCN tools in master soon.

BlGene:
@shelhamer OK, I fixed the commit messages. 

I'm happy to have FCN functionality merged too. Thanks for the attention to detail. :+1:

elezar:
Just a quick question from my side, can the ND crop be used to implement a rectangular crop as in #1980?

BlGene:
This layer crops blob B to have the same shape as Blob A. So its a bit hacky, but you could create a dummy blob (A) that has the desired shape and crop the (B) blob to this size.

BR, Max

Issue #2688, opened 2015-07-03
cuDNN max-pooling is not compatible with in-place dropout


mfigurnov:
This is the same issue as https://github.com/BVLC/caffe/issues/117, except that now the problem arises only when max-pooling layer uses cuDNN engine. The reason is probably the same: cuDNN's max-pooling compares inputs and outputs of the layer to perform backward pass.

How to reproduce: train @mavenlin's [network for CIFAR-10](https://gist.github.com/mavenlin/e56253735ef32c3c296d) using cuDNN-enabled Caffe. The training gets stuck after few thousand iterations. However, if pool1 layer is switched to Caffe engine, the training proceeds just fine.

longjon:
This is a symptom of #2015, so I'll note it there as a special case.

Issue #2015, opened 2015-03-02
In-place computation can break gradient computation


cdoersch:
For instance, MVNLayer reads data from its top blob during the backward pass, under the assumption that this data is exactly the same as the output it created.  If it's been modified by a later layer that does in-place computation, the gradient will be computed incorrectly.

In general, caffe should not rely on the user to know under what citcumstances a layer can safely be done in-place.

seanbell:
Note: you may want to coordinate any suggested changes with #1979.

longjon:
According to @mfigurnov, cuDNN max pooling is also a layer that requires its top data during backward.

seanbell:
It's probably worth adding a mechanism to each layer that says whether it (a) does in-place computation and (b) can support the next layer doing in-place computation.  Then, the net could check that all of the layers are compatible upon startup.

shelhamer:
Further thoughts from Sean Bell in #2853:

> My understanding is that right now there is no specification -- you basically need to study the layer implementation to decide whether or not you can put an in-place layer after it. Getting it wrong will lead to incorrect results, but you won't get any error or warning about it.
> 
> A better solution would be to have each layer declare whether or not it allows for in-place computation, as well as whether the next layer can have in-place computation. Then, caffe could check these flags and raise errors as necessary. This isn't implemented, but it would be great if someone did.

PR #3574, opened 2016-01-20
Workaround for inplace max pooling issue

thatguymike:
CuDNN assumes layers are not being modified in place, thus breaking our index tracking for updates in some cases in Caffe.  Until there is a workaround in Caffe (index management) or cuDNN, use Caffe
engine for max pooling, or don't use in place layers after max pooling layers

### DIFF
--- a/src/caffe/layer_factory.cpp
+++ b/src/caffe/layer_factory.cpp
@@ -91,11 +91,11 @@ shared_ptr<Layer<Dtype> > GetPoolingLayer(const LayerParameter& param) {
                 << "Using Caffe's own pooling layer.";
       return shared_ptr<Layer<Dtype> >(new PoolingLayer<Dtype>(param));
     }
-// CuDNN assumes layers are not being modified in place, thus breaking
-// our index tracking for updates in some cases in Caffe.  Until there
-// is a workaround in Caffe (index management) or cuDNN, use Caffe
-// layer to max pooling, or don't use in place layers after max
-// pooling layers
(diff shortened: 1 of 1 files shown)


shelhamer:
This solution forces all max pooling to be done by the Caffe engine without offering an override. Correctness is the best default, so I'm fine with that, but I wanted to check that this global switch is desired.

Note this is a particular instance of #2015 highlighted in #2688.

thatguymike:
For now, I think this is the best solution until we understand the way we want to fundamentally handle in-place more generally.  There is a performance difference, but for most folks doing training, shouldn't matter much.  The fundamental issue with inplace operations is going to bite us again and again I fear.  We can choose to try to address the fundamental issue, but without this change or similar workaround we are breaking people's nets.

shelhamer:
@thatguymike agreed. Please adjust the indentation (4 spaces to align with code) for merge -- for some reason lint is not checking the style.

Thanks for the workaround.

thatguymike:
The whole file appears to be 2 space indent for code, which is what I have...  Or do you just mean you want the comment block indented?  Lint is enforcing style, it made me add the other style changes you prefer like spaces after if's.

shelhamer:
Ok, merged in 37066eb07af936da0ced1564e78c91002ebb6add (with changes squashed into a single commit). Closing this PR.

@thatguymike It was the comments. I thought our lint check caught out-of-line comments but apparently not.

PR #3587, opened 2016-01-22
Copy proto definition to distribute directory

junshi15:
To ensure Caffe to be executable on JVM, caffe.prototxt is required by JVM layers to access solver and network configurations. For that purpose, we make caffe.prototxt available in distribute directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -651,6 +651,8 @@ superclean: clean supercleanfiles
 $(DIST_ALIASES): $(DISTRIBUTE_DIR)
 
 $(DISTRIBUTE_DIR): all py | $(DISTRIBUTE_SUBDIRS)
+	# add proto
+	cp -r src/caffe/proto $(DISTRIBUTE_DIR)/
 	# add include
 	cp -r include $(DISTRIBUTE_DIR)/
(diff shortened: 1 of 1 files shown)


shelhamer:
It's slightly odd to copy the `caffe.proto` source for distribution. Are you compiling the proto definition for java separately or do you have different bindings than those generated by the usual `protoc`? The java protobuf compilation could be added to the build.

On the other hand shipping `caffe.proto` seems harmless.

junshi15:
Yes, we process the proto files in java as well, mainly for the input data layer. For example, we get the source location from the memory data layer. The location is typically an HDFS URI and is the base for a Spark RDD.

anfeng:
@shelhamer In our use case, we need Java class generated from caffe.proto at this point. Python binding may be needed later. It is okay for us if Caffe distribute those bindings for caffe.proto. However, it will be easier to let users handle them directly.

shelhamer:
Ok this is fine, but to quote my earlier comment

> The java protobuf compilation could be added to the build

in the same way the [Python proto bindings are built](https://github.com/BVLC/caffe/blob/master/Makefile#L623-L626).

PR #3588, opened 2016-01-22
Refine P2PSync

junshi15:
This PR makes the following changes into P2PSync class.

    P2PSync::GetInitIter() ... a new getter function for initial iteration number.
    P2PSync::run() ... refactored into 2 steps: prepare(), and thread launch/execution
    P2PSync::prepare() ... establish the GPU pairs within a process for reduce operations

### DIFF
--- a/include/caffe/parallel.hpp
+++ b/include/caffe/parallel.hpp
@@ -93,7 +93,10 @@ class P2PSync : public GPUParams<Dtype>, public Solver<Dtype>::Callback,
     return solver_;
   }
 
-  void run(const vector<int>& gpus);
+  void Run(const vector<int>& gpus);
+  void Prepare(const vector<int>& gpus,
+               vector<shared_ptr<P2PSync<Dtype> > >* syncs);
+  inline const int initial_iter() const { return initial_iter_; }
 
  protected:
   void on_start();
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -380,7 +380,8 @@ void P2PSync<Dtype>::on_gradients_ready() {
 }
 
 template<typename Dtype>
-void P2PSync<Dtype>::run(const vector<int>& gpus) {
+void P2PSync<Dtype>::Prepare(const vector<int>& gpus,
+            vector<shared_ptr<P2PSync<Dtype> > >* syncs) {
   // Pair devices for map-reduce synchronization
   vector<DevicePair> pairs;
   DevicePair::compute(gpus, &pairs);
@@ -391,15 +392,14 @@ void P2PSync<Dtype>::run(const vector<int>& gpus) {
   LOG(INFO)<< "GPUs pairs " << s.str();
 
   SolverParameter param(solver_->param());
-  vector<shared_ptr<P2PSync<Dtype> > > syncs(gpus.size());
 
   // Build the GPU tree by finding the parent for each solver
   for (int attempts = 0; attempts < pairs.size(); ++attempts) {
     for (int i = 1; i < pairs.size(); ++i) {
-      if (!syncs[i].get()) {
+      if (!syncs->at(i).get()) {
         P2PSync<Dtype>* parent = NULL;
-        for (int j = 0; j < syncs.size(); ++j) {
-          P2PSync<Dtype>* sync = j == 0 ? this : syncs[j].get();
+        for (int j = 0; j < syncs->size(); ++j) {
+          P2PSync<Dtype>* sync = j == 0 ? this : syncs->at(j).get();
           if (sync) {
             const SolverParameter& p = sync->solver()->param();
             if (p.device_id() == pairs[i].parent()) {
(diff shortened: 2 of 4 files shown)


shelhamer:
Please squash the lint fixes -- we like to keep the history clear.

include/caffe/parallel.hpp
```diff
@@ -94,6 +94,9 @@ class P2PSync : public GPUParams<Dtype>, public Solver<Dtype>::Callback,
   }
 
   void run(const vector<int>& gpus);
+  void prepare(const vector<int>& gpus,
```
longjon: This should probably be `Prepare` according to the style guide and existing style. Since we already have `run`, however, I guess we can postpone that to a future commit.

include/caffe/parallel.hpp
```diff
@@ -94,6 +94,9 @@ class P2PSync : public GPUParams<Dtype>, public Solver<Dtype>::Callback,
...
+               vector<shared_ptr<P2PSync<Dtype> > >* syncs);
+  inline const int GetInitIter() const { return initial_iter_; }
```
longjon: And this should be `initial_iter`, or rename `initial_iter_` to `init_iter_` and call it `init_iter`. See other getters throughout Caffe; they are `underscore_case` and do not start with `Get` and match the member name.

src/caffe/parallel.cpp
```diff
@@ -391,15 +392,14 @@ void P2PSync<Dtype>::run(const vector<int>& gpus) {
   LOG(INFO)<< "GPUs pairs " << s.str();
 
   SolverParameter param(solver_->param());
-  vector<shared_ptr<P2PSync<Dtype> > > syncs(gpus.size());
 
   // Build the GPU tree by finding the parent for each solver
   for (int attempts = 0; attempts < pairs.size(); ++attempts) {
     for (int i = 1; i < pairs.size(); ++i) {
-      if (!syncs[i].get()) {
+      if (!syncs->at(i).get()) {
```
longjon: Note that `at` is not the same as `operator[]`; this diff adds bounds checking. Presumably it should have been this way all along, though.

longjon:
Thanks, this patch looks good except for the getter name issue above, an optionally fix up existing naming issues (member functions that do nontrivial work should be capitalized CamelCase.)

src/caffe/parallel.cpp
junshi15: This git diff is misleading. In the original code, syncs is a vector, in the revised code, syncs is a pointer to the vector. The reason for the change is because Caffe Lint discourages passing variables by non-constant reference.

junshi15:
Thanks for the review. We have revised the code accordingly.

junshi15:
It looks there is a problem with travis-CI PYTHON-3.0 builds. It's not a problem with this PR. I will try to build again once that is fixed.

src/caffe/parallel.cpp
longjon: This is not my point. While it's clear that we've switched to a pointer type, we've _also_ switched from calling `operator[]` to the `at` method. (You could just as well have written `(*syncs)[i]` in the new code, or `syncs.at(i)` in the old code.) Please note that these are not the same; the former does not perform bounds checking, while the latter does. 

Now presumably bounds checking _is_ the right thing to do here; there's not going to be a performance issue here, and the indexing is not very obviously within bounds.

Furthermore, we do indeed disallow non-`const` references. However, the right thing to do here might be to pass a `const` reference. Even though we use pointer types for output arguments, note that the output arguments here are really the `P2PSync` elements of the vector, not the vector itself (e.g., it's not okay for `Prepare` to add or remove elements from this vector, even though that's allowed in the current patch). See #945 for a lot more discussion on this point.

So I'd recommend changing the type of `syncs` to a `const` reference, but _keeping_ the calls to `at`, if you think that makes sense.

longjon:
Okay, this looks good except for the (more subtle) point in the comment above; feel free to adjust accordingly and rebase on master which should fix the Travis issue.

src/caffe/parallel.cpp
junshi15: Thanks for reiterating the difference between [i] and at(i). It is clear.

The output arguments in Prepare() is however different from what's described in https://github.com/BVLC/caffe/pull/945. Prepare() in this patch corresponds to the original code here: https://github.com/BVLC/caffe/blob/master/src/caffe/parallel.cpp#L394-L412

syncs starts with a vector of empty shared_ptrs (i.e. null). Inside Prepare(), those shared pointers gets assigned. In https://github.com/BVLC/caffe/pull/945, vector<Blob<Dtype>* > is already initialized (i.e. the pointers inside the vector points valid objects, not null), those Forward or Backward functions only change the content of the blobs, but not replace or delete the blobs.

I had tried const reference in the first place, but had issues with this reset function. 
https://github.com/BVLC/caffe/pull/3588/files#diff-55c214d759a1a3d26cfdaf807eb86b10L412
Hence it is now a pointer to the vector.

longjon: Ah, yes, you're right, `const &` won't work here, so a pointer is correct.

longjon:
Thanks for the patch, the diff and history look clean now, merging.

PR #1416, opened 2014-11-07
Added opencv vector<Mat> to memory data layer with tests

mtamburrano:
- AddMatVector  to load Opencv Mat into MemoryDataLayer
- MemoryDataLayer accepts blobs with dynamic batch size,  especially useful when you need to predict a number of images that is not known a priori

PR #3589, opened 2016-01-22
Enhance memory data layer

junshi15:
This PR enhances memory data layer with:

```
forward_gpu() is added to avoid unneeded data copy
parameter "share_in_parallel" for controlling memory sharing or not (default: true)
parameter "source" for specifying the location of data source, e.g. hdfs URI
```

This PR introduce an optional parameter for LayerParameter:

```
"source_class" ... Name of custom class for parsing input data source
```

### DIFF
--- a/include/caffe/layers/memory_data_layer.hpp
+++ b/include/caffe/layers/memory_data_layer.hpp
@@ -27,6 +27,7 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
   virtual inline const char* type() const { return "MemoryData"; }
   virtual inline int ExactNumBottomBlobs() const { return 0; }
   virtual inline int ExactNumTopBlobs() const { return 2; }
+  virtual inline bool ShareInParallel() const { return share_in_parallel_; }
 
   virtual void AddDatumVector(const vector<Datum>& datum_vector);
 #ifdef USE_OPENCV
@@ -47,6 +48,8 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
  protected:
   virtual void Forward_cpu(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top);
+  virtual void Forward_gpu(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
 
   int batch_size_, channels_, height_, width_, size_;
   Dtype* data_;
@@ -56,6 +59,7 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
   Blob<Dtype> added_data_;
   Blob<Dtype> added_label_;
   bool has_new_data_;
+  bool share_in_parallel_;
 };
 
 }  // namespace caffe
--- a/src/caffe/layers/memory_data_layer.cpp
+++ b/src/caffe/layers/memory_data_layer.cpp
@@ -15,6 +15,8 @@ void MemoryDataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>& bottom,
   channels_ = this->layer_param_.memory_data_param().channels();
   height_ = this->layer_param_.memory_data_param().height();
   width_ = this->layer_param_.memory_data_param().width();
+  share_in_parallel_ = this->layer_param_.memory_data_param()
+                       .share_in_parallel();
   size_ = channels_ * height_ * width_;
   CHECK_GT(batch_size_ * size_, 0) <<
       "batch_size, channels, height, and width must be specified and"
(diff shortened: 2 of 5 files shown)


src/caffe/proto/caffe.proto
```diff
@@ -347,6 +347,9 @@ message LayerParameter {
   // Parameters shared by loss layers.
   optional LossParameter loss_param = 101;
 
+  // Specify name of source class
+  optional string source_class = 143;
```
shelhamer: This should be a field in `MemoryDataParameter`, not `LayerParameter` if I understand this patch.

include/caffe/layers/memory_data_layer.hpp
```diff
@@ -27,6 +27,7 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
   virtual inline const char* type() const { return "MemoryData"; }
   virtual inline int ExactNumBottomBlobs() const { return 0; }
   virtual inline int ExactNumTopBlobs() const { return 2; }
+  virtual inline bool ShareInParallel() const { return share_in_parallel_; }
```
longjon: As in #3588, getters should normally be underscore_case according to the style guide and existing usage.

src/caffe/layers/memory_data_layer.cpp
```diff
@@ -117,6 +119,19 @@ void MemoryDataLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
     has_new_data_ = false;
 }
 
+template <typename Dtype>
+void MemoryDataLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  CHECK(data_) << "MemoryDataLayer needs to be initalized by calling Reset";
+  top[0]->Reshape(batch_size_, channels_, height_, width_);
+  top[1]->Reshape(batch_size_, 1, 1, 1);
```
longjon: I can't figure out why this Reshape code is in Forward... it looks like it was added in https://github.com/BVLC/caffe/commit/cedefd77625db978623fcf00ecc32d899eed954b, along with #1416, but I don't see why we don't just do the reshape at the time the batch size is changed? That would avoid duplicating these lines between CPU and GPU.

src/caffe/proto/caffe.proto
longjon: Agree. Also, this parameter (along with `source`) needs some more explanation, as right now they seem unused in Caffe. Why are we adding these?

longjon:
Comments as noted; also, the logic here does not seem quite right.

In this patch we step through device memory in GPU mode, and host memory in CPU mode. But what if we give this layer a host pointer in GPU mode? (Or (should this even be allowed?) a device pointer in CPU mode?) It looks like this patch will break things, even though (at least the former case) works fine as-is (and is a pretty common case!).

So it seem like the action of Forward should really depend on the pointer type, not the mode.

src/caffe/proto/caffe.proto
anfeng: This pull request is motivated by making Caffe work on Spark (via CaffeOnSpark). source_class was used by CaffeOnSpark to decide which particular Scala class to use for handling a given data source. See [CaffeOnSpark code](https://github.com/yahoo/CaffeOnSpark/blob/master/caffe-grid/src/main/scala/com/yahoo/ml/caffe/DataSource.scala#L114). 

Our original idea is to support custom sources for data layers beyond MemoryDataLayer. For example, to support NLP use cases, we have introduced a VWDataLayer. That's why we propose an optional field at LayerParameter. By having source_class at LayerParemeter, we will enable CaffeOnSpark users to introduce customized parser for different data layers.

FYI, VWDataParameter is defined as below:
message VWDataParameter {
  optional string source = 1;
  optional uint32 batch_size = 2;
  optional bool shuffle = 3 [default = false];
  message TopLayer{
    optional string name = 1;
    optional string type = 2;
    optional uint32 channels = 3;
    optional uint32 height = 4;
    optional uint32 width = 5;
    optional string vw_namespace = 6;
  }
  repeated TopLayer top = 4;
  optional bool sparse = 5 [default = false];
  optional bool memory_input = 6[default = false];
}

junshi15:
Regarding the last comment, we assume memory layer gets a host pointer in the CPU mode (hence Forward_cpu) and a device pointer in the GPU mode (hence Forward_gpu). Is this a good assumption? Maybe we should add a few assertions to check that.

longjon:
> Regarding the last comment, we assume memory layer gets a host pointer in the CPU mode (hence Forward_cpu) and a device pointer in the GPU mode (hence Forward_gpu). Is this a good assumption?

No. As noted above, this assumption is violated in practice. The way I've most commonly used `MemoryDataLayer` is with a host pointer (the only supported option, right now) in GPU mode. That's desirable in the common case where some data already exists in host memory, and we want to run a GPU computation on it without bothering to copy all of it device (perhaps because it won't fit, which is why we're batching through it with this layer in the first place). I don't think it's reasonable to break that use case, and I don't see any reason why the layer can't do the right thing for either mode and either pointer type.

junshi15:
Can we assume the following:
1) host memory is always allocated by CaffeMallocHost(), not by an arbitrary malloc().
2) in CPU mode, this layer does not take a device pointer, i.e. Forward_cpu() won't handle device pointer.
If those are o.k., then we can use cudaPointerGetAttributes() to detect the pointer type in the GPU mode.

shelhamer:
Closing according to #5528. Thanks for the effort to improve this troubled layer.

Borracho27 (commented):
## - O[]()~~**jggrj**~~

PR #3590, opened 2016-01-22
Add functions to check and grab GPU

junshi15:
This PR introduces capability for Caffe applications to programmatically obtain GPU resources in a cloud environment. In such an environment, multiple Caffe applications may be allocated to a server, and these apps should be bound to different set of GPUs.

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -153,6 +153,11 @@ class Caffe {
   static void SetDevice(const int device_id);
   // Prints the current GPU status.
   static void DeviceQuery();
+  // Check if specified device is available
+  static bool CheckDevice(const int device_id);
+  // Search from start_id to the highest possible device ordinal,
+  // return the ordinal of the first available device.
+  static int FindDevice(const int start_id = 0);
   // Parallel training info
   inline static int solver_count() { return Get().solver_count_; }
   inline static void set_solver_count(int val) { Get().solver_count_ = val; }
--- a/src/caffe/common.cpp
+++ b/src/caffe/common.cpp
@@ -70,6 +70,15 @@ void Caffe::DeviceQuery() {
   NO_GPU;
 }
 
+bool Caffe::CheckDevice(const int device_id) {
+  NO_GPU;
+  return false;
+}
+
+int Caffe::FindDevice(const int start_id) {
+  NO_GPU;
+  return -1;
+}
 
 class Caffe::RNG::Generator {
  public:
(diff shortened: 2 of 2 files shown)


src/caffe/common.cpp
```diff
@@ -192,6 +201,26 @@ void Caffe::DeviceQuery() {
   return;
 }
 
+bool Caffe::CheckDevice(const int device_id) {
+  bool r =  ((cudaSuccess == cudaSetDevice(device_id)) &&
+            (cudaSuccess == cudaFree(0)));
```
longjon: Can you explain why we're calling `cudaFree` here?

src/caffe/common.cpp
```diff
@@ -192,6 +201,26 @@ void Caffe::DeviceQuery() {
...
+  cudaGetLastError();
```
longjon: Presumably this is to reset any error that may have occurred; if so that's worthy of a comment, since right now it looks like we're just discarding the error.

src/caffe/common.cpp
```diff
@@ -192,6 +201,26 @@ void Caffe::DeviceQuery() {
   return;
 }
 
+bool Caffe::CheckDevice(const int device_id) {
+  bool r =  ((cudaSuccess == cudaSetDevice(device_id)) &&
```
longjon: There's an extra space on this line.

src/caffe/common.cpp
```diff
@@ -192,6 +201,26 @@ void Caffe::DeviceQuery() {
...
+            (cudaSuccess == cudaFree(0)));
+  cudaGetLastError();
+  return r;
+}
+
+int Caffe::GrabDevice(const int start_id) {
+  int count = 0;
+  int r = -1;
+  if (cudaSuccess == cudaGetDeviceCount(&count)) {
```
longjon: Rather than putting the whole body of this method in a conditional, I think the whole thing would read better if we failed immediately if this call breaks (`CUDA_CHECK`, or is there a reason we need an error return here?), and then we can drop the `r` temporary which is making this look more complicated than it is.

src/caffe/common.cpp
```diff
@@ -192,6 +201,26 @@ void Caffe::DeviceQuery() {
...
 }
 
+bool Caffe::CheckDevice(const int device_id) {
+  bool r =  ((cudaSuccess == cudaSetDevice(device_id)) &&
+            (cudaSuccess == cudaFree(0)));
+  cudaGetLastError();
+  return r;
+}
+
+int Caffe::GrabDevice(const int start_id) {
```
longjon: Would `FindDevice` be a better name here? To me "grab" implies that this method is claiming the device, but it's merely a passive check.

include/caffe/common.hpp
```diff
@@ -149,6 +149,10 @@ class Caffe {
   static void SetDevice(const int device_id);
   // Prints the current GPU status.
   static void DeviceQuery();
+  // Check if specified device is available
+  static bool CheckDevice(const int device_id);
+  // Get the first available device id since start_id
```
longjon: "that's at least start_id", right? "Since" is ambiguous as to inclusion (and doesn't read well anyway).

longjon:
Thanks, this looks fine after the minor issues above have been addressed.

src/caffe/common.cpp
```diff
@@ -192,6 +201,26 @@ void Caffe::DeviceQuery() {
...
+  int count = 0;
+  int r = -1;
+  if (cudaSuccess == cudaGetDeviceCount(&count)) {
+    for (int i = start_id; i < count; i++) {
+      if (CheckDevice(i)) {
+        r = i;
+        break;
+      }
+    }  // for
```
longjon: This comment is spurious.

junshi15:
thanks for the comments. We have updated accordingly.

junshi15:
How do I trigger the build again? The problems were due to PYTHON3, not related to any changes in this PR.

longjon:
You'll need to rebase on the latest master to re-run the fixed Travis build. (Sorry about the breakage, that was due to a change upstream to protobuf.)

(By the way, is there a reason you've stuck with the `GrabDevice` name? I still find it misleading...)

junshi15:
We use EXCLUSIVE_PROCESS mode at Yahoo. In this mode, GrabDevice() does claim the device and prohibit others from using the device. But this is not the case when the devices are in the default mode. I am fine with FindDevice().

longjon:
I see, that makes sense (although note that `CheckDevice` is actually doing the grabbing in an exclusive mode, which is not implied by the name). The current naming makes sense in the default mode (which is the one always used by BVLC, as far as I know) (although perhaps this function is less useful in that mode).

The history and diff are now clean and passing so I'm merging as-is and we can revise names later on if it makes sense to do so. Thanks!

PR #3605, opened 2016-01-27
Remove incorrect cast of gemm int arg to Dtype in BiasLayer

jeffdonahue:
Oops, just noticed this.


jeffdonahue:
Merging this minor fix.

Issue #670, opened 2014-07-11
Transposed weights sharing == tied weights for AE


ducha-aiki:
Hi,

Is there any possibility to use weights sharing (like in [#546](https://github.com/BVLC/caffe/pull/546) ), but transposed? It is used for [autoencoders](http://deeplearning.net/tutorial/dA.html) and [deconvnets](http://www.matthewzeiler.com/pubs/iccv2011/iccv2011.pdf)?

Macbull:
@shelhamer I can't find solution to transposed weights there ! Can you point me to the particular comment you are talking about ?

SBodenstedt:
I am also looking in transposed weight sharing, but I can't find the solution mentioned above either... is there an example? Thanks!

Macbull:
@SBodenstedt see if #2565 is of some help to you !

SBodenstedt:
@Macbull Thanks for the tip! I checked it out and compiled, but it doesn't seem to work. I used the Transpose layer to share weights between two InnerProduct layers, but when trying to train the net, a seg fault occurred:
```
I0811 18:41:12.985070  2152 net.cpp:489] Collecting Learning Rate and Weight Decay.
I0811 18:41:12.985090  2152 net.cpp:247] Network initialization done.
I0811 18:41:12.985100  2152 net.cpp:248] Memory required for data: 9931188
I0811 18:41:12.985183  2152 solver.cpp:42] Solver scaffolding done.
caffe-d: /usr/include/boost/smart_ptr/shared_ptr.hpp:653: typename boost::detail::sp_member_access<T>::type boost::shared_ptr<T>::operator->() const [with T = caffe::Blob<float>; typename boost::detail::sp_member_access<T>::type = caffe::Blob<float>*]: Assertion `px != 0' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff699ccc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56      ../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff699ccc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff69a00d8 in __GI_abort () at abort.c:89
[...]

PR #3612, opened 2016-01-29
Tied weights with transpose flag for InnerProduct layer

kashefy:
I wanted to train an autoencoder where the deocder uses the tranpose of the encoder's weight matrix. This is first discussed in [#670](https://github.com/BVLC/caffe/issues/670) and followed up in [#1211 (comment)](https://github.com/BVLC/caffe/issues/1211#issuecomment-58031848). But it seemed this wasn't resolved. I found @jeffdonahue 's suggestion in [this comment](https://github.com/BVLC/caffe/issues/1211#issuecomment-58038944) to just add a transpose flag to the InnerProduct layer quite reasonable.

This PR adds a transpose flag to the InnerProduct layer as well as its params protobuf message.
When set to true for the deocder, in the forward pass, the call to the matrix multiplication routine is instructed to **NOT** transpose the weight matrix. Which is what you want in the usual case and for the encoder.

Tying the weights between encoder and decoder requires:

1. sharing the encoder's weight params with the inner product layer that is the decoder.
2. setting the share_mode to true in both ip layers (otherwise it won't allow for the shape mismatch)
3. Adding 'transpose: true' to the decoder's inner_product_param

A sample [trainval.prototxt](https://gist.github.com/kashefy/72205d7c2da39520f893) to demonstrate usage.

I haven't written unit tests around this yet. Open to suggestions to what makes sense to test for here.

Thanks for reviewing and looking forward to the feedback.

### DIFF
--- a/src/caffe/layers/inner_product_layer.cpp
+++ b/src/caffe/layers/inner_product_layer.cpp
@@ -11,6 +11,7 @@ void InnerProductLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
   const int num_output = this->layer_param_.inner_product_param().num_output();
   bias_term_ = this->layer_param_.inner_product_param().bias_term();
+  transpose_ = this->layer_param_.inner_product_param().transpose();
   N_ = num_output;
   const int axis = bottom[0]->CanonicalAxisIndex(
       this->layer_param_.inner_product_param().axis());
@@ -27,10 +28,15 @@ void InnerProductLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
     } else {
       this->blobs_.resize(1);
     }
-    // Intialize the weight
+    // Initialize the weights
     vector<int> weight_shape(2);
-    weight_shape[0] = N_;
-    weight_shape[1] = K_;
+    if (transpose_) {
+      weight_shape[0] = K_;
+      weight_shape[1] = N_;
+    } else {
+      weight_shape[0] = N_;
+      weight_shape[1] = K_;
+    }
     this->blobs_[0].reset(new Blob<Dtype>(weight_shape));
     // fill the weights
     shared_ptr<Filler<Dtype> > weight_filler(GetFiller<Dtype>(
@@ -80,7 +86,8 @@ void InnerProductLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
   const Dtype* bottom_data = bottom[0]->cpu_data();
   Dtype* top_data = top[0]->mutable_cpu_data();
   const Dtype* weight = this->blobs_[0]->cpu_data();
-  caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasTrans, M_, N_, K_, (Dtype)1.,
+  caffe_cpu_gemm<Dtype>(CblasNoTrans, transpose_ ? CblasNoTrans : CblasTrans,
+      M_, N_, K_, (Dtype)1.,
       bottom_data, weight, (Dtype)0., top_data);
   if (bias_term_) {
     caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, M_, N_, 1, (Dtype)1.,
@@ -97,8 +104,17 @@ void InnerProductLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
     const Dtype* top_diff = top[0]->cpu_diff();
     const Dtype* bottom_data = bottom[0]->cpu_data();
     // Gradient with respect to weight
-    caffe_cpu_gemm<Dtype>(CblasTrans, CblasNoTrans, N_, K_, M_, (Dtype)1.,
-        top_diff, bottom_data, (Dtype)1., this->blobs_[0]->mutable_cpu_diff());
+    if (transpose_) {
+      caffe_cpu_gemm<Dtype>(CblasTrans, CblasNoTrans,
+          K_, N_, M_,
+          (Dtype)1., bottom_data, top_diff,
+          (Dtype)1., this->blobs_[0]->mutable_cpu_diff());
+    } else {
+      caffe_cpu_gemm<Dtype>(CblasTrans, CblasNoTrans,
+          N_, K_, M_,
+          (Dtype)1., top_diff, bottom_data,
+          (Dtype)1., this->blobs_[0]->mutable_cpu_diff());
+    }
   }
   if (bias_term_ && this->param_propagate_down_[1]) {
     const Dtype* top_diff = top[0]->cpu_diff();
@@ -110,9 +126,17 @@ void InnerProductLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
   if (propagate_down[0]) {
     const Dtype* top_diff = top[0]->cpu_diff();
     // Gradient with respect to bottom data
-    caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, M_, K_, N_, (Dtype)1.,
-        top_diff, this->blobs_[0]->cpu_data(), (Dtype)0.,
-        bottom[0]->mutable_cpu_diff());
+    if (transpose_) {
+      caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasTrans,
+          M_, K_, N_,
+          (Dtype)1., top_diff, this->blobs_[0]->cpu_data(),
+          (Dtype)0., bottom[0]->mutable_cpu_diff());
+    } else {
+      caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans,
+          M_, K_, N_,
+          (Dtype)1., top_diff, this->blobs_[0]->cpu_data(),
+          (Dtype)0., bottom[0]->mutable_cpu_diff());
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/caffe/layers/inner_product_layer.cu
```diff
@@ -19,8 +19,9 @@ void InnerProductLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
       caffe_gpu_axpy<Dtype>(N_, bias_multiplier_.cpu_data()[0],
                             this->blobs_[1]->gpu_data(), top_data);
   } else {
-    caffe_gpu_gemm<Dtype>(CblasNoTrans, CblasTrans, M_, N_, K_, (Dtype)1.,
-                          bottom_data, weight, (Dtype)0., top_data);
+      caffe_gpu_gemm<Dtype>(CblasNoTrans, transpose_ ? CblasNoTrans : CblasTrans,
+                            M_, N_, K_, (Dtype)1.,
+                            bottom_data, weight, (Dtype)0., top_data);
```
jeffdonahue: remove added indent

src/caffe/layers/inner_product_layer.cpp
```diff
@@ -80,8 +81,11 @@ void InnerProductLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
   const Dtype* bottom_data = bottom[0]->cpu_data();
   Dtype* top_data = top[0]->mutable_cpu_data();
   const Dtype* weight = this->blobs_[0]->cpu_data();
-  caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasTrans, M_, N_, K_, (Dtype)1.,
+
+  caffe_cpu_gemm<Dtype>(CblasNoTrans, transpose_ ? CblasNoTrans : CblasTrans,
+      M_, N_, K_, (Dtype)1.,
       bottom_data, weight, (Dtype)0., top_data);
+
```
jeffdonahue: remove added empty line here and on 84

jeffdonahue:
Thanks @kashefy! This looks pretty good to me.

> setting the share_mode to true in both ip layers (otherwise it won't allow for the shape mismatch)

This shouldn't be needed -- instead the weight param should be set to the correct shape by swapping `N_`/ & `K_`, changing lines 32-33 of `inner_product_layer.cpp` to be conditioned on `transpose`.

Besides that, please see the style nitpicks and squash your history to a single commit.

Re testing: it would be good to have a few unit tests:

- [ ] verify the correct shape of the parameter with and without `transpose` set
- [ ] a gradient check with `transpose` set
- [ ] a forward check, for example: initialize an IP layer without `transpose` and the parameter randomly initialized, run `Forward`, save the result; initialize another IP layer *with* `transpose`, manually copy and transpose the value of the parameter from the first IP layer, then run `Forward` on the same input and check that the result is the same

kashefy:
@jeffdonahue, thanks for the feedback. Will fix styling (travis build failed because of it) and add the unit tests.

jeffdonahue:
Great, thanks.  Also just noticed you didn't change backward -- pretty sure that will need a different `CBlasTrans` setting as well.  (But no need to think about it once you write the gradient check :)

kashefy:
quick update:
Wrote unit tests around forward, backward, blob shape (with and without transpose). Tests pass. Fixed styling.
Transposing shared weights needs more work. The transposing ip stores the weight shape in the post-transpose form but the transposing doesn't happen until the multiplication is called. However setting the num_outputs of the decoder collides with these assumptions.
Maybe I need a second flag next to the transpose one to hold off on switching the weight shapes in case the weights are tied with another 'encoder' layer.

src/caffe/test/test_inner_product_layer.cpp
```diff
@@ -148,4 +265,127 @@ TYPED_TEST(InnerProductLayerTest, TestGradient) {
   }
 }
 
+TYPED_TEST(InnerProductLayerTest, TestGradientTransposeFalse) {
```
jeffdonahue: Shouldn't this test be `TestGradientTransposeTrue` (with the corresponding change from `set_transpose(false)` to `set_transpose(true)`)?  This test is effectively a duplicate of the existing test above (TestGradient), I'd think.

And if this test is changed to be done with `transpose` on, is there still anything additionally tested by `TestBackwardTranspose`?  I would think the combination of `TestForward` with the gradient check would cover all functionality.

src/caffe/test/test_inner_product_layer.cpp
```diff
@@ -91,6 +135,79 @@ TYPED_TEST(InnerProductLayerTest, TestForward) {
...
+    inner_product_param->set_transpose(true);
+    shared_ptr<InnerProductLayer<Dtype> > ip_t(
+        new InnerProductLayer<Dtype>(layer_param));
+    ip_t->SetUp(this->blob_bottom_vec_, this->blob_top_vec_);
+    const int count_w = layer->blobs()[0]->count();
+    EXPECT_EQ(count_w, ip_t->blobs()[0]->count());
+    // manually copy and transpose the weights from 1st IP layer into 2nd
+    const Dtype* w = layer->blobs()[0]->cpu_data();
+    Dtype* w_t = ip_t->blobs()[0]->mutable_cpu_data();
+    const int WIDTH = layer->blobs()[0]->shape(1);
```
jeffdonahue: variable names `WIDTH` and `WIDTH_T` should be lowercased

jeffdonahue:
@kashefy the tests and code look good but see comments/nitpicks above.  Once you've addressed these, please squash your history to a single commit (or, if you prefer, two commits -- one for the style fixes of existing code, and another for your added feature and tests), and I can merge this.  Thanks!

src/caffe/test/test_inner_product_layer.cpp
kashefy: I capitalized them because their constants. Overkill?

kashefy:
@jeffdonahue thanks for the feedback. Will go over the redundant test. The PR as it is now only adds a transpose feature to the ip layer. Tiying weights in an auto encoder doesn't work yet. If you think the transpose feature is useful on its own, I can the tying part in another PR.

src/caffe/test/test_inner_product_layer.cpp
jeffdonahue: In general we follow the Google style guide, but I think our convention is to use regular lowercase with underscore variable names (`my_variable_name`) for `const`s initialized from other variables (as in this case), but camel case with `k` prefix (`kMyVariableName`) for `const`s directly initialized with their values known at compile time per the [Google style guide](https://google.github.io/styleguide/cppguide.html#Constant_Names).  We use all caps names only for macros AFAIK.

jeffdonahue:
I'm not sure I understand why shared weights between an "encoder" and "decoder" layer wouldn't work in the current form.  Both the shape and memory layout of the weight matrix would be the same between a normal IP layer (`transpose = false`) that takes D-dimensional input and produces N-dimensional output, as a transposed IP layer (`transpose = true`) that takes N-dimensional input and produces D-dimensional output.  Given that, I would think the only other thing that *should* need to be done in the `transpose=true` case (and which you have done here) is to change the BLAS transpose settings in forward/backward when reading from/writing to the weights.

I could certainly be missing something though.

src/caffe/test/test_inner_product_layer.cpp
kashefy: Got it, will fix. Thanks for clarifying.

kashefy:
The decoders weights are shaped KxN which is a correct because the encoder
has NxK. The individual weight elements are still in layed out as NxK,
they're not properly transposed yet. All weights are still present in the
decoder's reference to the weight blob but arranged wrong and lead to the
wrong transpose.
The unit test doesn't show this because it does a manual transpose. I'll
add a unit test with weight sharing to cover this case and help me in
debugging.

I'm not sure I understand why shared weights between an "encoder" and
"decoder" layer wouldn't work in the current form. Both the shape and
memory layout of the weight matrix would be the same between a normal IP
layer (transpose = false) that takes D-dimensional input and produces
N-dimensional output, as a transposed IP layer (transpose = true) that
takes N-dimensional input and produces D-dimensional output. Given that, I
would think the only other thing that *should* need to be done in the
transpose=true case (and which you have done here) is to change the BLAS
transpose settings in forward/backward when reading from/writing to the
weights.

I could certainly be missing something though.

—
Reply to this email directly or view it on GitHub
<https://github.com/BVLC/caffe/pull/3612#issuecomment-179505664>.

kashefy:
Failures are due to import errors when running python nose tests. Possible solution in #3638

kashefy:
Travis job passing now, can't really explain why. But glad it the import errors are gone now.

kashefy:
Hello @jeffdonahue, I think this is ready. The transpose worked for shared weights after all as is.

jeffdonahue:
@kashefy thanks, looks like this is almost there!  But could you add a simple `TestGradientTranspose` test? It should be exactly the same as the existing `TestGradient` but of course have one extra line that does `set_transpose(true)`.  And with that test added I'm inclined to say `TestBackwardTranspose` should be removed, unless you think there is something additionally tested in that which isn't covered by the gradient checker.

kashefy:
Hello @jeffdonahue, I've added `TestGradientTranspose` (bascially TestGradient + set transpose to true, as you suggested). You're right, `TestBackwardTranspose` is somewhat redundant. It's not covering anything that the gradient checker isn't already covering. However, I find it to be helpful in narrowing down where a failure could come from. The test was actually very helpful in setting up the backward computation, so I'm a bit reluctant in throwing it out. I tend to use tests as a debug aid, so I usually end up writing more to better understand where a failure is coming from at the expense of increasing redundancy and length of code.

kashefy:
Hello @jeffdonahue, do you think the current tests are sufficient? Anything else you think should go into this PR? Thanks.

jeffdonahue:
@kashefy thanks for adding the gradient check; I suppose it can't hurt much to have the backward test as it's presumably very quick (relative to the full gradient check).  LGTM -- thanks for this work.

PR #3613, opened 2016-01-30
Python/net spec coordinate map and crop computation

longjon:
This PR provides an updated version of #1975 (see also #1976; this is the new version described there). This is meant for use along with #3570 (new ND crop layer).

This version has several _advantages_ over #1975, which make it a better candidate for merge.

1. Unlike #1975, layers do not need a pointer back to their owning net, and are not responsible for inspecting each other. This preserves the clean separation between `Net` and `Layer`, whereas the strategy employed by #1975 allowed for layers to interact in arbitrary ways that are difficult to reason about and make certain kinds of functionality difficult to implement. (This was the primary objection blocking the merge of #1975.) The alternate strategy used here is to perform the net-level analysis at net definition time (through pycaffe's net spec), and bake the results into the parameters of the net, through a much more straightforward (e.g.) crop layer (now #3570).
2. This alternate strategy makes the computed parameters user-visible and user-modifiable, which has value for debugging and understanding.
2. Cropping issues appear at specification time rather than run time.
2. Rather than adding C++ code scattered throughout layers, this patch adds a single Python file.
4. The C++ `Layer` interface remains unchanged; layers don't gain an extra method which may be obscure to some users.

In addition, the functionality of this code is _significantly expanded_ from that of #1975. In particular,

1. The assumption that common ancestors can be reached by descending through first bottoms is removed. Any two spatially-mappable blobs with a common ancestor should work for cropping or otherwise computing coordinate maps.
2. Layers with multiple bottoms (e.g., `Eltwise` or `Convolution` with multiple inputs/outputs) should now be supported.
2. Rectangular filters should now be supported.
3. ND de/convolution should be supported.
4. Dilated de/convolution should be supported.

Like #1975, the coordinate maps computed with this code can be used for other purposes in addition to cropping.

The main _disadvantage_ of this approach compared to #1975 is that the coordinate mapping formulae are kept apart from the definitions of their corresponding layers, which have to be manually kept in sync. Given that these formulae only come in a few basic forms, this seems like a reasonable approach for now.

This PR is not complete. The latest version of this code has just been written and has not been well-tested. Work still to be done:
- [x] nudge this PR/#3570 as needed for compatibility
- [x] add a test for basic functionality
- [x] add a test for rectangular filters
- [x] add a test for ND convolution
- [x] settle on names for functions (`coord_map_from_to` is more awkward than it needs to be)
- [x] flesh out docstrings and comments (in particular, explain how to use `coord_map_from_to`)
- [x] fix style issues
- [x] consider whether any currently unsupported layers should be supported

### DIFF
--- a/python/caffe/coord_map.py
+++ b/python/caffe/coord_map.py
@@ -1,0 +1,95 @@
+from __future__ import division
+import numpy as np
+from caffe import layers as L
+
+PASS_THROUGH_LAYERS = ['AbsVal', 'ReLU', 'PReLU', 'Dropout', 'LRN', 'Eltwise',
+        'BatchNorm', 'BNLL', 'Log', 'Exp', 'MVN', 'Power', 'Sigmoid', 'Split',
+        'TanH', 'Threshold']
+
+def conv_params(fn):
+    params = fn.params.get('convolution_param', fn.params)
+    axis = params.get('axis', 1)
+    ks = np.array(params['kernel_size'], ndmin=1)
+    dilation = np.array(params.get('dilation', 1), ndmin=1)
+    assert len({'pad_h', 'pad_w', 'kernel_h', 'kernel_w', 'stride_h',
+        'stride_w'} & set(fn.params)) == 0, \
+                'cropping does not support legacy _h/_w params'
+    return (axis, np.array(params.get('stride', 1), ndmin=1),
+            (ks - 1) * dilation + 1,
+            np.array(params.get('pad', 0), ndmin=1))
+
+class UndefinedMapException(Exception):
+    pass
+
+def coord_map(fn):
+    if fn.type_name in ['Convolution', 'Pooling', 'Im2col']:
+        axis, stride, ks, pad = conv_params(fn)
+        return axis, 1 / stride, (pad - (ks - 1) / 2) / stride
+    elif fn.type_name == 'Deconvolution':
+        axis, stride, ks, pad = conv_params(fn)
+        return axis, stride, (ks - 1) / 2 - pad
+    elif fn.type_name in PASS_THROUGH_LAYERS:
+        return None, 1, 0
+    elif fn.type_name == 'Crop':
+        axis = fn.params.get('axis')
+        return axis, 1, - fn.params['crop']
+    else:
+        raise UndefinedMapException
+
+class AxisMismatchException(Exception):
+    pass
+
+def compose((ax1, a1, b1), (ax2, a2, b2)):
+    if ax1 is None:
+        ax = ax2
+    elif ax2 is None or ax1 == ax2:
+        ax = ax1
+    else:
+        raise AxisMismatchException
+    return ax, a1 * a2, a1 * b2 + b1
+
+def inverse((ax, a, b)):
+    return ax, 1 / a, -b / a
+
+def coord_map_from_to(top_from, top_to):
+    # We need to find a common ancestor of top_from and top_to.
+    # We'll assume that all ancestors are equivalent here (otherwise the graph
+    # is an inconsistent state (which we could improve this to check for)).
+    # For now use a brute-force algorithm.
+
+    # walk back from top_from, keeping the coord map as we go
+    from_maps = {top_from: (None, 1, 0)}
+    frontier = {top_from}
+    while frontier:
+        top = frontier.pop()
+        try:
+            for bottom in top.fn.inputs:
+                from_maps[bottom] = compose(from_maps[top], coord_map(top.fn))
+                frontier.add(bottom)
+        except UndefinedMapException:
+            pass
+
+    # now walk back from top_to until we hit a common blob
+    to_maps = {top_to: (None, 1, 0)}
+    frontier = {top_to}
+    while frontier:
+        top = frontier.pop()
+        if top in from_maps:
+            return compose(to_maps[top], inverse(from_maps[top]))
+        try:
+            for bottom in top.fn.inputs:
+                to_maps[bottom] = compose(to_maps[top], coord_map(top.fn))
+                frontier.add(bottom)
+        except UndefinedMapException:
+            continue
+
+    # if we got here, we did not find a blob in common
+    raise RuntimeError, 'Could not compute map between tops; are they connected ' \
+            'by spatial layers?'
+
+def crop(top_from, top_to):
+    ax, a, b = coord_map_from_to(top_from, top_to)
+    assert (a == 1).all(), 'scale mismatch on crop (a = {})'.format(a)
+    assert (b <= 0).all(), 'cannot crop negative width (b = {})'.format(b)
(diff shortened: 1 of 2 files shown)


longjon:
By the way (this should become evident when tests are added): basic usage is like this:
```python
from caffe.coord_map import crop
from caffe import layers as L

data = L.Data(...)
...
output = L.Convolution(...)
cropped_output = crop(output, data)
```

python/caffe/coord_map.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import division
+import numpy as np
+from caffe import layers as L
+
+PASS_THROUGH_LAYERS = ['AbsVal', 'ReLU', 'PReLU', 'Dropout', 'LRN', 'Eltwise',
+        'BatchNorm', 'BNLL', 'Log', 'Exp', 'MVN', 'Power', 'Sigmoid', 'Split',
+        'TanH', 'Threshold']
+
+def conv_params(fn):
+    params = fn.params.get('convolution_param', fn.params)
```
seanbell: If `fn.type_name == 'Pooling'`, then this should check `pooling_param` I think

AmandaYingYiWu24:
How to resolve the error message "F0131 23:28:03.316512   929 compute_image_mean.cpp:120] This tool requires OpenCV; compile with USE_OPENCV."?

python/caffe/coord_map.py
shelhamer: A reminder to one day settle #1318

longjon: Not quite; perhaps this warrants a comment. What's going on here is that the parameters are normally read from `fn.params`, which contains the conv params for a conv layer and the pooling params for a pooling layer. However, for layers which _share_ the `ConvolutionParameter` message type (i.e., currently only deconv layer), we need to explicitly ask for `convolution_param`, since `Convolution` is not the name of the layer.

SvenTwo:
Does this replace the "crop" layer types used in some fully convolutional networks? How do you perform training of fully convolutional networks if there is no crop layer to bring the per-pixel labels and the network output into the same coordinate space? Will you have to do the cropping manually as a pre-processing step on the training data?

longjon:
@SvenTwo: #3570 becomes the new `Crop` layer; this code replaces #1975 in order to automatically determine crop parameters. There is no regression of functionality compared to any previous branches; see details above.

ahundt:
@longjon will #3570 and #3613 together provide the functionality in https://github.com/longjon/caffe/? I created https://github.com/longjon/caffe/pull/11 to keep things going until the requisite parts are done. Also, is this close to merging? Things seem to have stalled in both for a month or so. Apologies if this is the wrong place to post this.

python/caffe/coord_map.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import division
+import numpy as np
+from caffe import layers as L
+
+PASS_THROUGH_LAYERS = ['AbsVal', 'ReLU', 'PReLU', 'Dropout', 'LRN', 'Eltwise',
```
shelhamer: + ELU

shelhamer:
@ahundt right, python coord map #3613/this and crop layer #3570 together deliver the same functionality and efficiency of `longjon/caffe:future` but with less intrusion into the framework and layers.

ahundt:
@shelhamer thanks! Also are these two patches applied to master compatible with the exiting trained .caffemodel and .prototext of the version in https://github.com/longjon/caffe?

ahundt:
The answer to my prior question is that yes, they are compatible.

From the model zoo:
 - fcn-32s runs with expected output
 - fcn-8s crashes [Here are the relevant logs](https://gist.github.com/ahundt/8bb0cc0396f136eee9f6)


printed output in python:
```
I0229 14:37:44.157938 2077495296 layer_factory.hpp:77] Creating layer fuse
I0229 14:37:44.157945 2077495296 net.cpp:91] Creating Layer fuse
I0229 14:37:44.157949 2077495296 net.cpp:425] fuse <- upscore2_upscore2_0_split_1
I0229 14:37:44.157953 2077495296 net.cpp:425] fuse <- score-pool4c
I0229 14:37:44.157958 2077495296 net.cpp:399] fuse -> score-fused
F0229 14:37:44.157968 2077495296 eltwise_layer.cpp:34] Check failed: bottom[i]->shape() == bottom[0]->shape() 
```

key lines of stack trace:
```
7   libglog.0.dylib               	0x000000011fa91015 google::LogMessageFatal::~LogMessageFatal() + 15
8   libglog.0.dylib               	0x000000011fa8e363 google::LogMessageFatal::~LogMessageFatal() + 9
9   libcaffe.so.1.0.0-rc3         	0x000000011acf1f4a caffe::EltwiseLayer<float>::Reshape(std::__1::vector<caffe::Blob<float>*, std::__1::allocator<caffe::Blob<float>*> > const&, std::__1::vector<caffe::Blob<float>*, std::__1::allocator<caffe::Blob<float>*> > const&) + 202
10  libcaffe.so.1.0.0-rc3         	0x000000011ad466b9 caffe::Net<float>::Init(caffe::NetParameter const&) + 3449
11  libcaffe.so.1.0.0-rc3         	0x000000011ad47de6 caffe::Net<float>::Net(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, caffe::Phase, caffe::Net<float> const*) + 454
```

longjon:
@ahundt, no (unless @shelhamer has made updates I'm not aware of), existing prototxts are not compatible with these two PRs. They will probably run with `Crop` layers doing nothing, so the FCN-8s dimension mismatch is exactly as expected.

shelhamer:
@ahundt @longjon no, existing models are not compatible but we plan to bundle up nets in a standard format once master is sorted out. It's only an architectural/configuration difference however and the same weights will be fine.

shelhamer:
With this PR @ 775a5c7 and #3570 I have been able to reproduce FCN experiments. Once there are pytests for coord map this can merge.

ahundt:
@shelhamer Is it possible to post new or updated pre-trained models using this code in the model zoo?

weiliu620:
Does pycaffe support 'Crop' layer? I run caffe.Net to load model

net = caffe.Net(model_file, model_weights, caffe.TEST)

and got error: 

F0419 10:14:49.936854 10131 layer_factory.hpp:77] Check failed: registry.count(type) == 1 (0 vs. 1) Unknown layer type: Crop (known types: AbsVal, Accuracy, ArgMax, BNLL, Concat, ContrastiveLoss, Convolution, Data, Deconvolution, Dropout, DummyData, Eltwise, EuclideanLoss, Exp, Flatten, HDF5Data, HDF5Output, HingeLoss, Im2col, ImageData, InfogainLoss, InnerProduct, LRN, MVN, MemoryData, MultinomialLogisticLoss, Pooling, Power, ReLU, Sigmoid, SigmoidCrossEntropyLoss, Silence, Slice, Softmax, SoftmaxWithLoss, Split, TanH, Threshold, WindowData)

PR #3619, opened 2016-02-01
Dynamic linking bugfix

### LINKED ISSUE #3459: Python fails when import caffe, Library not loaded: @rpath/libcaffe.so
OSX 10.11.12, cud 7.5, make runtest succeeded.
When I try to import caffe in python, got error

```
  File "<stdin>", line 1, in <module>
  File "caffe/__init__.py", line 1, in <module>
    from .pycaffe import Net, SGDSolver, NesterovSolver, AdaGradSolver, RMSPropSolver, AdaDeltaSolver, AdamSolver
  File "caffe/pycaffe.py", line 13, in <module>
    from ._caffe import Net, SGDSolver, NesterovSolver, AdaGradSolver, \
ImportError: dlopen(caffe/_caffe.so, 2): Library not loaded: @rpath/libcaffe.so
  Referenced from: /Users/kiwi/Developer/caffe/distribute/python/caffe/_caffe.so
  Reason: image not found
```

Already tried:
1. set DYLD_LIBRARY_PATH
Not working, and import procedure will pend

2. check python path in makeconfig file, and my system python & anaconda python have the same version

3. hdf5 installed via anaconda not brew.

4. Turned off System Integrity Protection

5. Tried otool -L caffe/libcaffe.so/test_all.testbin, all linked right to lib*.7.5.dylib

### LINKED ISSUE #3597: libcaffe.so.1.0.0-rc3 problem
is this a problem?
ImportError: dlopen(caffe/_caffe.so, 2): Library not loaded: libcaffe.so.1.0.0-rc3

zfogg:
Confirmed working for both OS X (El Capitan) and Linux (Ubuntu LTS)

This fixes #3597 and fixes #3459.

About the Linux dynamic linking setup.. perhaps it would be better practice if we added libcaffe to the OS's dynamic library cache? We could do something like this:

```bash
~/src/caffe$ ldconfig -r $HOME -p ./build/lib
~/src/caffe$ ldconfig -p | grep caffe
        libcaffe.so.1.0.0-rc3 (libc6,x86-64) => /home/ubuntu/src/caffe/build/lib/libcaffe.so.1.0.0-rc3
        libcaffe.so (libc6,x86-64) => /home/ubuntu/src/caffe/build/lib/libcaffe.so
```

So, it works well right now using `ld.so's` [magic '$ORIGIN' variable](http://linux.die.net/man/8/ld.so), but maybe switching to `ldconfig` is the right move. Any preference? Let me know! :)


lukeyeager:
Your fix doesn't seem to work on TravisCI (Ubuntu 12.04).

This comment might be relevant/helpful:
> I'm not sure how the linking bug was introduced, but running the following two lines after make alland make test fixes the problem for good:
```
install_name_tool -change libcaffe.so.1.0.0-rc3 @executable_path/../lib/libcaffe.so.1.0.0-rc3 .build_release/tools/caffe
install_name_tool -change libcaffe.so.1.0.0-rc3 @executable_path/../lib/libcaffe.so.1.0.0-rc3 .build_release/test/test_all.testbin
```
@safrooze https://github.com/BVLC/caffe/issues/3597#issuecomment-178898935

zfogg:
hacking away at the issue.. p sure I know what the problem is; please hold and I'll push a fix :)

shelhamer:
Thanks for taking a stab at a fix @zfogg but this is addressed by smaller patches in #3695 and #3696. If there are further cmake changes you would like to contribute please open a separate PR.

PR #3623, opened 2016-02-02
IP layer (GPU): use gpu_data() instead of cpu_data() for bias_mul

kashefy:
The cuda code for IP layer's Forward_gpu() was calling bias_multiplier_.cpu_data() instead of calling gpu_data(). Switched it to gpu_data().

### DIFF
--- a/src/caffe/layers/inner_product_layer.cu
+++ b/src/caffe/layers/inner_product_layer.cu
@@ -16,7 +16,7 @@ void InnerProductLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
     caffe_gpu_gemv<Dtype>(CblasNoTrans, N_, K_, (Dtype)1.,
                          weight, bottom_data, (Dtype)0., top_data);
     if (bias_term_)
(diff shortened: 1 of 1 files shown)


jeffdonahue:
Thanks @kashefy, but `cpu_data()` is correct here as the memory is directly accessed (by the `[0]`).  Changing to `gpu_data()[0]` should cause a segfault.

seanbell:
@jeffdonahue why not use `1` here?  By calling both `cpu_data` and `gpu_data` each pass, the layer unnecessarily copies the vector of `1`s to and from the GPU.

seanbell:
@jeffdonahue why not use `1` here?  By calling both `cpu_data()` and `gpu_data()` each pass, the layer unnecessarily copies the vector of `1`s to and from the GPU.

jeffdonahue:
Yes, using 1 would be fine and probably saves a couple of cycles. But there shouldn't be any extra copying since there is no `mutable_*` calls -- the copy only occurs on a first call (mutable or not) to an accessor on the opposite device from the last `mutable_*` call.  So since the only mutable call is on CPU in reshape, calling `cpu_data` should never trigger a copy.

PR #3632, opened 2016-02-04
Precise accuracy calculation.

BlGene:
Add the option to return the counts that contribute to an accuracy score.

Why do we need this?
Together with PR #3631 this adds the options of precise accuracy calculations (e.g. over the entire test set.) It fixes the problem that during the test phase accuracies are averaged without respecting `counts` variable between batches, withing one test. An equivalent thing happens during train when `iter_size` is set.

This may only be needed in certain cases, e.g. a high number or ignore_labels in the dataset, different image sizes during segmentation, e.c.t. A minimal solution to the problem is to let the user do the accuracy accumulation ( respecting weighting ) using a simple layer. These two pull requests
add the necessary functionality to easily implement this layer.

Here is also a python layer that does the accumulation respecting counts.
```
# A layer that accumulates outputs and outputs weighted mean

# The use case is for this layer to accumulate values over epochs,
# hoever "epoch" can be an arbitrary number of iterations.
# To make sure that you see the ouput of this layer check that
# iter_per_epoch % display == 0
class OutputAccum(caffe.Layer):
    def setup(self, bottom, top):
        assert(len(top) == len(bottom))
        params = json.loads(self.param_str)

        # How many iteration to collect before printing output
        self.iter_per_epoch = params["iter_per_epoch"]

        # Offset self.iteration to -1 so that it starts with 1
        self.iteration = 0
        self.epoch = 0
        self.accum = ([]*len(bottom),)
        self.print_now = False

        self.top_shape = []
        for i in range(len(bottom)):
            blob_shape = list(bottom[i].data.shape)
            blob_shape[-1] -= 1
            self.top_shape.append(blob_shape)

        # Reshape in forward, we need to do this since only here do we
        # know what the iteration number is, hence if we should output
        # This is quite off label use, but I don't know where this would cuase
        # truble
        pass

    def forward(self,bottom,top):

        # see reshape function
        for i in range(len(self.top_shape)):
            top[i].reshape(*self.top_shape[i])

        if self.iteration % self.iter_per_epoch == 0:
            self.print_now = True
        else:
            self.print_now = False
            top[0].reshape(0)
        # I'll behave now

        for i in range(len(bottom)):
            #Make sure to append a copy there
            self.accum[i].append(bottom[0].data.copy())

        if  self.print_now:
            #make sure we print exactly when we
            accum = np.array(self.accum)
            values = accum[:,:,:-1]
            weights = accum[:,:,-1]
            for i in range(accum.shape[0]):
                means = np.average(values[i],axis=0,weights=weights[i])
                top[i].data[...] = means

            self.print_now = False
            self.print_next = False
            self.accum = ([]*len(bottom),)

        self.iteration += 1
``` 

Example layer specification
```
layer {
  type: "Python"
  bottom: "acc"
  top: "epoch_acc"
  python_param {
    module: "pylayer"
    layer: "OutputAccum"
    param_str: '{"iter_per_epoch":18}'
  }
  #include {
  #  phase: TRAIN
  #}
}
```

### DIFF
--- a/include/caffe/layers/accuracy_layer.hpp
+++ b/include/caffe/layers/accuracy_layer.hpp
@@ -86,6 +86,9 @@ class AccuracyLayer : public Layer<Dtype> {
   bool has_ignore_label_;
   /// The label indicating that an instance should be ignored.
   int ignore_label_;
+  /// Whether to return count of valid entries
+  bool return_count_;
+
   /// Keeps counts of the number of samples per class.
   Blob<Dtype> nums_buffer_;
 };
--- a/src/caffe/layers/accuracy_layer.cpp
+++ b/src/caffe/layers/accuracy_layer.cpp
@@ -17,6 +17,7 @@ void AccuracyLayer<Dtype>::LayerSetUp(
   if (has_ignore_label_) {
     ignore_label_ = this->layer_param_.accuracy_param().ignore_label();
   }
+  return_count_ = this->layer_param_.accuracy_param().return_count();
 }
 
 template <typename Dtype>
@@ -34,7 +35,11 @@ void AccuracyLayer<Dtype>::Reshape(
       << "label count (number of labels) must be N*H*W, "
       << "with integer values in {0, 1, ..., C-1}.";
   vector<int> top_shape(0);  // Accuracy is a scalar; 0 axes.
+  if (return_count_) {
+    top_shape = vector<int>(1, 2);
+  }
   top[0]->Reshape(top_shape);
+
   if (top.size() > 1) {
     // Per-class accuracy is a vector; 1 axes.
     vector<int> top_shape_per_class(1);
@@ -92,6 +97,9 @@ void AccuracyLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
 
   // LOG(INFO) << "Accuracy: " << accuracy;
   top[0]->mutable_cpu_data()[0] = accuracy / count;
+  if (return_count_) {
+    top[0]->mutable_cpu_data()[1] = count;
+  }
   if (top.size() > 1) {
     for (int i = 0; i < top[1]->count(); ++i) {
       top[1]->mutable_cpu_data()[i] =
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -464,6 +464,9 @@ message AccuracyParameter {
 
   // If specified, ignore instances with the given label.
   optional int32 ignore_label = 3;
+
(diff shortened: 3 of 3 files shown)


BlGene:
In case this correctly identified a problem, do we have the same situation when averaging loss between batches when 'iter_size' is set?

svobora:
This is a BUG, since 

1) the accuracy is computed wrong when using ignore_label (I couldnt believe someone actually did this bug when I first seen the code.) in the first place.
2) (follows from (1)) the accuracy can become NAN in case all data is ignored in one of the test_iter. (Happens in multi-loss setup for example, when one of the loss layer is fed with only ignore class.)

The only possibility I see is to change the retval of AccuracyLayer from accuracy/count to accuracy and count and aggregate the result after all test_iter batches.

BlGene:
@svobora 

BUG is such a harsh word ;), its just a small inconsistency stemming from incremental feature addition.

Actually I was of the same opinion and started to try to fix by changing the retval of AccuracyLayer, solver for Tests, ect. see: https://github.com/BlGene/caffe/tree/precise_accuracy. However I noticed that this would be quite an intrusive change and stopped. Strictly speaking with no valid data the accuracy *should* be NaN. If we have NaN loss this is potentially more problematic, maybe the solver could just ignore it, in case it does not so already?

Since this will only affect a small minority of users I find the idea of managing this inside the network(user space) better because it is less intrusive and more flexible.

svobora:
@BIGene

Thank you for the reply. 

The problem here is that accuracy is dependent on the order of data. In my case, I have like 5% of data with one of the labels missing. For measurement of accuracy, I process batches of size 512 over ~200 test_iter to cover my whole test set of ~100k data. Now, if only a single of those 200 batches consists of all ignore labels, I have NaN accuracy of the given batch (that is correct) but also of the whole test set (that is not correct).

BlGene:
@svobora OK, technicality a bug in that case, but it could still be fixed in an un-intrusive way by having the solver not average NaN values*. The intrusive ( always passing counts to solver ) way maybe it would be a bit cleaner, but its up to the devs to decide whats better.

For now I would encourage you to circumvent this problem via your network as this PR suggests.

*In that case the accuracy layer wold return the mean mini-batch accuracy instead of the mean dataset accuracy. If its not computational too expensive set `test_iter:1` to get mean dataset accuracy.

jenifferYingyiWu:
How to test the images in val.txt, and calculate the precistion and recall?

BlGene:
I am closing this PR the trend has shifted to do evaluation in python. In case you are still evaluating with the method built in to the solver, please be aware of this issue. In most cases it probably won't matter.

PR #3638, opened 2016-02-05
explicit install of libgfortran via conda

kashefy:
Nose tests have been failing with import errors due to not linking to gfortran libs.
This is strange because travis installs gfortran via apt-get.
Having conda install libgfortran resolves the problem.


kashefy:
Looks like the linking errors are gone. Can't tell why, but doesn't look like this is needed any more.

PR #3665, opened 2016-02-11
Add support for windows build

pavlejosipovic:
Necessary changes to enable widndows build:
Visual studio project files.
Minor code chnages.
3rd party dependencies are resolved through NuGet.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,63 @@
+###############################################################################
+# Set default behavior to automatically normalize line endings.
+###############################################################################
+* text=auto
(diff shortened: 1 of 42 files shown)


olesalscheider:
Why do we need to add Visual studio project files? Can't we just generate them with CMake?

There are already two build systems and IMO adding another one is a bad idea since it will be difficult to keep them all in sync.

pavlejosipovic:
Experimental Windows branch was created and all the changes from PR are integrated there.

PR #3676, opened 2016-02-16
Remove useless LevelDB include

flx42:
The tests could not compile with USE_LEVELDB=0 and LevelDB missing from the system


flx42:
Obviously this include is missing a `#ifdef USE_LEVELDB`, but as far as I can tell, this include is just not needed.
@shelhamer: the tests are passing, should be a no-brainer (even if Travis doesn't test all the possible build combinations).

jeffdonahue:
Thanks @flx42

PR #3696, opened 2016-02-20
Fix Library Install name on OS X for @rpath Linking

shelhamer:
For linking of the caffe tools, tests, etc. the library install name needs to include the @rpath set for the executables and interfaces.

@lukeyeager this should fix a number of the OSX complaints since #3311 -- sorry for not catching it then.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -283,9 +283,8 @@ ifeq ($(OSX), 1)
 	# boost::thread is called boost_thread-mt to mark multithreading on OS X
 	LIBRARIES += boost_thread-mt
 	# we need to explicitly ask for the rpath to be obeyed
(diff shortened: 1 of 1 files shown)


lukeyeager:
:+1: Thanks for figuring this out!

DaveRichmond-:
Hi,

I'm trying to install Caffe on OSX 10.11, and having the same linking issues as many other people.  I can see that this is the solution, but I'm afraid I don't know how to implement it.  I see that there is a commit associated with this -- has it been fixed in Caffe?

I still get the error (dyld: Library not loaded: libhdf5_hl.10.dylib) when making runtest, or if I try to import caffe into python.  Otherwise everything seemed to build fine (only warnings).

Apologies for my confusion.  Any suggestions would be much appreciated.

Thanks,
Dave

PR #3703, opened 2016-02-23
Supporting Blobs with num_axes() != 4 in pycaffe interface

shaibagon:
Currently, Blobs (especially inputs to net) with `num_axes() != 4` are not supported by pycaffe interface. 
This is because pycaffe uses `.num` property of `Blob` that is only supported for 4D Blobs.

This PR attempts to correct this by removing all references to `Blob.num` property (that assumes Blob is 4D). 
Replacing it with accessing `Blob.shape[0]` that is good also for Blobs with `num_axes() != 4`

**This is a very minor change code-wise**

### DIFF
--- a/python/caffe/pycaffe.py
+++ b/python/caffe/pycaffe.py
@@ -98,7 +98,7 @@ def _Net_forward(self, blobs=None, start=None, end=None, **kwargs):
         # Set input according to defined shapes and make arrays single and
         # C-contiguous as Caffe expects.
         for in_, blob in kwargs.iteritems():
-            if blob.shape[0] != self.blobs[in_].num:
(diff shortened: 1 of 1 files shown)


shelhamer:
Thanks for this small patch to make the pycaffe keyword args work.

More precisely, this is only an issue for `num_axes > 4` and when passing `blobs=` or `diffs=` keyword args to the pycaffe `forward(), backward()` respectively. The pycaffe blob interface, the `Net.{blobs,params}` properties, and `Net.{forward(),backward()}` methods all work in N-D.

shaibagon:
@shelhamer 
Thanks for accepting this PR.

PR #3715, opened 2016-02-24
CMake: Do not include "${PROJECT_BINARY_DIR}/include" with SYSTEM option

olesalscheider:
This is important for the include order. Without this patch, a
previously installed caffe.pb.h might be included instead of the one
that is generated during the build.


shelhamer:
@jczaja @Nerei or anyone else familiar with CMake, please check this for merge.

Nerei:
Yes. that's right patch. I comes from copypast and that I didn't take this into account that caffe may be installed to system paths.

LTGM

PR #3716, opened 2016-02-24
Use six library to ensure pycaffe.py python3 compliance

ttdt:
By replacing _itervalues_, _next_ and _iteritems_ with six library functions, pycaffe.py is now python3 compliant.

### DIFF
--- a/python/caffe/pycaffe.py
+++ b/python/caffe/pycaffe.py
@@ -14,6 +14,8 @@ from ._caffe import Net, SGDSolver, NesterovSolver, AdaGradSolver, \
         RMSPropSolver, AdaDeltaSolver, AdamSolver
 import caffe.io
 
+import six
+
 # We directly update methods from Net here (rather than using composition or
 # inheritance) so that nets created by caffe (e.g., by SGDSolver) will
 # automatically have the improved interface.
@@ -97,7 +99,7 @@ def _Net_forward(self, blobs=None, start=None, end=None, **kwargs):
             raise Exception('Input blob arguments do not match net inputs.')
         # Set input according to defined shapes and make arrays single and
         # C-contiguous as Caffe expects.
-        for in_, blob in kwargs.iteritems():
+        for in_, blob in six.iteritems(kwargs):
             if blob.shape[0] != self.blobs[in_].shape[0]:
                 raise Exception('Input is not batch sized')
             self.blobs[in_].data[...] = blob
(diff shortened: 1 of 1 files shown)


python/caffe/pycaffe.py
```diff
@@ -256,10 +259,10 @@ def _Net_batch(self, blobs):
     ------
     batch: {blob name: list of blobs} dict for a single batch.
     """
-    num = len(blobs.itervalues().next())
-    batch_size = self.blobs.itervalues().next().num
+    num = len(next(itervalues(blobs)))
+    batch_size = next(itervalues(self.blobs)).num
     remainder = num % batch_size
-    num_batches = num / batch_size
+    num_batches = int(num / batch_size)
```
seanbell: You can use `//` to get integer divide for both `python2` and `python3`.

ttdt:
Use // for integer division (python2 and python3 compliant)

ttdt:
CI build fails because of random 404 on package download from repo.continuum.io: https://repo.continuum.io/pkgs/free/linux-64/setuptools-20.1.1-py35_0.tar.bz2
I'll try again (close/reopen PR) in a few hours, hoping the bug will be fixed.

python/caffe/pycaffe.py
```diff
@@ -14,6 +14,9 @@
         RMSPropSolver, AdaDeltaSolver, AdamSolver
 import caffe.io
 
+# py3 compliance
+from six import itervalues, next, iteritems
```
longjon: I guess I'd rather we make our use of `six` explicit by just doing `import six` here... I think that's how we already do it in other files, and `six` is pretty short, and I think it's easier to read that way.

You can also drop the comment, IMO... we all know what `six` does, and anyone who is not familiar with it can find out easily

longjon:
Looks good except as noted. We should really have tests that cover these cases...

longjon:
Great, the diff looks good to me now. If you could clean up the history by squashing into a single commit, this will be ready for merge.

ttdt:
PR #3746 opened with a clean history.

shelhamer:
@ttdt in the future do not close the PR, but instead open the PR by force pushing to your branch. Making new PRs adds noise to the tracker.

longjon:
Thanks for the Python 3 update, merging!

(By the way, I think you mean "integer division" or "floor division" in the commit message (but it's not worth fixing, just a note for the future)).

PR #3541, opened 2016-01-12
Fixed pycaffe build for Python 3.4 on OSX

Austriker:
Hi,

I have been working on using pycaffe with python 3.4. I tried using Homebrew for boost and protobuf but with no luck.
I am on a MacBook Air under OSX 10.11.2

I added an option in Cmake to select openblas instead of vecLib since I was having an error : 

``` sh
./caffe/include/caffe/util/mkl_alternate.hpp:11:10: fatal error: 'cblas.h' file not found
```

I installed openblas with Homebrew.

I updated the pip3 requirements.txt since one of them wasn't working with python 3.4.

For the build, I had to build boost and protobuf : 
#### Boost : 1.60

``` sh
./bootstrap.sh --with-python=/Library/Frameworks/Python.framework/Versions/3.4/bin/python3.4
```

You have to modify the project-config.jam : 

```
    using python : 3.4 : /Library/Frameworks/Python.framework/Versions/2.7 ;
```

to : 

```
    using python : 3.4 : /Library/Frameworks/Python.framework/Versions/3.4 ;
```

Then : 

``` sh
./b2 toolset=clang cxxflags="-stdlib=libc++ -std=c++11" linkflags="-stdlib=libc++ -std=c++11" -j4 install
```

But I still had an issue with the python includes path so I added an user config file ~/user-config.jam : 

```
using python
    : 3.4
    : /Library/Frameworks/Python.framework/Versions/3.4/bin/python3.4
    : /Library/Frameworks/Python.framework/Versions/3.4/include/python3.4m
    : /Library/Frameworks/Python.framework/Versions/3.4/lib
    ;
```
#### Protobuf : 3.0b2

I dowloaded the source from the github repo and followed the build instructions. 
For the python package I followed the instructions here : https://github.com/google/protobuf/tree/master/python

Finally to build caffe I did : 

``` sh
mkdir build
cmake -DCPU_ONLY=ON -DCMAKE_BUILD_TYPE=Release -Dpython_version=3 -DBLAS=Open ..
make -j6 all
make test
make install
```

The build works fine and the command `python3 -c "import caffe"` works.

Best regards

PR #3720, opened 2016-02-24
Fixed Makefile/CMake for Mac OS X

cepiross:
Hi,
I have used caffe on OS X El Capitan (10.11.3). I slightly changed Makefile for supporting ViennaCL without CUDA as follows:
* Macbook Pro 13' (2015) provides Intel Iris Graphics 6100, which can be used with OpenCL
* In addition, Apple LLVM does not currently support openmp

In the meantime, I modified FindvecLib.cmake to avoid locating wrong path from obsolete vecLib.h, which is reported from #3541.

Based on this modification, Caffe build is successful on OS X El Capitan by configuring either makefile.config or CMake.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,8 @@ ifeq ($(CPU_ONLY),1)
 	USE_GREENTEA := 0
 endif
 
-CXXFLAGS += -std=c++11 -fopenmp -Wno-deprecated-declarations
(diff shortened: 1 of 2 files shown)


cepiross:
[Makefile.config.txt](https://github.com/BVLC/caffe/files/145639/Makefile.config.txt) is an example of Makefile.config for Mac OS X El Capitan.

naibaf7:
Works as expected. Will merge.

PR #3732, opened 2016-02-26
[WIP] Allow for pooling over the first blob axis based on a key

elezar:
Internally, we have found that treating images as a collection and pooling over them results in improved performance for some tasks.

In an attempt to generalize this, I have started work on a key-based pooling layer. The idea is that in addition to the bottom blob representing the input data, a second blob is provided that indicates the keys associated with each of the entries in the batch (first dimension). An additional pooling step is then used to pool over all images withe the same key.

This can be seen as something similar to a reduce-by-key operation.

Although I am implementing this as an additional layer (which has a pooling layer member), there are some open questions which may indicate that it would be worth it to extend the pooling layer itself. As the initial proof of concept, I am performing the MAX pooling over the keys as a hard-coded step. Ideally, this should be handled by the pooling layer itself. Does it make sense to add further parameters to the pooling layer that handles these options? Is it feasible to pool over color channels, for example?

### DIFF
--- a/include/caffe/layers/key_pooling_layer.hpp
+++ b/include/caffe/layers/key_pooling_layer.hpp
@@ -52,10 +52,7 @@ class KeyPoolingLayer : public Layer<Dtype> {
   vector<Dtype> has_keys_;
   // Store the start and end indices for each key in the input array.
   vector<int> key_start_;
-  vector<int> key_end_;
-
-  int largest_key_set_;
-
+  vector<int> key_len_;
 
 };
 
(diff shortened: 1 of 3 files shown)


elezar:
I have a minimal working (CPU) implementation for MAX pooling over keys. Any feedback would be appreciated.

I still need to improve the generality of the layer with regards to other pooling parameters, but before then I have some questions with regards to MAX pooling. Here, the indices of the maximum elements are returned in a second top blob (or stored internally). One thing that is a bit problematic with regards to this pull request is that index returned is always per-channel. Does it not make sense to change this so that the indices are always global (w.r.t. a batch)? I would assume that this is required for Nd pooling, and also seems to be in keeping with other pooling implementations (such as https://www.tensorflow.org/versions/r0.7/api_docs/python/nn.html#max_pool_with_argmax). For the time being, I am able to work around this by adjusting the indices manually, after calling the internal pooling layer, but this really is not ideal.

elezar:
Closing pull request due to lack of interest/time.

PR #3742, opened 2016-02-27
Add new orthogonal weight filler (initializer)

mattdawkins:


ducha-aiki:
#3548 
But c++ is better, of course :)

mattdawkins:
Yeah, the implementation seems to be about the same (I think) though with the C++ version it can be used with the caffe train tool instead of just python. Only downside is the OCV dependency.

seanbell:
:+1:  Thanks for the PR; I've had to write a similar filler in the past.

PR #3747, opened 2016-02-29
Little fix for NetSpec: allow setting top blob names using string

Nerei:
Some top blob names (ex. in GoogleNet) contain symbols that can't be used as python variable name. 

```python
n = caffe.NetSpec()
n.conv1/7x7_s2 = L.Convolution(n.data, name='conv1/7x7_s2', ...) 
```
This patch allows the following:
```python
n = caffe.NetSpec()
n['conv1/7x7_s2'] = L.Convolution(n.data, name='conv1/7x7_s2', ...) 
```
The patch is just a simple workaround that doesn't break original NetSpec style to specify top names

### DIFF
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -175,6 +175,12 @@ class NetSpec(object):
     def __getattr__(self, name):
         return self.tops[name]
 
+    def __setitem__(self, key, value):
+        self.__setattr__(key, value)
+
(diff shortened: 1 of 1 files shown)


shelhamer:
Right, the dictionary style NetSpec is definitely useful in these cases and we've done this in side branches without trouble. Thanks for the patch!

PR #3749, opened 2016-02-29
Don't force datum.label=0 in array_to_datum

lukeyeager:
`datum.label=None` is preferable to `datum.label=0`.


lukeyeager:
Ping? I feel like the odds of breaking existing workflows should be pretty low. If they're not specifying a label, then I can't see why they would be reading it later.

seanbell:
LGTM.

@shelhamer

shelhamer:
Looks good. Thanks for this and the IO care in general @lukeyeager, and thanks for signaling this @seanbell.

Issue #3750, opened 2016-02-29
upgrade_net_proto_binary segfaults


dieterichlawson:
`upgrade_net_proto_binary` segfaults when upgrading certain caffemodel files.

Steps to reproduce:

1) Download VGG16 from model zoo
2) run `tools/upgrade_net_proto_binary VGG_ILSVRC_16_layers.caffemodel vgg16.caffemodel`
3) It segfaults, giving the following stacktrace in gdb.
```
libdc1394 error: Failed to initialize libdc1394
[libprotobuf WARNING google/protobuf/io/coded_stream.cc:505] Reading dangerously large protocol message.  If the message turns out to be larger than 2147483647 bytes, parsing will be halted for security reasons.  To increase the limit (or to disable these warnings), see CodedInputStream::SetTotalBytesLimit() in google/protobuf/io/coded_stream.h.
[libprotobuf WARNING google/protobuf/io/coded_stream.cc:78] The total number of bytes read was 553432081
[libprotobuf FATAL /usr/include/google/protobuf/repeated_field.h:613] CHECK failed: (index) < (size()):
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: (index) < (size()):

Program received signal SIGABRT, Aborted.
0x00007ffff63b1cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff63b1cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff63b50d8 in __GI_abort () at abort.c:89
#2  0x00007ffff69b6535 in __gnu_cxx::__verbose_terminate_handler() ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff69b46d6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff69b4703 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff69b4922 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff0af7b58 in google::protobuf::internal::LogMessage::Finish() ()
[...]

Issue #3381, opened 2015-11-24
UpgradeNetAsNeeded should refuse to process nets with mixed versions


seanbell:
The upgrade code (`UpgradeNetAsNeeded`) should refuse to process a net that has a mix of layer versions (some from the old format `layers`, and some from the new format `layer`), and give a message explaining the problem and how to manually fix it.  Otherwise, you end up with some layers being silently deleted (e.g. #3375).

I thought this was already an issue, but I couldn't find it.

PR #3755, opened 2016-03-01
Fix Upgrade Net Tools

shelhamer:
- output logging content from upgrade tools
- check all net upgrade conditions all the time
- fix input upgrade logic for binaryproto/caffemodel #3750
- refuse to process nets with mixed versions #3381

@jeffdonahue any thoughts on the harsher reaction of #3381? I think it is safer and more explicit about the issue.

### DIFF
--- a/src/caffe/util/upgrade_proto.cpp
+++ b/src/caffe/util/upgrade_proto.cpp
@@ -13,7 +13,8 @@
 namespace caffe {
 
 bool NetNeedsUpgrade(const NetParameter& net_param) {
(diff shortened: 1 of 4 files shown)


shelhamer:
I'm going ahead and merging this since a number of users have run into this issue -- whether to die on mixed version definitions can be amended later if we decide against it.

shelhamer:
Follow-ups:

- [ ] add more upgrade tests to `test_upgrade_proto` for data transformer, input
- [ ] test both proto text and proto binary

PR #3756, opened 2016-03-01
MKL support improvments

jczaja:
- Update in documentation on how to use MKL with caffe
- Fix to cmake based Linux build of caffe using MKL eg. cmake *\<path to CMakeLists.txt\> -DBLAS=mkl*


shelhamer:
Thanks for the MKL details.

PR #3757, opened 2016-03-01
Merge master@{2016-03-01} into windows 

pavlejosipovic:


pavlejosipovic:
That means that nobody should fork this branch.
If somebody wants to use caffe as-is on Windows that is OK.

But if they want to develop on top of this on Windows they should stick with Microsoft/caffe?

shelhamer:
Oh sorry, I read this wrong the first time: by the title I thought this was a PR from the windows branch to master. PRs to the BVLC/caffe:windows branch totally make sense. However I'd say that changes from master into windows should be incorporated by rebase, and not merge, as we like to have a rebase workflow to keep history clear and make future merges simpler.

For routine commits to keep up with master I would simply push them to the windows branch. For distinct development in the windows branch I would open PRs to make the changes more obvious, gather comments, and so on.

Does that sound like a good process? Happy to hear suggestions.

Sorry for the confusion.

pavlejosipovic:
I prefer to use rebase on feature branches, that are short lived.
I'm not sure how long it will take before we get windows build into master.

I would like for others to be able to fork this branch and enable them to develop on top of it, without forcing them to rebase each time they pull from bvlc/windows. 

Once we are done with work in windows branch, and we decide to integrate back to master, we can then do the rebase to keep the history clean.

Does this sound reasonable?

pavlejosipovic:
@Shelhamer can please share your thoughts about my proposal above?

shelhamer:
@pavlejosipovic I defer to you for your choice of workflow in the `windows` branch as you will be handling the issues and PRs.

> I would like for others to be able to fork this branch and enable them to develop on top of it, without forcing them to rebase each time they pull from bvlc/windows.

Makes sense.

> Once we are done with work in windows branch, and we decide to integrate back to master, we can then do the rebase to keep the history clean.

Right, the requirements for merge are (1) no conflicts (2) sane history -- that is, no trail of fixup commits like ab158fa at the end, but commits that at least mostly work at each step.

My personal preference and the workflow for `master` is that everyone tracks `BVLC/caffe:master` in their fork, branches from master to do their own work, then sends PRs to `master` after rebasing onto the latest as outlined in the [development guide](http://caffe.berkeleyvision.org/development.html). I dislike non-trivial merge commits since they obscure issues with the real commits, like conflicts, that would be caught by rebase.

Regarding keeping `master` and `windows` in sync, or in general keeping long-lived feature branches, I find it simplest to keep rebasing with the help of [git rerere](https://git-scm.com/docs/git-rerere). However, as you noted, this isn't a workflow for everybody and we don't want to discourage community participation in the `windows` branch. For the routine sync of `windows` to `master`, I'd rather you push the changes directly to `BVLC/caffe:windows` than make a PR unless there is a particular conflict or suchlike that you want to discuss, but if you think it's more visible/useful/whatever so be it (just tag these PRs as "windows").

In conclusion, do what you want as long as the end result is a fast-forward merge of legible commits. Thanks for asking about the workflow details!

PR #3770, opened 2016-03-04
Removed lint script reference to non-existant caffe_memcpy function.

BlGene:


shelhamer:
Thanks for the cleanup (disgregard CI fail due to python3 package issue fixed in #3773).

PR #3771, opened 2016-03-04
Update versions for Travis build

shelhamer:
There are new releases of Miniconda, boost, and protobuf.
I'm partly checking whether recent travis failures like #3613 are
due to a version issue or what.

It seems like Travis is out again due to some CMake or Python 3 issue.


shelhamer:
Closing in favor of @longjon's investigation in #3773.

PR #3773, opened 2016-03-04
Travis: force protobuf 3.0.0b2 for Python 3

longjon:
This is temporary measure to avoid the apparent upstream issue with protobuf 3.0.0b2.post1 which has broken Travis (previously, the most recent available version of the Python package was installed).

This needs to be tested against Travis; do not merge until passing.


longjon:
Merging this simple fix to Travis.

However, we should keep in mind that this patch should be reverted (or switched to a non-beta version) once upstream stabilizes.

PR #3777, opened 2016-03-05
Added automatic installation script for Linux!

hack-r:
This is tested on Ubuntu 15. Just run the script and (py) Caffe is ready
to go for Python 3 using CPU. Add 1 line of code or edit makefile if you
want to use GPU; just change python3 to python and pip3 to pip if you 
want to use  the historical version of Python.


shelhamer:
Sorry, I think these kinds of scripts are better placed in the wiki since they are liable to rot. For packaging we have Debian #5111, Snap #5300, and docker.

PR #3778, opened 2016-03-06
Only check the dimensions start from start_axis

wk910930:
crop_offset may not be suitable or needed for dimensions before start_axis.


MalteOeljeklaus:
+1, I also need this! The present code in the master branch does not allow different numbers of channels for the two input blobs even when only the spatial dimensions are cropped. This does not affect the original FCN models but is obstructive when implementing models like in the hypercolumn paper. I've yet to do more testing but for now this pull request seems to work for me.

shelhamer:
@wk910930 Sorry, you're totally right! Thanks for this and thanks @MalteOeljeklaus for confirming. This was fixed in #3993. Sorry this was lost in the deadline whirlwind of March.

PR #3797, opened 2016-03-09
Fix Travis-CI conda/boost issue

shelhamer:
This simply clears the conda dir before installation.
Thanks to report by @sasagalic-MSFT and recommended workaround by
@anfeng.

### DIFF
--- a/scripts/travis/travis_install.sh
+++ b/scripts/travis/travis_install.sh
@@ -60,6 +60,8 @@ rm -f $LMDB_FILE
 # Install the Python runtime dependencies via miniconda (this is much faster
 # than using pip for everything).
 export PATH=$CONDA_DIR/bin:$PATH
+# clear any cached conda (see #3786)
+rm -rf $CONDA_DIR
 if [ ! -d $CONDA_DIR ]; then
   if [ "$PYTHON_VERSION" -eq "3" ]; then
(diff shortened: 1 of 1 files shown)


lukeyeager:
You could also just stop caching the directory:
https://github.com/BVLC/caffe/blob/rc3/.travis.yml#L17-L20

shelhamer:
@lukeyeager see the doc links here https://github.com/BVLC/caffe/issues/3798#issue-139682670 because I'm confused whether or not it should even be caching. I can't access the caching interface in the travis docs for clearing the cache, either because we're not a private/paying repo or because we're not on the new infrastructure.

lukeyeager:
Oh yeah, you're right.

> https://docs.travis-ci.com/user/caching
The features described here are currently only available for private repositories on travis-ci.com and our new container-based infrastructure.

> Note that APT caching is not available for the container-based infrastructure.

Strange that removing the cache still accomplishes something, though...

> I'd rather not removing caching since in general it seems more efficient, we just want to clear the cache when switching versions.

So are you going to revert this commit later today then? How is what you did different than removing caching?

shelhamer:
> So are you going to revert this commit later today then? How is what you did different than removing caching?

I'd rather keep the travis config intact for now while we figure out #3798 and keep the workaround in a script. No, I'm not going to break it again.

PR #3799, opened 2016-03-09
Tests: fix incorrect order of layer_param.set_phase() and DropoutLayer(layer_param)

AlexeyKruglov:
In TestDropoutForward(...) constructor, layer_param.set_phase(...) function call and DropoutLayer(layer_param) constructor order was mixed.  This is not a big problem by itself, but may become a bug if the code is reused.


AlexeyKruglov:
The failed Travis check has nothing to do with my commit. It says

> CMake Error at /usr/share/cmake-3.2/Modules/FindBoost.cmake:1182 (message):
>   Unable to find the requested Boost libraries.

My commit didn't introduce any new dependencies / includes.

PR #3804, opened 2016-03-11
mnist get data script bug fix

dongzhuoyao:
In some instances,if we use ctrl+c to stop the downloading,the wget tool will automatically  append some string  like '.0','.1' which gunzip cannot open.

so,i add -O option to customize the download file name.


lukeyeager:
See https://github.com/BVLC/caffe/pull/3297

PR #3821, opened 2016-03-15
Update Makefile: Changed MKL_DIR to MKLROOT

jreniecki:
MKLROOT variable should be used by Makefile, because it's set as environment variable by MKL scripts.
Similar change in CMake was already done in #3756


jczaja:
Looks good to me. eg. change is fine!

PR #3827, opened 2016-03-16
Fix a crash when running caffe time utility

jainanshul:
No kernels or only kernel prototypes found when build executable

This fixes naibaf7/caffe#23

@naibaf7 ?


naibaf7:
Thanks a lot. I implemented the fix slightly different. Hopefully it works.

PR #3845, opened 2016-03-17
Add Dockerfile from centos7 image

pedrohml:


seanbell:
1.  Is there an advantage of using centos7 instead of ubuntu14.04 (in the current images here https://github.com/BVLC/caffe/tree/master/docker)?

2.  It looks like this doesn't support CUDA?

elezar:
@pedrohml if we decide to include both Ubuntu and Centos images, I would suggest modifying the folder structure accordingly (see for example https://github.com/NVIDIA/nvidia-docker).

pedrohml:
Sorry, for delay... 
1) Yes, there are advantages. I see centos increasing quickly the number of production servers. I think one big reason for that is to be a cleaner OS than Ubuntu and have a big community support too. A dockerfile could help people to learn how to create dedicated centos servers to run caffe pipelines.

2) I agree. The proposed path wasn't the best one. We can proceed with the nvidia-docker one. 

Please,  keep me updated about the interest of caffe running on centos. I can make the proper adjustments in that case.

shelhamer:
I'd like to keep the bundled Dockerfile on Ubuntu since that's our most common platform. Please feel free to link to alternative Dockerfiles on the wiki though. Thanks for the installation effort.

PR #3848, opened 2016-03-18
Opencl some misc fixes

gongzg:
@naibaf7 
This PR includes some misc fixes. As the FFT domain kernel still has some issues which could not pass all the tests, thus I have to defer it to next week.


naibaf7:
@gongzg 
Nice, thanks!
Note that I added some testing mechanisms to test the compilation of OpenCL kernels to the branch yesterday. Might be handy.

PR #3853, opened 2016-03-20
Fix build in Windows 10

ctapang:
The cmd script file BinPlaceCudaDependencies.cmd fails in Windows 10
without this fix.


ctapang:
Windows script file fixed to allow builds in Windows 10, VS2013.

pavlejosipovic:
@ctapang I think that this issue has been resolved in the meantime, does it work for you know?

PR #3736, opened 2016-02-27
Expose net stages to pycaffe

lukeyeager:
Together with https://github.com/BVLC/caffe/pull/3211 (Input layer), this should solve https://github.com/BVLC/caffe/issues/1245 (all-in-one nets).

Doesn't handle stage exclusion
Doesn't test all possible permutations

PR #3863, opened 2016-03-21
Expose all netstate options (for all-in-one nets)

lukeyeager:
*Was #3736*

There are 3 configurable dimensions to a network's [NetState](https://github.com/BVLC/caffe/blob/rc3/src/caffe/proto/caffe.proto#L257-L261): **phase** (TRAIN or TEST), **level** and **stage**. Currently, only **phase** is exposed through common interfaces. This PR exposes **level** and **stage** to the user, further enabling all-in-one nets (#1245)

New CLI:
```
build/tools/caffe time -model network.prototxt -level 1 -stage deploy
```
New pycaffe interface:
```python
net = caffe.Net('network.prototxt', weights='weights.caffemodel',
  phase=caffe.TEST, level=1, stages=['deploy'])
```

---

*EDIT: PR has been updated not to break old code*

~~This is a **breaking change** (see https://github.com/BVLC/caffe/pull/1790#issuecomment-199420059 for an explanation). Users will need to transform pycaffe code like:~~
```python
# old
net = caffe.Net('network.prototxt', 'weights.caffemodel', caffe.TEST)
```
~~to:~~
```python
# new
net = caffe.Net('network.prototxt', weights='weights.caffemodel', phase=caffe.TEST)
# or
net = caffe.Net('network.prototxt', caffe.TEST, weights='weights.caffemodel')
```
~~With some effort, this could probably be done without breaking the interface (but I'd need some help with the Boost.Python code). However, I am actually in favor of breaking the interface in order to make it simpler and more "Pythonic".~~

### DIFF
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -25,6 +25,7 @@ class Net {
  public:
   explicit Net(const NetParameter& param, const Net* root_net = NULL);
   explicit Net(const string& param_file, Phase phase,
+      const int level = 0, const vector<string>* stages = NULL,
       const Net* root_net = NULL);
   virtual ~Net() {}
 
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -86,19 +86,42 @@ void CheckContiguousArray(PyArrayObject* arr, string name,
   }
 }
 
-// Net constructor for passing phase as int
-shared_ptr<Net<Dtype> > Net_Init(
-    string param_file, int phase) {
-  CheckFile(param_file);
+// Net constructor
+shared_ptr<Net<Dtype> > Net_Init(string network_file, int phase,
+    const int level, const bp::object& stages,
+    const bp::object& weights) {
+  CheckFile(network_file);
 
-  shared_ptr<Net<Dtype> > net(new Net<Dtype>(param_file,
-      static_cast<Phase>(phase)));
+  // Convert stages from list to vector
+  vector<string> stages_vector;
+  if (!stages.is_none()) {
+    for (int i = 0; i < len(stages); i++) {
+      stages_vector.push_back(bp::extract<string>(stages[i]));
+    }
+  }
+
+  // Initialize net
+  shared_ptr<Net<Dtype> > net(new Net<Dtype>(network_file,
+        static_cast<Phase>(phase), level, &stages_vector));
+
+  // Load weights
+  if (!weights.is_none()) {
+    std::string weights_file_str = bp::extract<std::string>(weights);
+    CheckFile(weights_file_str);
+    net->CopyTrainedLayersFrom(weights_file_str);
+  }
+
   return net;
 }
 
-// Net construct-and-load convenience constructor
+// Legacy Net construct-and-load convenience constructor
 shared_ptr<Net<Dtype> > Net_Init_Load(
     string param_file, string pretrained_param_file, int phase) {
+  LOG(WARNING) << "DEPRECATION WARNING - deprecated use of Python interface";
+  LOG(WARNING) << "Use this instead (with the named \"weights\""
+    << " parameter):";
+  LOG(WARNING) << "Net('" << param_file << "', " << phase
+    << ", weights='" << pretrained_param_file << "')";
   CheckFile(param_file);
   CheckFile(pretrained_param_file);
 
@@ -245,7 +268,12 @@ BOOST_PYTHON_MODULE(_caffe) {
 
   bp::class_<Net<Dtype>, shared_ptr<Net<Dtype> >, boost::noncopyable >("Net",
     bp::no_init)
-    .def("__init__", bp::make_constructor(&Net_Init))
+    // Constructor
+    .def("__init__", bp::make_constructor(&Net_Init,
+          bp::default_call_policies(), (bp::arg("network_file"), "phase",
+            bp::arg("level")=0, bp::arg("stages")=bp::object(),
+            bp::arg("weights")=bp::object())))
+    // Legacy constructor
     .def("__init__", bp::make_constructor(&Net_Init_Load))
     .def("_forward", &Net<Dtype>::ForwardFromTo)
     .def("_backward", &Net<Dtype>::BackwardFromTo)
--- a/python/caffe/test/test_net.py
+++ b/python/caffe/test/test_net.py
@@ -72,7 +72,11 @@ class TestNet(unittest.TestCase):
         f.close()
         self.net.save(f.name)
         net_file = simple_net_file(self.num_output)
-        net2 = caffe.Net(net_file, f.name, caffe.TRAIN)
+        # Test legacy constructor
+        #   should print deprecation warning
+        caffe.Net(net_file, f.name, caffe.TRAIN)
+        # Test named constructor
+        net2 = caffe.Net(net_file, caffe.TRAIN, weights=f.name)
         os.remove(net_file)
         os.remove(f.name)
         for name in self.net.params:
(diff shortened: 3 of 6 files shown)


seanbell:
While I agree that the new interface is more Pythonic, I personally think that the old interface can be maintained with a little bit of massaging.  All that needs to be edited is the `Net.__init__` function, which can be done in the same style as the rest of the `_Net_X` functions.

Modify the end of `pycaffe.py` by adding:
```
# Original Boost.Python Net.__init__ function
_Net_init_boost = Net.__init__


def _Net_init(self, *args, **kwargs):
    """ Transform the arguments to handle the legacy API """
    if len(args) == 2 and not kwargs:
        # Net(param_file, phase)
        _Net_init_boost(self, args[0], weights_file=None, phase=args[1])
    elif len(args) == 3 and not kwargs:
        # Net(param_file, weights_file, phase)
        _Net_init_boost(self, args[0], weights_file=args[1], phase=args[2])
    else:
        _Net_init_boost(self, *args, **kwargs)


# Attach methods to Net.
Net.__init__ = _Net_init
...
```

The above code makes `make pytest` pass.

One other note: It seems that if you wanted to use a kwarg for every argument, then in some place it's named `param_file` and in others it's named `network_file`.  Maybe this should be harmonized to use the same name?

seanbell:
While I agree that the new interface is more Pythonic, I personally think that the old interface can be maintained with a little bit of massaging.  All that needs to be edited is the `Net.__init__` function, which can be done in the same style as the rest of the `_Net_X` functions.

Modify the end of `pycaffe.py` by adding:
```python
# Original Boost.Python Net.__init__ function
_Net_init_boost = Net.__init__


def _Net_init(self, *args, **kwargs):
    """ Construct a caffe.Net """
    # Transform the arguments to handle the legacy API
    if len(args) == 2 and not kwargs:
        # Handle case: Net(param_file, phase)
        _Net_init_boost(self, args[0], weights_file=None, phase=args[1])
    elif len(args) == 3 and not kwargs:
        # Handle case: Net(param_file, weights_file, phase)
        _Net_init_boost(self, args[0], weights_file=args[1], phase=args[2])
    else:
        # Handle new API:
        # Net('net.prototxt', weights_file='weights.caffemodel', phase=caffe.TEST, level=1, stages=['deploy'])
        _Net_init_boost(self, *args, **kwargs)


# Attach methods to Net.
Net.__init__ = _Net_init
...
```

The above code makes `make pytest` pass, but I haven't tested it thoroughly.

One other note: It seems that if you wanted to use a kwarg for every argument, then in some place it's named `param_file` and in others it's named `network_file`.  Maybe this should be harmonized to use the same name?

include/caffe/net.hpp
```diff
@@ -25,6 +25,7 @@ class Net {
  public:
   explicit Net(const NetParameter& param, const Net* root_net = NULL);
   explicit Net(const string& param_file, Phase phase,
+      const int level = 0, const vector<string>* stages = NULL,
```
ajtulloch: Could you also insert the new arguments (level, stages) after root_net? Otherwise, this breaks calling code that instantiates Net with root_net non-null:

```
Net(param_file, phase, root_net);
```

I'm not sure how common this is, but it's pretty trivial to avoid it by just inserting the new optional arguments at the end of the declaration?

lukeyeager: I agree, that seems like the safer choice. I wonder why no tests failed? Seems like a test hole.

lukeyeager:
> One other note: It seems that if you wanted to use a kwarg for every argument, then in some place it's named param_file and in others it's named network_file. Maybe this should be harmonized to use the same name?

I agree, the naming change is annoying. `network_file` and `weights_file` seem more meaningful to me than `param_file` and `pretrained_param_file`. They do a better job of conveying the fact that the first file is a textual representation of the network, while the second file is a binary file containing numerical weights.

I thought it might be appropriate to rename them here since this is the transition from internal code to external interface. Other common interfaces use words like [`solver.net=...`](https://github.com/BVLC/caffe/blob/rc3/src/caffe/proto/caffe.proto#L119) and [`caffe -weights=...`](https://github.com/BVLC/caffe/blob/rc3/tools/caffe.cpp#L39-L41).

include/caffe/net.hpp
lukeyeager: Does anyone use the `root_net` parameter? I'd like to keep the `phase`, `level` and `stages` parameters next to each other since they're naturally grouped. If anything, I'd prefer to move `root_net` before `phase`, but that would definitely break existing code.

lukeyeager:
> While I agree that the new interface is more Pythonic, I personally think that the old interface can be maintained with a little bit of massaging. All that needs to be edited is the `Net.__init__` function, which can be done in the same style as the rest of the _Net_X functions.

@seanbell I think I've found a nice compromise between our solutions. I've left the old `Net_Init_Load()` function in place, and I'm using Boost.Python's pattern matching to distinguish between `Net(string, string, int)` and `Net(string, int, int)`.

Now all of these constructors work as expected:
```python
### Legacy style
net = caffe.Net('network.prototxt', caffe.TRAIN)
net = caffe.Net('network.prototxt', 'weights.caffemodel', caffe.TEST)
  # pattern matches to Net(network_file, weights_file, phase)
  # prints deprecation warning

### New style
net = caffe.Net('network.prototxt', phase=caffe.TRAIN)
net = caffe.Net('network.prototxt', weights_file='weights.caffemodel', phase=caffe.TEST)
net = caffe.Net('network.prototxt', phase=caffe.TEST, weights_file='weights.caffemodel')
net = caffe.Net('network.prototxt', caffe.TRAIN, 0)
  # pattern matches to Net(network_file, phase, level)
net = caffe.Net('network.prototxt', caffe.TEST, 0, [], 'weights.caffemodel')
```

Made changes and rebased to address merge conflict from https://github.com/BVLC/caffe/pull/3982.

lukeyeager:
Is there anything in particular holding back this PR?

Does the re-arranging of the `root_net` parameter break any of the @BVLC team's crazy [cool] workflows?

Is my Python constructor compromise not satisfactory?

lukeyeager:
* Updated Net constructor to not switch order of arguments - good idea @ajtulloch 
* Added `-phase` flag to `tools/caffe` so that you can time the `TEST` phase
* Rebased to `master`

ajtulloch:
LGTM. Cool stuff @lukeyeager!

shelhamer:
Will review after NIPS deadline 05/20. Thanks for this @lukeyeager and thanks for review @ajtulloch!

python/caffe/_caffe.cpp
```diff
@@ -237,7 +260,12 @@ BOOST_PYTHON_MODULE(_caffe) {
 
   bp::class_<Net<Dtype>, shared_ptr<Net<Dtype> >, boost::noncopyable >("Net",
     bp::no_init)
-    .def("__init__", bp::make_constructor(&Net_Init))
+    // Constructor
+    .def("__init__", bp::make_constructor(&Net_Init,
+          bp::default_call_policies(), (bp::arg("network_file"), "phase",
+            bp::arg("level")=0, bp::arg("stages")=bp::object(),
+            bp::arg("weights_file")=bp::object())))
+    // XXX Legacy constructor
```
shelhamer: No need for `XXX`

python/caffe/_caffe.cpp
```diff
@@ -237,7 +260,12 @@ BOOST_PYTHON_MODULE(_caffe) {
 
   bp::class_<Net<Dtype>, shared_ptr<Net<Dtype> >, boost::noncopyable >("Net",
     bp::no_init)
-    .def("__init__", bp::make_constructor(&Net_Init))
+    // Constructor
+    .def("__init__", bp::make_constructor(&Net_Init,
+          bp::default_call_policies(), (bp::arg("network_file"), "phase",
+            bp::arg("level")=0, bp::arg("stages")=bp::object(),
+            bp::arg("weights_file")=bp::object())))
```
shelhamer: How about `weights` instead of `weights_file`? Including `_file` is more explicit but I like short arg names and every example will be a path to a `.caffemodel`.

tools/caffe.cpp
```diff
@@ -34,6 +34,13 @@ DEFINE_string(solver, "",
     "The solver definition protocol buffer text file.");
 DEFINE_string(model, "",
     "The model definition protocol buffer text file.");
+DEFINE_string(phase, "",
+    "Optional; network phase (TRAIN or TEST). Only used for 'time'.");
+DEFINE_int32(level, 0,
```
shelhamer: Why are `level` and `stage` not honored for `caffe train` on the command line? I think it's reasonable for command line args to override the definition, just as the `-gpu` arg takes precedence over `solver_mode`.

python/caffe/test/test_net.py
```diff
@@ -72,10 +72,235 @@ def test_save_and_read(self):
...
+        # Test named constructor
+        net2 = caffe.Net(net_file, caffe.TRAIN, weights_file=f.name)
         os.remove(net_file)
         os.remove(f.name)
         for name in self.net.params:
             for i in range(len(self.net.params[name])):
                 self.assertEqual(abs(self.net.params[name][i].data
                     - net2.params[name][i].data).sum(), 0)
+
+class TestLevels(unittest.TestCase):
```
shelhamer: There should be corresponding tests on the core, C++ side since not everything is in Python (though I use plenty of Python myself). Perhaps most of the tests should actually be core tests and then more abbreviated tests could be done through pycaffe.

shelhamer:
Re: `root_net` in https://github.com/BVLC/caffe/pull/3863#discussion_r60153570 and https://github.com/BVLC/caffe/pull/3863#discussion_r59980841, this is a pretty obscure arg that should never really be used outside of the framework itself. Furthermore I hope it can be eventually refactored out through further work on the parallel implementation. With that in mind, I think it is reasonable to keep it as the last, optional arg as to not complicate the normal instantiation of `Net`s. Thoughts?

tools/caffe.cpp
lukeyeager: Would we just set the `train_state` and not override any of the `test_states`?
https://github.com/BVLC/caffe/blob/rc3/src/caffe/proto/caffe.proto#L135-L136

shelhamer:
Although I originally voted for net states and these varied ways of expressing them, I'm now of two minds. On one hand phase, level, and stage states make all-in-one definitions possible and potentially make variations easier to express, but on the other these are just more special cases in the definition of nets. That said, they do seem to address a need in practice and since they are in the library they should at least be properly exposed to its interfaces.

While I now use and suggest net spec, net states have their purposes so thanks for extending them throughout the interfaces @lukeyeager.

For merge:

- [ ] extend `caffe train` to accept `level, stage` or decide not to and warn/die if they are given
- [ ] migrate net state tests to C++ side of testing, or at least add a basic test there to not rely on Python
- [ ] switch order of args with `root_net`?

Let me know if there are other concerns.

tools/caffe.cpp
shelhamer: That is what I had in mind, but perhaps that's counterintuitive too and we should instead catch when these args are given to `caffe train` and complain. It's reasonable to not support overriding, and my concern is more about silently running a different net than intended.

lukeyeager: I've updated the PR to change `train_state`. That seems reasonably intuitive since the command is "caffe train" after all.

lukeyeager:
I've addressed (1) and (2).

(3) Gulp. Time to dive into gtest ...

(4) Don't care

lukeyeager:
Added a basic C++ test

lukeyeager:
Ready for review. Do you want more extensive tests? Would you like me to switch the location of the `root_net` arg? I'll squash whenever you approve.

shelhamer:
@lukeyeager I'd rather switch the order of `root_net` as to not have the `NULL` arg for many standard instantiation of `Net`s, so please go ahead with that.

It would be best to have a C++ test that checks that the layers for do in fact exist, for instance by checking the layer name in the instantiated net (or differentiate the `train` and `val` dimensions of the `data` top  for an alternative check). The current test checks that instantiating with `stage` does not crash, but does not check what is actually instantiated.

Thanks.

include/caffe/net.hpp
lukeyeager: As per Evan's comments below, I reverted back to the original behavior of moving `root_net` to the end.

python/caffe/test/test_net.py
lukeyeager: Done

shelhamer:
@lukeyeager re: root net, make sure to update `solver.cpp` accordingly:

- https://github.com/BVLC/caffe/blob/master/src/caffe/solver.cpp#L107
- https://github.com/BVLC/caffe/blob/master/src/caffe/solver.cpp#L186-L187

At some point all of that code needs to be simplified, but it's here for now and not a problem to be solved by this PR.

lukeyeager:
> make sure to update solver.cpp accordingly:

None of those invocations need updating - they use a different constructor which expects a `NetParameter` object instead of a filename.

lukeyeager:
> I'd rather switch the order of root_net

Done

> It would be best to have a C++ test that checks that the layers for do in fact exist

Done

shelhamer:
https://github.com/BVLC/caffe/pull/3863#issuecomment-223065222 Right, sorry about that.

With https://github.com/BVLC/caffe/pull/3863#issuecomment-223070737 this looks good to me. @longjon will double-check for merge.

lukeyeager:
Rebased after conflict from https://github.com/BVLC/caffe/pull/4227

lukeyeager:
> this looks good to me. @longjon will double-check for merge.

ping @longjon

shelhamer:
Merging after offline discussion. Thanks @lukeyeager!

p.s. @jeffdonahue noted that this not strictly necessary since the `NetState` of the `NetParameter` can be set before instantiating a `Net`. However, this PR contributes a one-step interface to instantiating any state and enables exposing the different states to languages lacking protobuf bindings (such as matlab, as far as I know).

lukeyeager:
> this not strictly necessary since the NetState of the NetParameter can be set before instantiating a Net.

Sure, but it's nice to be able to change it without editing `.prototxt`.

> Merging after offline discussion.

Thanks!

PR #3575, opened 2016-01-20
Fixbug #3494 No to_python (by-value) converter found for C++ t…

errord:
fixbug #3494 , register shared_ptr Blob Dtype

PR #3866, opened 2016-03-22
Added check before register ptr to python

gongenhao:
This PR added check before register ptr to python for Net,Blob,Layers which is able to solve the warning related to: 
#3575 
and 
https://groups.google.com/forum/#!topic/caffe-users/C_air48cISU

On master branch, there are warnings when run pycaffe.
/home/ubuntu/caffe/python/caffe/pycaffe.py:13: RuntimeWarning: to-Python converter for boost::shared_ptrcaffe::Net<float > already registered; second conversion method ignored.
from ._caffe import Net, SGDSolver, NesterovSolver, AdaGradSolver, \
/home/ubuntu/caffe/python/caffe/pycaffe.py:13: RuntimeWarning: to-Python converter for boost::shared_ptrcaffe::Blob<float > already registered; second conversion method ignored.
from ._caffe import Net, SGDSolver, NesterovSolver, AdaGradSolver, \
/home/ubuntu/caffe/python/caffe/pycaffe.py:13: RuntimeWarning: to-Python converter for boost::shared_ptrcaffe::Solver<float > already registered; second conversion method ignored.
from ._caffe import Net, SGDSolver, NesterovSolver, AdaGradSolver, \

### DIFF
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -255,7 +255,10 @@ BOOST_PYTHON_MODULE(_caffe) {
     .def("_set_input_arrays", &Net_SetInputArrays,
         bp::with_custodian_and_ward<1, 2, bp::with_custodian_and_ward<1, 3> >())
     .def("save", &Net_Save);
-  bp::register_ptr_to_python<shared_ptr<Net<Dtype> > >();
+  if (boost::python::converter::registry::query(
+    boost::python::type_id <shared_ptr<Net<Dtype> > >()) == NULL) {
+    bp::register_ptr_to_python<shared_ptr<Net<Dtype> > >();
+  }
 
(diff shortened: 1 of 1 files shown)


shelhamer:
This doesn't seem quite right. With boost 1.60 I have this kind of error on `make pytest`:

> TypeError: No to_python (by-value) converter found for C++ type: boost::shared_ptr<caffe::Net<float> >

This is why #3575 was merged, though I agree the current warnings with other versions are a hassle and I would like to resolve the issue.

ZhengRui:
My warnings disappeared after using this fix, my boost version 1.58, osx10.10.

shelhamer:
Closing as fixed by #4069 (which works across boost versions without warnings).

PR #3868, opened 2016-03-22
Fix the bug of incorrect gradient when loss_weight set to zero

Cysu:
This PR aims at fixing the bug raised in #2895.

A flag (diff_written_) is added to the Blob class to indicate whether its diff has been used. When BP, clear the bottom diffs (if necessary) of the layers that do not propagate down. The computation and memory overhead are negligible.

### DIFF
--- a/include/caffe/blob.hpp
+++ b/include/caffe/blob.hpp
@@ -24,7 +24,8 @@ template <typename Dtype>
 class Blob {
  public:
   Blob()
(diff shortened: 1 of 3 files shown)


shaibagon:
thanks!

dangweili:
Hope this bug could be solved soon

dangweili:
Does any other loss functions meets the same bug when the loss_weight is setted to be zero?

PR #3869, opened 2016-03-22
Add makefile option for Anaconda Python3

Coldmooon:
While the current makefile.config has the option of `Python3`,  `Anaconda python 3.5` is not included. By adding `Anaconda 3` path to `PYTHON_INCLUDE`, `Caffe` can be compiled by `Anaconda python 3`.


Coldmooon:
To use `Anaconda python 3.5` in `Caffe`, I posted my operations in the issue [Python 3.4+ support](https://github.com/google/protobuf/issues/1276#issuecomment-199359063)

PR #3893, opened 2016-03-26
Update info about MKL licensing

rscohn2:
The instructions say that MKL is free for students, but as of 8/2015, MKL is free for everyone with community licensing.


shelhamer:
Thanks for noting the change in licensing.

PR #3913, opened 2016-03-30
small bugs and small improvements

antran89:
This PR addresses small improvements in Caffe codes such as small programming bugs, typos and etc.
- [x] There is a small bug in pooling_layer.cu line 141. The `cumsum` should be initialized as `0.` for cumulative sum. I have not tested it, but logically, it should be initialized with `0.`.
- [x] Improvements in compute_image_mean.cpp program: Print output to stderr (while still logging). It was tested on my LDMB database.

### DIFF
--- a/src/caffe/layers/pooling_layer.cu
+++ b/src/caffe/layers/pooling_layer.cu
@@ -138,7 +138,7 @@ __global__ void StoPoolForwardTest(const int nthreads,
     const int wstart = pw * stride_w;
     const int wend = min(wstart + kernel_w, width);
     // We set cumsum to be 0 to avoid divide-by-zero problems
-    Dtype cumsum = FLT_MIN;
+    Dtype cumsum = 0.;
     Dtype cumvalues = 0.;
     const Dtype* const bottom_slice =
         bottom_data + (n * channels + c) * height * width;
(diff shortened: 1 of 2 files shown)


seanbell:
@Yangqing You were the last to edit this file (https://github.com/BVLC/caffe/commit/c8e7cce7316c29b32d0b2dcdc98bff75afa0ab40) -- why did you change it to `FLT_MIN`?

seanbell:
@antran89 looking at this function, it seems that the last line of the kernel should also be updated to:
`top_data[index] = (cumsum > 0.) ? (cumvalues / cumsum) : 0.;` to avoid divide-by-zero.

Yangqing:
@seanbell Well, c8e7cce was a bug :)

seanbell:
@Yangqing no worries -- I was just checking to make sure, thanks :)

src/caffe/layers/pooling_layer.cu
```diff
@@ -149,7 +149,7 @@ __global__ void StoPoolForwardTest(const int nthreads,
         cumvalues += bottom_slice[h * width + w] * bottom_slice[h * width + w];
       }
     }
-    top_data[index] = cumvalues / cumsum;
+    top_data[index] = (cumsum > 0.) ? cumvalues / cumsum : 0.;
```
elezar: Is the test for `cumsum > 0.` correct here. Do we not allow negative values?

antran89: @elezar You are right? It should be `cumsum != 0.` for safe. 
@seanbell What do you think?

seanbell: Negative values aren't valid for stochastic pooling.  The input represents values in a probability distribution (though it doesn't have to be normalized to sum to 1).

PR #3919, opened 2016-03-30
CUB Memory Manager + cuDNN v4 and v5 support

drnikolaev:
This PR adds two non-separatable features to Caffe: high performance CUB Memory Manager and long awaited upgrade from cuDNN v3 to cuDNN v4 (and upcoming v5) libraries.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ env:
     - WITH_CUDA=true WITH_CMAKE=true WITH_IO=true
     - WITH_CUDA=false WITH_CMAKE=false WITH_IO=false
     - WITH_CUDA=false WITH_CMAKE=true WITH_IO=false PYTHON_VERSION=3
+#  Currently there is no way to install cudnn via apt-get. Uncomment when it's available.
+#    - WITH_CUDA=true WITH_CMAKE=false WITH_IO=true WITH_CUDNN=true
+#    - WITH_CUDA=true WITH_CMAKE=true WITH_IO=true WITH_CUDNN=true
 
 language: cpp
 
--- a/Makefile
+++ b/Makefile
@@ -7,12 +7,19 @@ $(error $(CONFIG_FILE) not found. See $(CONFIG_FILE).example.)
 endif
 include $(CONFIG_FILE)
 
+# Rectify input parameters
+ifeq ($(CPU_ONLY),1)
+  USE_CUDNN=0
+endif
+
+PROJECT_DIR=$(PWD)
+
 BUILD_DIR_LINK := $(BUILD_DIR)
 ifeq ($(RELEASE_BUILD_DIR),)
-	RELEASE_BUILD_DIR := .$(BUILD_DIR)_release
+	RELEASE_BUILD_DIR := $(PROJECT_DIR)/.$(BUILD_DIR)_release
 endif
 ifeq ($(DEBUG_BUILD_DIR),)
-	DEBUG_BUILD_DIR := .$(BUILD_DIR)_debug
+	DEBUG_BUILD_DIR := $(PROJECT_DIR)/.$(BUILD_DIR)_debug
 endif
 
 DEBUG ?= 0
@@ -23,6 +30,8 @@ else
 	BUILD_DIR := $(RELEASE_BUILD_DIR)
 	OTHER_BUILD_DIR := $(DEBUG_BUILD_DIR)
 endif
+
+THIRDPARTY_DIR=$(PROJECT_DIR)/3rdparty
 
 # All of the directories containing code.
 SRC_DIRS := $(shell find * -type d -exec bash -c "find {} -maxdepth 1 \
@@ -171,7 +180,7 @@ ifneq ("$(wildcard $(CUDA_DIR)/lib64)","")
 endif
 CUDA_LIB_DIR += $(CUDA_DIR)/lib
 
-INCLUDE_DIRS += $(BUILD_INCLUDE_DIR) ./src ./include
+INCLUDE_DIRS += $(BUILD_INCLUDE_DIR) ./src ./include $(THIRDPARTY_DIR)
 ifneq ($(CPU_ONLY), 1)
 	INCLUDE_DIRS += $(CUDA_INCLUDE_DIR)
 	LIBRARY_DIRS += $(CUDA_LIB_DIR)
@@ -325,6 +334,8 @@ endif
 # cuDNN acceleration configuration.
 ifeq ($(USE_CUDNN), 1)
 	LIBRARIES += cudnn
+	INCLUDE_DIRS += ${CUDNN_DIR}/include
+	LIBRARY_DIRS += ${CUDNN_DIR}/install/cuda/lib64
 	COMMON_FLAGS += -DUSE_CUDNN
 endif
 
@@ -440,7 +451,7 @@ endif
 # Define build targets
 ##############################
 .PHONY: all lib test clean docs linecount lint lintclean tools examples $(DIST_ALIASES) \
-	py mat py$(PROJECT) mat$(PROJECT) proto runtest \
+	py mat py$(PROJECT) mat$(PROJECT) thirdparty proto runtest \
 	superclean supercleanlist supercleanfiles warn everything
 
 all: lib tools examples
--- a/include/caffe/layer.hpp
+++ b/include/caffe/layer.hpp
@@ -38,7 +38,8 @@ class Layer {
    * layer.
    */
   explicit Layer(const LayerParameter& param)
-    : layer_param_(param), is_shared_(false) {
+    : layer_param_(param), is_shared_(false),
+      forward_passed_(false), backward_passed_(false) {
       // Set phase and copy blobs (if there are any).
       phase_ = param.phase();
       if (layer_param_.blobs_size() > 0) {
@@ -316,6 +317,21 @@ class Layer {
     param_propagate_down_[param_id] = value;
   }
 
+  bool IsForwardPassed() const {
+    return forward_passed_;
+  }
+
+  void ForwardPassed(bool passed) {
+    forward_passed_ = passed;
+  }
+
+  bool IsBackwardPassed() const {
+    return backward_passed_;
+  }
+
+  void BackwardPassed(bool passed) {
+    backward_passed_ = passed;
+  }
 
  protected:
   /** The protobuf that stores the layer parameters */
@@ -430,6 +446,9 @@ class Layer {
  private:
   /** Whether this layer is actually shared by other nets*/
   bool is_shared_;
+
+  bool forward_passed_;
+  bool backward_passed_;
 
   /** The mutex for sequential forward if this layer is shared */
   shared_ptr<boost::mutex> forward_mutex_;
--- a/include/caffe/layers/batch_norm_layer.hpp
+++ b/include/caffe/layers/batch_norm_layer.hpp
@@ -7,6 +7,9 @@
 #include "caffe/layer.hpp"
 #include "caffe/proto/caffe.pb.h"
 
+#define BN_VARIANCE_CLIP_START 200
+#define BN_VARIANCE_CLIP_CONST 4.0
+
 namespace caffe {
 
 /**
@@ -63,17 +66,30 @@ class BatchNormLayer : public Layer<Dtype> {
   virtual void Backward_gpu(const vector<Blob<Dtype>*>& top,
      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
 
-  Blob<Dtype> mean_, variance_, temp_, x_norm_;
+  virtual void multicast_cpu(int N, int C, int S, const Dtype *x, Dtype *y);
+  virtual void compute_sum_per_channel_cpu(int N, int C, int S,
+      const Dtype *x, Dtype *y);
+  virtual void compute_mean_per_channel_cpu(int N, int C, int S,
+      const Dtype *x, Dtype *y);
+#ifndef CPU_ONLY
+  virtual void compute_sum_per_channel_gpu(int N, int C, int S,
+      const Dtype *x, Dtype *y);
+  virtual void multicast_gpu(int N, int C, int S, const Dtype *x, Dtype *y);
+  virtual void compute_mean_per_channel_gpu(int N, int C, int S,
+      const Dtype *x, Dtype *y);
+#endif
+
+  Blob<Dtype> mean_, variance_, inv_variance_, x_norm_;
   bool use_global_stats_;
   Dtype moving_average_fraction_;
   int channels_;
   Dtype eps_;
-
-  // extra temporarary variables is used to carry out sums/broadcasting
-  // using BLAS
-  Blob<Dtype> batch_sum_multiplier_;
-  Blob<Dtype> num_by_chans_;
-  Blob<Dtype> spatial_sum_multiplier_;
+  int iter_;
+  // auxiliary arrays
+  Blob<Dtype> ones_N_, ones_HW_, ones_C_;
+  Blob<Dtype> temp_;
+  Blob<Dtype> temp_C_;
+  Blob<Dtype> temp_NC_;
 };
 
 }  // namespace caffe
--- a/include/caffe/layers/cudnn_lcn_layer.hpp
+++ b/include/caffe/layers/cudnn_lcn_layer.hpp
@@ -9,6 +9,9 @@
 
 #include "caffe/layers/lrn_layer.hpp"
 #include "caffe/layers/power_layer.hpp"
+#ifndef CPU_ONLY
+#include "caffe/util/gpu_memory.hpp"
+#endif
 
 namespace caffe {
 
@@ -17,8 +20,7 @@ template <typename Dtype>
 class CuDNNLCNLayer : public LRNLayer<Dtype> {
  public:
   explicit CuDNNLCNLayer(const LayerParameter& param)
-      : LRNLayer<Dtype>(param), handles_setup_(false), tempDataSize(0),
-        tempData1(NULL), tempData2(NULL) {}
+    : LRNLayer<Dtype>(param), handles_setup_(false), tempDataSize_(0) {}
   virtual void LayerSetUp(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top);
   virtual void Reshape(const vector<Blob<Dtype>*>& bottom,
@@ -32,15 +34,14 @@ class CuDNNLCNLayer : public LRNLayer<Dtype> {
       const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
 
   bool handles_setup_;
-  cudnnHandle_t             handle_;
   cudnnLRNDescriptor_t norm_desc_;
   cudnnTensorDescriptor_t bottom_desc_, top_desc_;
 
   int size_, pre_pad_;
   Dtype alpha_, beta_, k_;
 
-  size_t tempDataSize;
-  void *tempData1, *tempData2;
+  size_t             tempDataSize_;
+  gpu_memory::buffer temp1_, temp2_;
 };
 #endif
 
--- a/include/caffe/parallel.hpp
+++ b/include/caffe/parallel.hpp
@@ -57,6 +57,8 @@ class GPUParams : public Params<Dtype> {
   using Params<Dtype>::size_;
   using Params<Dtype>::data_;
   using Params<Dtype>::diff_;
+ private:
+  int buffer_device_;
 };
 
 class DevicePair {
--- a/include/caffe/util/cudnn.hpp
+++ b/include/caffe/util/cudnn.hpp
@@ -8,7 +8,7 @@
 #include "caffe/proto/caffe.pb.h"
 
 #define CUDNN_VERSION_MIN(major, minor, patch) \
-    (CUDNN_VERSION >= (major * 1000 + minor * 100 + patch))
+     (CUDNN_VERSION >= (major * 1000 + minor * 100 + patch))
 
 #define CUDNN_CHECK(condition) \
   do { \
@@ -91,8 +91,13 @@ template <typename Dtype>
 inline void createFilterDesc(cudnnFilterDescriptor_t* desc,
     int n, int c, int h, int w) {
   CUDNN_CHECK(cudnnCreateFilterDescriptor(desc));
+#if CUDNN_VERSION_MIN(5, 0, 0)
   CUDNN_CHECK(cudnnSetFilter4dDescriptor(*desc, dataType<Dtype>::type,
-      n, c, h, w));
+                                         CUDNN_TENSOR_NCHW, n, c, h, w));
+#else
+  CUDNN_CHECK(cudnnSetFilter4dDescriptor_v4(*desc, dataType<Dtype>::type,
+                                            CUDNN_TENSOR_NCHW, n, c, h, w));
+#endif
 }
 
 template <typename Dtype>
@@ -123,8 +128,15 @@ inline void createPoolingDesc(cudnnPoolingDescriptor_t* pool_desc,
     LOG(FATAL) << "Unknown pooling method.";
   }
   CUDNN_CHECK(cudnnCreatePoolingDescriptor(pool_desc));
-  CUDNN_CHECK(cudnnSetPooling2dDescriptor(*pool_desc, *mode, h, w,
-        pad_h, pad_w, stride_h, stride_w));
+#if CUDNN_VERSION_MIN(5, 0, 0)
+  CUDNN_CHECK(cudnnSetPooling2dDescriptor(*pool_desc, *mode,
+                                          CUDNN_PROPAGATE_NAN, h, w,
+                                          pad_h, pad_w, stride_h, stride_w));
+#else
+  CUDNN_CHECK(cudnnSetPooling2dDescriptor_v4(*pool_desc, *mode,
+                                          CUDNN_PROPAGATE_NAN, h, w,
+                                          pad_h, pad_w, stride_h, stride_w));
+#endif
 }
 
 }  // namespace cudnn
--- a/include/caffe/util/gpu_memory.hpp
+++ b/include/caffe/util/gpu_memory.hpp
@@ -1,0 +1,121 @@
+#ifndef CAFFE_UTIL_GPU_MEMORY_HPP_
+#define CAFFE_UTIL_GPU_MEMORY_HPP_
+
+#include <vector>
+#include "caffe/common.hpp"
+
+namespace caffe {
+
+class gpu_memory {
+ public:
+  enum PoolMode {
+    NoPool,     // Straight CUDA malllc/free. May be very expensive
+    CubPool,     // CUB caching allocator
+#ifdef CPU_ONLY
+    DefaultPool = NoPool
+#else
+    DefaultPool = CubPool     // CUB pool is able to use unified memory properly
+#endif
+  };
+
+  static const char* getPoolName();
+  static bool usingPool() {
+    return mode_ != NoPool;
+  }
+
+  class arena {
+   public:
+    arena(const std::vector<int>& gpus,
+          PoolMode m = DefaultPool, bool debug = false) {
+      init(gpus, m, debug);
+    }
+    ~arena() {
+      destroy();
+     }
+  };
+
+#ifndef CPU_ONLY
+  class buffer {
+   public:
+    // Construction/destruction
+    buffer(): ptr_(NULL), stream_(), size_(0) {}
+    buffer(size_t size, cudaStream_t s = cudaStreamDefault): stream_(s) {
+      reserve(size);
+    }
+    ~buffer() { gpu_memory::deallocate(ptr_, stream_); }
+
+    // Accessors
+    void* data() const { return ptr_; }
+    size_t size() const { return size_; }
+
+    // Memory allocation/release
+    void reserve(size_t size) {
+      if (size > size_) {
+        if (ptr_)
+          gpu_memory::deallocate(ptr_, stream_);
+        gpu_memory::allocate(&ptr_, size, stream_);
+        size_ = size;
+      }
+    }
+
+    /*
+     * This method behaves differently depending on pool availability:
+     * If pool is available, it returns memory to the pool and sets ptr to NULL
+     * If pool is not available, it does nothing (retaining memory)
+     */
+    void release() {
+      if (gpu_memory::usingPool()) {
+        gpu_memory::deallocate(ptr_, stream_);
+        ptr_ = NULL;
+        size_ = 0;
+      }
+      // otherwise (no pool) - we retain memory in the buffer
+    }
+
+   private:
+    void*         ptr_;
+    cudaStream_t  stream_;
+    size_t        size_;
+  };
+  static void update_dev_info(int device);
+
+# endif
+
+ private:
+  static void init(const std::vector<int>&, PoolMode, bool);
+  static void destroy();
+
+  static bool             initialized_;
+  static PoolMode         mode_;
+  static bool             debug_;
+
+#ifndef CPU_ONLY
+  struct MemInfo {
+    MemInfo()  {
+      free = total = flush_count = 0;
+    }
+    size_t   free;
+    size_t   total;
+    unsigned flush_count;
+  };
+
+  static vector<MemInfo>  dev_info_;
+
+ public:
+  typedef void* pointer;
+
+  static void allocate(pointer* ptr, size_t size,
+                       cudaStream_t stream = cudaStreamDefault);
+  static void deallocate(pointer ptr, cudaStream_t = cudaStreamDefault);
+
+  static void getInfo(size_t *free_mem, size_t *used_mem);
+
+ private:
+  static void initMEM(const std::vector<int>& gpus, PoolMode m);
+
+#endif
+};
+
+}  // namespace caffe
+
+# endif
--- a/include/caffe/util/math_functions.hpp
+++ b/include/caffe/util/math_functions.hpp
@@ -33,6 +33,16 @@ template <typename Dtype>
 void caffe_cpu_axpby(const int N, const Dtype alpha, const Dtype* X,
     const Dtype beta, Dtype* Y);
 
+// y[i]= max(a*x[i], b*y[i])
+template <typename Dtype>
+void caffe_cpu_eltwise_max(const int N, const Dtype alpha, const Dtype* X,
+    const Dtype beta, Dtype* Y);
+
+// y[i]= min(a*x[i], b*y[i])
+template <typename Dtype>
+void caffe_cpu_eltwise_min(const int N, const Dtype alpha, const Dtype* X,
+    const Dtype beta, Dtype* Y);
+
 template <typename Dtype>
 void caffe_copy(const int N, const Dtype *X, Dtype *Y);
 
@@ -245,6 +255,17 @@ void caffe_gpu_fabs(const int n, const Dtype* x, Dtype* y);
 
 template <typename Dtype>
 void caffe_gpu_scale(const int n, const Dtype alpha, const Dtype *x, Dtype* y);
+
+// y[i]= max(a*x[i], b*y[i])
+template <typename Dtype>
+void caffe_gpu_eltwise_max(const int n, const Dtype alpha, const Dtype* x,
+    const Dtype beta, Dtype* y);
+
+// y[i]= min(a*x[i], b*y[i])
+template <typename Dtype>
+void caffe_gpu_eltwise_min(const int n, const Dtype alpha, const Dtype* x,
+    const Dtype beta, Dtype* y);
+
 
 #define DEFINE_AND_INSTANTIATE_GPU_UNARY_FUNC(name, operation) \
 template<typename Dtype> \
--- a/src/caffe/common.cpp
+++ b/src/caffe/common.cpp
@@ -3,6 +3,7 @@
 #include <cmath>
 #include <cstdio>
 #include <ctime>
+#include <vector>
 
 #include "caffe/common.hpp"
 #include "caffe/util/rng.hpp"
@@ -105,7 +106,11 @@ void* Caffe::RNG::generator() {
 #else  // Normal GPU + CPU Caffe.
 
 Caffe::Caffe()
-    : cublas_handle_(NULL), curand_generator_(NULL), random_generator_(),
+    : cublas_handle_(NULL), curand_generator_(NULL),
+#ifdef USE_CUDNN
+    cudnn_handle_(NULL),
+#endif
+    random_generator_(),
     mode_(Caffe::CPU), solver_count_(1), root_solver_(true) {
   // Try to create a cublas handler, and report an error if failed (but we will
   // keep the program running as one might just want to run CPU code).
@@ -119,6 +124,11 @@ Caffe::Caffe()
       != CURAND_STATUS_SUCCESS) {
     LOG(ERROR) << "Cannot create Curand generator. Curand won't be available.";
   }
+#ifdef USE_CUDNN
+  if (cudnnCreate(&cudnn_handle_) != CUDNN_STATUS_SUCCESS) {
+    LOG(ERROR) << "Cannot create cuDNN handle. cuDNN won't be available.";
+  }
+#endif
 }
 
 Caffe::~Caffe() {
@@ -126,6 +136,9 @@ Caffe::~Caffe() {
   if (curand_generator_) {
     CURAND_CHECK(curandDestroyGenerator(curand_generator_));
   }
+#ifdef USE_CUDNN
+  if (cudnn_handle_) CUDNN_CHECK(cudnnDestroy(cudnn_handle_));
+#endif
 }
 
 void Caffe::set_random_seed(const unsigned int seed) {
@@ -164,6 +177,10 @@ void Caffe::SetDevice(const int device_id) {
       CURAND_RNG_PSEUDO_DEFAULT));
   CURAND_CHECK(curandSetPseudoRandomGeneratorSeed(Get().curand_generator_,
       cluster_seedgen()));
+#ifdef USE_CUDNN
+  if (Get().cudnn_handle_) CUDNN_CHECK(cudnnDestroy(Get().cudnn_handle_));
+  CUDNN_CHECK(cudnnCreate(&Get().cudnn_handle_));
+#endif
 }
 
 void Caffe::DeviceQuery() {
--- a/src/caffe/layers/batch_norm_layer.cpp
+++ b/src/caffe/layers/batch_norm_layer.cpp
@@ -235,5 +235,4 @@ STUB_GPU(BatchNormLayer);
 #endif
 
 INSTANTIATE_CLASS(BatchNormLayer);
-REGISTER_LAYER_CLASS(BatchNorm);
 }  // namespace caffe
--- a/src/caffe/layers/batch_norm_layer.cpp
+++ b/src/caffe/layers/batch_norm_layer.cpp
@@ -1,6 +1,7 @@
 #include <algorithm>
 #include <vector>
 
+#include "caffe/filler.hpp"
 #include "caffe/layers/batch_norm_layer.hpp"
 #include "caffe/util/math_functions.hpp"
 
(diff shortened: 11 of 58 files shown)


seanbell:
Right now Github refuses to render the diff, so the only thing that one can see are the added `3rdparty` files.  A workaround could be to add all `3rdparty` files in one commit, and put all other changes in a second commit.

seanbell:
Right now Github refuses to render the entire diff (too many changes), so the only thing that one can see are the added `3rdparty` files.  A workaround could be to add all `3rdparty` files in one commit, and put all other changes in a second commit.

drnikolaev:
OK, thank you. Will do 2 commits in 2 PRs...

include/caffe/layer.hpp
```diff
@@ -316,6 +317,21 @@ class Layer {
     param_propagate_down_[param_id] = value;
   }
 
+  bool IsForwardPassed() const {
```
ajtulloch: Is it possible to not have these as a part of `Layer`? AFAICT you can have exactly the same effect by just putting these forward_passed/backward_passed as instance variables in CuDNNBatchNormalizationLayer, and you don't add (essentially) completely unused base functions to a core class. 

IMO it's preferable to keep the core classes like Layer, Blob as small as possible.

src/caffe/util/gpu_memory.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    switch (m) {
+    case CubPool:
+      initMEM(gpus, m);
+      break;
+    default:
+      break;
+    }
+    if (debug)
+      std::cout << "gpu_memory initialized with "
```
ajtulloch: Can you just make these VLOG(1) instead of having a custom debug variable? Then users can just set --vmodule=gpu_module=1 to get the lines. 

Additionally, can you avoid logging to stdout and log to stderr (which you get with vlog)?

include/caffe/layer.hpp
thatguymike: This is part of the annoyance of Caffe's lazy allocation.  There are allocations that happen late, specifically setup of cuRand, so we need to be able to delay final setup of algorithm choice until we know how much memory we really have.  (There is an upcoming PR to move to the findEx paths in cudnnv5 instead of get that make this even more challenging).  We think we need this more generally.  Would love a better solution, but we start going down a "plan" style path pretty quickly.

src/caffe/test/test_gradient_based_solver.cpp
```diff
@@ -465,7 +465,7 @@ class GradientBasedSolverTest : public MultiDeviceTest<TypeParam> {
     for (int devices = 1; devices <= available_devices; ++devices) {
       // Configure batch size for single / multi device equivalence.
       // Constant data is needed for multi device as for accumulation.
-      num_ = kNum * devices;
+      num_ = kNum * devices * devices;
```
ajtulloch: Just curious, what was the reason for the changes in this file?

thatguymike: We have kept going back in forth on this.  This enables multiGPU solver tests to work properly.  The reason this needs to be done is that num_ is a global and not passed through the tests and the solver tests read num to know how many participants they have.  With multiple solvers, this gets the correct batch size initialized.  There is likely a cleaner but more invasive way to do this.  Happy to hear about a better solution.

src/caffe/util/gpu_memory.cpp
drnikolaev: Done. Thank you.

src/caffe/test/test_gradient_based_solver.cpp
ajtulloch: Cool, thanks!

cuihenggang:
That's cool that we can do batch normalization using cuDNN. So when we use the CudnnBatchNorm layer, do we still need to append a Scale layer after it?

borisgin:
No you don;t have to add scale/shift layers. CuDNN BN layer has both scale and shift inside.

drnikolaev:
Removed unnecessary files from 3rdparty directory. Just 5 of them left - those are the only required.

antran89:
Cool! How do I make changes to use this module? Do I need to add a new Layer before any ReLU layer? Or it will automatically do BN when I specify a variable in protxt file.

mfernezir:
Regarding BN layer usage. I've recently posted in https://github.com/NVIDIA/DIGITS/issues/629 about some differences between NVIDIA Caffe and BVLC version.

Since this PR has the same BatchNormParameter message in caffe.proto like the current NVIDIA's one, the following example should work in BVLC Caffe as well:

```
## BatchNorm
layer {
  bottom: "conv1/7x7_s2"
  name: "conv1/7x7_s2/bn"
  top: "conv1/7x7_s2/bn"
  type: "BatchNorm"
  param {
    lr_mult: 1
    decay_mult: 0
  }
  param {
    lr_mult: 1
    decay_mult: 0
  }
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
  }
}
layer {
  bottom: "conv1/7x7_s2/bn"
  top: "conv1/7x7_s2/bn"
  name: "conv1/relu_7x7"
  type: "ReLU"
}
```

There's no need to specify use_global_stats since Cafffe automatically infers the correct state TEST or TRAIN. The same layer definition is used for both train_val and deploy prototxt.

drnikolaev:
Added "Returned shift and scale back to BN layer" commit to make BN layer implementation consistent with the paper, cuDNN and other frameworks implementation.

borisgin:
Today If you want to use BN layer without scale and shift, then you can initialize these two parameters with 1 and 0, and set lr and weight decay to 0 in the train_val.prototxt. I can add a new parameter, which can do this automatically.
Boris

src/caffe/layers/batch_norm_layer.cu
```diff
@@ -6,166 +6,205 @@
...
+  caffe_gpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, N*C, S , 1,
+      1., temp_NC.gpu_data(), ones_HW_.gpu_data(), 0., y);
+}
+
+// y[c] = sum x(.,c,...)
+template <typename Dtype>
+void BatchNormLayer<Dtype>::compute_sum_per_channel_gpu(int N, int C, int S,
+    const Dtype *x, Dtype *y ) {
+  // assume that x.shape(1)==C
+  Blob<Dtype> temp_NC;
```
ajtulloch: It looks like this was meant to be `temp_NC_`? Do you want to `s/temp_NC/temp_NC_/g` in these files?

src/caffe/layers/batch_norm_layer.cu
```diff
@@ -6,166 +6,205 @@
...
   const Dtype* bottom_data = bottom[0]->gpu_data();
   Dtype* top_data = top[0]->mutable_gpu_data();
-  int num = bottom[0]->shape(0);
-  int spatial_dim = bottom[0]->count()/(channels_*bottom[0]->shape(0));
-
-  if (bottom[0] != top[0]) {
-    caffe_copy(bottom[0]->count(), bottom_data, top_data);
-  }
 
+//  if (bottom[0] != top[0]) {
```
ajtulloch: rm instead of commenting out?

src/caffe/layers/batch_norm_layer.cu
borisgin: Originally it was temp_NC_, but then I decided that it can be more safe to allocate temp_NC inside of these 2 functions to avoid potential hidden over-write (for example if I will use temp_NC_ outside) . I will re-check if this have noticable effect on time.

src/caffe/layers/batch_norm_layer.cu
borisgin: Right. I left it as a reminder to test BN in-place with potential overwrite ( e.g. con. layer --> BN --> relu)

drnikolaev: Redundant comments removed

src/caffe/layers/batch_norm_layer.cpp
```diff
@@ -22,17 +23,29 @@ void BatchNormLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
   if (this->blobs_.size() > 0) {
     LOG(INFO) << "Skipping parameter initialization";
   } else {
-    this->blobs_.resize(3);
+    this->blobs_.resize(5);
```
wlike: Why is 5 blobs? blobs_[4] is not used in the current code.

src/caffe/layers/batch_norm_layer.cpp
```diff
@@ -43,197 +56,207 @@ void BatchNormLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
     CHECK_EQ(bottom[0]->shape(1), channels_);
```
wlike: When `bottom[0]->num_axes() == 1`, access of `bottom[0]->shape(1)` is out of range.
The condition should be `bottom[0]->num_axes() > 1`.

borisgin: right. I will add this check.

borisgin:
I updated batch normalization layer. The new code:

- fixed the "... if (bottom[0]->num_axes() > 1) ... " in Reshape(.)
- changed initialization of global mean/ and variance
- add "variance clipping"

Most significant addition is a "variance clipping" . I added it to improve the stability of original BN algorithm, where very high or very low variance can result in 2 types of problem: 
1. large gradients , and as a result training accuracy --> 0 
2. corruption of global mean / variance. 
The "gradient clipping" can solve the first problem, but not the second: he training accuracy stay high, but the test accuracy is damaged since global mean& variance are corrupted. 
To prevent both issues I clip variance both from below and above using global mean and variance from previous iteration to remove outliers.
```
// clip variance 
    if (iter_ > BN_VARIANCE_CLIP_START) { 
      // temp_C_[c] = average_var + gobal_var[c]
      Dtype y = caffe_cpu_asum(C, this->blobs_[3]->cpu_data());
      caffe_cpu_scale(C, Dtype(y/C), ones_C_.cpu_data(), temp_C_.mutable_cpu_data());
      caffe_cpu_axpby(C, Dtype(1.0), this->blobs_[3]->cpu_data(), 
           Dtype(1.0), temp_C_.mutable_cpu_data());
      caffe_cpu_eltwise_min(C, Dtype(BN_VARIANCE_CLIP_CONST), temp_C_.cpu_data(),
            Dtype(1.0), variance_.mutable_cpu_data());
      // clip from below
      caffe_cpu_eltwise_max(C,   Dtype((1.)/BN_VARIANCE_CLIP_CONST), 
             this->blobs_[3]->cpu_data(), Dtype(1.0), variance_.mutable_cpu_data());
    }
```

src/caffe/layers/batch_norm_layer.cpp
seanbell: `blobs_[4]` is referenced in a few places, but I agree that it looks like it's not used anywhere.

seanbell: blobs_[4] is referenced in a few places, but I agree that it looks like it's not used anywhere.

@borisgin can you comment on this?

.travis.yml
```diff
@@ -8,6 +8,9 @@ env:
     - WITH_CUDA=true WITH_CMAKE=true WITH_IO=true
     - WITH_CUDA=false WITH_CMAKE=false WITH_IO=false
     - WITH_CUDA=false WITH_CMAKE=true WITH_IO=false PYTHON_VERSION=3
+#  Currently there is no way to install cudnn via apt-get. Uncomment when it's available.
+#    - WITH_CUDA=true WITH_CMAKE=false WITH_IO=true WITH_CUDNN=true
+#    - WITH_CUDA=true WITH_CMAKE=true WITH_IO=true WITH_CUDNN=true
```
longjon: This can go in its own commit.

Makefile
```diff
@@ -7,12 +7,19 @@ $(error $(CONFIG_FILE) not found. See $(CONFIG_FILE).example.)
 endif
 include $(CONFIG_FILE)
 
+# Rectify input parameters
+ifeq ($(CPU_ONLY),1)
+  USE_CUDNN=0
+endif
```
longjon: This can go in its own commit/PR, and please follow existing style (e.g., use space-padded `:=` assignment).

Makefile
```diff
@@ -7,12 +7,19 @@ $(error $(CONFIG_FILE) not found. See $(CONFIG_FILE).example.)
...
+
+PROJECT_DIR=$(PWD)
```
longjon: What's the reason for doing this? Is this necessary for the cuDNN changes, or can we do this in another PR (if it's really needed)?

Makefile
```diff
@@ -24,6 +31,8 @@ else
 	OTHER_BUILD_DIR := $(DEBUG_BUILD_DIR)
 endif
 
+THIRDPARTY_DIR=$(PROJECT_DIR)/3rdparty
```
longjon: Again, please use consistent style, `:=` here.

Makefile
```diff
@@ -325,6 +334,8 @@ endif
 # cuDNN acceleration configuration.
 ifeq ($(USE_CUDNN), 1)
 	LIBRARIES += cudnn
+	INCLUDE_DIRS += ${CUDNN_DIR}/include
```
longjon: Again, consistency... please use `$(VAR)` for variable expansion.

Makefile
```diff
@@ -440,7 +451,7 @@ endif
 # Define build targets
 ##############################
 .PHONY: all lib test clean docs linecount lint lintclean tools examples $(DIST_ALIASES) \
-	py mat py$(PROJECT) mat$(PROJECT) proto runtest \
+	py mat py$(PROJECT) mat$(PROJECT) thirdparty proto runtest \
```
longjon: I'm confused as to why this is here, as `thirdparty` doesn't seem to be referenced later on. Have I missed something?

include/caffe/layer.hpp
```diff
@@ -316,6 +317,21 @@ class Layer {
...
+    return forward_passed_;
+  }
+
+  void ForwardPassed(bool passed) {
```
longjon: Per Google style guide and existing Caffe usage, accessors are spelled in `underscore_case`, and take the name of the member variable (so, `forward_passed_` should be accessed with a function called `forward_passed`, and set with a function called `set_forward_passed`).

This is an FYI orthogonal to the question of whether these functions should be here, which I'll now consider!

include/caffe/layer.hpp
longjon: Unless I've misunderstood, it's hard to believe that there is any real performance implication due to 1.

If I'm imagining it correctly, what @thatguymike is calling a '"plan" style path' sounds right to me, though admittedly difficult to implement in current Caffe. [In fact, I have code of my own, (not as a Caffe branch) that takes that path, so it might be nice to discuss that elsewhere.]

Currently, however, this is a hack. Note that in current usage, memory usage can change (potentially drastically) after the initial forward-backward pass due to reshaping. So it's better not to push such things into the core of Caffe. If it becomes useful to share this kind of code among cuDNN layers, that can be done with another (intermediate) class, or some other kind of helper. This is also a (minor?) violation of modularity; layers are not really supposed to know anything about other layers or nets or whoever is running them.

I agree with @ajtulloch here -- I see no compelling reason to add these functions to all layers.

include/caffe/layers/batch_norm_layer.hpp
```diff
@@ -7,6 +7,9 @@
 #include "caffe/layer.hpp"
 #include "caffe/proto/caffe.pb.h"
 
+#define BN_VARIANCE_CLIP_START 200
+#define BN_VARIANCE_CLIP_CONST 4.0
```
longjon: I'll think later about what these constants are doing, but for now: don't use `#define` when you could use a `const` variable.

include/caffe/layers/batch_norm_layer.hpp
```diff
@@ -63,17 +66,30 @@ class BatchNormLayer : public Layer<Dtype> {
...
-
-  // extra temporarary variables is used to carry out sums/broadcasting
-  // using BLAS
-  Blob<Dtype> batch_sum_multiplier_;
-  Blob<Dtype> num_by_chans_;
-  Blob<Dtype> spatial_sum_multiplier_;
+  int iter_;
+  // auxiliary arrays
+  Blob<Dtype> ones_N_, ones_HW_, ones_C_;
+  Blob<Dtype> temp_;
```
longjon: Can we have a more descriptive name here? (Is there a reason we lost the descriptive names in the original?)

include/caffe/layers/batch_norm_layer.hpp
```diff
@@ -63,17 +66,30 @@ class BatchNormLayer : public Layer<Dtype> {
   virtual void Backward_gpu(const vector<Blob<Dtype>*>& top,
      const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
 
-  Blob<Dtype> mean_, variance_, temp_, x_norm_;
+  virtual void multicast_cpu(int N, int C, int S, const Dtype *x, Dtype *y);
```
longjon: Caffe's type style is `Dtype* x`, not `Dtype *x`.

include/caffe/layers/cudnn_lcn_layer.hpp
```diff
@@ -32,15 +34,14 @@ class CuDNNLCNLayer : public LRNLayer<Dtype> {
...
-  cudnnHandle_t             handle_;
   cudnnLRNDescriptor_t norm_desc_;
   cudnnTensorDescriptor_t bottom_desc_, top_desc_;
 
   int size_, pre_pad_;
   Dtype alpha_, beta_, k_;
 
-  size_t tempDataSize;
-  void *tempData1, *tempData2;
+  size_t             tempDataSize_;
```
longjon: We do not normally columnize declarations in Caffe.

include/caffe/parallel.hpp
```diff
@@ -57,6 +57,8 @@ class GPUParams : public Params<Dtype> {
   using Params<Dtype>::size_;
   using Params<Dtype>::data_;
   using Params<Dtype>::diff_;
+ private:
```
longjon: Put a blank line before private.

include/caffe/util/cudnn.hpp
```diff
@@ -8,7 +8,7 @@
 #include "caffe/proto/caffe.pb.h"
 
 #define CUDNN_VERSION_MIN(major, minor, patch) \
-    (CUDNN_VERSION >= (major * 1000 + minor * 100 + patch))
+     (CUDNN_VERSION >= (major * 1000 + minor * 100 + patch))
```
longjon: I assume this whitespace change is an error and should not be here? (If there is some reason for it, please do not put it in the same commit as unrelated code.)

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,121 @@
+#ifndef CAFFE_UTIL_GPU_MEMORY_HPP_
+#define CAFFE_UTIL_GPU_MEMORY_HPP_
+
+#include <vector>
+#include "caffe/common.hpp"
+
+namespace caffe {
+
+class gpu_memory {
```
longjon: Classes are `CamelCase` in Google style/Caffe. This is also easily confused with `SyncedMem`...

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,121 @@
...
+ public:
+  enum PoolMode {
+    NoPool,     // Straight CUDA malllc/free. May be very expensive
+    CubPool,     // CUB caching allocator
+#ifdef CPU_ONLY
+    DefaultPool = NoPool
+#else
+    DefaultPool = CubPool     // CUB pool is able to use unified memory properly
+#endif
+  };
```
longjon: Indentation is broken starting from this line...

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+#include <vector>
+#include "caffe/common.hpp"
+
+namespace caffe {
+
+class gpu_memory {
+ public:
+  enum PoolMode {
+    NoPool,     // Straight CUDA malllc/free. May be very expensive
```
longjon: Two spaces before comments, unless they're part of a multi-line block. You may also wish to spell `malloc` correctly.

include/caffe/layer.hpp
drnikolaev: Now please look at net.cpp:

+  for (int i = start; i <= end; ++i) {
+    layers_[i]->ForwardPassed(true);
+  }

Here layers[i] is a pointer to the *base* class Layer. How would we call ForwardPassed here?

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,121 @@
...
+    CubPool,     // CUB caching allocator
+#ifdef CPU_ONLY
+    DefaultPool = NoPool
+#else
+    DefaultPool = CubPool     // CUB pool is able to use unified memory properly
+#endif
+  };
+
+  static const char* getPoolName();
```
longjon: As earlier, accessors are `underscore_case` and do not start with `get`. (Cheap methods may also be `underscore_case`. Most methods are `CamelCase`. No methods are `lowercaseCamel`.)

include/caffe/util/gpu_memory.hpp
longjon: Constants in Caffe are generally `LOUD_CASE`. (This is not the same as the Google style guide in this case.)

include/caffe/util/math_functions.hpp
```diff
@@ -33,6 +33,16 @@ template <typename Dtype>
 void caffe_cpu_axpby(const int N, const Dtype alpha, const Dtype* X,
     const Dtype beta, Dtype* Y);
 
+// y[i]= max(a*x[i], b*y[i])
```
longjon: Please space-pad operators, even in comments (`y[i] = max(a * x[i], b * y[i])`).

src/caffe/common.cpp
```diff
@@ -3,6 +3,7 @@
 #include <cmath>
 #include <cstdio>
 #include <ctime>
+#include <vector>
```
longjon: Why is this being added here? Is this relevant to these changes?

src/caffe/layers/cudnn_batch_norm_layer.cu
```diff
@@ -0,0 +1,112 @@
...
+      mode_,
+      cudnn::dataType<Dtype>::one,
+      cudnn::dataType<Dtype>::zero,
+      bottom_desc_,
+      bottom_data,
+      bottom_desc_,
+      top_data,
+      scale_bias_mean_var_desc_,
+      scale_data,
+      bias_data,
```
longjon: Please don't add all these linebreaks unless you think there's a really good reason for it; it doesn't agree with existing style, and it makes the code harder to read by stretching it out. It's okay to group arguments logically.

src/caffe/layers/cudnn_batch_norm_layer.cu
```diff
@@ -0,0 +1,112 @@
...
+      1-this->moving_average_fraction_,
```
longjon: Spaces around operators; it's particularly easy to misread this one.

src/caffe/layers/cudnn_conv_layer.cpp
```diff
@@ -2,25 +2,23 @@
...
-#define CUDNN_STREAMS_PER_GROUP 3
+namespace caffe {
 
 /**
  * TODO(dox) explain cuDNN interface
  */
 template <typename Dtype>
 void CuDNNConvolutionLayer<Dtype>::LayerSetUp(
     const vector<Blob<Dtype>*>& bottom, const vector<Blob<Dtype>*>& top) {
+
```
longjon: This is an erroneous whitespace change.

src/caffe/layers/cudnn_conv_layer.cpp
```diff
@@ -2,25 +2,23 @@
...
 
-namespace caffe {
 
-// Set to three for the benefit of the backward pass, which
-// can use separate streams for calculating the gradient w.r.t.
-// bias, filter weights, and bottom data for each group independently
-#define CUDNN_STREAMS_PER_GROUP 3
+namespace caffe {
 
 /**
```
longjon: By the way, this whole file has a pretty messed up style... everything is indented inside the `namespace` (it shouldn't be), and there are lots of unneeded blank lines. (Caffe style is generally pretty compact wrt blank lines, which I can see arguments either way, but there's no reason to be inconsistent here.)

If you do fix these style issues, however, please do so in a separate commit.

src/caffe/layers/cudnn_conv_layer.cpp
```diff
@@ -137,93 +131,42 @@ void CuDNNConvolutionLayer<Dtype>::Reshape(
...
-    CUDNN_CHECK(cudnnGetConvolutionForwardWorkspaceSize(handle_[0],
+    CUDNN_CHECK(cudnnGetConvolutionForwardWorkspaceSize(Caffe::cudnn_handle(),
       bottom_descs_[i],
       filter_desc_,
       conv_descs_[i],
       top_descs_[i],
       fwd_algo_[i],
       &(workspace_fwd_sizes_[i])));
 
+    //
```
longjon: This appears to be spurious.

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -1,126 +1,182 @@
...
+#include "caffe/util/gpu_memory.hpp"
+#include "caffe/util/im2col.hpp"
+#include "caffe/util/math_functions.hpp"
+
+// Those defines serve single purpose to keep sane C++ formatting
+// in presence of <80 characters rule
+#define cudnnConvFwd                       cudnnConvolutionForward
+#define cudnnConvBwdBias                   cudnnConvolutionBackwardBias
+#define cudnnConvBwdFilter                 cudnnConvolutionBackwardFilter
+#define cudnnConvBwdData                   cudnnConvolutionBackwardData
```
longjon: Is this really necessary? If you think so, please remove the huge floating space in the middle.

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -1,126 +1,182 @@
...
-    for (int g = 0; g < this->group_; g++) {
-      // Gradient w.r.t. bias.
-      if (this->bias_term_ && this->param_propagate_down_[1]) {
-        CUDNN_CHECK(cudnnConvolutionBackwardBias(handle_[0*this->group_ + g],
-              cudnn::dataType<Dtype>::one,
-              top_descs_[i],  top_diff + top_offset_ * g,
-              cudnn::dataType<Dtype>::one,
-              bias_desc_, bias_diff + bias_offset_ * g));
-      }
+  template <typename Dtype>
```
longjon: Did we gain an indent here?

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -1,126 +1,182 @@
...
+      gpu_memory::getInfo(&workspace_limit_bytes, &total_memory);
+      if (workspace_fwd_sizes_[i] > workspace_limit_bytes) {
+          this->Reshape(bottom, top);
       }
 
-      // Gradient w.r.t. bottom data.
-      if (propagate_down[i]) {
-        if (weight == NULL) {
-          weight = this->blobs_[0]->gpu_data();
+      // !!!! Not safe if group_ > 1 !!!!
```
longjon: Uh, that's frightening? What does that mean?

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -1,126 +1,182 @@
...
+      workspace.reserve(workspace_fwd_sizes_[i]);
+
+      // Forward through cuDNN in parallel over groups.
+      for (int g = 0; g < this->group_; g++) {
+          // Filters.
```
longjon: These comments seem spurious to me.

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -1,126 +1,182 @@
...
+                                 bottom_data + bottom_offset_ * g,
+                                 filter_desc_,
+                                 weight + this->weight_offset_ * g,
+                                 conv_descs_[i],
+                                 fwd_algo_[i],
+                                 workspace.data(),
+                                 workspace.size(),
+                                 cudnn::dataType<Dtype>::zero,
+                                 top_descs_[i],
+                                 top_data + top_offset_ * g));
```
longjon: Again, could we maybe write these in logical groups instead of taking up thirteen lines?

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -1,126 +1,182 @@
...
+      // Synchronize the work across groups, each of which went into its own
+      // stream, by launching an empty kernel into the default (null) stream.
+      // NOLINT_NEXT_LINE(whitespace/operators)
+      CUDA_CHECK(cudaStreamSynchronize(cudaStreamLegacy));
+    }
+  }
+
+  template <typename Dtype>
+  void
+  CuDNNConvolutionLayer<Dtype>::
```
longjon: Don't split method names this way; that style does not exist in Caffe.

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -1,126 +1,182 @@
...
+        // Synchronize the work across groups, each of which went into its own
+        // stream, by launching an empty kernel into the default (null) stream.
+        // NOLINT_NEXT_LINE(whitespace/operators)
+        CUDA_CHECK(cudaStreamSynchronize(cudaStreamLegacy));
+    }
   }
-}
 
-INSTANTIATE_LAYER_GPU_FUNCS(CuDNNConvolutionLayer);
+  INSTANTIATE_LAYER_GPU_FUNCS(CuDNNConvolutionLayer);
```
longjon: This is spurious (per overall indentation error).

src/caffe/layers/cudnn_lcn_layer.cpp
```diff
@@ -35,36 +34,21 @@ void CuDNNLCNLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
...
-    cudaFree(tempData1);
-    cudaFree(tempData2);
-
-    // allocate new buffers
-    CUDA_CHECK(cudaMalloc(&tempData1, totalSizeInBytes));
-    CUDA_CHECK(cudaMalloc(&tempData2, totalSizeInBytes));
-  }
+  // size for temp buffers
+  tempDataSize_ = sizeof(Dtype)*bottom[0]->num()* \
+    this->channels_*this->height_*this->width_;
```
longjon: Spaces around operators.

longjon: The blackslash appears to be totally spurious here.

src/caffe/layers/dropout_layer.cu
```diff
@@ -24,11 +24,11 @@ void DropoutLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
     unsigned int* mask =
         static_cast<unsigned int*>(rand_vec_.mutable_gpu_data());
     caffe_gpu_rng_uniform(count, mask);
+    CUDA_POST_KERNEL_CHECK;
```
longjon: If this should be here, it should be in another PR.

src/caffe/layers/dropout_layer.cu
```diff
@@ -24,11 +24,11 @@ void DropoutLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
...
     // set thresholds
     // NOLINT_NEXT_LINE(whitespace/operators)
     DropoutForward<Dtype><<<CAFFE_GET_BLOCKS(count), CAFFE_CUDA_NUM_THREADS>>>(
         count, bottom_data, mask, uint_thres_, scale_, top_data);
-    CUDA_POST_KERNEL_CHECK;
```
longjon: As above, this does not seem to have any relevance to cuDNN (but why was this change made anyway?).

longjon:
Okay, I did a(n incomplete) pass on this. A few overall comments:

1. Something I've said before and will keep saying: Make PRs _as small as possible_. Make commits _as small as possible_. It's okay for one PR to depend on another. It is very difficult to review and accept, all at once, a set of changes this large. I see _at least_ four major PRs here: one which adds and builds against CUB, one which adds CUB support, one which updates existing cuDNN layers for versions four and five, and one which adds cuDNN batch norm support. Each of those have major discussion points on their own, and each could be merged on their own.

    In addition, there's a bunch of smaller changes which seem like they could be broken out into their own PRs. Every unrelated change is a multiplicative factor in merge waiting time. No PR is too small. So long as it passes Travis and makes some kind of improvement.

    That goes for commits as well! Even if many changes critically depend on each other, they should be broken into the smallest steps that make logical sense. For a patch like this, I would expect a double-digit number of commits. History should always be rewritten to tell a clean, compelling story. This should not be too onerous a demand; it should be a natural part of constructing and explaining a PR.

    Okay, lecture over.
2. Forgive me if this has been discussed already elsewhere, but: why are dumping CUB code into Caffe here, instead of taking CUB (or some part of it) as a dependency? This is not the right place to maintain a memory allocator...
3. Please follow existing style, as documented (mostly) by the Google style guide. Passing lint does not mean your style is correct; it just means it didn't fail any of the checks we happen to have. This code is pretty divergent from Caffe style in many places.
4. Per 1, I did not make a complete pass; in particular, I have not fully reviewed changes to batch normalization or parallel.

My recommendation here is to split this up. As soon as we're clear on point 2 above, we can merge in CUB (or support for CUB), after which I think the changes to existing cuDNN layers are pretty straightforward, and could be merged by themselves, while the discussion on the right batch norm interface resolves, and so forth...

include/caffe/layer.hpp
longjon: You can't, of course. Instead, you'd need to set `forward_passed_` in layer code, presumably at the end of `Forward_*`.

longjon: (Perhaps the intention here was to keep track not simply of whether an individual layer's forward has completed, but whether _all_ layers' forwards' have completed. But note that that's not what this code does anyway.)

longjon:
One option re: 2 above that has come up offline is to use a git submodule. I don't personally have experience with that, so I'm open to hearing opinions about what the pros and cons are; right now that feels like a better option to me than dumping in the code.

seanbell:
@longjon: I've used submodules.

**Pros:** They're a clean way of separating out third-party dependencies.  Submodules are tied to a specific commit, so if the third party library makes updates, you will need to manually update if you want to use a later version in caffe.

**Cons:** All install scripts (that check out caffe), tutorials, and docker scripts would need to be updated with an extra line: `git submodule update --init --recursive`.  If you're using submodules, you need to think about them as an extra step; they are not automagic.  This logic could be built into the Makefile, so it may not be a problem.

flx42:
@seanbell Actually, you can also use the following option with `git clone`:
```
--recursive, --recurse-submodules
    After the clone is created, initialize all submodules within, using their default settings. This is equivalent to running git submodule update --init --recursive immediately after the clone is finished. This option is ignored if the cloned repository
    does not have a worktree/checkout (i.e. if any of --no-checkout/-n, --bare, or --mirror is given)
```

drnikolaev:
We can simply call `wget https://raw.githubusercontent.com/NVlabs/cub/1.5.2/cub/util_allocator.cuh` 
Looks like the simplest way to get those tiny 5 headers.

seanbell:
@flx42 My point was that `git clone` by itself no longer works, and install scripts will still need to be updated.  I agree that `git clone --recursive` is a nice shorthand.

@drnikolaev `wget` seems like it could work as well, if there is a good way of keeping track of versions.

.travis.yml
flx42: It's actually possible if you have the right repo:
http://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1404/x86_64

flx42:
So, what are our options here? I see 3 so far:
1) Integrate a subset of CUB inside Caffe. This is the current code.

2) Use git submodules or its alternative: git subtrees. It might be easier to pin versions this way. The build systems/install scripts would need to be modified to hide the change from the user. Using `wget` could also work for CUB, but this solution might not scale well if we want to start doing the same for other projects that are small enough to be compiled inside Caffe.
I think this solution is quite common.

3) Make CUB optional. While it's nice to have the choice, we would add yet another build combination, making the test matrix even bigger: Makefile/CMake, CPU/GPU, cuDNN/no cuDNN, CUB/no CUB. Also, the code would have even more ifdefs, possibly nested, for instance `#if USE_CUDNN` and then `#if USE_CUB`.

gcp:
> 3) Make CUB optional. While it's nice to have the choice, we would add yet another build combination

Not if you require it for CuDNN support. That would also remove the #ifdef nesting. Yes, it's an extra install step for the user, but the same user already has to install CUDA (Toolkit) and CuDNN so why is this different?

FWIW, I have a bunch of inference code that's using Caffe as a C++ library, where GPU acceleration stops working on this branch and Caffe silently falls back to CPU code (without errors). My best bet is that this branch added some extra required initialization steps that aren't being called in all paths.

Is CUB a hard requirement to use CuDNNv5 at all, or just for the new features?

flx42:
> Not if you require it for CuDNN support. That would also remove the #ifdef nesting. Yes, it's an extra install step for the user, but the same user already has to install CUDA (Toolkit) and CuDNN so why is this different?

You do have to stop at one point ;).
CUB is indeed not too difficult to install:
```
curl -fsSL https://github.com/NVlabs/cub/archive/1.5.2.tar.gz | tar -C /usr/local -zxvf -
```
But it's still an extra step and it will break the build of many many users. I think option 2) is cleaner, it's easier to hide for users.

```
FWIW, I have a bunch of inference code that's using Caffe as a C++ library, where GPU acceleration stops working on this branch and Caffe silently falls back to CPU code (without errors). My best bet is that this branch added some extra required initialization steps that aren't being called in all paths.
```
This is a bit off-topic, but do you have a `#define USE_CUDNN 1` before the `#include <caffe/caffe.hpp>`? If not, you're in for some troubles (Caffe doesn't have a config header in order to know how it was compiled).

drnikolaev:
@flx42 one more way (I'm going to begin with it):

4) wget 5 files from github. 

This wget would be executed in Makefile (CMake is even simpler cause it has target for that).
It means that users shouldn't do anything at all.

flx42:
@drnikolaev that's actually my 2)

shelhamer:
re: submodules, in my experience they are a reasonable way to handle dependencies. However as @seanbell points out install and docs will have the handle submodule init, but that's fine. (Note that this will discourage tarball/zip downloads from github, but these are problematic anyway.)

Including CUB as a submodule seems like the preferred choice:

- it is versioned, but keeps CUB state separated from Caffe
- it is automatic (or will be in our usage)
- it scales to other dependencies (whether 3rd party libraries or our own examples) with the tool we already have (git). a git url seems a little more canonical than a list of urls for `wget`.

Let's package the minimal part of CUB as a submodule if it can't be handled as a regular dependency like everything else. Note that the [NVlabs/cub](https://github.com/NVlabs/cub) is unsuitable as-is since it includes all kinds of generated files for over 150(!) mb.

@drnikolaev could you change this PR accordingly to have a separate commit for a CUB submodule?

flx42:
@shelhamer: it looks like you can have a shallow submodule:
```
$ git submodule add --depth 1 https://github.com/NVlabs/cub cub
Cloning into 'cub'...
remote: Counting objects: 785, done.
remote: Compressing objects: 100% (438/438), done.
remote: Total 785 (delta 443), reused 550 (delta 342), pack-reused 0
Receiving objects: 100% (785/785), 6.21 MiB | 820.00 KiB/s, done.
Resolving deltas: 100% (443/443), done.
Checking connectivity... done.
```
Much smaller. But not sure what's the impact of doing a shallow submodule clone.

drnikolaev:
@shelhamer Hi Evan,
wget imho equally brittle as git submodule. We need 2 extra lines in Makefile, something like CUB_VERSION := 1.5.2 plus a line calling a script which takes that version variable as an argument. When new CUB comes we just need to change the version. Let me do that, then I'll file a PR and we see how it goes. If it won't be accepted I'll re-do it as a submodule.

cuihenggang:
Why the CuDNNBatchNormLayer has five blobs. What are blobs_[3] and blobs_[4]?

https://github.com/drnikolaev/caffe/blob/bvlc_cub_v4_v5/src/caffe/layers/cudnn_batch_norm_layer.cpp#L32

cuihenggang:
I was trying to use this code to train an inception network on the CIFAR-10 data, but I find it doesn't converge. I have trained the same network with the original BatchNormLayer of Caffe's master branch, and it converged.

I don't know what could be wrong. Do you have a train_val file that uses BatchNorm layers that we can try?

borisgin:
On Thu, May 12, 2016 at 1:45 PM, Henggang Cui <notifications@github.com>
wrote:

> I was trying to use this code to train an inception network on the
> CIFAR-10 data, but I find it doesn't converge. I have trained the same
> network with the original BatchNormLayer of Caffe's master branch, and it
> converged.
>
> I don't know what could be wrong. Do you have a train_val file that uses
> BatchNorm layers that we can try?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/BVLC/caffe/pull/3919#issuecomment-218880175>
>
I attached 2 files for cifar-10.

borisgin:
and these two for alexnet with BN

cuihenggang:
Hi @borisgin, sorry, I cannot see your attachments?

borisgin:
and these for inceptionv1 ( 5x5 are replaced by double 3x3)

borisgin:
blobs_[0] - scale
blobs_[1] - shift
blobs_[2] global mean  (accumulated during TRAIN, used in TEST)
blobs_[3] global varinace (accumulated during TRAIN, used in TEST)
blobs_[4] used to be iteration coutner, now void (for cudnn compatibility)

cuihenggang:
I figured out why my network didn't converge. I forgot to set the scale_filler and bias_filler for my BN layers. After I set the fillers, it converges.

flx42:
Please rebase and rename the issue now that #4159 was merged.

seanbell:
Ping?  Now that NIPS passed on 5/20, is there any more progress on this?

mathmanu:
If this BN method is used, will I be able to load a cafemodel (for finetuning) trained using the old BN method. I tried this in the nvidia/cafe repo and load of caffemodel exited complaining that the number of blobs doesn't match. This could be a critical need as there are several older caffemodels out there that we need to use.

borisgin:
No. Old BN has 2 blobs: global mean and variance., New one has 5: scale, bias, gloabl_mean, and global_variance, global_counter

mathmanu:
This could be a problem since several popular models become un-usable. (eg. https://github.com/KaimingHe/deep-residual-networks)

Do we really have to break the backward compatibility this way? In the old BN we used to have a  separate scaling layer and that was fine. 

Kindly re-consider this and try to keep the compatibility.

mathmanu:
Alternatively - you could provide an up gradation method as well - for example, while loading the older mode for finetuning, the scale and bias blobs could be forced to one and zero respectively. 

I also saw that the shape of the blobs (global_mean, and global_variance)  were also different - although they were of same size - this also creates problem.

mathmanu:
I agree that there is merit in what you did - by combining the normalization and scaling. 

Yet another easy fix is to make this a different layer - say "BatchNormScale", and keep the older layer as it is as a separate layer for backward compatibility. This is so far the simplest (minimal code changes) solution that I could come up with.

borisgin:
Agree,  Adding new layer would be the simplest way.

mathmanu:
So do you think you can do it the name change (and keep the old layer) right away? I would love to use this new layer with CUDNN as CUDNN gives me 2x boost in speed.

mathmanu:
Do you have any convergence issue with CUDNN BatchNorm used in this PR? In the nvidia/caffe, I had to set the engine as CAFFE for BatchNorm to get convergence.

I was struggling with the convergence issue, but finally the following worked for me. Specifying the engine as CAFFE is important. CUDNN BatchNorm doesn't converge for me.

The following is the configuration that I used in nvidia/caffe version. I am posting it here because I think the underlying implementation is same.

layer {
  name: "bn2"
  bottom: "conv2"
  top: "conv2"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CAFFE
  }
}

mathmanu:
If the CUDNN BatchNorm converged, that would have given me overall 4x boost in speed, but now with the CAFFE engine for BatchNorm, I get only 2x boost in speed overall!

mathmanu:
I have couple of more comments:

1. If you change the oder of the blobs to:  gloabl_mean, and global_variance, scale, bias, global_counter, then I don't have to specify 4 param fields for lr_mult and decay_mult - but only 2. 

2. If the definition of scale and bias fields in BatchNormParameter is changed to:
  optional float scale_filler = 5 [default = 1];
  optional float bias_filler = 6 [default = 0];
Then I don't have to specify these also in the prototxt.

These changes will help someone who is trying to use this layer for the first time - apart from saving some space in the prototxt.

mathmanu:
Need urgent help - any suggestions to solve the issue of non convergence with CUDNN BatchNorm? Are CUDNN BatchNorm implementations same in nvcaffe-0.15, nvcaffe-0.16 and this PR? (I tried nvcaffe-0.16) Any suggestions would help.

borisgin:
I will re-check the convergence issue with cuDNN_BN.

borisgin:
The idea to change parameter order is very good. I will do this.

mathmanu:
Thanks. How about the fillers - can you provide default values for them too?

mathmanu:
When you check the convergence issue with cuDNN_BN, kindly do so with a network that has many BN layers. For example ResNet18.

borisgin:
can you send me your solver.ptototxt and train_test.prototxt files for the
model which converg with caffe-engine and diverge with cudnn please?

mathmanu:
I couldn't attach for some reason. I have copied the combined (solver + train - you don't need two files) prototxt below. If you change the engine to CAFFE, it will start to converge.

#Sover parameters
test_iter: 200
test_interval: 1000
test_initialization: true
display: 100
base_lr: 0.01
lr_policy: "multistep"
stepvalue: 25000
stepvalue: 50000
stepvalue: 75000
stepvalue: 100000
gamma: 0.1
max_iter: 125000
momentum: 0.9
weight_decay: 1e-4
regularization_type: "L2" #"L1"
snapshot: 1000
snapshot_prefix: "training/resnet18"
solver_mode: GPU
random_seed: 33


#Net parameters
net_param {

name: "ResNet-18(1024)"

layer {
  name: "data"
  type: "Data"
  top: "data"
  top: "label"
  include {
    phase: TRAIN
  }
# mean pixel / channel-wise mean instead of mean image
  transform_param {
    crop_size: 224
    mean_value: 128
    mean_value: 128
    mean_value: 128
    mirror: true
  }
  data_param {
    source: "/user/a0393608/files/data/datasets/object-detect/other/ilsvrc/2012/ilsvrc12_train_lmdb"
    batch_size: 64
    backend: LMDB
  }
}

layer {
  name: "data"
  type: "Data"
  top: "data"
  top: "label"
  include {
    phase: TEST
  }
# mean pixel / channel-wise mean instead of mean image
  transform_param {
    crop_size: 224
    mean_value: 128
    mean_value: 128
    mean_value: 128
    mirror: false
  }
  data_param {
    source: "/user/a0393608/files/data/datasets/object-detect/other/ilsvrc/2012/ilsvrc12_val_lmdb"
    batch_size: 64
    backend: LMDB
  }
}
layer {
  name: "conv1"
  bottom: "data"
  top: "conv1"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  param { lr_mult: 2 decay_mult: 0 }
  convolution_param {
    num_output: 64
    kernel_size: 7
    pad: 3
    stride: 2
    weight_filler { type: "msra" std: 0.010 }
    bias_filler { type: "constant" value: 0 }
  }
}
layer {
  name: "bn_conv1"
  bottom: "conv1"
  top: "conv1"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "conv1_relu"
  bottom: "conv1"
  top: "conv1"
  type: "ReLU"
}
layer {
  name: "pool1"
  bottom: "conv1"
  top: "pool1"
  type: "Pooling"
  pooling_param {
    pool: MAX
    kernel_size: 3
    stride: 2
  }
}
layer {
  name: "res2a_branch2a"
  bottom: "pool1"
  top: "res2a_branch2a"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 64
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn2a_branch2a"
  bottom: "res2a_branch2a"
  top: "res2a_branch2a"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res2a_branch2a_relu"
  bottom: "res2a_branch2a"
  top: "res2a_branch2a"
  type: "ReLU"
}
layer {
  name: "res2a_branch2b"
  bottom: "res2a_branch2a"
  top: "res2a_branch2b"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 64
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn2a_branch2b"
  bottom: "res2a_branch2b"
  top: "res2a_branch2b"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
	name: "res2a"
	bottom: "pool1"
	bottom: "res2a_branch2b"
	top: "res2a"
	type: "Eltwise"
}
layer {
  name: "res2a_relu"
  bottom: "res2a"
  top: "res2a"
  type: "ReLU"
}
layer {
  name: "res2b_branch2a"
  bottom: "res2a"
  top: "res2b_branch2a"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 64
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn2b_branch2a"
  bottom: "res2b_branch2a"
  top: "res2b_branch2a"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res2b_branch2a_relu"
  bottom: "res2b_branch2a"
  top: "res2b_branch2a"
  type: "ReLU"
}
layer {
  name: "res2b_branch2b"
  bottom: "res2b_branch2a"
  top: "res2b_branch2b"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 64
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn2b_branch2b"
  bottom: "res2b_branch2b"
  top: "res2b_branch2b"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
	name: "res2b"
	bottom: "res2a"
	bottom: "res2b_branch2b"
	top: "res2b"
	type: "Eltwise"
}
layer {
  name: "res2b_relu"
  bottom: "res2b"
  top: "res2b"
  type: "ReLU"
}
layer {
  name: "res3a_branch2a"
  bottom: "res2b"
  top: "res3a_branch2a"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 128
    kernel_size: 3
    pad: 1
    stride: 2
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn3a_branch2a"
  bottom: "res3a_branch2a"
  top: "res3a_branch2a"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res3a_branch2a_relu"
  bottom: "res3a_branch2a"
  top: "res3a_branch2a"
  type: "ReLU"
}
layer {
  name: "res3a_branch2b"
  bottom: "res3a_branch2a"
  top: "res3a_branch2b"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 128
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn3a_branch2b"
  bottom: "res3a_branch2b"
  top: "res3a_branch2b"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res3a_branch1"
  bottom: "res2b"
  top: "res3a_branch1"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 128
    kernel_size: 1
    pad: 0
    stride: 2
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn3a_branch1"
  bottom: "res3a_branch1"
  top: "res3a_branch1"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
	name: "res3a"
	bottom: "res3a_branch1"
	bottom: "res3a_branch2b"
	top: "res3a"
	type: "Eltwise"
}
layer {
  name: "res3a_relu"
  bottom: "res3a"
  top: "res3a"
  type: "ReLU"
}
layer {
  name: "res3b_branch2a"
  bottom: "res3a"
  top: "res3b_branch2a"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 128
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn3b_branch2a"
  bottom: "res3b_branch2a"
  top: "res3b_branch2a"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res3b_branch2a_relu"
  bottom: "res3b_branch2a"
  top: "res3b_branch2a"
  type: "ReLU"
}
layer {
  name: "res3b_branch2b"
  bottom: "res3b_branch2a"
  top: "res3b_branch2b"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 128
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn3b_branch2b"
  bottom: "res3b_branch2b"
  top: "res3b_branch2b"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
	name: "res3b"
	bottom: "res3a"
	bottom: "res3b_branch2b"
	top: "res3b"
	type: "Eltwise"
}
layer {
  name: "res3b_relu"
  bottom: "res3b"
  top: "res3b"
  type: "ReLU"
}
layer {
  name: "res4a_branch2a"
  bottom: "res3b"
  top: "res4a_branch2a"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 256
    kernel_size: 3
    pad: 1
    stride: 2
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn4a_branch2a"
  bottom: "res4a_branch2a"
  top: "res4a_branch2a"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res4a_branch2a_relu"
  bottom: "res4a_branch2a"
  top: "res4a_branch2a"
  type: "ReLU"
}
layer {
  name: "res4a_branch2b"
  bottom: "res4a_branch2a"
  top: "res4a_branch2b"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 256
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn4a_branch2b"
  bottom: "res4a_branch2b"
  top: "res4a_branch2b"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res4a_branch1"
  bottom: "res3b"
  top: "res4a_branch1"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 256
    kernel_size: 1
    pad: 0
    stride: 2
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn4a_branch1"
  bottom: "res4a_branch1"
  top: "res4a_branch1"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
	name: "res4a"
	bottom: "res4a_branch1"
	bottom: "res4a_branch2b"
	top: "res4a"
	type: "Eltwise"
}
layer {
  name: "res4a_relu"
  bottom: "res4a"
  top: "res4a"
  type: "ReLU"
}
layer {
  name: "res4b_branch2a"
  bottom: "res4a"
  top: "res4b_branch2a"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 256
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn4b_branch2a"
  bottom: "res4b_branch2a"
  top: "res4b_branch2a"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res4b_branch2a_relu"
  bottom: "res4b_branch2a"
  top: "res4b_branch2a"
  type: "ReLU"
}
layer {
  name: "res4b_branch2b"
  bottom: "res4b_branch2a"
  top: "res4b_branch2b"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 256
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn4b_branch2b"
  bottom: "res4b_branch2b"
  top: "res4b_branch2b"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
	name: "res4b"
	bottom: "res4a"
	bottom: "res4b_branch2b"
	top: "res4b"
	type: "Eltwise"
}
layer {
  name: "res4b_relu"
  bottom: "res4b"
  top: "res4b"
  type: "ReLU"
}
layer {
  name: "res5a_branch2a"
  bottom: "res4b"
  top: "res5a_branch2a"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 512
    kernel_size: 3
    pad: 1
    stride: 2
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn5a_branch2a"
  bottom: "res5a_branch2a"
  top: "res5a_branch2a"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res5a_branch2a_relu"
  bottom: "res5a_branch2a"
  top: "res5a_branch2a"
  type: "ReLU"
}
layer {
  name: "res5a_branch2b"
  bottom: "res5a_branch2a"
  top: "res5a_branch2b"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 512
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn5a_branch2b"
  bottom: "res5a_branch2b"
  top: "res5a_branch2b"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res5a_branch1"
  bottom: "res4b"
  top: "res5a_branch1"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 512
    kernel_size: 1
    pad: 0
    stride: 2
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn5a_branch1"
  bottom: "res5a_branch1"
  top: "res5a_branch1"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
	name: "res5a"
	bottom: "res5a_branch1"
	bottom: "res5a_branch2b"
	top: "res5a"
	type: "Eltwise"
}
layer {
  name: "res5a_relu"
  bottom: "res5a"
  top: "res5a"
  type: "ReLU"
}
layer {
  name: "res5b_branch2a"
  bottom: "res5a"
  top: "res5b_branch2a"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 512
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn5b_branch2a"
  bottom: "res5b_branch2a"
  top: "res5b_branch2a"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
  name: "res5b_branch2a_relu"
  bottom: "res5b_branch2a"
  top: "res5b_branch2a"
  type: "ReLU"
}
layer {
  name: "res5b_branch2b"
  bottom: "res5b_branch2a"
  top: "res5b_branch2b"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  convolution_param {
    num_output: 512
    kernel_size: 3
    pad: 1
    stride: 1
    bias_term: false
    weight_filler { type: "msra" std: 0.010 }
    dilation: 1    
    group: 1
  }
}
layer {
  name: "bn5b_branch2b"
  bottom: "res5b_branch2b"
  top: "res5b_branch2b"
  type: "BatchNorm"
  param { #scale
    lr_mult: 1
    decay_mult: 1
  }
  param { #shift/bias
   lr_mult: 1
    decay_mult: 1
  } 
  param { #global mean
    lr_mult: 0
    decay_mult: 0
  }
  param { #global var
   lr_mult: 0
    decay_mult: 0
  }     
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
    engine: CUDNN
  }
}
layer {
	name: "res5b"
	bottom: "res5a"
	bottom: "res5b_branch2b"
	top: "res5b"
	type: "Eltwise"
}
layer {
  name: "res5b_relu"
  bottom: "res5b"
  top: "res5b"
  type: "ReLU"
}
layer {
  name: "pool5"
  bottom: "res5b"
  top: "pool5"
  type: "Pooling"
  pooling_param {
    pool: AVE
    kernel_size: 7
    stride: 1
  }
}
layer {
  name: "conv6"
  bottom: "pool5"
  top: "conv6"
  type: "Convolution"
  param { lr_mult: 1 decay_mult: 1 }
  param { lr_mult: 2 decay_mult: 0 }
  convolution_param {
    num_output: 1024
    kernel_size: 1
    pad: 0
    stride: 1
    weight_filler { type: "msra" std: 0.010 }
    bias_filler { type: "constant" value: 0 }
  }
}
layer {
  name: "loss"
  type: "SoftmaxWithLoss"
  bottom: "conv6"
  bottom: "label"
  propagate_down: 1
  propagate_down: 0
  top: "loss"
  loss_weight: 1
}

layer {
  name: "accuracy"
  type: "Accuracy"
  bottom: "conv6"
  bottom: "label"
  top: "accuracy"
  include {
    phase: TEST
  }
}
layer {
  name: "accuracy/top-5"
  type: "Accuracy"
  bottom: "conv6"
  bottom: "label"
  top: "accuracy/top-5"
  include {
    phase: TEST
  }
  accuracy_param {
    top_k: 5
  }
}

}

mathmanu:
Were you able to reproduce the behavior?

borisgin:
I was able to reproduce the problem: caffe engine is converging, but cudnn
BN diverges
Btw, I would add two parameters to BN layer definition:
  moving_average_fraction: 0.9
  eps: 0.0001

mathmanu:
Thankyou so much for confirming. I'll wait for further information from you.

mathmanu:
Btw, I can see that the additional parameters that you mentioned have default values - so technically I don't need to specify them. Do the values suggested by you produce better accuracy?

  // How much does the moving average decay each iteration?
  optional float moving_average_fraction = 2 [default = .999];
  // Small value to add to the variance estimate so that we don't divide by
  // zero.
  optional float eps = 3 [default = 1e-5];

mathmanu:
Is there a quick fix or a workaround that can solve this issue?
Thanks,

borisgin:
Yes. CUDNN_BN requires different top and bottom since it needs bottom for backward. I attached the fixed train_val.prototxt (I also did a few additional minor changes: change last layer from conv 1x1 to IP with num_of_outputs=1000). 

[train_val.txt](https://github.com/BVLC/caffe/files/501984/train_val.txt)

mathmanu:
Thanks. This helped a lot.

Btw, I have an observation. I have a network trained with the CAFFE BN engine. When I tried to TEST it using CUDNN BN engine, Caffe exited saying that the shapes of blobs in BN mismatch. But since the blobs are same sime (but different shapes), I was able to forcefully reshape the blobs and do the TEST - and it gave correct results!

borisgin:
My bug :(

mathmanu:
Don't worry - CUDNN is great - it gives me 4x speed boost - that makes a huge difference.

You just needs to do bit more testing before submitting a PR - cross compatibility test to CAFE engine and backward compatibility test.

mathmanu:
See another thread with a similar issue being reported: https://github.com/NVIDIA/DIGITS/issues/629

achaiah:
@borisgin   Out of curiosity, have you tried larger networks than Resnet-18 on NVIDIA?  Your resnet-18 is the only one that converges for me while I can't find a single example of a larger Resnet that does.

PR #3929, opened 2016-04-01
Fix compile warning with -Wpedantic flag

oliviersoares:
Small change to fix warnings like these:
[...]/caffe/include/caffe/util/mkl_alternate.hpp:34:47: warning: extra ‘;’ [-Wpedantic]
 DEFINE_VSL_UNARY_FUNC(Sqr, y[i] = a[i] \* a[i]);

when compiling some code including caffe headers and using gcc -Wpedantic or -pedantic flag.
Thanks.


shelhamer:
Thanks for contributing the fix! An equivalent patch was merged in #5380.

PR #3937, opened 2016-04-04
Issue with the exp layer when the base is e

emaggiori:
This fixes a bug in the exp layer. In the case of the base e (set by the default parameter base==-1), the outer_scale_ variable was computed by taking -1 as a base instead of e.

The new code uses exp(input_shift) instead of pow(base, input_shift) when the base is set to -1.

### DIFF
--- a/src/caffe/layers/exp_layer.cpp
+++ b/src/caffe/layers/exp_layer.cpp
@@ -23,7 +23,8 @@ void ExpLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
   const Dtype input_scale = this->layer_param_.exp_param().scale();
   const Dtype input_shift = this->layer_param_.exp_param().shift();
   inner_scale_ = log_base * input_scale;
-  outer_scale_ = (input_shift == Dtype(0)) ? Dtype(1) : pow(base, input_shift);
+  outer_scale_ = (input_shift == Dtype(0)) ? Dtype(1) :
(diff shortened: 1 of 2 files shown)


jeffdonahue:
Thanks @emaggiori, this does seem like a bug, yet I did write [tests](https://github.com/BVLC/caffe/blob/master/src/caffe/test/test_neuron_layer.cpp#L379) for what I think should be this exact case, and those are clearly passing.  So I'm not sure what's going on yet...any ideas?  Ideally this would include a new test demonstrating the bug and that it's now fixed & passing.

jeffdonahue:
Ah... I see now -- the bug is only triggered when `shift` is non-zero, which is not tested.  (That makes more sense because I don't think I've ever used that parameter...).  Could you add a new `ExpLayer` test for that case, after the existing ones I linked in the previous comment?

(This is just one more piece of evidence in favor of modularity -- and in particular in favor of splitting off the scale/shift params from `Exp` and `Power` layer, as @longjon has argued.

emaggiori:
Hi, I added the tests. Let me know if this is okay.

jeffdonahue:
Great, thanks @emaggiori! Please squash your changes to a single commit and I'll merge this.

emaggiori:
Thanks @jeffdonahue, I squashed them to a single commit as suggest.

jeffdonahue:
Looks good, thanks again @emaggiori.

PR #3942, opened 2016-04-04
Set propagate_down=true to force backprop to a particular bottom

jeffdonahue:
Currently setting propagate_down=true has no effect -- only setting propagate_down=false has an effect (disabling backpropagation when it would otherwise be enabled).  This PR changes the behavior of propagate_down=true -- if set, backpropagation is forced even if it's not needed, sort of like a per-bottom `force_backward=true` setting.

### DIFF
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -427,12 +427,11 @@ int Net<Dtype>::AppendBottom(const NetParameter& param, const int layer_id,
   bottom_vecs_[layer_id].push_back(blobs_[blob_id].get());
   bottom_id_vecs_[layer_id].push_back(blob_id);
   available_blobs->erase(blob_name);
-  bool propagate_down = true;
+  bool need_backward = blob_need_backward_[blob_id];
   // Check if the backpropagation on bottom_id should be skipped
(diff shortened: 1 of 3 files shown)


jeffdonahue:
It's hard to tell from the reordered history here on GitHub, but the `add TestForcePropagateDown` commit (dee01c8) is first and the test should fail since the behavior isn't implemented.  The second commit (0f36571) is the one that actually updates the behavior and the tests should then pass.

jeffdonahue:
Hm, well, for some reason the tests here both passed, but on my repo's Travis, as expected, the first commit (dee01c8) [failed](https://travis-ci.org/jeffdonahue/caffe/builds/120697346) and the second commit (0f36571) [passed](https://travis-ci.org/jeffdonahue/caffe/builds/120697950).

shelhamer:
Thanks Jeff, this looks good except that the proto comment for `propagate_down` should reflect that it now forces both cases whether propagating by `true` or skipping by `false`. Feel free to merge once that is included.

jeffdonahue:
Thanks for the review Evan! Proto comment updated.

Issue #3420, opened 2015-12-06
Is this a BUG in blob->offset(n) ?


Coldmooon:
Say the input shape is 10 * 192 * 32 * 32.
Then:
bottom[0]->offset(0) will be 0.
bottom[0]->offset(1) will be 196608.
bottom[0]->offset(2) will be 393216.
...
bottom[0]->offset(9) will be 1769472.
bottom[0]->offset(10) will be 1966080.
There isn't bottom[0]->offset(11) .

The last one in a batch is in offset(9) because we start at zero. So,  I think there shouldn't be offset(10). 
If we read bottom[0]->cpu_data()[1966080], the code will exit without any error prompt.

seanbell:
Yes, this is a bug and it should be `CHECK_LT(n, num())`.  However, in order to fix this, you will need to also fix a few other places that depend on the current behavior of `offset`.  While this bug has been pointed out many times, this is the reason why it's not yet fixed.  See #792, #1681, #2835.

Coldmooon:
@seanbell This is a big project and needs programmers to well understand how layers work. I just found this bug in "prelu_layer.cu". By the way, could you tell me how you track issues and quickly know what issues are related to the current issue and should be referred? Because #792, #1681, #2835 you listed are very helpful to me. Thank you ~~

seanbell:
I just searched the project for issues, using the builtin GitHub search.  I forget what search terms I used.

Coldmooon:
@seanbell Very helpful~Thanks~

Noiredd:
There are two open PRs which aim to fix this: #2835 and #6021. There was also #792 which IMHO was better as it also checked `>=0` (currently the check is placed on the blob dimensions instead of arguments, which I think is a mistake).  
Both open PRs contain an identical patch, which does not pass the tests. Turns out that other code depends on this broken behaviour and the change is not as simple as that.

Chessing234:
Yeah, CHECK_LE(n, num()) should be CHECK_LT — same for c/h/w. Opened a PR for it.

Issue #1681, opened 2015-01-06
offset method in blob.hpp needs more precise checking


anguyen8:
I came across the offset method in the **blob.hpp** and realised that the checking in the **inline int offset(...);** method are a bit loose:
```
inline int offset(const int n, const int c = 0, const int h = 0,
      const int w = 0) const {
    CHECK_GE(n, 0);
    CHECK_LE(n, num_);    // <----
    CHECK_GE(channels_, 0);  
    CHECK_LE(c, channels_);   // <----
    CHECK_GE(height_, 0);  
    CHECK_LE(h, height_);    // <----
    CHECK_GE(width_, 0); 
    CHECK_LE(w, width_);   // <----
    return ((n * channels_ + c) * height_ + h) * width_ + w;
  }
```

A channel index can be from 0 to *channels_ - 1* so it should check if the index is ```>= 0 && < channel_```. I think a series of CHECKs here need to be revised to be more precise, for example:

```
CHECK_GE(channels_, 0);
CHECK_LT(c, channels_);   // <----
```

shelhamer:
See #792.

PR #3951, opened 2016-04-06
Fix Blob Offset Function

D-X-Y:
n, c,h,w should be strictly less than num() channels() height() width()


seanbell:
See https://github.com/BVLC/caffe/issues/3420#issuecomment-162336911.

See also #3420, #2835, #1681, #792.

longjon:
Closing per @seanbell's comment, thanks anyway.

PR #3954, opened 2016-04-06
Fix initialization of deconvolution layer parameters in python net_spec interface

SvenTwo:
Currently, you cannot initialize deconvolution layer parameters in the same way you can for a convolution layer, because the associated _param structure is called convolution_param instead of deconvolution_param.

E.g. the following throws an unknown attribute error:
deconv_layer = L.Deconvolution(bottom_layer, kernel_size=4, stride=2, num_output=25)

Solve the problem by adding the special case to the dictionary mapping layer types to parameters in param_name_dict().

### DIFF
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -37,7 +37,10 @@ def param_name_dict():
     # strip the final '_param' or 'Parameter'
     param_names = [s[:-len('_param')] for s in param_names]
     param_type_names = [s[:-len('Parameter')] for s in param_type_names]
-    return dict(zip(param_type_names, param_names))
+    result = dict(zip(param_type_names, param_names))
+    # convolution param is re-used in deconvolution layer
+    result['Deconvolution'] = 'convolution'
+    return result
 
 
 def to_proto(*tops):


SvenTwo:
Btw, simply doing convolution_param=dict(...) works of course. But I find it a bit annoying that the conv and deconv layers cannot be used symmetrically.

seanbell:
@shelhamer Since this is causing a reasonable amount of confusion, perhaps it could be merged in?

longjon:
It's possible to special case things like this, but I'm not sure it's a good idea... it's contrary to the way it's written in the prototxt, and this is not the only case where this happens, and the user has to figure out where these special cases exist, and the code has to be synchronized between different places, and it will become moot (or have to be revisited) if conv/pool params are unified.

That said this is a common and sensible case, so I'm happy to hear other opinions.

nitnelave:
Where is it written that the deconvolution uses the convolution param? In the prototxt? In that case, it could be nice to inspect it instead of just manipulating the names to create the param names.

SvenTwo:
@nitnelave You can look it up in the automatically generated documentation from the source code here: http://caffe.berkeleyvision.org/doxygen/classcaffe_1_1DeconvolutionLayer.html#details

I usually read things directly in the source code because layer-specific caffe documentation isn't that great ;-)

nitnelave:
@SvenTwo Yeah, I mean, where in the code is it written that Deconvolution uses the convolution param? Is it just in the prototxt? I had a quick look, I didn't find anything relevant.

SvenTwo:
@nitnelave It's implicit because the deconv layer class inherits from the convolution layer class, which handles the setup from parameters.

nitnelave:
Would it be possible to add some code to `REGISTER_LAYER_CLASS` to have a list of layers accessible somewhere (ultimately to python), with their corresponding parameters where relevant?
The default would be `LayerParameter`, of course, but we could add a specific parameter name for cases such as this.

SvenTwo:
Isn't that a bit overengineered for that one case? I just proposed one line of code change that shortens the definition of deconv layers a bit...

nitnelave:
I know and I agree with you, that's a good fix for now. But in the future,
you would have to list all the layers that share parameters, which
complicates the definition of a new layer.

So, to be more error resilient, I think we need some registration at the
layer level. Plus, a list of existing layers would be useful in python.

melody-rain:
Any other better implementation for
```
# convolution param is re-used in deconvolution layer
    result['Deconvolution'] = 'convolution'
```
?

PR #3955, opened 2016-04-06
Add external library data layer

sasagalic-MSFT:
Currently there are a number of data layers available in Caffe supporting different input formats. However, there is always one more format which is optimal for the given training.  
One cannot expect Caffe to support all the formats, but fairly close solution would be to implement extensible architecture which enables easy integration of external formats with Caffe.  
In this change external library data layer is implemented. This data layer expects external library (Shared Object on Linux or DLL on Windows) to provide data for Caffe network.  
Using this data layer has several benefits:
- Decouples data source from Caffe. New formats are easily consumable and do not require changes in Caffe codebase
- Enables in-memory synthetic data generation
- Enables multiple top blobs (current prefetching data layers only support 2 tops, which is a limitation for some trainings)
- Opens possibility to implement some of the existing data layers (LMDB, LevelDB etc.) as external libraries to reduce number of dependencies Caffe has

This change is implemented with compatibility in mind. All existing data layers maintain the same behaviour. External library adapters are provided for Windows and Linux to localize platform specific code.

### DIFF
--- a/include/caffe/util/external_lib_data_source.hpp
+++ b/include/caffe/util/external_lib_data_source.hpp
@@ -1,0 +1,51 @@
+/**
+* @brief    This file contains interface declarations that external data source
+*           needs to implement to be able to plug into caffe pipeline. External
+*           library needs to export the factory method with the signature
+*           defined below.
+*           To instruct caffe to use external lib as data source one needs to
+*           use external lib data layer.
+*/
+#ifndef EXTARNAL_LIB_DATA_SOURCE_H_
+#define EXTARNAL_LIB_DATA_SOURCE_H_
+
+class IExternalLibDataSource;
+
+// External library data source factory method signature.
+typedef IExternalLibDataSource* (*ExternalLibDataSourceFactoryMethod)(
+  const char* external_lib_params);
+
+enum BlobType { BlobTypeFLOAT, BlobTypeDOUBLE };
+
+/**
+* @brief    Defines interface for one data point.
+*/
+class IDatum {
+ public:
+  // Returns datum shape for the blob with the given name.
+  virtual void GetBlobShape(const char* blob_name, const int** shape,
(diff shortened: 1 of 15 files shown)


willyd:
Could this be made even more general? Allowing not only plugin-like data layers, but also compute layers?

include/caffe/util/external_lib_data_source.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+  // Returns data of the given type for the blob with the given name.
+  virtual void GetBlobData(const char* blob_name, void* blob_data,
+    BlobType type) = 0;
+};
+
+/**
+* @brief    Defines interface for data source.
+*/
+class IExternalLibDataSource {
```
hgaiser: Why did you create additional interfaces? Caffe already has an interface for data layers: `BaseDataLayer<Dtype>`. Why not let the loaded module create instances of that class instead of the new interfaces? I'd expect that would save a lot of code overhead. On top of that I think it would also simplify the process for supporting compute layers.

include/caffe/util/external_lib_data_source_util.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+class IExternalLibDataSource;
+
+/**
+* @brief    Class that abstracts away platform dependent library handling.
+*/
+class IExternalLib {
+ public:
+  virtual ~IExternalLib() {}
+  virtual IExternalLibDataSource* GetDataSource() = 0;
```
hgaiser: It would be nice if the class for loading symbols/functions from a library could be generalized (possibly through templating) to support loading functions with any signature. Then it could be reused for loading compute layers as well.

shelhamer:
Closing in favor of the more general approach in #5294. Thanks @sasagalic-MSFT for your concern about how to handle all the different varieties of data out there.

PR #3977, opened 2016-04-11
Fix protobuf message generation

tpwrules:
The latest versions of protobuf do not reveal empty message fields with dir(). This uses the documented way of determining all of a message's fields and so is compatible with past and future versions of protobuf.


ajtulloch:
LGTM. Any objections here @shelhamer ? 

http://stackoverflow.com/questions/24639562/getting-all-field-names-from-a-protocol-buffer
https://developers.google.com/protocol-buffers/docs/reference/python/google.protobuf.descriptor.Descriptor-class

have some more documentation on this if anyone was interested.

longjon:
Ah, nice. I was not aware that there was a documented way to do this. Thanks!

PR #3978, opened 2016-04-12
Fix create_mnist_data.exe for windows execution

ChaseLewis:
1) Adjust default lmdb size since it currently allocates 1TB of memory for the mnist database.
2) Add .bat file for downloading and unpacking the mnist database.
3) Add .bat file to run create_mnist_database.exe from default build directories
4) Add .bat file to run caffe.exe on the newly created lmdb database.


shelhamer:
@ChaseLewis Caffe is only compatible with Windows through its `windows` branch. You should coordinate any windows contributions with @pavlejosipovic on that branch and send PRs there rather than `master`.

Closing since this PR is directed to `master`.

PR #3984, opened 2016-04-12
Generalize bilinear filler to for N-D multilinear filler

christianpayer:
This branch changes the bilinear filler (#2213) to support an arbitrary number of dimensions. Also, different scaling factors for each dimension are supported. I changed the name of the filler from "bilinear" to "linear", as it suits better for its multi-dimensional functionality.

The usage of this filler is the same as before (see comments, or #2213). Tests are not included, but I plan to write some using python scikit-image.

### DIFF
--- a/include/caffe/filler.hpp
+++ b/include/caffe/filler.hpp
@@ -208,11 +208,11 @@ class MSRAFiller : public Filler<Dtype> {
 };
 
 /*!
-@brief Fills a Blob with coefficients for bilinear interpolation.
+@brief Fills a Blob with coefficients for multilinear interpolation.
 
 A common use case is with the DeconvolutionLayer acting as upsampling.
-You can upsample a feature map with shape of (B, C, H, W) by any integer factor
-using the following proto.
+You can upsample a feature map with shape of (B, C, S_n,..., S_1) by any
+integer factor using the following proto.
 \code
 layer {
   name: "upsample", type: "Deconvolution"
@@ -221,7 +221,7 @@ layer {
     kernel_size: {{2 * factor - factor % 2}} stride: {{factor}}
     num_output: {{C}} group: {{C}}
     pad: {{ceil((factor - 1) / 2.)}}
-    weight_filler: { type: "bilinear" } bias_term: false
+    weight_filler: { type: "multilinear" } bias_term: false
   }
   param { lr_mult: 0 decay_mult: 0 }
 }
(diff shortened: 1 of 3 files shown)


shelhamer:
While there is no reason this filler should be limited to 4-D, as it is now, this breaks backward compatibility as-is since there are existing definitions that make use of `bilinear`. See our [upgrade proto](https://github.com/BVLC/caffe/blob/master/src/caffe/util/upgrade_proto.cpp) code for an idea of how to automatically upgrade old definitions -- this is our practice for schema changes.

See #2213 for some earlier discussion of the naming. `linear` doesn't feel quite right since that could be nearly anything, although `bilinear` is admittedly over specific. Perhaps `interp` would do? The combination of a linear layer like `Convolution` and an `interp` filler might be clear enough.

Thoughts @longjon?

longjon:
In the usual interpolation sense (i.e., as a function of _coordinates_), _linear_ is not correct; _bilinear_ functions are _nonlinear_ functions which are separately linear in each of their arguments. `interp` is too generic, as there are non-bilinear interpolation methods (e.g., biquadratic or bicubic) which can be implemented using fillers and convolutions.

The generalization of bilinear is _multilinear_. Perhaps these could be accepted as synonyms.

christianpayer:
Thanks for your suggestions. I agree that `linear` is not the best name and that `multilinear` is much better.
If you agree to use `multilinear` I will change it and adapt the code for upgrade proto.

include/caffe/filler.hpp
```diff
@@ -229,32 +229,39 @@ layer {
...
-class BilinearFiller : public Filler<Dtype> {
+class LinearFiller : public Filler<Dtype> {
  public:
-  explicit BilinearFiller(const FillerParameter& param)
+  explicit LinearFiller(const FillerParameter& param)
       : Filler<Dtype>(param) {}
   virtual void Fill(Blob<Dtype>* blob) {
-    CHECK_EQ(blob->num_axes(), 4) << "Blob must be 4 dim.";
-    CHECK_EQ(blob->width(), blob->height()) << "Filter must be square";
+    CHECK_GT(blob->num_axes(), 3) << "Blob must have at least 3 dimensions.";
```
seanbell: Shouldn't this be `CHECK_GE`, if you want to encode "at least 3"?

include/caffe/filler.hpp
```diff
@@ -229,32 +229,39 @@ layer {
...
-      float y = (i / blob->width()) % blob->height();
-      data[i] = (1 - fabs(x / f - c)) * (1 - fabs(y / f - c));
+      unsigned int stride = 1;
+      Dtype weight = 1;
+      for (int axis = 0; axis < blob->num_axes() - 2; ++axis) {
+        unsigned int shape = blob->shape(axis + 2);
+        unsigned int sampling_factor = std::ceil(shape / 2.0f);
+        float center = (2 * sampling_factor - 1 - sampling_factor % 2)
+            / (2.0f * sampling_factor);
+        float coordinate = (i / stride) % shape;
```
seanbell: Since this gets saved into `data` which has type `Dtype`, maybe the `float` should all be `Dtype`?

include/caffe/filler.hpp
christianpayer: You are right, thanks for pointing that out. I will change it. I never tested the code for 1d inputs, but it should work, too.

include/caffe/filler.hpp
christianpayer: Thanks, I agree. Using Dtype would remove the implicit conversion for weight in the following line.

include/caffe/filler.hpp
```diff
@@ -229,32 +229,39 @@ layer {
...
-    float c = (2 * f - 1 - f % 2) / (2. * f);
     for (int i = 0; i < blob->count(); ++i) {
-      float x = i % blob->width();
-      float y = (i / blob->width()) % blob->height();
-      data[i] = (1 - fabs(x / f - c)) * (1 - fabs(y / f - c));
+      unsigned int stride = 1;
+      Dtype weight = 1;
+      for (int axis = 0; axis < blob->num_axes() - 2; ++axis) {
+        unsigned int shape = blob->shape(axis + 2);
+        unsigned int sampling_factor = std::ceil(shape / 2.0f);
```
seanbell: Calling this `factor` instead of `sampling_factor` would better fit with the above block comment that explains how it works.

shelhamer:
@christianpayer I second `multilinear`, so if you agree then go ahead with the new name.
Once done, please squash your commits into a new commit with a message that reflects the new filler (in particular the new name and generality of dimensions). Thanks!

christianpayer:
@shelhamer Thanks for the comments. I changed the name to `multilinear`, squashed the commit and changed the commit message.
I did not edit upgrade_proto.cpp, as I am not sure, where/how to change the code there. As far as I can see an upgrade of a filler name has not been done before. Do you want me to implement something like
```
bool NetNeedsWeightFillerUpgrade(const NetParameter& net_param);
void UpgradeNetWeightFiller(NetParameter* net_param);
```
in upgrade_proto.hpp/cpp?

shelhamer:
@christianpayer adding the check `NetNeedsWeightFIllerUpgrade()` and update `UpgradeNetWeightFiller()` is the idea. You'll need to be a bit familiar with the C++ interface to protobuf, but hopefully the existing methods can be a guide. Here's a quick sketch to illustrate key pieces:

```
for (int i = 0; i < net_param->layer_size(); ++i) {
      if (net_param->layer(i).has_convolution_param()) {
          FillerParameter* filler_param = net_param->mutable_layer(i)->mutable_convolution_param()->mutable_filler_parameter();
          if (filler_param.type() == "bilinear") {
              filler_param->set_type("multilinear");
          }
      }
}
```

Care to take a stab at it and we'll go from there?

christianpayer:
@shelhamer Thanks for the code snippet. I just implemented upgrade_proto for weight fillers and hope that this is what you had in mind.

christianpayer:
Please do not merge this PR, as I implemented in the meantime a more general interpolation filler, that allows linear/cubic/lanczos interpolation and rewrites most of the code of this PR.
Next week I will create a new PR with this more general filler.

christianpayer:
Closed in favour of #4198.

PR #3992, opened 2016-04-14
Update ipython notebook to do better visualization of weights.

warmspringwinds:
I think that by making the visualization of weights using 'nearest' interpolation will make it better.

Example of current visualization:

caffe_pull_before

With 'nearest' interpolation:

caffe_pull_after


shelhamer:
Refer to cell no. 1 of the example notebook https://github.com/BVLC/caffe/blob/master/examples/00-classification.ipynb and in particular `plt.rcParams['image.interpolation'] = 'nearest'  # don't interpolate: show square pixels`.

This makes all the plotting in the notebook use nearest neighbor interpolation. Is this not happening in your environment?

warmspringwinds:
@shelhamer Oops. I have just pulled that function from original ipython notebook to my code.

Sorry for bothering then :)

PR #3993, opened 2016-04-15
Fix Crop layer dimension checking to only check cropped dimensions

shelhamer:
The `Crop` layer should only check the actual dimensions to be cropped for shape compatibility given the crop size and offset.

The current, over-general check results in issues like refusing to crop the output for a binary classification task given color input, since the data has 3 channels while the output has 2. As the channels are not cropped (in this use case) that is irrelevant and wrong.

This PR fixes the check, makes the message clearer, and revises the comments here and there.


shelhamer:
Reported with fix in #3778 by wk910930—thanks for noting the issue!

Issue #3506, opened 2016-01-03
Extend pycaffe to expose LayerParameters in Python layer


shaibagon:
Is there a way to access LayerParameters in python code implementing a "Python" layer?
Specifically, I would like to know the "phase" I'm in while instantiating the layer.

Please see my question at [stackoverflow][1], and in the [caffe users][2] google group.

 [1]: http://stackoverflow.com/q/34549743/1714410
 [2]: https://groups.google.com/forum/#!topic/caffe-users/uiZ2jy0QOdI

rohrbach:
Closing as this is a usage issue (and you correctly asked on stackoverflow and caffe users, where it has been answered). Please don't post these things here.

Please ask usage questions on caffe-users: https://groups.google.com/forum/#!forum/caffe-users. 
If you do think there is a bug in Caffe, you're welcome to open a new issue according to https://github.com/BVLC/caffe/wiki/Reporting-Bugs-and-Other-Issues. Thanks!

shaibagon:
@rohrbach I am aware of caffe's "Reporting Bugs and Other Issues" policy. Getting access to LayerParameter in python is a feature request and not a usage issue. The answer provided at stack overflow is a very useful workaround.
It would still be nice to have access to LayerParameters in python. I am aware this is a very low priority issue.

Thanks.

shelhamer:
@shaibagon thanks for the clarification this is a feature request -- I've slightly re-titled and re-opened this issue. See #2149 for one approach.

shaibagon:
@shelhamer thank you for reviving this issue.

shaibagon:
I see [this PR](https://github.com/BVLC/caffe/pull/3995) related to this issue: exposing phase to python code.

shelhamer:
Closing as merged in #3995. `PythonLayer` can now check phase.

shaibagon:
@shelhamer thank you very much!

PR #3995, opened 2016-04-15
Allow the python layer have attribute "phase"

ZhouYzzz:
I noticed that the python layer cannot tell whether the phase is TRAIN or TEST. This change may bring convenience in some cases. e.g. to me.

    class MyLayer(caffe.Layer):
        ... ...
            if (self.phase == caffe.TRAIN):
                # do sth.
            else:
                # do sth. else
        ... ...

### DIFF
--- a/python/caffe/test/test_python_layer.py
+++ b/python/caffe/test/test_python_layer.py
@@ -44,6 +44,18 @@ class ParameterLayer(caffe.Layer):
     def backward(self, top, propagate_down, bottom):
         self.blobs[0].diff[0] = 1
 
+class PhaseLayer(caffe.Layer):
+    """A layer for checking attribute `phase`"""
+
+    def setup(self, bottom, top):
+        pass
+
+    def reshape(self, bootom, top):
+        top[0].reshape()
+
+    def forward(self, bottom, top):
+        top[0].data[()] = self.phase
+
 def python_net_file():
     with tempfile.NamedTemporaryFile(mode='w+', delete=False) as f:
         f.write("""name: 'pythonnet' force_backward: true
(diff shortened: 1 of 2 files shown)


ajtulloch:
Looks fine to me.  Would you be able to add a test that demonstrates reading out the phase with TRAIN/TEST values as well?

shaibagon:
related to issue #3506 -- thanks for implementing this PR!

ajtulloch:
Looks great to me. Can you squash the commits? @longjon , @shelhamer  does this look good to you?

python/caffe/test/test_python_layer.py
```diff
@@ -44,6 +44,23 @@ def forward(self, bottom, top):
...
         self.blobs[0].diff[0] = 1
 
+class PhaseLayer(caffe.Layer):
+    """A layer for checking attribute `phase`"""
+
+    def setup(self, bottom, top):
+        pass
+
+    def reshape(self, bootom, top):
+        top[0].reshape(1)
```
longjon: Scalars should be 0d; use `top[0].reshape()`.

python/caffe/test/test_python_layer.py
```diff
@@ -44,6 +44,23 @@ def forward(self, bottom, top):
...
+
+    def forward(self, bottom, top):
+        if (self.phase == caffe.TRAIN):
+            top[0].data[...] = 0 # caffe.TRAIN
```
longjon: Please never hardcode the values of symbolic constants (note that these really need to be correct, because they're checked later).

Why not replace this whole forward with `top[0].data[()] = self.phase`?

python/caffe/test/test_python_layer.py
```diff
@@ -140,3 +165,14 @@ def test_parameter(self):
         self.assertEqual(layer.blobs[0].data[0], 1)
 
         os.remove(net_file)
+
+    def test_phase(self):
+        net_file = phase_net_file()
+        # Test on phase TRAIN
```
longjon: You can drop these comments; they don't add anything to the code.

(And if you like, make this block a short loop: `for phase in caffe.TRAIN, caffe.TEST:`...)

longjon:
I never really got much into phase myself, but this is a clear improvement for the completeness of Python layers, so it seems worthwhile.

However, there are a few style issues, as noted. I'd be happy to merge once these are fixed (and squashed into the existing two commits).

ZhouYzzz:
@longjon Thanks for review and I've fixed it. I'm glad this pr is helpful.

longjon:
Looks very clean now, thanks!

PR #4013, opened 2016-04-19
Python LMDB Creator

ssudholt:
This PR adds a LMDB creation class in python. This allows for (almost) arbitrary Numpy arrays to be inserted in this lightweight database, bypassing the need to first save the arrays as images and then creating a LMDB by calling convert_imageset.

For tasks in which multi-dimensional labels are needed (e.g. multi-label classification, semantic segmentation,...), the LMDBCreator offers the possibility to create a dual LMDB. Internally, this wraps creating two LMDBs and inserting corresponding elements with identical keys. These two LMDBs can then be included in standard proto files.

The PR calls for an additional dependency in pylmdb which can be easily installed by pip.

### DIFF
--- a/python/caffe/lmdb_creator.py
+++ b/python/caffe/lmdb_creator.py
@@ -1,0 +1,281 @@
+import os
+import shutil
+import logging
+
+import numpy as np
+import lmdb
+import caffe.io
+
+class LMDBCreator(object):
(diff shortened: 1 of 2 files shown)


ssudholt:
If adding the pylmdb dependency is of no concern, this should be read for review

PR #4020, opened 2016-04-20
added pyo files to gitignore

ssudholt:
When running python with PYTHONOPTIMIZED=1, it generates pyo files which this PR adds to the global ignore file.


seanbell:
LGTM

seanbell:
It looks like you added some unnecessary extra commits -- please squash them back into a single commit.

ssudholt:
Oops sry, something broke during rebase

seanbell:
I think this is ready to merge.

@shelhamer

longjon:
Well, I don't see why we can't add this line... but may I ask: are these optimizations actually serving any useful purpose? I do not normally compile to `pyo`, and in fact the extension is removed by PEP 488 (and I'd rather not encourage running our Python code without assertions...)

ssudholt:
We use it to deactivating __debug__ parts of our code and I was generally under the impression that it ran a bit faster even when not considering debug related stuff. If I understand correctly, PEP 488 only applies for 3.4+ (I'm using 2.7).
You make a valid point though about deactivating asserts. Personally, I have never experienced any problems with pycaffe and an optimized Python environment.

PR #3731, opened 2016-02-26
LMDB map size - double when full

lukeyeager:
*Close #2404, close #3644, close #3728, close #3730*
*Also #1298, #1861, #2293, #2709*

This change helps users on Windows and/or 32-bit systems by removing the 1TB hard-coded map size.

PR #4021, opened 2016-04-20
Fixed integer overflow issue with LMDB on 32bit systems.

nickfraser:
I tried compiling Caffe on the Xilinx ZC706 evaluation board and noticed it wouldn't pass all of the tests. I then noticed that I have the same error when compiling Caffe on a 32bit Ubuntu 14.04 system (x86). It occurs with and without debugging enabled. It's likely this bug occurs in any system which uses an address size less than 41bits. I traced back the 1TB LMDB map size allocation in a few files. This 1TB allocation assumes that the underlying size_t parameter needed by LMDB is at least 41bits - this is not true on 32bit systems.

In order to address (if you'll excuse the pun) this, I did the following:
1) I created a macro to saturate the DB allocation size to the largest value available on the current system, if the user specifies a larger value. (Perhaps this should also include a warning? Either way, this is more desirable than overflow in my opinion...)
2) I added an extra variable in Makefile.config.example to allow the user to specify the desired allocation size. I added the definition of this symbol to the COMMON_FLAGS parameter in the Makefile.

The code should be self-explanatory, but let me know if you want any more information.

Also, if you have any suggestions on how I can improve my code - please let me know. Even if it's just a style preference - I'd like to hear about it.

Cheers,
Nick

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,12 +184,14 @@ LIBRARIES += glog gflags protobuf boost_system boost_filesystem m hdf5_hl hdf5
 USE_LEVELDB ?= 1
 USE_LMDB ?= 1
 USE_OPENCV ?= 1
+LMDB_MAP_SIZE ?= 1099511627776
 
 ifeq ($(USE_LEVELDB), 1)
 	LIBRARIES += leveldb snappy
 endif
 ifeq ($(USE_LMDB), 1)
 	LIBRARIES += lmdb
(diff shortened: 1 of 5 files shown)


seanbell:
Please see #3731, which is a more general solution to the problem.

nickfraser:
Apologies, I didn't see that pull request. I agree it's a better solution. However, the simplicity of this solution is less likely to create insidious bugs - if it were used temporarily until #3731 is merged.

longjon:
Closing as #3731 has been merged... comment if there's an outstanding issue that I've missed.

PR #4024, opened 2016-04-20
Explicitly point out -weights flag in fine-tuning tutorial

achalddave:
The -weights flag is somewhat easy to miss as it's only in one command,
but is the crucial thing that anyone searching for 'how to finetune' is
looking for. Hopefully this more clearly points out the '-weights' flag,
which might otherwise be overlooked in this tutorial.


shelhamer:
Sure, why not highlight it more. Thanks for helping with the documentation.

PR #4032, opened 2016-04-22
add transpose layer and its tests.

tianzhi0549:
As its name suggests, the layer could transpose/permute a blob into a new blob, which supports any number of axes.

Thank you:-)!


src/caffe/layers/transpose_layer.cu
```diff
@@ -0,0 +1,57 @@
...
+    int from_index = index, to_index = 0;
+    for (int i = 0; i < num_axes; i++) {
+    from_inds[i] = from_index / from_counts[i];
+    from_index = from_index % from_counts[i];
+  }
+  for (int i = 0; i < num_axes; i++) {
+    to_index += from_inds[map[i]] * to_counts[i];
+  }
+
+  *(to_data + to_index) = *(from_data + index);
```
seanbell: The indentation for lines 14-22 is off by a few spaces.

src/caffe/proto/caffe.proto
```diff
@@ -1372,3 +1373,11 @@ message PReLUParameter {
...
   optional bool channel_shared = 2 [default = false];
 }
+
+message TransposeParameter {
+  // For example, if you want to transpose NxCxHxW into WxNxHxC,
+  // the parameter should be the following:
+  // transpose_param { dim: 3 dim: 0 dim: 2 dim: 1 }
+  // ie, if the i-th dim has value n, then the i-th axis of top is equal to the n-th axis of bottom.
+  repeated int32 dim=1;
+}
```
seanbell: To be consistent with other layer parameters (which are ordered alphabetically), this should be put earlier in the file  (on line 1131).

seanbell:
Note that for certain transpose patterns, you could optimize it further (with copy operations), but this looks like a good general solution to me.

tianzhi0549:
Hello @seanbell, I want to add this `geam` support you mentioned. But I find that it seems to only support a 2D array(http://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-geam). In Caffe, 2D blobs are rare. And the overall improvement on speed is somewhat negligible (<10ms). Thank you for your suggestions. 

But I find the CPU implementation of the layer is very slow. Do you have any suggestion? Thank you again.

YutingZhang:
I am wondering if this is going to be merged. Thanks

tianzhi0549:
@YutingZhang They don't merge it probably because of the conflicts. I will fix them in the near future:-).

YutingZhang:
@tianzhi0549 Thanks! For now, let me just merge it into my code as it. This layer will be very helpful.

PR #4033, opened 2016-04-23
there is a very very small mistake in the LeNet MNIST tutorial

HeGaoYuan:
Hi,
In your caffe homepage, in the LeNet MNIST Tutorial, there is a sentence in the Training and Testing the Model section,

> Based on the solver setting, we will print the training loss function every 100 iterations, and test the network every 1000 iterations. You will see messages like this:

You said test the network every **1000** iterations, but actually you test the network every **500** iterations.

This is a very very small mistake, but I hope it is helpful for newer.


seanbell:
LGTM

jeffdonahue:
Thanks for the fix @HeGaoYuan! And thanks @seanbell for the review.

PR #4040, opened 2016-04-24
Read the data as a binary

ebadawy:
Appending `b` in the file mode as `hashlib` functions require to pass in bytes.


seanbell:
LGTM.  I believe this only makes a difference on non-Unix platforms.

ebadawy:
Yes, I'm using archlinux+python3 and I couldn't open the file properly before this fix.

longjon:
Right, this makes a difference on some platforms for Python 2, and all platforms for Python 3, where it is required.

Merging, thanks.

PR #3990, opened 2016-04-14
Windows CMake Build

willyd:
Hi,

This PR is not yet completely ready for merge so I am just starting a discussion here to see if there is any interest in supporting the CMake build for Windows. 

## TL;DR;
- Added support for building using CMake.
- Dependencies (boost, opencv, etc.) are automatically downloaded as pre-built binaries
- The project builds a DLL without adding the declspec decorations (see below for details)

## Long Version

Personally, I find that when using CMake with the Ninja generator build times are often much lower than with msbuild (thanks to  @LitingLin for mentionning that in https://github.com/BVLC/caffe/pull/2816). Especially when CUDA files are involved. However, this could be only a CMake issue because of the way it builds CUDA enabled libraries (I don't know if msbuild can built CUDA files in parallel?). Also supporting CMake is useful for other CMake based projects that would link to caffe.

The two main issues that need to be solved to support CMake are:
- Getting the dependencies 
- Handling the static registration of the caffe layers

**Getting the dependencies**
For the dependencies I updated my caffe dependency build tool https://github.com/willyd/caffe-builder so that it downloads, builds and installs all the required libraries in just a single command using CMake and Ninja (building all the required depedencies takes about one hour on [appveyor](https://ci.appveyor.com/project/willyd/caffe-builder). It also distributes the required Find*.cmake modules so that the right libraries are found all the time. To make it easier for every one I opted for an approach like the nuget approach where all the dependencies are downloaded automatically at cmake time. So any one that wants to build the caffe project just needs to do something like:
```cmd
git clone https://github.com/BVLC/caffe.git
cd caffe
git checkout windows
mkdir build
cd build
call "%VS120COMNTOOLS%..\..\VC\vcvarsall.bat" amd64
cmake -GNinja -DBLAS=Open -DCMAKE_BUILD_TYPE=Release ..\
ninja
```
One could use other generators but the downloaded binaries only contain the release configuration currently so building the Debug configuration would fail in VS. This limitation comes from the fact that building both release and debug configs on appveyor is not possible due to build time limits. I do intend to provide binary builds for both Debug and Release configs of the required dependencies through caffe-builder releases. caffe-builder offers other options to support cmake less intrusively (via the -C command line option). 

**Static regsitration of the caffe layers**
[...]

PR #4043, opened 2016-04-25
Patch the windows branch with CMake support

recheliu:
This pull request will patch Caffe so its source code and CMake file can work on both Linux and Windows. We have tested the same files on both Windows and Linux as seen below. 

This patch does not address the building of 3rd party libraries though. 

The followings are the tests we conducted for this pull request:

Build via CMake on Windows. CPU-only:

- Unit tests: PASSED.
- Mnist: PASSED.
- Cifar10/quick: PASSED.

Build via CMake on Windows. CPU & GPU:

- Unit tests: PASSED.
- Mnist: PASSED.
- Cifar10/quick: PASSED.

Build via the original windows/Caffe.sln:

- Unit tests (Cannot finished, but the original windows branch has the same issue).
- Mnist: PASSED.
- Cifar 10(quick): PASSED.
 
Build via CMake on Linux. CPU & GPU:

- Mnist: PASSED.
- Cifar10 (quick): PASSED.

Note that the LMDB for mnist/cifar10 databases were generated on Linux. 

The following is our test environment;

[Windows]
Windows 7 x64 with Visual Studio 2013. 
CUDA: 7.0 for CMake and 7.5 for the original .sln file.
boost: 1.55
CUDNN: 3.0
BLAS: OpenBLAS.
Python: WinPython 2.7.10.3.

[Linux]
Ubuntu 14.04.
CUDA: 7.0.
CUDNN: 3.0
BLAS: Atlas.
Python: 2.7.


willyd:
@recheliu I have submitted a similar PR in https://github.com/BVLC/caffe/pull/3990. You might be interested in this.

recheliu:
Hi @willyd,

Thanks for your reference. I compared your PR with mine. While our fixes to CMake are pretty similar, the main differences are listed  below:

1. Mine can only build static libraries. 
2. Mine does not need Ninja.
3. Mine does not handling the building of 3rd party libraries. 

On the other hand, I saw that you almost finished your PR. The Travis CI failures are mainly related to styles. AppVeyor failure is that the file solver_factory.cpp is not compiled. If you can fixed them, I think that we can just use your PR.

willyd:
@recheliu 

> the main differences are listed below:
> 
>     Mine can only build static libraries.
>     Mine does not need Ninja.
>     Mine does not handling the building of 3rd party libraries.

I would say that another major difference is that the forced symbol linking is handled differently too. Your PR parses the sources files, while mine lists the symbols from the generated caffe.lib. 

I don't actually require Ninja, but I haven't tested the Visual Studio generator so it might not work but it should not be too hard to support it as well.

> On the other hand, I saw that you almost finished your PR.

Note quite finished, it still needs more testing and proper CI support but where getting close you are right.

> AppVeyor failure is that the file solver_factory.cpp is not compiled. If you can fixed them, I think that we can just use your PR.

Just updated the PR with a fix for this.

Any contribution is welcome.

PR #4056, opened 2016-04-27
Fix an error in the example of ReshapeParameter.

wk910930:
One of reshape_param specifications cannot produce desired result.


longjon:
This patch looks correct to me; I'm not sure how @jeffdonahue came up with the fourth example as is. @jeffdonahue, can you confirm this diff?

jeffdonahue:
Oops -- yup, my mistake. Thanks for the correction!

PR #4058, opened 2016-04-27
Opencl zero copy syncedmem memory and spatial convolution enhancement

gongzg:
This PR consist of the following improvement:

1. Implement zero copy syncedmem for OCL device.
2. Make the spatial convolution kernel compatible with other OCL devices.
3. Refine the auto-tuning mechanism and get much faster tuning speed.

Also one patch to enable the libdnn with cmake build method. I did a performance comparison
between current libdnn and the spatial convolution kernel by using the models/bvlc_alexnet/deploy.prototxt on a BDW GT2 laptop. The spatial convolution kernel get
much better forwarding performance than libdnn engine, say 180ms vs 400ms. So I choose the
spatial domain convolution kernel as the default engine for Intel GPU device.


gongzg:
@naibaf7 please check and review. Thanks.

naibaf7:
@gongzg OK merged.

PR #4059, opened 2016-04-27
handle spaces in image file names

crazytan:


seanbell:
Can you add a unit test that verifies this behavior?

Also, before we can merge this, it should be squashed into a single commit.

crazytan:
sure. just found that CI failed :(

crazytan:
I add a unit test and a new image in examples.

seanbell:
Thanks for the test case; LGTM!

bchu:
Overlaps with these other 2 pull requests: 

https://github.com/BVLC/caffe/pull/3433

https://github.com/BVLC/caffe/pull/1971

I'm a bit perplexed why this issue hasn't been fixed yet. It's a pretty big bug.

src/caffe/layers/image_data_layer.cpp
```diff
@@ -37,10 +37,13 @@ void ImageDataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>& bottom,
   const string& source = this->layer_param_.image_data_param().source();
   LOG(INFO) << "Opening file " << source;
   std::ifstream infile(source.c_str());
-  string filename;
+  std::string line;
```
longjon: Why did we switch from `string` to `std::string` here? If we're going to `using` things (and we do `using std::string` in `common.hpp`), then I'd rather consistently drop the namespace.

src/caffe/layers/image_data_layer.cpp
```diff
@@ -37,10 +37,13 @@ void ImageDataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>& bottom,
...
+  size_t pos;
   int label;
-  while (infile >> filename >> label) {
-    lines_.push_back(std::make_pair(filename, label));
+  while (getline(infile, line)) {
```
longjon: In the other direction, how did `getline` end up in our namespace? Existing usage is `std::getline`, so I'd rather continue with that unless we explicitly import `getline` in some way that I've missed.

src/caffe/test/test_image_data_layer.cpp
```diff
@@ -34,16 +34,24 @@ class ImageDataLayerTest : public MultiDeviceTest<TypeParam> {
     std::ofstream outfile(filename_.c_str(), std::ofstream::out);
     LOG(INFO) << "Using temporary file " << filename_;
     for (int i = 0; i < 5; ++i) {
-      outfile << EXAMPLES_SOURCE_DIR "images/cat.jpg " << i;
+      outfile << EXAMPLES_SOURCE_DIR "images/cat.jpg " << i << std::endl;
```
longjon: So I take we were writing out very broken-looking test files before?

longjon:
Just added some minor stylistic comments, and then I'd like to proceed with this PR; #3433 is essentially similar, but this one fixes some issues there; we'll credit @bchu for the original implementation in the merge message.

#1971 is more complex and not presently ready for merge, so let's wait to consider a possible future rebase of that.

src/caffe/layers/image_data_layer.cpp
crazytan: sorry I didn't notice that we're `using std::string`, I'll remove `std::`

src/caffe/layers/image_data_layer.cpp
crazytan: it's surprising, but I did a little experiment and it seems that we don't have to ``using std::getline` to use the function. The function is not imported indeed.

src/caffe/test/test_image_data_layer.cpp
crazytan: I think so... but it works fine with previous implementation

src/caffe/layers/image_data_layer.cpp
longjon: Ah, it's argument-dependent lookup. The explicit version (as it is now) seems reasonable to me, especially since we don't use `getline` frequently in Caffe.

src/caffe/test/test_image_data_layer.cpp
```diff
@@ -34,16 +34,24 @@ class ImageDataLayerTest : public MultiDeviceTest<TypeParam> {
...
+    reshapefile << EXAMPLES_SOURCE_DIR "images/cat.jpg " << 0 << std::endl;
+    reshapefile << EXAMPLES_SOURCE_DIR "images/fish-bike.jpg " << 1
+                << std::endl;
     reshapefile.close();
+    // Create test input file for images with space in names
+    MakeTempFilename(&filename_space_);
+    std::ofstream spacefile(filename_space_.c_str(), std::ofstream::out);
+    LOG(INFO) << "Using temporary file " << filename_space_;
+    spacefile << EXAMPLES_SOURCE_DIR "images/cat.jpg " << 0 << std::endl;
+    spacefile << EXAMPLES_SOURCE_DIR "images/cat gray.jpg " << 1 << std::endl;
```
longjon: As a side note (don't fix it here): it seems awkward to me to keep test-dependent files in the `examples/` directory...

longjon:
This looks good, thanks for addressing this long-standing issue!

bchu:
Could you close this issue: https://github.com/BVLC/caffe/issues/1935?

PR #4065, opened 2016-04-28
avoid non-integer array indices

drewabbot:
non-integer indices cause `VisibleDeprecationWarning` in recent versions of numpy


longjon:
Looks good, thanks.

PR #4070, opened 2016-04-28
fix problems in net_surgery.ipynb

crazytan:
Hi,
I was running this notebook and I found several problems:

1. Instead of `import Image` it should be `from PIL import Image`, which is again an obsolete package. So I delete it and use `caffe.io.load_image`.
2. the comment `# set first filter bias to 10` should be `# set first filter bias to 1`.
3. `caffe` is imported again in input 7.

I corrected these small mistakes. Hope it helps!


ajtulloch:
LGTM, thanks!

longjon:
These changes look good, thanks. However, I'd prefer if you could remove irrelevant changes from the commit (e.g., object address changes), so that we maintain a clean, readable history. (And please do squash into a single commit.)

crazytan:
Thanks for the comment! I have reverted the unnecessary changes but kept the change of output diff (I think it's because the files themselves changed).

longjon:
Okay, thanks!

PR #4071, opened 2016-04-29
draw_net: accept prototxt without name

### LINKED ISSUE #3819: Bug: draw_net utility fails when prototxt has no "name" attribute
Consider the following simple prototxt

> layer {
  name: "data"
  type: "DummyData"
  top: "X"
  top: "y"
  dummy_data_param: { 
    shape: {dim:50 dim:1024 }
    shape: {dim:50}
  }
}
layer {
  name: "fc"
  type: "InnerProduct"
  bottom: "X"
  top: "fc"
  inner_product_param {
    num_output: 100
  }
}
layer {
  name: "relu"
  type: "ReLU"
  bottom: "fc"
  top: "fc" # in place
}

if I try to "draw" it

    ~$ python $CAFFE_ROOT/python/draw_net.py dummy.prototxt dummy.png

I get the following error:

> Drawing net to output.png  
Warning: /tmp/tmpOMoJNS: syntax error in line 1 near '{'

However, adding 

> name: "dummy net"

To the prototxt, solves this issue.

Why is this an issue? 
I noticed that when I use `caffe.NetSpec()` to create models' prototxt the `"name"` attribute is omitted.
So please, it would be nice if
1. `draw_net.py` will not fail if `"name"` is missing
2. `caffe.NetSpec()` will have a `"name"` attribute in the resulting `to_proto()` output.

Thanks!

mnogu:
Fixes #3819

### DIFF
--- a/python/caffe/draw.py
+++ b/python/caffe/draw.py
@@ -142,7 +142,7 @@ def get_pydot_graph(caffe_net, rankdir, label_edges=True):
     -------
     pydot graph object
     """
-    pydot_graph = pydot.Dot(caffe_net.name,
+    pydot_graph = pydot.Dot(caffe_net.name if caffe_net.name else 'Net',
                             graph_type='digraph',
                             rankdir=rankdir)
     pydot_nodes = {}


python/caffe/draw.py
```diff
@@ -142,9 +142,10 @@ def get_pydot_graph(caffe_net, rankdir, label_edges=True):
     -------
     pydot graph object
     """
-    pydot_graph = pydot.Dot(caffe_net.name,
-                            graph_type='digraph',
-                            rankdir=rankdir)
+    dot_args = {'graph_type': 'digraph', 'rankdir': rankdir}
+    if caffe_net.name:
+        dot_args['graph_name'] = caffe_net.name
+    pydot_graph = pydot.Dot(**dot_args)
```
ajtulloch: imo 

```
pydot_graph = pydot.Dot(
  graph_name=caffe_net.name if caffe.net.name else 'Net', 
  graph_type='digraph', 
  rankdir=rankdir) 
```

is cleaner.

ajtulloch:
Looks great, thanks for this. @longjon / @shelhamer?

python/caffe/draw.py
seanbell: I agree; this is more readable than the proposed change.
(typo: `caffe.net` --> `caffe_net`)

mnogu:
I addressed your comments.
Thank you for the reviews.

seanbell:
LGTM!  Please squash into a single commit.

mnogu:
Squashed.

PR #4075, opened 2016-04-30
fix CUDA version detection in Makefile to accomodate OS X's grep

szha:
OS X grep doesn't support \d by default (unfortunately). This causes the CUDA version detection to fail:
```
[~/caffe] make clean
awk: syntax error at source line 1
 context is
	{exit  >>>  < <<<  7.0;}
awk: illegal statement at source line 1
```
If you run the version check manually:
```
[~/caffe] echo $(/usr/local/cuda/bin/nvcc -V | grep -o 'release \d')

```
After the change
```
[~/caffe] echo $(/usr/local/cuda/bin/nvcc -V | grep -o 'release [0-9.]*' | grep -o '[0-9.]*')
7.5
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -272,7 +272,7 @@ endif
 ifeq ($(OSX), 1)
 	CXX := /usr/bin/clang++
 	ifneq ($(CPU_ONLY), 1)
-		CUDA_VERSION := $(shell $(CUDA_DIR)/bin/nvcc -V | grep -o 'release \d' | grep -o '\d')
+		CUDA_VERSION := $(shell $(CUDA_DIR)/bin/nvcc -V | grep -o 'release [0-9.]*' | grep -o '[0-9.]*')
 		ifeq ($(shell echo | awk '{exit $(CUDA_VERSION) < 7.0;}'), 1)
 			CXXFLAGS += -stdlib=libstdc++
 			LINKFLAGS += -stdlib=libstdc++


szha:
Some related env info for reproducing the issue:
```
[~/caffe] uname -r
15.4.0

[~/caffe] /usr/local/cuda/bin/nvcc -V
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2015 NVIDIA Corporation
Built on Thu_Sep_24_00:26:39_CDT_2015
Cuda compilation tools, release 7.5, V7.5.19

[~/caffe] /usr/bin/grep --version
grep (BSD grep) 2.5.1-FreeBSD
```

longjon:
On my systems, OS X `egrep` does support `\d`, but you're right that `grep` does not; on the other hand, my Linux `egrep` does _not_ support `\d`.

The POSIX version which works with both `egrep`s is `[[:digit:]]`, which would be an alternative here.

Note that the patched code doesn't do the same thing as the original code, as the original code only matches the first digit (and only then if your `grep` is actually `egrep`).

This version seems both actually correct and I prefer this way of writing it so we don't have to remember the character classes and what flags are needed to support them.

The original version was introduced by https://github.com/BVLC/caffe/commit/33a56e08d9a2714c2492bb7999a350636ed142be (@shelhamer, do you alias `grep` to `egrep`?)

Thanks, merging.

shelhamer:
Thanks for the compatibility change @szsha and review @longjon. I'd forgotten this particularity of my environment—I'm running a different `grep`—so the OS X standard is indeed a better choice for the Makefile.

PR #4082, opened 2016-05-02
Pin the base image version for the GPU Dockerfile

flx42:
The previous Dockerfile can break if image nvidia/cuda:cudnn is updated to any of the following:
- Ubuntu 16.04 LTS (already released)
- cuDNN v5 (soon)
- CUDA 8.0 (soon)


seanbell:
LGTM; version pinning is usually a good thing.

shelhamer:
Thanks, this makes sense and we can update the version once #3919 is resolved/divided into separate PRs.

PR #4085, opened 2016-05-03
replace open by fopen

hevy-CC4477:


src/caffe/util/io.cpp
```diff
@@ -32,35 +33,35 @@ using google::protobuf::io::CodedOutputStream;
 using google::protobuf::Message;
 
 bool ReadProtoFromTextFile(const char* filename, Message* proto) {
-  int fd = open(filename, O_RDONLY);
-  CHECK_NE(fd, -1) << "File not found: " << filename;
-  FileInputStream* input = new FileInputStream(fd);
+  FILE* fp = fopen(filename, "r");
+  CHECK_NOTNULL(fp);
```
seanbell: This should give a more user-friendly error message.  People will file issues over this.

seanbell: Something like this:
`CHECK_NOTNULL(fp) << "Could not open file: '" << filename << "' (does not exist or insufficient permissions)";`

src/caffe/util/io.cpp
```diff
@@ -32,35 +33,35 @@ using google::protobuf::io::CodedOutputStream;
...
-  close(fd);
+  fclose(fp);
 }
 
 bool ReadProtoFromBinaryFile(const char* filename, Message* proto) {
-  int fd = open(filename, O_RDONLY);
-  CHECK_NE(fd, -1) << "File not found: " << filename;
-  ZeroCopyInputStream* raw_input = new FileInputStream(fd);
+  FILE* fp = fopen(filename, "r");
+  CHECK_NOTNULL(fp);
```
seanbell: Same here

seanbell:
Is this any faster?

The old method used a `FileInputStream`.  According to the docs, too many layers of buffering harms performance (https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.io.zero_copy_stream_impl):
> FileInputStream is preferred over using an ifstream with IstreamInputStream. The latter will introduce an extra layer of buffering, harming performance. Also, it's conceivable that FileInputStream could someday be enhanced to use zero-copy file descriptors on OSs which support them.

seanbell:
It seems like you're using `fopen` and then pulling out the underlying raw file that you could've obtained with `open`.  Is the only benefit portability?

Doesn't that also allocate a buffer which is unused?

src/caffe/util/io.cpp
```diff
@@ -32,35 +33,35 @@ using google::protobuf::io::CodedOutputStream;
...
+  ZeroCopyInputStream* raw_input = new FileInputStream(fileno(fp));
   CodedInputStream* coded_input = new CodedInputStream(raw_input);
```
ajtulloch: CodedInputStream already handles buffering:

```
If you want to be able to read the old-fashion way, you can create a CodedInputStream or CodedOutputStream wrapping these objects and use their ReadRaw()/WriteRaw() methods. These will, of course, add a copy step, but Coded*Stream will handle buffering so at least it will be reasonably efficient.
```

so I don't see the use of this diff. As @seanbell  points out, you're creating a buffered abstraction, pulling out the fd anyway, then passing it to another buffering abstraction, whereas the previous behavior was just to pass the fd to the buffering abstraction.

ajtulloch:
I vote to close this PR - it doesn't appear to be necessary.

longjon:
Having heard nothing further from the author, and seeing as this PR does not seem to achieve its goals, I'm closing.

Thanks for the review @seanbell and @ajtulloch.

PR #4101, opened 2016-05-06
Allow reshaping blobs to size 0.

erictzeng:
The current implementation of `Blob::Reshape` has a check to ensure that Blob sizes do not exceed `INT_MAX` that goes awry whenever the size of the Blob is 0. (More precisely, the check causes a floating point exception if a 0 appears in the shape anywhere except for the last element.) This PR disables that check if `count_ == 0`.

While this seems a little silly upon first glance, I think there is a very real use case for supporting this behavior. I'm training a detection network that uses ground truth boxes, which are typically represented by 5 values, but some images have no ground truth boxes at all. In this case, having a ground truth blob of shape (0, 5) seems perfectly sensible to me.

This PR also includes a simple test that reshapes a Blob to shape (0, 5), which would have failed prior to the fix.

### DIFF
--- a/src/caffe/blob.cpp
+++ b/src/caffe/blob.cpp
@@ -30,7 +30,9 @@ void Blob<Dtype>::Reshape(const vector<int>& shape) {
   int* shape_data = static_cast<int*>(shape_data_->mutable_cpu_data());
   for (int i = 0; i < shape.size(); ++i) {
     CHECK_GE(shape[i], 0);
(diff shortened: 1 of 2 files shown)


jeffdonahue:
Thanks @erictzeng, LGTM (can merge once Travis passes).  The current behavior (allowing empty blobs only in the case that 0 is the last (and only the last) dimension, and otherwise failing with a floating point exception) is clearly wrong.  An alternative to this which would also at least make behavior consistent would be to die on all empty blobs, but I don't see a compelling reason to go that way.

PR #4118, opened 2016-05-10
CUB Memory Manager header files download

drnikolaev:
This is a follow up on #3919 - as it was suggested, we are about to file 4 PRs. This is the first one including 3 commits:
1. The script calling `wget` in a loop.
2. Makefile change making call to the script.
3. CMake support.
We did consider `git submodule` and `git subtree` approaches and found that this one is the most maintainable. We are open to any other ideas, of course. Thank you.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,10 @@ else
 	OTHER_BUILD_DIR := $(DEBUG_BUILD_DIR)
 endif
 
+THIRDPARTY_DIR := ./3rdparty
+CUB_DIR := $(THIRDPARTY_DIR)/cub
+CUB_VERSION := 1.5.2
+
 # All of the directories containing code.
 SRC_DIRS := $(shell find * -type d -exec bash -c "find {} -maxdepth 1 \
 	\( -name '*.cpp' -o -name '*.proto' \) | grep -q ." \; -print)
@@ -581,7 +585,7 @@ $(PROTO_BUILD_DIR)/%.pb.o: $(PROTO_BUILD_DIR)/%.pb.cc $(PROTO_GEN_HEADER) \
 		|| (cat $@.$(WARNS_EXT); exit 1)
 	@ cat $@.$(WARNS_EXT)
 
-$(BUILD_DIR)/cuda/%.o: %.cu | $(ALL_BUILD_DIRS)
+$(BUILD_DIR)/cuda/%.o: %.cu | $(ALL_BUILD_DIRS) $(CUB_DIR)
 	@ echo NVCC $<
 	$(Q)$(CUDA_DIR)/bin/nvcc $(NVCCFLAGS) $(CUDA_ARCH) -M $< -o ${@:.o=.d} \
 		-odir $(@D)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -568,7 +572,7 @@ $(STATIC_NAME): $(OBJS) | $(LIB_BUILD_DIR)
 	@ echo AR -o $@
 	$(Q)ar rcs $@ $(OBJS)
 
-$(BUILD_DIR)/%.o: %.cpp | $(ALL_BUILD_DIRS)
+$(BUILD_DIR)/%.o: %.cpp | $(ALL_BUILD_DIRS) $(CUB_DIR)
```
longjon: Is CUB going to be needed for (non-CUDA) object files? It would nice if this line were not necessary, so that CUB does not need to be downloaded for (e.g.) CPU-only builds.

Makefile
```diff
@@ -638,6 +642,7 @@ $(PY_PROTO_INIT): | $(PY_PROTO_BUILD_DIR)
 	touch $(PY_PROTO_INIT)
 
 clean:
+	@- $(RM) -rf $(CUB_DIR)
```
longjon: Perhaps this should go in `superclean`, or just be omitted? It's not that uncommon to `make clean` and seems a bit odd to me to force a re-download upon doing so.

longjon:
Okay, this looks pretty good to me, two comments above.

As long as the mechanism is contained within the build process (so that it can be seamlessly changed later), it doesn't matter too much to me whether it's `wget`, `submodule`, or `subtree`... we'll find out soon enough if there are serious pain points.

With a resolution of the comments above I'd be happy to merge this alongside a second PR that depends on it to add functionality.

lukeyeager:
I think you want to add `3rdparty/cub/` to `.gitignore`

Makefile
drnikolaev: Removed this one and left only %.cu rule - it get's triggered when NVCC is running.

Makefile
drnikolaev: Moved this to superclean

drnikolaev:
Notes addressed, `.gitignore` entry added. Thank you.

Makefile
```diff
@@ -667,6 +671,10 @@ superclean: clean supercleanfiles
 		echo $(SUPERCLEAN_FILES) | tr ' ' '\n'; \
 		$(RM) $(SUPERCLEAN_FILES); \
 	fi
+	@ \
+	if [ -d $(CUB_DIR) ]; then \
```
longjon: Hm, is there a reason we switched from `@- $(RM)...` syntax to this `if` block? Both should have the same effect (right?) and we're already using the former above...

drnikolaev: Yeah, I was thinking about to make it consistent with the previous case... Never mind, reverted to `@- $(RM)...`.

drnikolaev:
Addressed one more comment and also fixed Cmake issue preventing from building out of source.

drnikolaev:
@longjon  are we good to go with this one?

longjon:
This looks good to me. Since this PR doesn't add any functionality that depends on CUB, what I'd like to do is merge this along with a follow-up PR that does; just build that branch on top of this one, and no further action should be needed here unless additional comments arise.

I'm adding a label to indicate this. Thanks!

PR #4122, opened 2016-05-10
Add jupyter to the docker image

vmalarcon:


docker/standalone/cpu/Dockerfile
```diff
@@ -33,11 +34,13 @@ RUN git clone -b ${CLONE_TAG} --depth 1 https://github.com/BVLC/caffe.git . && \
     for req in $(cat python/requirements.txt) pydot; do pip install $req; done && \
     mkdir build && cd build && \
     cmake -DCPU_ONLY=1 .. && \
-    make -j"$(nproc)"
+    make all -j"$(nproc)" && \
+    make pycaffe
```
lukeyeager: CMake builds pycaffe automatically for me when Python is detected. Does it not for you?

flx42:
You are only modifying the CPU Dockerfile here, please modify the template file and regenerate the two Dockerfiles.

flx42:
Also, you need a minimum jupyter configuration file to set `open_browser = False` and to modify the IP you are listening on.

flx42:
I said a **minimum** configuration file :)
And you still don't need to modify the `make` command.

flx42:
I think the following is enough:
```
c.NotebookApp.ip = '*'
c.NotebookApp.port = 8888
c.NotebookApp.open_browser = False
```
Also, ping @elezar

elezar:
Thanks for the addition @vmalarcon!

If the consensus is that the images should include Jupyter, then I agree with @flx42 that the config file should be kept minimal, and in this case, we could consider creating it in the Dockerfile:
```
RUN echo "c.NotebookApp.ip = '*'" > configfile
RUN echo "c.NotebookApp.port = 8888" >> configfile
RUN echo "c.NotebookApp.open_browser = False" >> configfile
```
(I generally specify these on the command line when starting a notebook server)
but I don't know whether this is in keeping with Docker best practices and separating the config into a separate file is probably better.

Another option is to have the jupyter version in a separate docker image? I use such an image locally for building and testing caffe. I do this using the following files:
- [Dockerfile.txt](https://github.com/BVLC/caffe/files/258684/Dockerfile.txt)
- [start_caffe_docker.sh.txt](https://github.com/BVLC/caffe/files/258683/start_caffe_docker.sh.txt)
(note that the `.txt` extension should be removed in each case)

vmalarcon:
Thank you guys for the comments/help @flx42 @elezar! 

I like @elezar 's idea of using a separate Dockerfile for jupyter and inherit from caffe:[g|c]pu. I think I'll do that on my end. So I'll close this PR. Actually, I never intended this to be merged to the main repo.

vmalarcon:
https://github.com/vmalarcon/caffe/pull/1

elezar:
Glad we could help!

PR #4128, opened 2016-05-11
a comment misses a space char

gdh1995:
One assert message is:
* `"mkdir " << source << "failed"`

I think `"failed"`'s a typo and should be `" failed"`.


longjon:
Thanks!

PR #4165, opened 2016-05-17
conversion from old to new net style

jeremy-rutman:
This add a python file 'alteneue_converter.py' which converts an 'old style' net prototext to the new style.
It succesfully converts the vgg nets  [here](https://gist.github.com/ksimonyan/fd8800eeb36e276cd6f9#file-readme-md) to the new format.

### DIFF
--- a/python/alteneue_converter.py
+++ b/python/alteneue_converter.py
@@ -1,0 +1,55 @@
+__author__ = 'jeremy'
+import argparse
+
+'''
+#This takes an 'old timey' net and converts to the new format 
+#layers->layer
+#old data format to new format
+'''
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


ajtulloch:
What is the difference between https://github.com/BVLC/caffe/blob/master/tools/upgrade_net_proto_text.cpp and this?

shelhamer:
Closing as an incomplete Python implementation of the existing C++ upgrade logic—handled by the command line tools `upgrade_net_proto_text` and `upgrade_net_proto_binary` and by `Net` init—as noted by @ajtulloch.

@jeremy-rutman please comment if there is a further use case or other reason for this.

jeremy-rutman:
ok, I had not realized there was existing C++ upgrade logic.

PR #4175, opened 2016-05-19
CUB Memory Manager Integration

drnikolaev:
This is a follow up PR after #4118 (see also #3919)

### DIFF
--- a/include/caffe/util/gpu_memory.hpp
+++ b/include/caffe/util/gpu_memory.hpp
@@ -1,0 +1,186 @@
+#ifndef CAFFE_UTIL_GPU_MEMORY_HPP_
+#define CAFFE_UTIL_GPU_MEMORY_HPP_
+
+#include <vector>
+#include "caffe/common.hpp"
+#ifndef CPU_ONLY
+
+namespace cub {
+  class CachingDeviceAllocator;
+}
+
+namespace caffe {
+
+struct GPUMemory {
+  static void GetInfo(size_t* free_mem, size_t* used_mem) {
+    return mgr_.GetInfo(free_mem, used_mem);
+  }
+
+  template <class Any>
+  static void allocate(Any** ptr, size_t size,
+      int device = INVALID_DEVICE,
+      cudaStream_t stream = cudaStreamDefault) {
+    if (!try_allocate(reinterpret_cast<void**>(ptr), size, device, stream)) {
+      CUDA_CHECK(cudaGetDevice(&device));
+      LOG(FATAL) << "Out of memory: failed to allocate " << size
+          << " bytes on device " << device;
+    }
+  }
+
+  static void deallocate(void* ptr, int device = INVALID_DEVICE,
+      cudaStream_t stream = cudaStreamDefault) {
+    mgr_.deallocate(ptr, device, stream);
+  }
+
+  static bool try_allocate(void** ptr, size_t size, int device = INVALID_DEVICE,
+      cudaStream_t stream = cudaStreamDefault) {
+    return mgr_.try_allocate(ptr, size, device, stream);
+  }
+
+  enum Mode {
+    CUDA_MALLOC,   // Straight CUDA malloc/free (may be expensive)
+    CUB_ALLOCATOR  // CUB caching allocator
+  };
+
+  // Scope initializes global Memory Manager for a given scope.
+  // It's instantiated in test(), train() and time() Caffe brewing functions
+  // as well as in unit tests main().
+  struct Scope {
+    Scope(const std::vector<int>& gpus, Mode m = CUB_ALLOCATOR,
+          bool debug = false) {
+      mgr_.init(gpus, m, debug);
+    }
+  };
+
+  // Workspace's release() functionality depends on global pool availability
+  // If pool is available, it returns memory to the pool and sets ptr to NULL
+  // If pool is not available, it retains memory.
+  // This is single GPU workspace. See MultiWorkspace for multi-GPU support.
+  struct Workspace {
+    Workspace()
+      : ptr_(NULL), size_(0), device_(INVALID_DEVICE),
+        stream_(cudaStreamDefault) {}
+    Workspace(size_t size, int device = INVALID_DEVICE,
+        cudaStream_t s = cudaStreamDefault)
+      : ptr_(NULL), size_(0), device_(device), stream_(s) {
+      reserve(size, device);
+    }
+    ~Workspace() { mgr_.deallocate(ptr_, device_, stream_); }
+
+    void* data() const { return ptr_; }
+    size_t size() const { return size_; }
+    int device() const { return device_; }
+
+    bool try_reserve(size_t size, int device = INVALID_DEVICE) {
+      bool status = true;
+      if (size > size_) {
+        if (ptr_ != NULL) {
+          mgr_.deallocate(ptr_, device_, stream_);
+        }
+        if (device != INVALID_DEVICE) {
+          device_ = device;  // switch from default to specific one
+        }
+        status = mgr_.try_allocate(&ptr_, size, device_, stream_);
+        if (status) {
+          size_ = size;
+        }
+      }
+      return status;
+    }
+
+    void reserve(size_t size, int device = INVALID_DEVICE) {
+      if (!try_reserve(size, device)) {
+        CUDA_CHECK(cudaGetDevice(&device));
+        LOG(FATAL) << "Out of memory: failed to allocate " << size
+            << " bytes on device " << device;
+      }
+    }
+
+    void release() {
+      if (mgr_.using_pool() && ptr_ != NULL) {
+        mgr_.deallocate(ptr_, device_, stream_);
+        ptr_ = NULL;
+        size_ = 0;
+      }
+      // otherwise (no pool) - we retain memory in the buffer
+    }
+
+   private:
+    void* ptr_;
+    size_t size_;
+    int device_;
+    cudaStream_t stream_;
+
+    DISABLE_COPY_AND_ASSIGN(Workspace);
+  };
+
+  // This implementation maintains workspaces on per-GPU basis.
+  struct MultiWorkspace {
+    bool try_reserve(size_t size) {
+      return current_workspace()->try_reserve(size);
+    }
+    void reserve(size_t size) {
+      current_workspace()->reserve(size);
+    }
+    void release() {
+      current_workspace()->release();
+    }
+    void* data() const {
+      return current_workspace()->data();
+    }
+    size_t size() const {
+      return current_workspace()->size();
+    }
+    int device() const {
+      return current_workspace()->device();
+    }
+
+   private:
+    shared_ptr<Workspace> current_workspace() const;
+    mutable vector<shared_ptr<Workspace> > ws_;
+  };
+
+ private:
+  struct Manager {
+    Manager();
+    ~Manager();
+    void GetInfo(size_t* free_mem, size_t* used_mem);
+    void deallocate(void* ptr, int device, cudaStream_t stream);
+    bool try_allocate(void** ptr, size_t size, int device, cudaStream_t);
+    const char* pool_name() const;
+    bool using_pool() const { return mode_ != CUDA_MALLOC; }
+    void init(const std::vector<int>&, Mode, bool);
+
+    Mode mode_;
+    bool debug_;
+
+   private:
+    struct DevInfo {
+      DevInfo() {
+        free_ = total_ = flush_count_ = 0;
+      }
+      size_t free_;
+      size_t total_;
+      unsigned flush_count_;
+    };
+    void update_dev_info(int device);
+    vector<DevInfo> dev_info_;
+    bool initialized_;
+    cub::CachingDeviceAllocator* cub_allocator_;
+
+    static const unsigned int BIN_GROWTH;  ///< Geometric growth factor
+    static const unsigned int MIN_BIN;  ///< Minimum bin
+    static const unsigned int MAX_BIN;  ///< Maximum bin
+    static const size_t MAX_CACHED_BYTES;  ///< Maximum aggregate cached bytes
+  };
+
+  static const int INVALID_DEVICE;  ///< Default is invalid: CUB takes care
+
+  static Manager mgr_;
+};
+
+}  // namespace caffe
+
+#endif
+
+#endif
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -11,6 +11,7 @@
 #include "boost/thread.hpp"
 #include "caffe/caffe.hpp"
 #include "caffe/parallel.hpp"
+#include "caffe/util/gpu_memory.hpp"
 
 namespace caffe {
 
@@ -82,14 +83,17 @@ GPUParams<Dtype>::GPUParams(shared_ptr<Solver<Dtype> > root_solver, int device)
 
   // Allocate device buffers
   CUDA_CHECK(cudaSetDevice(device));
-  CUDA_CHECK(cudaMalloc(&data_, size_ * sizeof(Dtype)));
+  buffer_device_ = device;
+  GPUMemory::allocate(reinterpret_cast<void **>(&data_),
+      size_ * sizeof(Dtype));
 
   // Copy blob values
   const vector<Blob<Dtype>*>& net =
       root_solver->net()->learnable_params();
   apply_buffers(net, data_, size_, copy);
 
-  CUDA_CHECK(cudaMalloc(&diff_, size_ * sizeof(Dtype)));
+  GPUMemory::allocate(reinterpret_cast<void **>(&diff_),
+      size_ * sizeof(Dtype));
   caffe_gpu_set(size_, Dtype(0), diff_);
 
   CUDA_CHECK(cudaSetDevice(initial_device));
@@ -101,8 +105,14 @@ GPUParams<Dtype>::GPUParams(shared_ptr<Solver<Dtype> > root_solver, int device)
 template<typename Dtype>
 GPUParams<Dtype>::~GPUParams() {
 #ifndef CPU_ONLY
-  CUDA_CHECK(cudaFree(data_));
-  CUDA_CHECK(cudaFree(diff_));
+  int initial_device;
+  CUDA_CHECK(cudaGetDevice(&initial_device));
+  CUDA_CHECK(cudaSetDevice(buffer_device_));
+  GPUMemory::deallocate(data_);
+  GPUMemory::deallocate(diff_);
+  data_ = NULL;
+  diff_ = NULL;
+  CUDA_CHECK(cudaSetDevice(initial_device));
 #endif
 }
 
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -245,7 +245,8 @@ P2PSync<Dtype>::P2PSync(shared_ptr<Solver<Dtype> > root_solver,
     }
     // Allocate receiving buffer on parent
     CUDA_CHECK(cudaSetDevice(peer));
-    CUDA_CHECK(cudaMalloc(&parent_grads_, size_ * sizeof(Dtype)));
+    GPUMemory::allocate(reinterpret_cast<void**>(&parent_grads_),
+        size_ * sizeof(Dtype));
     CUDA_CHECK(cudaSetDevice(self));
   }
 
@@ -258,22 +259,22 @@ P2PSync<Dtype>::P2PSync(shared_ptr<Solver<Dtype> > root_solver,
 template<typename Dtype>
 P2PSync<Dtype>::~P2PSync() {
 #ifndef CPU_ONLY
-  int initial_device;
-  CUDA_CHECK(cudaGetDevice(&initial_device));
-  const int self = solver_->param().device_id();
-  CUDA_CHECK(cudaSetDevice(self));
-
   if (parent_) {
-    CUDA_CHECK(cudaFree(parent_grads_));
+    int initial_device;
+    CUDA_CHECK(cudaGetDevice(&initial_device));
+    const int self = solver_->param().device_id();
     const int peer = parent_->solver_->param().device_id();
+    CUDA_CHECK(cudaSetDevice(peer));
+    GPUMemory::deallocate(parent_grads_);
+    parent_grads_ = NULL;
     int access;
+    cudaSetDevice(self);
     CUDA_CHECK(cudaDeviceCanAccessPeer(&access, self, peer));
     if (access) {
       CUDA_CHECK(cudaDeviceDisablePeerAccess(peer));
     }
+    CUDA_CHECK(cudaSetDevice(initial_device));
   }
-
-  CUDA_CHECK(cudaSetDevice(initial_device));
 #endif
 }
 
--- a/src/caffe/test/test_caffe_main.cpp
+++ b/src/caffe/test/test_caffe_main.cpp
@@ -1,5 +1,6 @@
 #include "caffe/caffe.hpp"
 #include "caffe/test/test_caffe_main.hpp"
+#include "caffe/util/gpu_memory.hpp"
 
 namespace caffe {
 #ifndef CPU_ONLY
@@ -32,6 +33,8 @@ int main(int argc, char** argv) {
   cout << "Current device id: " << device << endl;
   cudaGetDeviceProperties(&CAFFE_TEST_CUDA_PROP, device);
   cout << "Current device name: " << CAFFE_TEST_CUDA_PROP.name << endl;
+  caffe::GPUMemory::Scope gpu_memory_scope(devices);
+
 #endif
   // invoke the test.
   return RUN_ALL_TESTS();
--- a/src/caffe/test/test_caffe_main.cpp
+++ b/src/caffe/test/test_caffe_main.cpp
@@ -1,3 +1,4 @@
+#include <vector>
 #include "caffe/caffe.hpp"
 #include "caffe/test/test_caffe_main.hpp"
 #include "caffe/util/gpu_memory.hpp"
@@ -17,18 +18,30 @@ int main(int argc, char** argv) {
   caffe::GlobalInit(&argc, &argv);
 #ifndef CPU_ONLY
   // Before starting testing, let's first print out a few cuda defice info.
-  int device;
-  cudaGetDeviceCount(&device);
-  cout << "Cuda number of devices: " << device << endl;
+  std::vector<int> devices;
+  int device_count;
+
+  cudaGetDeviceCount(&device_count);
+  cout << "Cuda number of devices: " << device_count << endl;
+
   if (argc > 1) {
     // Use the given device
-    device = atoi(argv[1]);
-    cudaSetDevice(device);
-    cout << "Setting to use device " << device << endl;
+    devices.push_back(atoi(argv[1]));
+    cudaSetDevice(devices[0]);
   } else if (CUDA_TEST_DEVICE >= 0) {
     // Use the device assigned in build configuration; but with a lower priority
-    device = CUDA_TEST_DEVICE;
+    devices.push_back(CUDA_TEST_DEVICE);
   }
+
+  if (devices.size() == 1) {
+    cout << "Setting to use device " << devices[0] << endl;
+    cudaSetDevice(devices[0]);
+  } else {
+    for (int i = 0; i < device_count; ++i)
+      devices.push_back(i);
+  }
+
+  int device;
   cudaGetDevice(&device);
   cout << "Current device id: " << device << endl;
   cudaGetDeviceProperties(&CAFFE_TEST_CUDA_PROP, device);
--- a/src/caffe/util/gpu_memory.cpp
+++ b/src/caffe/util/gpu_memory.cpp
@@ -1,0 +1,204 @@
+#ifndef CPU_ONLY
+#include <algorithm>
+#include <vector>
+#include "caffe/common.hpp"
+#include "caffe/util/gpu_memory.hpp"
+
+#include "cub/util_allocator.cuh"
+
+namespace caffe {
+using std::vector;
+
+const int GPUMemory::INVALID_DEVICE =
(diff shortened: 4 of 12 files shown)


drnikolaev:
Travis build failed because #4118 is not merged yet.

longjon:
So what I actually had in mind in https://github.com/BVLC/caffe/pull/4118#issuecomment-219875656 was for these commits to be on top of the commits in #4118. I.e., this PR should _include_ #4118, i.e., all of the #4118 commits should appear here, as well as the new commits. That way Travis will pass and we can review the new commits; further comments on the old commits will remain at #4118.

This is what I meant earlier about having one PR "depend" on another; it's okay to build a PR off an unmerged branch, e.g., another PR. (We've done this in the past to break up complex changes.) GitHub will mark the dependency as merged once the dependent is merged.

Sorry if this is a bit confusing, as I know I ranted about small PRs earlier; a PR included in another doesn't count. This way we can avoid leaving the repo in an intermediate state, but also break up the changes into reviewable, testable units.

drnikolaev:
Got it, thanks. Will do. Meanwhile, could you please review this one?

lzhangzz:
Is a significant increase in GPU memory consumption expected? The code currently use the CUB memory manager for all allocations so that the allocations are rounded up to fit the bin sizes. 

I have tested with [NVIDIA/Caffe](https://github.com/NVIDIA/caffe/tree/caffe-0.15) and got a 25-30% increase in GPU memory on ResNet-50. As the code for memory management is almost the same, I assume the same thing will happen here.

Should we provide some mechanism to choose whether a memory allocation should be cached since GPU memory is a very demanding resource during training? e.g. in training, most top and bottom blobs don't need caching since they will persist in the whole training process.

drnikolaev:
@longjon I hope I did it right: bvlc_pr1_cub (4 commits) is for 4118, then bvlc_pr2_cub_integration (7 commits) was spawned out of it and used for the current 4175.

drnikolaev:
@lzhangzz Absolutely, just replace
`caffe::GPUMemoryManager::Arena arena(gpus);`
by
`caffe::GPUMemoryManager::Arena arena(gpus, caffe::GPUMemoryManager::NO_POOL);`

lzhangzz:
@drnikolaev That would also disable memory caching for the workspace used by CuDNNConvolutionLayer. Is it possible to disable caching for top and bottom blobs while enabling it for CUDNN workspace?

drnikolaev:
@lzhangzz It's possible in next PR

lukeyeager:
Oops I caused a merge conflict with https://github.com/BVLC/caffe/pull/4214

drnikolaev:
W: There is no public key available for the following key IDs:

1397BC53640DB551

W: Failed to fetch http://us.archive.ubuntu.com/ubuntu/dists/trusty-updates/universe/source/Sources  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.

longjon:
@drnikolaev: yeah, our Travis install seems to have become a bit sketchy lately, with random hash failures; I've restarted the build (as the author of the branch, maybe you can too?).

Yes, the history is nice now. I plan to do a pass over this before Monday.

drnikolaev:
@longjon thank you. BTW, I couldn't find "restart" button anywhere.

flx42:
Force push to the branch to restart a build.
If nothing has changed, you can just `--amend --reset-author` before pushing, it will change the hash of the last commit.

longjon:
Okay, a few overall comments:

1. It looks like most of the content is in the final commit -- shouldn't this be the first (new) commit?
2. This PR adds a bunch of process-level state in the form of `GPUMemoryManager`. We used to put process-level state in the static `Caffe` object, but that's now thread-level, so I guess there is no better place for this. I could also see an argument for this being `Net`-specific (shouldn't we be able to use one net with `malloc` alongside another with the pool?), but we should also be able to share a pool between nets, and `malloc` is already globally managed, so I guess this is fine.
3. In any case, it seems to me that rather than having a class with a bunch of `static` members, it's a better idea to have a class with non-`static` members and one `static` instance. That's how `Caffe` works now, and it affords more flexibility (e.g., for implementing the last parenthetical).
4. In this PR we also meet `Arena` and `Buffer`. `Arena` appears to provide scoped memory management. That's nifty, but what purpose is that intended to solve? In this code, we always just set the manager once, globally. Another option would be to have `GPUMemoryManager` handle `init`/`destroy` on construction/destruction, with a static call to ask for the initial construction, similar to how the `Caffe` object works. (Although this may cause trouble with `static` destructor order, and is a bit ugly itself... I could see an argument either way.)
5. Can you explain the `Buffer` class? I can't place where it is being used. (In any case, it's a good idea to write a short explanation of this kind of thing which each PR.)
6. It looks like this hardcodes the CUB allocator as default for all allocations when calling the `caffe` binary. Unless the overhead is really negligible, I'm don't think that's going to be tolerable. It's common to run nets which use memory up to the limit, and we shouldn't break those by default.
7. How does this interact with pycaffe?

src/caffe/util/gpu_memory.cpp
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  switch (m) {
+  case CUB_POOL:
+    InitMemory(gpus, m);
+    break;
+  default:
+    break;
+  }
+  VLOG_IF(1, debug) << "GPUMemoryManager initialized with " << pool_name();
```
longjon: Why not just use `DLOG`?

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+#include <vector>
+#include "caffe/common.hpp"
+
+namespace caffe {
+
+class GPUMemoryManager {
+ public:
+  enum PoolMode {
+    NO_POOL,  // Straight CUDA malloc/free (may be expensive)
```
longjon: Perhaps this should be called `MALLOC`, or something else? I could see other options being added which are not pools...

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,126 @@
...
+#ifdef CPU_ONLY
+    DEFAULT_POOL = NO_POOL
+#else
+    DEFAULT_POOL = CUB_POOL  // CUB pool is able to use unified memory properly
+#endif
+  };
+
+  static const char* pool_name();
+  static bool using_pool() {
+    return mode_ != NO_POOL;
```
longjon: Per above, perhaps this test should be inverted...

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  class Arena {
+   public:
+    Arena(const std::vector<int>& gpus, PoolMode m = DEFAULT_POOL, bool debug =
+        false) {
+      init(gpus, m, debug);
+    }
+    ~Arena() {
```
longjon: Write very short functions on one line, like `~Arena() { destroy(); }`.

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,126 @@
...
+      destroy();
+    }
+  };
+
+#ifndef CPU_ONLY
+  class Buffer {
+   public:
+    // Construction/destruction
```
longjon: No need for these kinds of comments -- any reader can tell what's here.

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,126 @@
...
+      free_ = total_ = flush_count_ = 0;
+    }
+    size_t free_;
+    size_t total_;
+    unsigned flush_count_;
+  };
+  static vector<MemInfo> dev_info_;
+
+ public:
+  typedef void* pointer;
```
longjon: Why do this? This is not a complicated type, and now the reader has to search for and keep in mind the definition of `pointer`.

src/caffe/util/gpu_memory.cpp
```diff
@@ -0,0 +1,186 @@
...
+  default:
+    CUDA_CHECK(cudaMalloc(ptr, size));
+    break;
+  }
+}
+
+void GPUMemoryManager::deallocate(void* ptr, cudaStream_t stream) {
+  // allow for null pointer deallocation
+  if (!ptr) {
+    return;
```
longjon: Should this be silent or checked? Is it useful to silently allow this call on a null pointer?

src/caffe/util/gpu_memory.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+void GPUMemoryManager::InitMemory(const std::vector<int>& gpus, PoolMode m) {
+  mode_ = m;
+  switch (mode_) {
+  case CUB_POOL:
+    try {
+      // Just in case someone installed 'no cleanup' arena before
+      delete cub_allocator;
+      cub_allocator = new cub::CachingDeviceAllocator(2, 6, 22, (size_t) -1,
```
longjon: Surely these parameters should not be hardcoded in the middle of this file, right?

src/caffe/util/gpu_memory.cpp
flx42: Being able to deallocate a null pointer is consistent with `cudaFree`, `free`, and `delete`.

src/caffe/util/gpu_memory.cpp
```diff
@@ -0,0 +1,186 @@
...
+          false, debug_);
+    } catch (...) {
```
longjon: I guess we are converting exceptions into the `CHECK` below? Do we want to do that? (If we do, it's worth a comment. But I don't really see why we should...)

src/caffe/parallel.cpp
```diff
@@ -82,14 +83,17 @@ GPUParams<Dtype>::GPUParams(shared_ptr<Solver<Dtype> > root_solver, int device)
 
   // Allocate device buffers
   CUDA_CHECK(cudaSetDevice(device));
-  CUDA_CHECK(cudaMalloc(&data_, size_ * sizeof(Dtype)));
+  buffer_device_ = device;
+  GPUMemoryManager::allocate(reinterpret_cast<void **>(&data_),
```
longjon: Surely we don't need `reinterpret_cast` here? (Actually I'm not sure why we need a cast at all...)

src/caffe/parallel.cpp
```diff
@@ -245,7 +245,8 @@ P2PSync<Dtype>::P2PSync(shared_ptr<Solver<Dtype> > root_solver,
     }
     // Allocate receiving buffer on parent
     CUDA_CHECK(cudaSetDevice(peer));
-    CUDA_CHECK(cudaMalloc(&parent_grads_, size_ * sizeof(Dtype)));
+    GPUMemoryManager::allocate(reinterpret_cast<void **>(&parent_grads_),
```
longjon: As above.

src/caffe/test/test_caffe_main.cpp
```diff
@@ -1,3 +1,7 @@
+// The main caffe test code. Your test cpp code should include this hpp
+// to allow a main function to be compiled into the binary.
```
longjon: Why is this comment appearing in this commit?

src/caffe/util/gpu_memory.cpp
longjon: Ah, right -- this is the correct behavior!

src/caffe/util/gpu_memory.cpp
```diff
@@ -0,0 +1,186 @@
...
+    cub_allocator = NULL;
+    break;
+  default:
+    break;
+  }
+  mode_ = NO_POOL;
+}
+
+void GPUMemoryManager::allocate(void** ptr, size_t size, cudaStream_t stream) {
+  CHECK((ptr) != NULL);
```
longjon: I believe we have `CHECK_NOTNULL` for this.

src/caffe/test/test_caffe_main.cpp
shelhamer: This PR needs rebasing. This is an artifact of the comment clean up in #4214.

include/caffe/util/gpu_memory.hpp
borisfom: We can rename PoolMode -> Mode, values:  CUDA_MALLOC, CUB_ALLOCATOR

include/caffe/util/gpu_memory.hpp
borisfom: RIght.

include/caffe/util/gpu_memory.hpp
borisfom: Agreed.

src/caffe/util/gpu_memory.cpp
borisfom: Well, those are CUB-specific. Alternative would be to hardcode them in the header as constants.

src/caffe/util/gpu_memory.cpp
borisfom: CUDA_CHECK is a standard macro to wrap all CUDA calls - one important part of its role is to call CudaGetLastError which clears global error flag.

src/caffe/parallel.cpp
borisfom: src/caffe/parallel.cpp:95:52: error: invalid conversion from ‘float**’ to ‘void**’ [-fpermissive]
   GPUMemory::allocate(&diff_, size_ * sizeof(Dtype));
                                                    ^
In file included from src/caffe/parallel.cpp:14:0:
./include/caffe/util/gpu_memory.hpp:16:15: error:   initializing argument 1 of ‘static void caffe::GPUMemory::allocate(void**, size_t, cudaStream_t)’ [-fpe\
rmissive]
   static void allocate(void** ptr, size_t size,

borisfom: src/caffe/parallel.cpp:95:52: error: invalid conversion from ‘float**’ to ‘void**’ [-fpermissive]
   GPUMemory::allocate(&diff_, size_ * sizeof(Dtype));
                                                    ^
In file included from src/caffe/parallel.cpp:14:0:
./include/caffe/util/gpu_memory.hpp:16:15: error:   initializing argument 1 of ‘static void caffe::GPUMemory::allocate(void**, size_t, cudaStream_t)’ [-fpe\
rmissive]
   static void allocate(void** ptr, size_t size,
               ^

flx42: @longjon is confused because the `cudaMalloc` call that is being replaced doesn't have the ugly `reinterpret_cast`, this is because in `cuda_runtime.h`, `cudaMalloc` has a template wrapper:
```
template<class T>
static __inline__ __host__ cudaError_t cudaMalloc(
  T      **devPtr,
  size_t   size
)
{ 
  return ::cudaMalloc((void**)(void*)devPtr, size);
}
```
If @longjon wants to have an API closer to `cudaMalloc`, we might need to the same.

borisfom: Right, I will add a convenience wrapper.

longjon: I see... the documentation declared `cudaMalloc` simply has the `void**` type, but of course there's more to it than that. It's also perhaps a bit surprising that `Dtype*` implicitly converts to `void*`, but `Dtype**` does not implicitly convert to `void**`... A wrapper seems reasonable.

drnikolaev:
@longjon we carefully refactored the code according to the comments made above, ran it through QA and here is the update. 
1. As of  "a class with a bunch of static members" vs. "a class with non-static members and one static instance" issue. We tried both. The only difference we see is writing "GPUMemory::allocate" vs. "GPUMemory::Get().allocate". We prefer the shorter version. If you insist on the other one, we can redo it.
2. As of "Buffer" class. Now it's called "Workspace". It's used to allocate a chunk of memory (usually quite large) as a workspace for memory-demanding algorithms (like cuDNN ones for Convolution). It also allows to call `reserve` function multiple times and memory reallocation doesn't happen when the amount requested is smaller than the one already allocated (see `gpu_memory.hpp:65`).
3. You write: _It looks like this hardcodes the CUB allocator as default for all allocations when calling the caffe binary. Unless the overhead is really negligible, I'm don't think that's going to be tolerable. It's common to run nets which use memory up to the limit, and we shouldn't break those by default._
Using memory up to the limit with minimal overhead is exactly the point of the new memory manager. Instead of making spectacularly expensive `cudaMalloc` we offer cache. We built a histogram of 682K memory allocation sizes (based on Inception v1 net) and found that 500K+ of them are smaller than 100KB in size each. This is the ideal application for cache. As of using memory up to the limit, the next PR will address that for cuDNN Convolution (see original #3919).
4. We verified that PyCaffe works as expected with the manager.

borisfom:
To add to Sergei's comment: with this change alone, we would expect most nets to actually run faster - as overhead of caching allocator is negligible while the overhead of real cudaMalloc is very substantial.
Also, this particular caching allocator has the feature of flushing the cache on out-of-memory situation, so in fact I would be very surprised if it would slow down any use case.

longjon:
Okay, I am traveling for a few days and won't be able to do a detailed read until Monday (@shelhamer may take a look before then), but wanted to make a few comments in the meantime:

1. Re: @drnikolaev's 1 above, the syntactic difference is besides the point; I'm fine with having a `static` helper function `allocate` which obtains a singleton and calls its (differently named) non-`static` allocation method, which would provide the same syntax. My concern is with modularity; in the future, somebody may want to have multiple memory managers in the same process, and then they'll have to refactor, instead of just doing that. (The latter has happened with the singleton `Caffe`.)

    Actually it looks like the new version has just one `static` data member in `GPUMemory` and `Manager` holds the management data non-staticly, so I think it's possible that actually you _did_ address my concern.
2. Re: @drnikolaev's 3 above and @borisfom's comment, my concern is solely with _memory_ overhead. I totally agree that no one should ever be calling `cudaMalloc` at runtime, and in fact no one ever does in normal use of Caffe. (So I don't agree that "most nets" should be faster -- all nets should run at the same speed, unless someone did something very bad (i.e., tried to call `cudaMalloc` during a layer forward/backward). Initialization I would expect to be faster, but no one should be relying on that.)

    So what sort of memory overhead should we expect for common nets? Is @lzhangzz's 25-30% above accurate? My concern is that (e.g.) 12 GB nets running on 12 GB cards are going to start crashing _by default_ with this PR, which would be a huge headache.
3. Re: Pycaffe, what does "works as expected" mean? There seem to be explicit code paths for the memory manager in the `caffe` tool and for tests, but not for pycaffe; so does pycaffe use `cudaMalloc`, or CUB? Can one switch between them? How?

drnikolaev:
@longjon thank you for the feedback. Here are few quick answers to the concerns raised. Especially important to note that 25-30% overhead is quite far from the reality. But first things first.

1. Yes, we did try to make it extensible. So, like you said, if we need to add some other manager (like pooling manager for example), we could easily do so.

2. I wrote a quick and dirty counter:
```c++
struct AllocationsCounter {
  // {allocated} -> {counter,overhead}
  typedef typename std::map<size_t, std::pair<size_t,size_t> > MapAllocs;
  MapAllocs map_allocs;

  void add_allocation(size_t allocated_size, size_t requested_size) {
    std::pair<size_t,size_t>& vals = map_allocs[allocated_size];
    vals.first++;
    vals.second += (allocated_size - requested_size);
  }

  AllocationsCounter() {}
  ~AllocationsCounter() {
    size_t total_allocated = 0UL, total_overhead = 0UL;
    MapAllocs::const_iterator it = map_allocs.begin();
    printf("Chunk Count Overhead\n");
    while (it != map_allocs.end()) {
      printf("%lu %lu %lu\n", it->first, it->second.first, it->second.second);
      total_overhead += it->second.second;
      total_allocated += it->first * it->second.first;
      ++it;
    }
    printf("Total allocated: %lu, total overhead: %lu\n",
        total_allocated, total_overhead);
  }
};
extern AllocationsCounter alloc_counter;
```
I added it to `util_allocator.cuh` file after `cudaMalloc` call:
```c++
    alloc_counter.add_allocation(search_key.bytes, bytes);
```
Note: here _we count real allocations from a device_, we do not look at the cache. Then I ran 1000 iterations of the Googlenet. Here are the results:

Chunk|Count|Overhead
-------|---|-------
64|12|360
128|24|192
256|51|552
512|54|1536
1024|39|8832
2048|15|10368
4096|15|864
16384|6|12288
32768|9|30720
65536|21|274176
131072|30|623616
262144|51|1634304
524288|45|6494208
1048576|28|9494528
2097152|35|18196480
4194304|38|34766848
4816896|6|0
5619712|4|0
6422528|14|0
7225344|2|0
8028160|4|0
8388608|6|0
9633792|2|0
10436608|21|0
11239424|2|0
12845056|10|0
14450688|2|0
16056320|2|0
19267584|4|0
24084480|10|0
25690112|34|0
26492928|10|0
30105600|3|0
38535168|16|0
41746432|2|0
51380224|16|0
96337920|2|0
154140672|4|0
205520896|2|0

**Total allocated: 5282599424, total overhead: 71549872**

Thus the overhead is **1.4%** on this particular example. Everyone who disagrees, please send us your examples -- we'll gladly consider them.

3. As of PyCaffe. Here is the snippet from `_caffe.cpp`:
```c++
#ifndef CPU_ONLY
shared_ptr<GPUMemory::Scope> gpu_memory_scope;
#endif

// Selecting mode.
void set_mode_gpu() {
  Caffe::set_mode(Caffe::GPU);
#ifndef CPU_ONLY
  vector<int> gpus;
  int count = 0;
  CUDA_CHECK(cudaGetDeviceCount(&count));
  for (int i = 0; i < count; ++i) {
    gpus.push_back(i);
  }
  CHECK_GT(gpus.size(), 0);
  gpu_memory_scope.reset(new GPUMemory::Scope(gpus));
#endif
}
```
In order to use CUB one needs to invoke `set_mode_gpu()`. Please let us know if some other call would be better to use here.

shelhamer:
> my concern is solely with memory overhead.
> [...] So what sort of memory overhead should we expect for common nets? Is @lzhangzz's 25-30% above accurate? My concern is that (e.g.) 12 GB nets running on 12 GB cards are going to start crashing by default with this PR, which would be a huge headache.

Right. In master Caffe all allocation is done during initialization or the first forward/backward so there is generally no time overhead, and this PR does not seem to add time overhead.

However, controlling memory overhead is crucial. We cannot suffer memory overhead since many current nets will crash as @longjon indicated. I benchmarked [VOC FCN-8s](https://github.com/shelhamer/fcn.berkeleyvision.org/blob/master/voc-fcn8s/deploy.prototxt) forward/backward and and CaffeNet solving without noting significant overhead: the difference in memory consumed was <10mb.

@drnikolaev @borisfom What kind of memory overhead are you observing in QA?

@lzhangzz could your overhead been due to different choices of engine/cuDNN method?

thatguymike:
I will note the point of this work is to allow us to stop hard coding workspace for cuDNN which should actually allow tolerance of larger networks when running cuDNN since we can waste memory in the current implementation.  For cuBLAS based routines that use small buffers, it may add some overhead, but as Sergei shows above, we have seen very little overhead even in near pathological cases.

As you can see in the histogram Sergei sent out, the overheads are in really small allocations which aren't generally parts of most layers.  So you would need something really pathological with MANY small iterations that force the caching allocator up a factor of 2 in buck size.   It's possible, but we haven't seen it.

lzhangzz:
@shelhamer the large overhead is caused by the large value of MAX_BIN in the version I tested, for the latest version in which the value is set to 22, the overhead is almost gone.

drnikolaev:
Added MultiWorkspace for multi-GPU setups. @longjon please have a look. Thank you.

longjon:
Okay, sorry for the delay folks, I've returned. Let me summarize the current state of this:

* I've done a cursory test myself and the branch built and ran seamlessly and without undue overhead (for CaffeNet, at least).
* The parallel changes are okay with me; I'll defer to others as I'm less intimate with that code.
* The `SyncedMemory` changes are trivial and as expected.
* The pycaffe integration is in `set_mode_gpu`; that's fine with me.
* Almost all other changes are isolated to `gpu_memory.*`.
* `cudaMalloc` is available as a fallback (via hardcoded default), and should revert to exactly pre-patch behavior for non-cuDNN layers.
* Debugging messages now use `DLOG`.

I'll make a few remaining inline comments, but then I don't see anything blocking merge, pending any further comments from core people.

include/caffe/util/gpu_memory.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+namespace caffe {
+
+struct GPUMemory {
+  static void GetInfo(size_t* free_mem, size_t* used_mem) {
+    return mgr_.GetInfo(free_mem, used_mem);
+  }
+
+  template <class Any>
+  static void allocate(Any** ptr, size_t size,
```
longjon: Non-trivial methods should be CamelCase, following the style guide (e.g., `GetInfo` is above, so why the inconsistency here?).

tools/caffe.cpp
```diff
@@ -303,7 +319,17 @@ int time() {
 
   // Set device id and mode
   vector<int> gpus;
+#ifndef CPU_ONLY
+  // Read flags for list of GPUs
   get_gpus(&gpus);
+  while (gpus.size() > 1) {
```
longjon: This reads a bit oddly to me... can't you just reconstruct the `vector` as a singleton (using its iterator)? And why bother to print the GPUs that we _aren't_ using?

src/caffe/test/test_caffe_main.cpp
```diff
@@ -17,18 +18,30 @@ int main(int argc, char** argv) {
...
-    cout << "Setting to use device " << device << endl;
+    devices.push_back(atoi(argv[1]));
+    cudaSetDevice(devices[0]);
   } else if (CUDA_TEST_DEVICE >= 0) {
     // Use the device assigned in build configuration; but with a lower priority
-    device = CUDA_TEST_DEVICE;
+    devices.push_back(CUDA_TEST_DEVICE);
   }
+
+  if (devices.size() == 1) {
```
longjon: The logic around here reads in a confusing way to me... e.g., doesn't `cudaSetDevice` get called twice if a device is given on the command line? And the conditional here isn't really about the number of devices, it's about whether either case above was invoked...

I think this would read better if the `else` below were joined to the `if` above, so we have just three mutually exclusive cases. And then I guess we can call `cudaSetDevice` on the first device always (what's the behavior of the third case now in that respect? Do we even need to call it at all? Why wasn't it called for the build option before?).

longjon:
Hm, I can't figure out why the last comment I made shows as being on an outdated diff, even though it appears in the current commit history (GitHub issue?). Anyway, take note, I believe that last comment is still relevant.

drnikolaev:
@longjon thanks for trying this out. I've just got back from vacation too and we discussed your notes. The problem is that every change (regardless of its size) requires few days of QA to run regression tests. This is quite expensive. Would it be ok to postpone the changes you proposed above till next PRs delivering real goods (like NCCL and FIndEx family of calls choosing really best convolution algorithms)? This would be real time and money saver for us and for the community.

longjon:
I'm not sure I understand the constraint there, since the QA you're mentioning is, as far as I know, opaque to BVLC/caffe (is it not?). I.e., caffe/master and whatever code you've QA'ed are necessarily divergent.

(Offline we discussed making these remaining changes ourselves to push this forward, which would, I presume, have the same issue.)

In general, I'm okay with merging code that makes any kind of improvement, even if future work is expected. But I don't think it's a good idea, e.g., to merge code that violates style rules; that essentially leaves the repo in an invalid state, and encourages other committers to break those rules.

thatguymike:
The QA cost is opaque to BVLC, not to us.  Since this is a significant change, we take it through pretty rigorous testing.  That has cost.  If these changes are required to move this PR forward, we will try to get to it after our current development and release cycle which is not a short term thing.  But this PR is holding up other changes and PRs we basically have ready to go to bring BVLC/Caffe performance up to being at least somewhat close to other frameworks and Nvidia's branch of Caffe.  (Performance delta is quite significant now on larger networks and multiGPU)

Your project, so your call here on what you want to do.  I certainly understand your concerns but the long review cycle, changes, and rebasing has real cost to us.  We have no issue with you making the changes and incuring the testing costs on your side and we'll pick up and continue with our next sets of PRs when we get around to rebasing those against changes you may make and retest and QA your changes with our updates.  

No ill will intended.  We are trying to be good citizens and give back to the BVLC master branch the enhancements we've made, but at some point it becomes onerous to do so.

longjon:
What I don't understand is why Nvidia's QA needs to track caffe/master. It seems difficult for us to make merging decisions based on a process we're not privy to, and in fact we don't, and Nvidia surely doesn't QA every commit we make.

Nvidia can PR to BVLC code which has not been evaluated according to Nvidia's QA so long as it passes BVLC's (i.e., Travis). Nvidia can internally use code which has passed its own QA even if it is not identical to BVLC's caffe/master. If Nvidia's QA subsequently finds issues with BVLC's caffe/master, Nvidia is welcome (and encouraged) to PR fixes.

BVLC does not require Nvidia to run a multi-day QA process to make changes to PRs, nor does it prohibit Nvidia from running code which is not a commit to BVLC's caffe/master. The coupling between these two is, as far as I can tell (and feel free to correct me if I am missing something) self-imposed by Nvidia.

So my suggestion is that Nvidia revise PRs against Travis, like any other committer, paying the same cost as any other committer. Whatever additional QA Nvidia wants to do internally is up to Nvidia, and might inform but need not track that process.

(Unless, of course, Nvidia wants to make its QA transparent; but that is another discussion.)

We of course thank Nvidia for its welcome contributions, but we do have to merge code according to the information available to BVLC/caffe, and we can't reasonably take into account costs which Nvidia is imposing on itself as part of that process...

At present my inclination is to continue with the previously discussed plan, i.e., BVLC will make final changes and merge this PR.

shelhamer:
> At present my inclination is to continue with the previously discussed plan, i.e., BVLC will make final changes and merge this PR.

Right, let's continue with that plan to make merge progress and reduce memory consumption. Thanks @longjon

shelhamer:
@drnikolaev did you intend to close this branch? I believe there were very few changes left for merge. Thanks.

drnikolaev:
We are making improvements and we also found&fixed some bugs. So, we'll better open a new one when we'd be sure it's good and stable.

shelhamer:
Ok, got it. Thanks for the update and continued work on memory pool integration.

PR #4194, opened 2016-05-21
Added the keyword argument `layer_name` in the method set_input_arrays. 

tiagofrepereira2012:
With this feature it is possible to set multiple MemoryData Layers in the same Network.
Unit tests are passing.
Hope everything is correct and useful


tiagofrepereira2012:
Ready to be merged

python/caffe/pycaffe.py
```diff
@@ -248,16 +248,23 @@ def _Net_forward_backward_all(self, blobs=None, diffs=None, **kwargs):
     return all_outs, all_diffs
 
 
-def _Net_set_input_arrays(self, data, labels):
+def _Net_set_input_arrays(self, data, labels, layer_name=""):
```
ajtulloch: can you use `layer_name=None`? I think that would be more idiomatic.

python/caffe/_caffe.cpp
```diff
@@ -115,10 +115,18 @@ void Net_Save(const Net<Dtype>& net, string filename) {
 }
 
 void Net_SetInputArrays(Net<Dtype>* net, bp::object data_obj,
-    bp::object labels_obj) {
+    bp::object labels_obj, const string &layer_name) {
+
+  shared_ptr<MemoryDataLayer<Dtype> > md_layer;
+  if (layer_name == "") {
```
ajtulloch: None in Python is `bp::object()` in C++, if that helps this comparison here. You can use `bp::extract<std::string>(layer_name)` to get the string out as well.

tiagofrepereira2012:
Thanks for the comments @ajtulloch 
Just pushed the modifications

ajtulloch:
I don't think that code is correct re. the testing against empty string. Could you add some unit tests that validate:

set_input_array with 

a) no layer name passed, no memory data layer as layers[0] => Exception
b) no layer name passed, memory data layer as layers[0] => layers[0] has inputs set (and verified with a forward pass)
c) layer name passed to arbitrary layer, memory data layer as that layer => that layer has inputs set (and verified with a forward pass)

tiagofrepereira2012:
Sorry for the delay @ajtulloch, I was busy with my exams.
Yes sure I can do it.
I will do it from scratch since there is no test cases for this method.

python/caffe/_caffe.cpp
```diff
@@ -115,10 +115,20 @@ void Net_Save(const Net<Dtype>& net, string filename) {
 }
 
 void Net_SetInputArrays(Net<Dtype>* net, bp::object data_obj,
-    bp::object labels_obj) {
+    bp::object labels_obj, bp::object layer_name) {
+
+  const string str_layer = bp::extract<std::string>(layer_name);
+
+  shared_ptr<MemoryDataLayer<Dtype> > md_layer;
+  if (str_layer == "") {
```
ajtulloch: e.g.g this should be comparing against bp::object() if layer_name is None.

tiagofrepereira2012:
Just pushed some code.

Basically I added some unit tests to test the method `set_input_arrays` and now the binding is comparing the `layer_name` with `Py_None` instead of the string.

python/caffe/_caffe.cpp
```diff
@@ -117,17 +117,17 @@ void Net_Save(const Net<Dtype>& net, string filename) {
 void Net_SetInputArrays(Net<Dtype>* net, bp::object data_obj,
     bp::object labels_obj, bp::object layer_name) {
 
-  const string str_layer = bp::extract<std::string>(layer_name);
-
   shared_ptr<MemoryDataLayer<Dtype> > md_layer;
-  if (str_layer == "") {
+  if (layer_name.ptr() == Py_None) {
```
ajtulloch: Can this be:

```
if (layer_name == bp::object()) {
  ...
}
```

python/caffe/_caffe.cpp
```diff
@@ -117,17 +117,17 @@ void Net_Save(const Net<Dtype>& net, string filename) {
...
     md_layer =
       boost::dynamic_pointer_cast<MemoryDataLayer<Dtype> >(net->layers()[0]);
   } else {
+    const string str_layer = bp::extract<std::string>(layer_name);
     md_layer =
       boost::dynamic_pointer_cast<MemoryDataLayer<Dtype> >
           (net->layer_by_name(str_layer));
   }
+
```
ajtulloch: rm extra newline.

python/caffe/test/test_memory_layer.py
```diff
@@ -0,0 +1,75 @@
...
+        self.solver.net.set_input_arrays(self.data, self.labels, 
+         'data') #Using the layer name
+        self.solver.net.forward()
+
+        self.assertTrue(np.any(np.isclose(self.data, 
+          self.solver.net.blobs['data'].data, rtol=1e-05, atol=1e-08))
+          ) #Testing after forward
+
+
+        self.solver.net.set_input_arrays(self.data,
```
ajtulloch: fix indentation:

```
# using net.layer[0]
self.solver.net.set_input_arrays(self.data, self.labels)
```

python/caffe/test/test_memory_layer.py
```diff
@@ -0,0 +1,75 @@
...
+        self.assertTrue(np.any(np.isclose(self.data, 
+          self.solver.net.blobs['data'].data, rtol=1e-05, atol=1e-08))
+          ) #Testing after forward
+
+
+        self.solver.net.set_input_arrays(self.data, 
+           self.labels)  #Using the layer [0]
+        self.solver.net.forward()
+
+        self.assertTrue(np.any(np.isclose(self.data,
```
ajtulloch: `np.testing.assert_array_equal(self.data, self.solver.net.blobs['data'].data)`

python/caffe/test/test_memory_layer.py
```diff
@@ -0,0 +1,75 @@
...
+        self.solver.net.forward()
+
+        self.assertTrue(np.any(np.isclose(self.data, 
+          self.solver.net.blobs['data'].data, rtol=1e-05, atol=1e-08))
+          ) #Testing after forward
+
+
+    def test_raise(self):
+        "Catching Exceptions"
+        self.assertRaises(RuntimeError, self.solver.net.set_input_arrays,self.data, self.labels, "conv")
```
ajtulloch: IMO

```
with self.assertRaises(RuntimeError):
  self.solver.net.set_input_arrays(self.data, self.labels, "conv")
```

would be cleaner.

python/caffe/_caffe.cpp
tiagofrepereira2012: I will fix that with your solution, but to compare with Py_None is more clean, since we are dealing with the python bindings.
The method `.ptr()` will returns a `PyObject*` and the comparison with Py_None is the right way to do it.

tiagofrepereira2012:
strange, I cannot reproduce this segmentation fault

tiagofrepereira2012:
Now is everything green

tiagofrepereira2012:
Any news about this one?

seanbell:
Before this can be merged, it should be squashed into a single commit with a rebase:
```
git fetch --all
git rebase -i origin/master
```

nitnelave:
Similar PR, different approach: https://github.com/BVLC/caffe/pull/4346

nitnelave:
Maybe change the test net to have two MemoryDataLayers? This is what we're trying to achieve, after all.

tiagofrepereira2012:
Shall I do the rebase?

tiagofrepereira2012:
Well, it is done

nitnelave:
You still have too many commits. You can rewrite your history with git's interactive-rebase, (`git rebase -i master`), to squash commits together and re-order them.

tiagofrepereira2012:
Sorry, now is properly squashed and all green

nitnelave:
Shouldn't the test have two memorydatalayers?

tiagofrepereira2012:
Hi guys, sorry for the delay.
I was in a conference this week.

Just pushed the requested test. Everything is green and ready to be merged.

tiagofrepereira2012:
Any news about this one?
Thanks

ajtulloch:
This LGTM. @longjon @shelhamer?

PR #4198, opened 2016-05-23
Generalize bilinear interpolation filler to N-D multilinear/multicubic/lanczos filler

christianpayer:
This branch implements an n-dimensional generalization of the bilinear filler (#2213) and adds cubic and lanczos fillers.

It makes #3984 obsolete, as it uses a new common base class for all interpolation fillers. The base class InterpolationFillerBase calculates the weight values by calling the virtual interpolation functions of its derived classes. This PR implements linear, cubic and lanczos interpolation fillers, while additional interpolation functions (e.g. Hermite, Mitchell, Gaussian) may be implemented easily by deriving the class InterpolationFillerBase and implementing the virtual functions f() and support().

I have written a simple test script in python, which tests the fillers with various scaling factors in x and y. I will upload it after some cleanup.

If you have comments/suggestions, let me know!

### DIFF
--- a/include/caffe/filler.hpp
+++ b/include/caffe/filler.hpp
@@ -296,6 +296,40 @@ class MultilinearFiller : public InterpolationFillerBase<Dtype> {
 };
 
 /**
+ * @brief Fills a Blob with coefficients for multicubic interpolation.
+ *
+ * Set type of weight_filler to 'multicubic'. Support is set to 2. See
+ * InterpolationFillerBase for a detailed explanation of how to set kernel_size,
+ * stride and pad.
+ * If you apply this to an image, this operation is equivalent to the following
+ * call in Python with Scikit.Image.
+ * \code{.py}
+ * out = skimage.transform.rescale(img, (factor_y, factor_x), mode='constant', order=3, clip=False, cval=0)
+ * \endcode
+ */
+template <typename Dtype>
+class MulticubicFiller : public InterpolationFillerBase<Dtype> {
+ public:
+  explicit MulticubicFiller(const FillerParameter& param)
+      : InterpolationFillerBase<Dtype>(param) {}
+ protected:
+  virtual Dtype f(Dtype x) {
+    const Dtype A = -0.5;
+    if (std::abs(x) <= 1) {
+      return (A + 2) * std::pow(std::abs(x), 3)
+          - (A + 3) * std::pow(std::abs(x), 2) + 1;
+    }
+    if (std::abs(x) < 2) {
+      return A * std::pow(std::abs(x), 3)
+          - 5 * A * std::pow(std::abs(x), 2)
+          + 8 * A * std::abs(x) - 4 * A;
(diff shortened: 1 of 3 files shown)


christianpayer:
This jupyter notebook tests the new fillers:
http://nbviewer.jupyter.org/gist/christianpayer/3ee54b5c6463927410fff0fbc7cf556d

Coldmooon:
Can this PR be used to downsample a feature map?

christianpayer:
@Coldmooon The fillers can also be used to downsample a feature map with fractional factors (1/2, 1/3, 1/4,...), as they work with Deconvolution and Convolution layers, where the strides define the up- and downsampling factors, respectively.
As these fillers are originally intended to be used as deconvolution fillers, the results for convolution may not give the desired results, i.e. the resulting feature map values are scaled by factor_x * factor_y. So if you use these fillers for downsampling, you need to append an additional layer (e.g. PowerLayer) that scales the values with 1 / (factor_x * factor_y) in order to get the same results as with Matlab, etc.

Coldmooon:
@christianpayer Thank you very much~! Recently, I've been studying resampling by caffe. Your PR helps me a lot.

ajtulloch:
Really nice work! Could you add some unit tests, following the existing examples in `test_filler.cpp`?

christianpayer:
@Coldmooon Your changes for downsampling look OK. You are right, there is a difference between the downsampling of this PR and the reference created with skimage. I don't know exactly, how skimage performs downsampling, as it can be implemented in multiple ways. Even if you compare the outputs of skimage's linear/cubic resampling and opencv's, you will see many differences.
With this PR I tried to resample (almost) exactly like Matlab's imresize. So if you compare the outputs pixel by pixel, you should compare them with Matlab (and hopefully see no difference).

Regarding your question of non-integer factors, this is not possible with a caffe convolution. With an integer factor, you will get the same kernel for every x/y-coordinate of the image. With non-integer factors, this is not the case. You would need different kernels for different coordinates of the image, which is not possible with the convolution layer.

If you want to get more insights into this, I suggest you to debug Matlab's imresize and look for the function 'contributions' and compare the outputs for integer and non-integer factors.

christianpayer:
@ajtulloch Thanks! I can add some unit test, but I don't know, what should be tested. I could check, whether the kernels sum up to a fixed value. Or I could compare with hardcoded weights (possibly created from Matlab?) and check for differences.
What would you suggest?

Coldmooon:
@christianpayer I tested upsampling and downsampling with non-integer factors. The resampled images for the two cases look like noise and look like the downsampled image with an integer factor but without appending the `Power` layer. Now I can see what you said about the non-integer case. Thank you~. Maybe it's possible to implement a new `nearest interpolation` layer for non-integer factor. By the way, it seems that `STN` could create a `bilinear` sampling kernel. I don't know if `STN` can work for non-integer factors.

PR #4214, opened 2016-05-25
Remove misleading comment from a test file

lukeyeager:
Looks like it was accidentally copy+pasted from here:
https://github.com/BVLC/caffe/blob/rc3/include/caffe/test/test_caffe_main.hpp#L1-L2


shelhamer:
Yeah, thanks for cleaning up.

Issue #4186, opened 2016-05-20
Segmentation fault training network with no image


doe300:
When trying to load a network for training (e.g. models/bvlc_reference_caffenet) with no images in the data-set, 
`ImageDataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>&, const vector<Blob<Dtype>*>&)`
 crashes in line 68 
`cv::Mat cv_img = ReadImageToCVMat(root_folder + lines_[lines_id_].first, new_height, new_width, is_color);` with a SIGSEGV (Segmentation fault)
since there are no entries in `lines_`.

Ergo: There should be a test before, to dump a useful error-message and gracefully exit the function.

Edit:
This also occurs in the today's master with DEBUG enabled

seanbell:
Sure, please submit a PR making this change.
Add a `CHECK_GT` macro to check the size of `lines_`.

malreddysid:
@seanbell 
Hey,
I've added a PR #4218  for the issue.

doe300:
Done, thx!

PR #4203, opened 2016-05-24
Add support for specifying a separator in the image data layer

elezar:
Although #3653 has been open for a while now, I noticed that #4059 (which adds support for file names with spaces in) means that my implementation may require a little more work. I therefore thought that I would break things up into multiple PRs so that I can get a better grip on the changes required.

This PR will add support for a specifying a separator (e.g. a comma) instead of assuming a space-separated list. This was also mentioned in https://github.com/BVLC/caffe/pull/3433#issuecomment-166772986.

I have also restructured the test class somewhat to make it easier to write tests for the image data layer.

PR #4218, opened 2016-05-26
Resolve SIGSEGV error in image_data_layer.cpp

malreddysid:
PR to resolve issue [#4186](https://github.com/BVLC/caffe/issues/4186).

Added check in image_data_layer.cpp to ensure that if an empty list is given as source to ImageData layer, segmentation fault won't occur.

### DIFF
--- a/src/caffe/layers/image_data_layer.cpp
+++ b/src/caffe/layers/image_data_layer.cpp
@@ -46,6 +46,8 @@ void ImageDataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>& bottom,
     lines_.push_back(std::make_pair(line.substr(0, pos), label));
   }
 
+  CHECK(!lines_.empty()) << "File is empty";
+
   if (this->layer_param_.image_data_param().shuffle()) {
     // randomly shuffle data
     LOG(INFO) << "Shuffling data";


src/caffe/layers/image_data_layer.cpp
```diff
@@ -46,6 +46,9 @@ void ImageDataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>& bottom,
     lines_.push_back(std::make_pair(line.substr(0, pos), label));
   }
 
+  // Check that image list is not empty.
+  CHECK_GT(lines_.size(), 0);
+
```
flx42: You should use method `empty()` instead.

malreddysid: Hey felix, how is that better? I think the functionality is same.

flx42: It's better practice, more idiomatic.
It doesn't matter for a vector, but for a list it does, in C++98, the complexity `list.size()` is not guaranteed to be constant.

malreddysid: Okay. Thanks. Should I make the change and submit another PR?

flx42: Just edit this PR. You need to `git commit --amend` your patch and then push force to your branch and it will automatically update this PR.

malreddysid:
Edited the code as per @flx42's suggestions.

src/caffe/layers/image_data_layer.cpp
```diff
@@ -46,6 +46,9 @@ void ImageDataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>& bottom,
     lines_.push_back(std::make_pair(line.substr(0, pos), label));
   }
 
+  // Check that image list is not empty.
```
longjon: You can omit this comment -- it does not add anything to the code.

malreddysid: Sure. Should I add a terminal output similar to line 73?

longjon:
Thanks, this looks good except one comment comment and then please squash (and force push) so your history appears as a single commit for this simple change (and perhaps use `git commit --amend` to write a slightly more descriptive one-line message, e.g., "check for non-empty ImageData filelist" (note that the information in the current message is already readily obtained with `git log --stat`)).

src/caffe/layers/image_data_layer.cpp
longjon: That would be a little nicer, yes.

malreddysid:
Edited the code as per @longjon's suggestions.

longjon:
Looks good, thanks!

ajtulloch:
Thanks @malreddysid

elezar:
Hi @malreddysid 

As part of #4203 I thought I would add a quick unit test for this PR and found the following:
- When the file is COMPLETELY empty then the code works as expected.
- When there is any whitespace in the file (e.g. a single newline), the following error is given:
```
[  DEATH   ] E0601 06:06:03.550549  5324 io.cpp:80] Could not open or find file 
[  DEATH   ] F0601 06:06:03.550745  5324 image_data_layer.cpp:77] Check failed: cv_img.data Could not load 
[  DEATH   ] *** Check failure stack trace: ***
[  DEATH   ]     @     0x7fc7170e7daa  (unknown)
[  DEATH   ]     @     0x7fc7170e7ce4  (unknown)
[  DEATH   ]     @     0x7fc7170e76e6  (unknown)
[  DEATH   ]     @     0x7fc7170ea687  (unknown)
[  DEATH   ]     @     0x7fc717ccedd5  caffe::ImageDataLayer<>::DataLayerSetUp()
[  DEATH   ]     @     0x7fc717cd8583  caffe::BasePrefetchingDataLayer<>::LayerSetUp()
[  DEATH   ]     @           0xad62ad  caffe::Layer<>::SetUp()
[  DEATH   ]     @           0xae04be  caffe::ImageDataLayerTest_WhitespaceFileRaiseError_Test<>::TestBody()
[  DEATH   ]     @           0xd4ed93  testing::internal::HandleExceptionsInMethodIfSupported<>()
[  DEATH   ]     @           0xd45817  testing::Test::Run()
[  DEATH   ]     @           0xd458be  testing::TestInfo::Run()
[  DEATH   ]     @           0xd459c5  testing::TestCase::Run()
[  DEATH   ]     @           0xd48718  testing::internal::UnitTestImpl::RunAllTests()
[  DEATH   ]     @           0xd489b7  testing::UnitTest::Run()
[  DEATH   ]     @           0x871b8f  main
[  DEATH   ]     @     0x7fc71179dec5  (unknown)
[  DEATH   ]     @           0x877d92  (unknown)
[  DEATH   ]     @              (nil)  (unknown)
[  DEATH   ] 
```
(the `[ DEATH ]` tags are from gtest).

Once #4203 is merged, could you look at adding some tests for this case?

For reference my two testcases are:
```
TYPED_TEST(ImageDataLayerTest, EmptyFileRaiseError) {
  typedef typename TypeParam::Dtype Dtype;

  // The following is needed to suppress the following warning:
  // [WARNING] ::Death tests use fork(), which is unsafe particularly in a
  //     threaded context.For this test, Google Test couldn't
  //     detect the number of threads.
  ::testing::FLAGS_gtest_death_test_style = "threadsafe";

  std::ofstream& output = this->stream();
  output.close();

  LayerParameter param;
  ImageDataParameter* image_data_param = param.mutable_image_data_param();
  image_data_param->set_source(this->filename_.c_str());
  ImageDataLayer<Dtype> layer(param);

  EXPECT_DEATH(layer.SetUp(this->blob_bottom_vec_, this->blob_top_vec_),
               "File is empty");

}

TYPED_TEST(ImageDataLayerTest, WhitespaceFileRaiseError) {
  typedef typename TypeParam::Dtype Dtype;

  // The following is needed to suppress the following warning:
  // [WARNING] ::Death tests use fork(), which is unsafe particularly in a
  //     threaded context.For this test, Google Test couldn't
  //     detect the number of threads.
  ::testing::FLAGS_gtest_death_test_style = "threadsafe";

  std::ofstream& output = this->stream();
  output << std::endl;
  output.close();

  LayerParameter param;
  ImageDataParameter* image_data_param = param.mutable_image_data_param();
  image_data_param->set_source(this->filename_.c_str());
  ImageDataLayer<Dtype> layer(param);

  EXPECT_DEATH(layer.SetUp(this->blob_bottom_vec_, this->blob_top_vec_),
               "File is empty");

}
```

malreddysid:
Hi @elezar,

Thanks for the suggestion. After #4203 gets merged, I'll add a check only the _'pos'_ value to prevent empty line being pushed into _'lines_'_.

elezar:
I don't know if it's as simple as that. If one has the following test:
```
std::ofstream& output = this->stream();
  // Insert 8 spaces into the test file.
  output << "        " << std::endl;
  output.close();
```
This also counts as an empty file, but simply checking pos will not be enough, as there are multiple spaces.

Note that this does not result in a segmentation violation, but does give a bit of a cryptic error message.

malreddysid:
#1971 could take care of all such bugs. But until that is implemented, I guess we could look for file extensions (.jpg, .png, etc) that are supported by OpenCV, along with **'pos**' value and length of label substring to check the validity of the file.

elezar:
Ok, if #1971 addressed this, then I will move the comments and tests there instead. Thanks!

PR #4236, opened 2016-05-30
fix spelling error in memory_data_layer.cpp

CDLuminate:
initalize -> initialize

applied in downstream in debian package:
http://anonscm.debian.org/cgit/debian-science/packages/caffe.git/tree/debian/patches/fix-spelling-error


shelhamer:
Thanks for the proofreading.

PR #4243, opened 2016-05-31
Rebase the ssd framework into Windows

rob1nu:


seanbell:
Was this an accidental PR?  I'm closing it because there is no message and it looks like an accident.

PR #4247, opened 2016-06-01
Fix vecLib search order

jklontz:
@CheeseWiz @shelhamer

Clients that have gone through OS X upgrades may have both the old `/System/Library/${__veclib_include_suffix}` and the new `/System/Library/Frameworks/Accelerate.framework/Versions/Current/${__veclib_include_suffix}`. The current behavior is to find `/System/Library/${__veclib_include_suffix}` first, resulting in two errors:

1. Compile time error missing `cblas.h` header file. It no longer lives alongside `vecLib.h` in `/System/Library/${__veclib_include_suffix}`.
2. Link time error no such framework `vecLib`. Should be using `Accelerate` framework.

Changing the order of the search paths allows Accelerate to be found first if it exists. I had to add `NO_DEFAULT_PATH` as otherwise `/System/Library/${__veclib_include_suffix}` is still found first, as CMake's default paths are checked before user-specified paths.

I believe this is the proper fix for [this Stack Overflow question](http://stackoverflow.com/questions/35544356/install-caffe-on-mac-error-fatal-error-cblas-h-file-not-found).

### DIFF
--- a/cmake/Modules/FindvecLib.cmake
+++ b/cmake/Modules/FindvecLib.cmake
@@ -14,9 +14,10 @@ set(__veclib_include_suffix "Frameworks/vecLib.framework/Versions/Current/Header
 
 find_path(vecLib_INCLUDE_DIR vecLib.h
           DOC "vecLib include directory"
-          PATHS /System/Library/${__veclib_include_suffix}
-                /System/Library/Frameworks/Accelerate.framework/Versions/Current/${__veclib_include_suffix}
-                /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk/System/Library/Frameworks/Accelerate.framework/Versions/Current/Frameworks/vecLib.framework/Headers/)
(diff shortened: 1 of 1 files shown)


jklontz:
Alternatively, searching for `cblas.h` instead of `vecLib.h` should resolve the issue.

cepiross:
For your information, in case of opencl branch, this issue was resolved by trying to find vecLibTypes.h instead of vecLib.h (https://github.com/BVLC/caffe/blob/opencl/cmake/Modules/FindvecLib.cmake#L15-19) without changing the order of the search paths. (https://github.com/BVLC/caffe/pull/3720)

elfprince13:
Is there any reason this hasn't been merged yet? I'm using Caffe as a submodule in a project and would prefer to (a) track the BVLC version, and (b) not have to manually patch.

shelhamer:
Thanks for the build fix @jklontz. It seems like this could potentially be simpler, as highlighted by @cepiross, but as this seems to find both the old library and new and works for others then it's fine for now.

PR #3068, opened 2015-09-15
merged pull#1271 for locally connected layer

raingo:
Merge #1271 to master branch. Everything is copied from there, except some little adjustments for the master branch

@jackculpepper, @shelhamer  please have a look.

PR #4251, opened 2016-06-01
Locally connected layer from raingo #3068

matthieudelaro:
Here is an attempt to finish the remaining work to merge raingo's implementation of locally connected layers into master branch:

- merge conflicts have been solved
- the filler (declared for test purposes only) has been moved inside the test itself

@jklontz
@raingo
@dougsouza

### DIFF
--- a/examples/siamese/mnist_siamese_local_solver.prototxt
+++ b/examples/siamese/mnist_siamese_local_solver.prototxt
@@ -1,0 +1,25 @@
+# The train/test net protocol buffer definition
+net: "examples/siamese/mnist_siamese_local_train_test.prototxt"
+# test_iter specifies how many forward passes the test should carry out.
+# In the case of MNIST, we have test batch size 100 and 100 test iterations,
+# covering the full 10,000 testing images.
+test_iter: 100
+# Carry out testing every 500 training iterations.
+test_interval: 500
+# The base learning rate, momentum and the weight decay of the network.
+base_lr: 0.01
+momentum: 0.9
+weight_decay: 0.0000
+# The learning rate policy
+lr_policy: "inv"
+gamma: 0.0001
+power: 0.75
+# Display every 100 iterations
+display: 100
+# The maximum number of iterations
+max_iter: 50000
+# snapshot intermediate results
+snapshot: 5000
+snapshot_prefix: "examples/siamese/mnist_siamese"
+# solver mode: CPU or GPU
+solver_mode: GPU
--- a/src/caffe/layers/local_layer.cu
+++ b/src/caffe/layers/local_layer.cu
@@ -1,0 +1,197 @@
+#include <vector>
+
+#include "caffe/filler.hpp"
+#include "caffe/layers/local_layer.hpp"
+#include "caffe/util/im2col.hpp"
+
+namespace caffe {
+
+template <typename Dtype>
+__global__ void local_update1_gpu_kernel(
+    const Dtype* data_A, const Dtype* data_B,
+    Dtype* data_R, const int filter_num,
+    const int location_num, const int output_num) {
+  int total = filter_num * location_num * output_num;
+  CUDA_KERNEL_LOOP(index, total) {
+    int p = index % location_num;
+    int n = (index / location_num) % filter_num;
+    int q = (index / location_num) / filter_num;
+    data_R[index] += data_A[q*location_num+p] * data_B[n*location_num+p];
+  }
+}
+
+template <typename Dtype>
+void local_update1_gpu(
+    const Dtype* data_A, const Dtype* data_B,
+    Dtype* data_R, const int filter_num,
+    const int location_num, const int output_num) {
+  // data_A is output_num x location_num
+  // data_B is filter_num x location_num
+  // data_R is output_num x filter_num x location_num,
+  //      the update performed is Rqnp += Aqp * Bnp
+  const int nthreads = filter_num * location_num * output_num;
+
+  local_update1_gpu_kernel<Dtype>  // NOLINT_NEXT_LINE(whitespace/operators)
+    <<<CAFFE_GET_BLOCKS(nthreads), CAFFE_CUDA_NUM_THREADS>>>(
+      data_A, data_B, data_R, filter_num, location_num, output_num);
+  CUDA_POST_KERNEL_CHECK;
+}
+
+// Explicit instantiation
+template void local_update1_gpu<float>(
+    const float* data_A, const float* data_B,
+    float* data_R, const int filter_num,
+    const int location_num, const int output_num);
+template void local_update1_gpu<double>(
+    const double* data_A, const double* data_B,
+    double* data_R, const int filter_num,
+    const int location_num, const int output_num);
+
+template <typename Dtype>
+__global__ void local_update2_gpu_kernel(
+    const Dtype* data_A, const Dtype* data_B,
+    Dtype* data_R, const int filter_num,
+    const int location_num, const int output_num) {
+  int total = filter_num * location_num;
+  CUDA_KERNEL_LOOP(index, total) {
+    int p = index % location_num;
+    int n = (index / location_num);
+    for (int q = 0; q < output_num; q++) {
+      data_R[index] +=
+        data_A[q*location_num+p] * data_B[(q*filter_num+n)*location_num+p];
+    }
+  }
+}
+
+template <typename Dtype>
+void local_update2_gpu(const Dtype* data_A, const Dtype* data_B,
+    Dtype* data_R, const int filter_num,
+    const int location_num, const int output_num) {
+  // data_A is output_num x location_num
+  // data_B is output_num x filter_num x location_num
+  // data_R is filter_num x location_num,
+  //    the update performed is Rnp += \sum_q(Aqp * Bqnp)
+  int nthreads = filter_num * location_num;
+
+
+  local_update2_gpu_kernel<Dtype>  // NOLINT_NEXT_LINE(whitespace/operators)
+    <<<CAFFE_GET_BLOCKS(nthreads), CAFFE_CUDA_NUM_THREADS>>>(
+        data_A, data_B, data_R, filter_num,
+        location_num, output_num);
+  CUDA_POST_KERNEL_CHECK;
+}
+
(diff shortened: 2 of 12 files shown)


shelhamer:
@matthieudelaro thanks for carrying on the prior work on the locally connected layer type.

Please rebase to squash all of @raingo's commits since lint fixes and merges are not informative history.  You can keep both of your commits even though "Fix Conflicts" is generic since you are continuing the earlier work and should get credit.

examples/siamese/mnist_siamese_local_solver.prototxt
```diff
@@ -0,0 +1,25 @@
+# The train/test net protocol buffer definition
```
shelhamer: I think the local LeNet example suffices, if any is needed at all, and there isn't a need to duplicate it for a siamese net too.

src/caffe/test/test_local_layer.cpp
```diff
@@ -0,0 +1,136 @@
...
+  convolution_param->set_stride(1);
+  convolution_param->set_num_output(1);
+  convolution_param->mutable_bias_filler()->set_type("constant");
+  convolution_param->mutable_bias_filler()->set_value(0.1);
+  shared_ptr<Layer<Dtype> > layer(
+      new LocalLayer<Dtype>(layer_param));
+  layer->SetUp(this->blob_bottom_vec_, this->blob_top_vec_);
+
+  // fill the weights of the layer
+  LOG(INFO) << "Doing mutable cpu";
```
shelhamer: Drop the `LOG`s: there's no need for logging in the test.

src/caffe/test/test_local_layer.cpp
```diff
@@ -0,0 +1,136 @@
...
+  EXPECT_EQ(this->blob_top_->num(), 2);
+  EXPECT_EQ(this->blob_top_->channels(), 3);
+  EXPECT_EQ(this->blob_top_->height(), 2);
+  EXPECT_EQ(this->blob_top_->width(), 1);
+}
+
+
+TYPED_TEST(LocalLayerTest, TestSimpleConvolution) {
+  typedef typename TypeParam::Dtype Dtype;
+  // We will simply see if the convolution layer carries out averaging well.
```
shelhamer: Out-of-date comment. This is checking whether the local layer with constant weights does averaging. Could you add a test that actually checks whether the weights are locally different? This does not actually check that kernel weighs are not identical.

src/caffe/test/test_local_layer.cpp
```diff
@@ -0,0 +1,136 @@
...
+  vector<Blob<Dtype>*> blob_bottom_vec_;
+  vector<Blob<Dtype>*> blob_top_vec_;
+};
+
+TYPED_TEST_CASE(LocalLayerTest, TestDtypesAndDevices);
+
+TYPED_TEST(LocalLayerTest, TestSetup) {
+  typedef typename TypeParam::Dtype Dtype;
+  LayerParameter layer_param;
+  LocalParameter* convolution_param =
```
shelhamer: Not a `convolution_param` but a `local_param`, though you could consider refactoring to make `LocalLayer` share the convolution param with `ConvolutionLayer` and `DeconvolutionLayer`.

elezar: :+1: for the idea to reuse `ConvolutionParameter`.

kevin-li:
Would really like to have this merged, my project needs this feature. Thank you a bunch!

src/caffe/layers/local_layer.cu
```diff
@@ -0,0 +1,197 @@
...
+template <typename Dtype>
+void LocalLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
+
+  Dtype* x_data = col_buffer_.mutable_gpu_data();
+  const Dtype* weight = this->blobs_[0]->gpu_data();
+  const Dtype* bottom_data = bottom[0]->gpu_data();
+  Dtype* top_data = top[0]->mutable_gpu_data();
+
+  Blob<Dtype> E;
```
ajtulloch: Can you cache these? This means that there's a malloc/free on every iteration (not sure if you care though).

src/caffe/layers/local_layer.cu
```diff
@@ -0,0 +1,197 @@
...
+  E.Reshape(1, 1, 1, K_);
+  FillerParameter filler_param;
+  filler_param.set_value(1);
+  ConstantFiller<Dtype> filler(filler_param);
+  filler.Fill(&E);
+
+  Blob<Dtype> intermediate;
+  intermediate.Reshape(1, 1, K_, N_);
```
ajtulloch: Ditto.

src/caffe/layers/local_layer.cu
```diff
@@ -0,0 +1,197 @@
...
+void LocalLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
+    const vector<Blob<Dtype>*>& top) {
+
+  Dtype* x_data = col_buffer_.mutable_gpu_data();
+  const Dtype* weight = this->blobs_[0]->gpu_data();
+  const Dtype* bottom_data = bottom[0]->gpu_data();
+  Dtype* top_data = top[0]->mutable_gpu_data();
+
+  Blob<Dtype> E;
+  E.Reshape(1, 1, 1, K_);
```
ajtulloch: Why not just do caffe_set(K, ..., 1.0)?

src/caffe/layers/local_layer.cu
```diff
@@ -0,0 +1,197 @@
...
+
+  Blob<Dtype> intermediate;
+  intermediate.Reshape(1, 1, K_, N_);
+  for (int n = 0; n < num_; n++) {
+    im2col_gpu(bottom_data + bottom[0]->offset(n), channels_, height_,
+        width_, kernel_size_, kernel_size_,
+        pad_, pad_, stride_, stride_,
+        dilation_, dilation_, x_data);
+
+    for (int m = 0; m < num_output_; m++) {
```
ajtulloch: Launching `num_output_` kernels per forward pass is suboptimal for performance, can you batch this?

src/caffe/test/test_local_layer.cpp
matthieudelaro: > Please rebase to squash all of @raingo's commits since lint fixes and merges are not informative history. 

What is the preferred way of doing it? If I squash in the same branch, I think that there will be conflict since it is already pushed. But if I do it in a new branch, then I should open a new PR?

matthieudelaro: local layer could you use convolution_param, but doesn't support all of its features. Then should it output a warning message when one tries to use such a feature?

For example, pad_h and pad_w are not supported. The padding is assumed to be the same in width and in height.

src/caffe/layers/local_layer.cu
matthieudelaro: Which line(s) are you talking about?

ajtulloch: `Blob<Dtype> E;` is stack-allocating E, so the memory it owns (allocated in E->Reshape/mutable_gpu_data()) is going to be allocated then freed in each iteration.

examples/siamese/mnist_siamese_local_solver.prototxt
matthieudelaro: done

src/caffe/test/test_local_layer.cpp
matthieudelaro: done

src/caffe/test/test_local_layer.cpp
matthieudelaro: done

src/caffe/test/test_local_layer.cpp
matthieudelaro: done
Now LocalLayer uses convolution_param, and can handled non-square padding, kernel size, etc.

src/caffe/layers/local_layer.cu
matthieudelaro: done

src/caffe/layers/local_layer.cu
matthieudelaro: done

src/caffe/layers/local_layer.cu
matthieudelaro: done

src/caffe/layers/local_layer.cu
matthieudelaro: You mean, create a dedicated kernel which would handle all the output channels at once?

matthieudelaro:
Hi all,

I tried to take all of your comments into account. The result is that LocalLayer inherits from BaseConvolutionLayer, can handle non-square parameters, goes through more tests, caches blobs for intermediate computations, and convolution_param is used instead local_param.

Now 
> Please rebase to squash all of @raingo's commits since lint fixes and merges are not informative history. 

What is the preferred way of doing it? If I squash in the same branch, I think that there will be conflict since it is already pushed. But if I do it in a new branch, then I should open a new PR?

Future performance bonus? As @ajtulloch commented about Forward_gpu:
> Launching num_output_ kernels per forward pass is suboptimal for performance, can you batch this?

ajtulloch:
Check out https://github.com/torch/nn/blob/master/lib/THNN/generic/SpatialConvolutionLocal.c for an implementation that uses im2col + a cublas batched gemm to do this in a single kernel per num-output.

jklontz:
@matthieudelaro What would it take to help incentivize completion of this? Please reach out to me, my email is listed in my profile. Thanks!

matthieudelaro:
Hi @jklontz. Sorry for late reply. I'll be offline in September, and I doubt I'll have time to work on this later.

Here is the thing: commits must be rebased (just one commit per author), and an optimization in CUDA is expected (I hope BVLC will accept it without it though, so that at least people can use it, and then anyone can improve) #4251. Sorry I hadn't time to finish this. If anyone is up for rebasing, he is welcome :)

As for the travis check that failed: none of them failed until I merged branch master into my branch. So I'm confident that removing what came from the up-to-date-2-month-ago master branch will solve the problem.

PR #4254, opened 2016-06-01
Cache protobuf3 build in TravisCI

lukeyeager:
This should shave a few minutes off the "python3-cmake" build. It's a similar solution to https://github.com/NVIDIA/DIGITS/pull/788, which is working great.

I'm not actually sure how TravisCI will handle this since only one of the nine build targets is building protobuf3. Here's the relevant documentation:
https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices

If I'm reading it correctly, each build target should have its own cache, so 8 of the targets should cache an empty directory, and the "python3-cmake" target should cache the build.

I can't really test this until it's merged, but it also can't break anything too seriously. Worst case is the python3 target stops working and we revert this commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,10 @@ env:
     - BUILD_NAME="cudnn-cmake" WITH_CMAKE=true WITH_CUDA=true WITH_CUDNN=true
 
 cache:
+  timeout: 604800  # 1 week
   apt: true
+  directories:
+    - ~/protobuf3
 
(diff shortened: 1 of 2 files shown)


shelhamer:
Ok, fingers-crossed this cooperates. Thanks for all the CI improvements.

lukeyeager:
Looks good. That build is slow, this actually cuts the build+test time in half!

Issue #4252, opened 2016-06-01
Maybe a bug in calculating the shape of output blob after pooling or convolution with stride


kgl-prml:
I found that the way of calculating the shape of feature map after pooling or convolution with stride is strange.

In `pooling_layer.cpp` line `90` :
```
  pooled_height_ = static_cast<int>(ceil(static_cast<float>(
      height_ + 2 * pad_h_ - kernel_h_) / stride_h_)) + 1;
  pooled_width_ = static_cast<int>(ceil(static_cast<float>(
      width_ + 2 * pad_w_ - kernel_w_) / stride_w_)) + 1;
```

I think the correct one should be:
```
  pooled_height_ = ceil((height_ + 2 * pad_h_ - kernel_h_ + 1) / stride_h_);
  pooled_width_ = ceil((width_ + 2 * pad_w_ - kernel_w_ +1) / stride_w_);
```

We can prove that the above two calculation methods are not equivalence. And we can list quite a few examples against the original calculation way defined in `pooling_layer.cpp` now.

For example (for simplicity, suppose no padding) : 
1) `heigh = width = 32, kernel_h = kernel_w = 1, stride_h = stride_w = 2`
The results for original one: `pooled_height_ = 17, pooled_width_ = 17`
But the results should be `pooled_height_ = 16, pooled_width_ = 16`

2) `heigh = width = 6, kernel_h = kernel_w = 2, stride_h = stride_w = 3`
The results for original one: `pooled_height_ = 3, pooled_width_ = 3`
But the results should be `pooled_height_ = 2, pooled_width_ = 2`
`... ...`

Actually, we can prove that when the remainder of `(height_ + 2 * pad_h_ - kernel_h_ + 1) / stride_h_` is zero or bigger than `1`, the results are not correct.

In `conv_layer.cpp`, the shape is calculated by a different way (using `floor` instead of `ceil`), but still not correct.

I did not read the `forward` and `backward` part, so I did not know if the following computations are correct. But if the shape of the output `Blob` is not correct, the results must be wrong.

I don't know if I have any misunderstanding. If so, please tell me. :)

shelhamer:
See #1318 for background and context and in particular https://github.com/BVLC/caffe/issues/1318#issuecomment-59596155. We'll likely switch to both pooling and convolution doing the same index and shape calculations, but we need to make sure we preserve backward compatibility, which is why it has not yet been resolved.

[...]

PR #4257, opened 2016-06-02
fix the pooling output reshape bug in pooling_layer.cpp

kgl-prml:
#4252


antran89:
Yeah, it seems to be weird to use `ceil` function here.

seanbell:
As mentioned in https://github.com/BVLC/caffe/issues/4252#issuecomment-224338172, the most important thing is to maintain backwards compatibility, which this PR does not.

One way to handle this could be to add a configuration parameter in the layer params (prototxt), where the default is the old behavior.

Issue #2216, opened 2015-03-27
-Wno-sign-compare revisited 


moskewcz:
following up on PR#488, based on running:
git log -G'sign-compare' -- Makefile

it looks like the warning settings for caffe were, in chrono order:

1) -Wall
2) -Wall  -Werror -Wno-sign-compare
3) -Wall -Wno-sign-compare

i just started noticing sign-compare warning when including caffe headers in another project using 'straight' -Wall. there aren't too many (11 by my count) sign-compare warnings in caffe *headers* (there are 101 more in .cpp files). if we can agree on a pattern to fix/supress them, what are the chances that a PR to that effect might get merged? my understanding that, in these cases the compiler is casting the int32_t to uint32_t and then doing uint32_t comparison, so the 'normal' fix is to explicity cast the signed value to unsigned.

moskewcz@orihime:~/git_work/caffe$ less make.log  | grep warning | grep .hpp | sort  -u
./include/caffe/blob.hpp:172:7: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/blob.hpp:57:18: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/layer.hpp:169:67: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/layer.hpp:176:5: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/layer.hpp:275:43: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/layer.hpp:282:5: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/layer.hpp:388:25: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/layer.hpp:412:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/layer.hpp:423:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/layer.hpp:462:16: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
./include/caffe/loss_layers.hpp:76:18: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
moskewcz@orihime:~/git_work/caffe$

moskewcz:
[...]

PR #4265, opened 2016-06-03
fix sign-compare warning

jklontz:
Self explanatory one-liner.


ajtulloch:
LGTM.

longjon:
See, e.g., #2216. In this case we can all see that the code is correct, but we should either not make this change here or make it everywhere, and there are arguments against the latter.

PR #4278, opened 2016-06-07
Fix for MemoryDataLayer memory leak (PyCaffe flow)

drnikolaev:
See https://github.com/NVIDIA/caffe/issues/154 to reproduce this leak. Problem: MemoryDataLayer<Dtype>::Reset makes a shallow copy. As a result, Python is unable to collect its garbage. This fix makes deep copy but it might defeat the MemoryDataLayer purpose.

### DIFF
--- a/include/caffe/layers/memory_data_layer.hpp
+++ b/include/caffe/layers/memory_data_layer.hpp
@@ -49,8 +49,8 @@ class MemoryDataLayer : public BaseDataLayer<Dtype> {
       const vector<Blob<Dtype>*>& top);
 
   int batch_size_, channels_, height_, width_, size_;
-  Dtype* data_;
-  Dtype* labels_;
+  Blob<Dtype> data_;
+  Blob<Dtype> labels_;
   int n_;
   size_t pos_;
   Blob<Dtype> added_data_;
--- a/src/caffe/layers/memory_data_layer.cpp
+++ b/src/caffe/layers/memory_data_layer.cpp
@@ -24,8 +24,6 @@ void MemoryDataLayer<Dtype>::DataLayerSetUp(const vector<Blob<Dtype>*>& bottom,
   top[1]->Reshape(label_shape);
   added_data_.Reshape(batch_size_, channels_, height_, width_);
   added_label_.Reshape(label_shape);
-  data_ = NULL;
-  labels_ = NULL;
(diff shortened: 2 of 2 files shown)


lukeyeager:
With this change, how is a `MemoryData` layer different from an `Input` layer? Should we just disable the `MemoryData` layer from pycaffe? Does any other workflow use it?

naibaf7:
@lukeyeager @drnikolaev 
The way we use the MemoryDataLayer in https://github.com/naibaf7/PyGreentea is that we have a class with a fixed amount of numpy arrays as buffers in there.

Through the custodian and ward method, the lifetime of the buffers is then tied to the lifetime of the network. This also avoids any issues with memory leaks.

I would prefer to document this rather than disabling the MemoryDataLayer in python or mutilate it by doing a deep copy.

drnikolaev:
@naibaf7 Thank you, good example. Closing this one.

enedene:
@drnikolaev Sorry for not responding earlier, I was not in a position to test your solution.

Unfortunately I don't think that providing a link to third party repository is a good way around this serious problem.  This is a basic usage scenario where user would like to have a control over filling the network during training and it leaks, it shouldn't.

@naibaf7  Btw `PyGreentea.py` seems to have bugs missing quotes around `float32`. Also `net.set_layer_input_arrays` in NetInputWrapper doesn't seem to exist in caffe, at least not in nvidia branch where I originally reported the problem, instead method `set_input_arrays` is used.

enedene:
@naibaf7 
```
import numpy as np
np.ascontiguousarray([0]).astype(float32)
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-2-6b2dfe4b4b68> in <module>()
----> 1 np.ascontiguousarray([0]).astype(float32)

NameError: name 'float32' is not defined

```
as in `NetInputWrapper`  class.

On the other hand, both:
```
np.ascontiguousarray([0]).astype("float32")
Out[3]: array([ 0.], dtype=float32)
```
and
```
np.ascontiguousarray([0]).astype(np.float32)
Out[4]: array([ 0.], dtype=float32)
```
work fine.

naibaf7:
@enedene 
Ah thanks, will fix that. Interestingly it does not complain on my version of numpy/python.

Issue #4304, opened 2016-06-13
OpenCL spatial convolution bug 


ngaloppo:
Hardware: SKL i5-6600U - Intel GPU
OS: Ubuntu 16.04

Testing AlexNet (batch 128), as defined by soumith's convnet benchmarks. Also tried 32

```
ViennaCL: Getting queue for device Intel(R) HD Graphics in context 0xbaa860
ViennaCL: Current queue id 0
ViennaCL: Setting handle kernel argument 0x10f22d0 at pos 0 for kernel U5_5_64_1_1_1_1_31_31_128_1_192_1_4_3_1_2
ViennaCL: Setting int precision kernel argument 0 at pos 1 for kernel U5_5_64_1_1_1_1_31_31_128_1_192_1_4_3_1_2
ViennaCL: Setting handle kernel argument 0x104e2d0 at pos 2 for kernel U5_5_64_1_1_1_1_31_31_128_1_192_1_4_3_1_2
ViennaCL: Setting int precision kernel argument 0 at pos 3 for kernel U5_5_64_1_1_1_1_31_31_128_1_192_1_4_3_1_2
ViennaCL: Setting handle kernel argument 0x2b48d80 at pos 4 for kernel U5_5_64_1_1_1_1_31_31_128_1_192_1_4_3_1_2
ViennaCL: Setting int precision kernel argument 0 at pos 5 for kernel U5_5_64_1_1_1_1_31_31_128_1_192_1_4_3_1_2
ViennaCL: Setting handle kernel argument 0x105ee20 at pos 6 for kernel U5_5_64_1_1_1_1_31_31_128_1_192_1_4_3_1_2
ViennaCL: Setting int precision kernel argument 17775936 at pos 7 for kernel U5_5_64_1_1_1_1_31_31_128_1_192_1_4_3_1_2
ViennaCL: Getting queue for device Intel(R) HD Graphics in context 0xbaa860
ViennaCL: Current queue id 0
ViennaCL: Getting current_context with id 0
ViennaCL: Getting queue for device Intel(R) HD Graphics in context 0xbaa860
ViennaCL: Current queue id 0
ViennaCL: Getting queue for device Intel(R) HD Graphics in context 0xbaa860
ViennaCL: Current queue id 0
ViennaCL: Getting queue for device Intel(R) HD Graphics in context 0xbaa860
ViennaCL: Current queue id 0
*** Aborted at 1465841354 (unix time) try "date -d @1465841354" if you are using GNU date ***
[...]

PR #4311, opened 2016-06-14
Use gpu_gemm to compute the verification data of spatial convolution

gongzg:
When enable the MKL the cpu_gemm code could not work correctly.
Don't know the root cause, just switch to use gpu_gemm to compute
the verification data to work around this problem.

Also clean up some verification code logic.

This patch is for issue #4304 

### DIFF
--- a/src/caffe/layers/conv_layer_spatial.cu
+++ b/src/caffe/layers/conv_layer_spatial.cu
@@ -971,8 +971,9 @@ void ConvolutionLayerSpatial<float>::setup_convolution(
       kernelQueue[x]->executionTime = timed_convolve(bottom, top, bottom_index_,
                                                      num_, kernelQueue[x]);
     } else {
+      // skip those kernels without a good local size.
       kernelQueue[x]->verified = false;
-      kernelQueue[x]->tested = false;
+      kernelQueue[x]->tested = true;
     }
 
   for (int_tp x = 0; x < kernelQueue.size(); x++)
@@ -993,12 +994,14 @@ void ConvolutionLayerSpatial<float>::setup_convolution(
           fastestTime = kernelQueue[x]->executionTime;
         }
       }
+      if (fastestKernel < 0) break;
       // Test fastest kernel
       bool verified = verify_result(bottom, top, bottom_index_, num_,
                                     verify_blob, kernelQueue[fastestKernel]);
       if (verified == true) {
         kernelQueue[fastestKernel]->verified = true;
         kernel_index_ = fastestKernel;
+        verification = true;
         break;
       } else {
         kernelQueue[fastestKernel]->tested = true;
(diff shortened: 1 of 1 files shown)


naibaf7:
@gongzg 
Sure that MKL (as CPU GEMM for cpu_gemm) is the issue here? It also seems to happen when using openBLAS.
I'll review later today.

gongzg:
@naibaf7 I tested with MKL/openBLAS/Atlas, and only MKL has issue with cpu_gemm. Both openBLAS and Atlas just work fine.

ngaloppo:
@gongzg I'm worried that the workaround is just hiding the problem. Perhaps we are triggering something in the GPU queues by running `gpu_gemm`, which flushes the pipeline and not showing some different, bigger problem.

naibaf7:
@ngaloppo @gongzg 
I agree. While it does not fail anymore now, and I accept this as a temporary fix, it is definitely not supposed to happen either way. The old version still fails for me when using OpenBLAS, contrary to what @gongzg observed.
Which makes me think pointer aliasing or/and an illegal memory offset gets accessed at some point.

gongzg:
@ngaloppo @naibaf7 , Agreed, I will investigate the real root cause tomorrow.

gongzg:
@ngaloppo @naibaf7 I spent some time to investigate this issue today. And can't find any buffer overflow or illegal memory issue.  And I found a potential MKL usage issue. If I applied the following patch, then using MKL will also pass the verification. Actually, the original segfault is triggered by failing on all cases and failed to get a valid config, thus get a illegal config kernel index, -1, then cause a segfault in verify_result. That segfault is not an unknown illegal pointer access.  I never met this issue with OpenBLAS or ATLAS. Fabian, I am guessing that the segfault you met may be caused by another platform issue which lead to no valid configuration thus trigger the known illegal config kernel index(-1) accessing. So for me, there seems no hiding big issue for this PR.

diff --git a/cmake/Modules/FindMKL.cmake b/cmake/Modules/FindMKL.cmake
index 5ab93b2..1e6d177 100644
--- a/cmake/Modules/FindMKL.cmake
+++ b/cmake/Modules/FindMKL.cmake
@@ -14,7 +14,7 @@


 # ---[ Options
-caffe_option(MKL_USE_SINGLE_DYNAMIC_LIBRARY "Use single dynamic library interface" ON)
+caffe_option(MKL_USE_SINGLE_DYNAMIC_LIBRARY "Use single dynamic library interface" OFF)^M
 caffe_option(MKL_USE_STATIC_LIBS "Use static libraries" OFF IF NOT MKL_USE_SINGLE_DYNAMIC_LIBRARY)
 caffe_option(MKL_MULTI_THREADED  "Use multi-threading"   ON IF NOT MKL_USE_SINGLE_DYNAMIC_LIBRARY)

naibaf7:
@gongzg 
Yes could be, since I ran it on a GT650M and on an Ivy Bridge chip...
I've heard my Skylake platform is on the way though, so I will be able to properly test & help you on optimizations around mid-July :)

gongzg:
@naibaf7 That's great. That will be much faster than the old IVB machine :)

naibaf7:
@gongzg 
I highly doubt it CPU wise, it's an i7-3632QM. But to test the new GPU architecture it will definitely be good :)

Issue #263, opened 2014-03-27
python can't import _caffe module


ghost:
I suspect this is some stupid set-up problem that I make since no one else seems to have trouble running it..

So I was trying to run the pre-trained network by following the instruction, when I tried to run the following command: 

python ../python/caffe/detection/detector.py --crop_mode=selective_search --pretrained_model=../examples/imagenet/caffe_reference_imagenet_model --model_def=../examples/imagenet/imagenet_deploy.prototxt _temp/cat.txt _temp/cat.h5

I got an error saying:

```
Traceback (most recent call last):
  File "detector.py", line 29, in <module>
    import caffe
  File "/python/caffe/__init__.py", line 1, in <module>
    from .pycaffe import Net
  File "/caffe/pycaffe.py", line 6, in <module>
    from ._caffe import CaffeNet
ImportError: No module named _caffe
```

It seems like python is not recognizing the _caffe.cpp file...Would anyone know how to solve this problem?

Thanks tons!

wguo68:
I am get the similiar error, but I don't know how to set the PATHONPATH .  How to do ?

abhisheksimhadri:
I am also facing a similar problem. I tried doing all of this. I am getting the following error:

ImportError: dlopen(caffe/_caffe.so, 2): Library not loaded: @rpath/libcaffe.so
  Referenced from: /Users/abhisheksimhadri/Documents/Github/caffe/distribute/python/python/caffe/_caffe.so
  Reason: image not found


make all
make test
make runtest
make pycaffe
make distribute

All work fine ... 

What am I doing wrong?

abhisheksimhadri:
And if I give the caffe module path as caffe/python instead of the one in distribute/python
Then the application crashes and I get the following error:


Process:               Python [6445]
Path:                  /usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python
Identifier:            Python
Version:               2.7.9 (2.7.9)
Code Type:             X86-64 (Native)
Parent Process:        bash [6438]
Responsible:           Terminal [320]
User ID:               501

Date/Time:             2015-04-25 20:21:57.900 +0530
OS Version:            Mac OS X 10.10.2 (14C109)
Report Version:        11
Anonymous UUID:        8D82BD11-4581-1BF6-F12B-DCA9FEDC2491


Time Awake Since Boot: 3300 seconds

[...]

PR #4314, opened 2016-06-14
Fix pyCaffe installation on Cmake for MacOS

cepiross:
Since default CMAKE_SHARED_LIBRARY_SUFFIX is `.dylib` at MacOS, pycaffe does not work. This PR will create `_caffe.so` so as to be correctly referred after cmake installation.


seansaito:
This workaround resolved my error when trying to import pycaffe. Simply running `make pycaffe` does not work for MacOS, as suggested in other solutions such as #263

cepiross:
Close this PR since it is too old to follow.

PR #4319, opened 2016-06-15
Bug fix for Concat layer in computing offsets for Crop

tiferet:
The coord_map function did not know how to treat Concatenation layers, and threw an UndefinedMapException when they were encountered without any warning to the user. I added Concat to the list of PASS_THROUGH_LAYERS, and added a warning in case other newer layers are encountered in the future. I am not 100% sure Concat should just be a PASS_THROUGH_LAYER, and does not require any special treatment, but from what I understood of how this code works I think this is correct, right?

### DIFF
--- a/python/caffe/coord_map.py
+++ b/python/caffe/coord_map.py
@@ -12,7 +12,7 @@ from caffe import layers as L
 PASS_THROUGH_LAYERS = ['AbsVal', 'BatchNorm', 'Bias', 'BNLL', 'Dropout',
                        'Eltwise', 'ELU', 'Log', 'LRN', 'Exp', 'MVN', 'Power',
                        'ReLU', 'PReLU', 'Scale', 'Sigmoid', 'Split', 'TanH',
-                       'Threshold']
+                       'Threshold', 'Concat']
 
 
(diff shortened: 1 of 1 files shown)


python/caffe/coord_map.py
```diff
@@ -76,6 +76,7 @@ def coord_map(fn):
         axis -= 1  # -1 for last non-coordinate dim.
         return axis, 1, - offset
     else:
+        print("Error: Encountered an unfamiliar layer: " + fn.type_name)
```
seanbell: Why not instead put this string in the exception as a message?  It's unconventional to log an error to stdout and then raise a separate exception.

seanbell:
Concat is only a pass-through layer if you concat along the channel or depth dimension.  If you concat along a spatial dimension, then no, concat is not a passthrough layer, from the perspective of `coord_map`.

tiferet:
@seanbell That's a good point. To be honest I haven't delved deeply into this code. Do you have a suggestion for how to handle the general case? I'm not sure it's possible to know what the offset should be for concatenation in a spatial dimension, without knowing what is being concatenated and why.

PR #4337, opened 2016-06-21
Fixed data/label dimensions when exporting batch containing one "element"

ShaggO:
Matlab drops singleton dimensions when ndims > 2, and thus exporting a chunk/batch of only one element potentially fails (it only succeeds if the remaining dimensions by coincidence match). I've added optional arguments to the store2hdf5 routine to specify the number of actual data and label dimensions.


ShaggO:
The dropping of singleton dimensions of course only applies to trailing dimensions.

PR #4342, opened 2016-06-21
Improve python interface for the solver

nitnelave:
With these additions, it becomes possible to use the solver to handle the learning rate while setting the inputs/diffs in python.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -73,6 +73,7 @@ class Solver {
     return test_nets_;
   }
   int iter() { return iter_; }
+  void set_iter(int iter) { iter_ = iter; }
 
   // Invoked at specific points during an iteration
   class Callback {
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -95,9 +95,10 @@ class Solver {
    */
   virtual inline const char* type() const { return ""; }
 
- protected:
   // Make and apply the update value for the current iteration.
   virtual void ApplyUpdate() = 0;
+
+ protected:
   string SnapshotFilename(const string extension);
   string SnapshotToBinaryProto();
   string SnapshotToHDF5();
--- a/python/caffe/test/test_solver.py
+++ b/python/caffe/test/test_solver.py
@@ -38,6 +38,11 @@ class TestSolver(unittest.TestCase):
         self.solver.solve()
         self.assertEqual(self.solver.iter, 100)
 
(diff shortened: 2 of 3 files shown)


python/caffe/test/test_solver.py
```diff
@@ -38,6 +38,25 @@ def test_solve(self):
...
+        data[...] = 0
+        # Set the diffs to 1
+        net.layers[1].blobs[0].diff[...] = 1
+        # Apply the update, the initial learning rate should be 0.01
+        self.solver.apply_update()
+        # Check that the new weights are -0.01, with a precision of 1e-7
+        self.assertTrue((data - -0.01 * np.ones(data.shape)).max() < 1e-7)
+
+
+
```
longjon: Why so much space?

python/caffe/test/test_solver.py
```diff
@@ -38,6 +38,25 @@ def test_solve(self):
...
+        self.assertEqual(self.solver.iter, 0)
+        self.solver.set_iter(50)
+        self.assertEqual(self.solver.iter, 50)
+
+    def test_apply_update(self):
+        net = self.solver.net
+        data = net.layers[1].blobs[0].data[...]
+        # Reset the weights of that layer to 0
+        data[...] = 0
+        # Set the diffs to 1
```
longjon: No need to write comments that don't say anything more than the code.

longjon:
Seems like a bit of an odd case, but I guess it makes sense; one might also want to change the iter for debugging or other purposes.

nitnelave:
@longjon updated as required

include/caffe/solver.hpp
```diff
@@ -73,6 +73,7 @@ class Solver {
     return test_nets_;
   }
   int iter() { return iter_; }
+  void SetIter(int iter) { iter_ = iter; }
```
longjon: Ah, I noticed this before but neglected to comment (sorry!): this should be `set_iter`; trivial functions use underscore_case (like the getter above).

nitnelave:
@longjon fixed function name.

Noiredd:
This has also been attempted in #6238, slightly more clever with the `iter++` IMO, but has no tests (for now).

shelhamer:
Thanks for this @nitnelave, but a fuller interface has been exposed with the merge of #6238. Note that it includes c5b5b55845c20930026f1f563136247c41f58075 for the update test, with your authorship of course.

PR #4343, opened 2016-06-21
improve top_names and bottom_names in pycaffe

nitnelave:
By rewriting these properties to proper OrderedDict, a lot of functionalities like iterations are added to the field. Plus, it gets computed only once, instead of partially at every element access.

### DIFF
--- a/python/caffe/pycaffe.py
+++ b/python/caffe/pycaffe.py
@@ -292,21 +292,31 @@ def _Net_batch(self, blobs):
                                                  padding])
         yield padded_batch
 
-
-class _Net_IdNameWrapper:
+def _Net_get_id_name(func, field):
     """
-    A simple wrapper that allows the ids propery to be accessed as a dict
-    indexed by names. Used for top and bottom names
-    """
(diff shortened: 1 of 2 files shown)


python/caffe/pycaffe.py
```diff
@@ -292,21 +292,30 @@ def _Net_batch(self, blobs):
...
+    Parameters
+    ----------
+    func: function id -> [id]
+    field: implementation field name (cache)
+
+    Returns
+    ------
+    A one-parameter function that can be set as a property.
+    """
+    def get_id_name(self):
```
longjon: Why not `@property` this and return a property directly? (That'll be more uniform with the rest of this file.)

longjon:
Thanks, I asked for this here: https://github.com/BVLC/caffe/pull/2865#issuecomment-169873421 :)

Looks good, just one comment above.

nitnelave:
Thanks for the catch, I forgot that I could decorate inner functions too.

longjon:
Code looks good, history looks good, passing test included: merging, thanks!

PR #4344, opened 2016-06-21
Expand the net.input list to all input layers

nitnelave:
Currently, the input list of a network is only used (AFAIK) to provide a nice interface in python/caffe. However, it only includes the layers of type "Input", not all the input layers.

I understand that the idea was that the list is only the layers that need to be set by hand, but when manipulating a network by hand, having those layers in the input list is usually easier.

For instance, the `forward` method in python can only set the input data if the layer is in the input list.

Moreover, it mirrors more closely the behaviour of the output list.

### DIFF
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -104,7 +104,7 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
     for (int top_id = 0; top_id < num_top; ++top_id) {
       AppendTop(param, layer_id, top_id, &available_blobs, &blob_name_to_idx);
       // Collect Input layer tops as Net inputs.
-      if (layer_param.type() == "Input") {
+      if (num_top && !layer_param.bottom_size()) {
         const int blob_id = blobs_.size() - 1;
         net_input_blob_indices_.push_back(blob_id);
         net_input_blobs_.push_back(blobs_[blob_id].get());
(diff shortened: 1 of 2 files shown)


src/caffe/net.cpp
```diff
@@ -104,7 +104,8 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
     for (int top_id = 0; top_id < num_top; ++top_id) {
       AppendTop(param, layer_id, top_id, &available_blobs, &blob_name_to_idx);
       // Collect Input layer tops as Net inputs.
-      if (layer_param.type() == "Input") {
+      if (layer_param.type() == "Input" ||
+          layer_param.type().find("Data") != std::string::npos) {
```
seanbell: This check seems fragile.  Wouldn't a more general definition of input be any layer that has a top but no bottom?

nitnelave:
Updated the input check to be more robust: it checks that the layer has outputs but no inputs.

longjon:
Is there a particular reason why you want to do this? The only input layers should be layers of type "Input"... otherwise they are layers that compute something or read data from somewhere. One can have, e.g., random numbers which have no bottoms but are not inputs.

nitnelave:
Well, I came across quite a few uses of network analysis (e.g. feature visualization) that are scripts in python that just take the "normal" prototxt of the network and set the input by hand (because they are optimizing the input to maximize the activation of a certain neuron, for instance). The code of the input layers is not called here, and the solver cannot be used, of course.

shelhamer:
I don't agree with this patch. Input layers are exactly that: bare inputs w/o logic that merely hold data. A solver can have a training net with `Input` layers that can be set manually between calls to `Step()`.

Closing, as I don't see what problem this solves. Thanks for thinking of potential interface improvements all the same.

PR #4346, opened 2016-06-21
add blob option to set_input_arrays

nitnelave:
This modification allows to set the input arrays even when the memory layer is not the first one (other first input layers, or two MemoryDataLayers).


tiagofrepereira2012:
This is quite similar to this PR (https://github.com/BVLC/caffe/pull/4194)
Would be really nice to have this feature

nitnelave:
@tiagofrepereira2012 Your PR has a nicer c++ interface, maybe? I don't know much about the c++ interface, I've been mostly using the python interface, so I don't know how easy it is to get a `bp::object` compared to an index, but I feel that it shouldn't be this function's responsibility to handle that. The python interface that I wrote has the same logic as other functions in the code, so there would be no surprise there.

PR #4347, opened 2016-06-21
add layer_dict to the python interface

nitnelave:
Similar to the blobs map, have a layer map.

This turns 
`net.layers[list(net._layer_names).index("ip2")]`
 into 
`net.layer_dict["ip2"]`

### DIFF
--- a/python/caffe/pycaffe.py
+++ b/python/caffe/pycaffe.py
@@ -43,6 +43,16 @@ def _Net_blob_loss_weights(self):
                                                        self._blob_loss_weights))
     return self._blob_loss_weights_dict
 
+@property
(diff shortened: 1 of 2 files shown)


ajtulloch:
Sure, seems reasonable to me. Could you add some unit tests?

nitnelave:
I'd be glad to, but where? :) (Sorry, it's my first time contributing to caffe, and I can't seem to find the tests)

nitnelave:
Nevermind, I found them in python/caffe/test

nitnelave:
Here are the tests. To check the equality of the pointed layers, I only checked the types, but since the layers of the test network are all different types anyway, it seemed reasonable.

PR #4348, opened 2016-06-22
Add clear_param_diffs to the python net interface

nitnelave:
It is the only thing missing to make a clean backward step. It is possible, but slow, to do it by hand by iterating over the layers and setting the diffs manually.


shelhamer:
Thanks for extending the interface @nitnelave !

PR #4351, opened 2016-06-22
add set_random_seed to the python interface

nitnelave:
There is currently (AFAIK) no way to set the random seed, making it impossible to have reproducible results.


python/caffe/_caffe.cpp
```diff
@@ -51,6 +51,9 @@ const int NPY_DTYPE = NPY_FLOAT32;
 void set_mode_cpu() { Caffe::set_mode(Caffe::CPU); }
 void set_mode_gpu() { Caffe::set_mode(Caffe::GPU); }
 
+// Set the random seed
```
longjon: Could you remove this redundant comment?

longjon:
Looks good except as noted, thanks.

nitnelave:
Comment removed as noted.

shelhamer:
Thanks for extending the Python interface @nitnelave.

PR #4358, opened 2016-06-24
Performance tuning for Intel Spatial engine.

gongzg:
This is another major performance boost PR for Intel Gen platform. The forwarding speed of convnet-benchmark's alexnet case increase to 290 images/second from 190 images/second.


naibaf7:
@gongzg 
Could you benchmark/time these (benchmark1 to benchmark128) with the current implementation: https://github.com/BVLC/caffe/tree/opencl/models/bvlc_alexnet
They are adjusted a bit to correctly give backprop timings & test different batch sizes for scaling. It's one of my libDNN reference benchmarks.

Looking forward to test this myself with a Skylake chip :)

gongzg:
@naibaf7 sure, will do that next week.

PR #4359, opened 2016-06-24
Fix for EmbedLayerTest failure

npanpaliya:
This test had a random failure due to floating point comparison. So, instead of a macro that does exact match, used EXPECT_FLOAT_EQ that tolerates some precision while comparing two float or double values.

### DIFF
--- a/src/caffe/test/test_embed_layer.cpp
+++ b/src/caffe/test/test_embed_layer.cpp
@@ -124,7 +124,7 @@ TYPED_TEST(EmbedLayerTest, TestForwardWithBias) {
     top_offset[4] = 0;
     bias_offset[0] = 0;
     for (int j = 0; j < kNumOutput; ++j) {
-      EXPECT_EQ(layer->blobs()[0]->data_at(weight_offset) +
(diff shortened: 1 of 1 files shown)


hartb:
I don't think the bugzilla link will be accessible to folks outside IBM.  Here's an example of the failure with some debug output:

[ RUN      ] EmbedLayerTest/0.TestForwardWithBias
src/caffe/test/test_embed_layer.cpp:129: Failure
Value of: this->blob_top_->data_at(top_offset)
  Actual: -9.15841
Expected: layer->blobs()[0]->data_at(weight_offset) + layer->blobs()[1]->data_at(bias_offset)
Which is: -9.15841

npanpaliya:
@seanbell, can this PR be merged?

seanbell:
LGTM

@shelhamer

longjon:
Looks good, but can you fix the commit message so that it does not refer to an issue in a different, non-accessible system?

npanpaliya:
@longjon : Thanks, I've changed the commit message.

PR #4360, opened 2016-06-24
Added PyForwardFromTo and PyBackwardFromTo to Net, for releasing GIL …

alessandroferrari:
…in pycaffe. Added ScopedGILRelease for easy GIL release. Modified _caffe.cpp in pycaffe accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -596,19 +596,19 @@ $(TEST_ALL_BIN): $(TEST_MAIN_SRC) $(TEST_OBJS) $(GTEST_OBJ) \
 		| $(DYNAMIC_NAME) $(TEST_BIN_DIR)
 	@ echo CXX/LD -o $@ $<
 	$(Q)$(CXX) $(TEST_MAIN_SRC) $(TEST_OBJS) $(GTEST_OBJ) \
-		-o $@ $(LINKFLAGS) $(LDFLAGS) -l$(LIBRARY_NAME) -Wl,-rpath,$(ORIGIN)/../lib
+		-o $@ $(LINKFLAGS) -l$(LIBRARY_NAME)  $(LDFLAGS) -Wl,-rpath,$(ORIGIN)/../lib
 
 $(TEST_CU_BINS): $(TEST_BIN_DIR)/%.testbin: $(TEST_CU_BUILD_DIR)/%.o \
 	$(GTEST_OBJ) | $(DYNAMIC_NAME) $(TEST_BIN_DIR)
 	@ echo LD $<
 	$(Q)$(CXX) $(TEST_MAIN_SRC) $< $(GTEST_OBJ) \
-		-o $@ $(LINKFLAGS) $(LDFLAGS) -l$(LIBRARY_NAME) -Wl,-rpath,$(ORIGIN)/../lib
+		-o $@ $(LINKFLAGS) -l$(LIBRARY_NAME) $(LDFLAGS) -Wl,-rpath,$(ORIGIN)/../lib
 
 $(TEST_CXX_BINS): $(TEST_BIN_DIR)/%.testbin: $(TEST_CXX_BUILD_DIR)/%.o \
 	$(GTEST_OBJ) | $(DYNAMIC_NAME) $(TEST_BIN_DIR)
 	@ echo LD $<
 	$(Q)$(CXX) $(TEST_MAIN_SRC) $< $(GTEST_OBJ) \
-		-o $@ $(LINKFLAGS) $(LDFLAGS) -l$(LIBRARY_NAME) -Wl,-rpath,$(ORIGIN)/../lib
+		-o $@ $(LINKFLAGS) -l$(LIBRARY_NAME) $(LDFLAGS) -Wl,-rpath,$(ORIGIN)/../lib
 
 # Target for extension-less symlinks to tool binaries with extension '*.bin'.
 $(TOOL_BUILD_DIR)/%: $(TOOL_BUILD_DIR)/%.bin | $(TOOL_BUILD_DIR)
--- a/Makefile
+++ b/Makefile
@@ -397,7 +397,8 @@ LIBRARY_DIRS += $(BLAS_LIB)
 
 LIBRARY_DIRS += $(LIB_BUILD_DIR)
 
-#for linking to the libpython2.7 for releasing scope in pycaffe (however should be linked even if pycaffe disabled)
+#for linking to the libpython2.7 for releasing scope in pycaffe 
+#(however should be linked even if pycaffe disabled)
 LIBRARIES += python2.7
 
 # Automatic dependency generation (nvcc is handled separately)
--- a/Makefile
+++ b/Makefile
@@ -397,6 +397,9 @@ LIBRARY_DIRS += $(BLAS_LIB)
 
 LIBRARY_DIRS += $(LIB_BUILD_DIR)
 
+#for linking to the libpython2.7 for releasing scope in pycaffe (however should be linked even if pycaffe disabled)
+LIBRARIES += python2.7
+
 # Automatic dependency generation (nvcc is handled separately)
 CXXFLAGS += -MMD -MP
 
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -18,6 +18,8 @@
 
 #include "caffe/util/device_alternate.hpp"
 
+#include "Python.h"
+
 // Convert macro to string
 #define STRINGIFY(m) #m
 #define AS_STRING(m) STRINGIFY(m)
@@ -180,6 +182,19 @@ class Caffe {
   Caffe();
 
   DISABLE_COPY_AND_ASSIGN(Caffe);
+};
+
+class ScopedGILRelease{
+    public:
+        ScopedGILRelease(){
+            m_thread_state = PyEval_SaveThread();
+        };
+        ~ScopedGILRelease(){
+            PyEval_RestoreThread(m_thread_state);
+            m_thread_state = NULL;
+        };
+    private:
+        PyThreadState * m_thread_state;
 };
 
 }  // namespace caffe
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -185,16 +185,16 @@ class Caffe {
 };
 
 class ScopedGILRelease{
-    public:
-        ScopedGILRelease(){
-            m_thread_state = PyEval_SaveThread();
-        };
-        ~ScopedGILRelease(){
-            PyEval_RestoreThread(m_thread_state);
-            m_thread_state = NULL;
-        };
-    private:
-        PyThreadState * m_thread_state;
+ public:
+  ScopedGILRelease() {
+    m_thread_state = PyEval_SaveThread();
+  }
+  ~ScopedGILRelease() {
+    PyEval_RestoreThread(m_thread_state);
+    m_thread_state = NULL;
+  }
+ private:
+  PyThreadState * m_thread_state;
 };
 
 }  // namespace caffe
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -18,8 +18,6 @@
 
 #include "caffe/util/device_alternate.hpp"
 
-#include "Python.h"
-
 // Convert macro to string
 #define STRINGIFY(m) #m
 #define AS_STRING(m) STRINGIFY(m)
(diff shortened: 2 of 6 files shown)


seanbell:
The tests all failed due to style issues -- please run `make lint`.

alessandroferrari:
Multithreading in python is doomed by GIL. Many users (as I do) are often using caffe for performing nets prediction by pycaffe, and often they are doing so from multithreaded applications. Pycaffe interface is not releasing GIL, which means that if caffe is doing predictions, other python threads are stuck. Since predictions are time consuming, this can lead to terrible slow down of the application. By means of this pull-request, GIL is released for forward pass and backward pass of the network. Within the pass, it is required temporarily only for coping with python layers, that require to hold GIL. 
Thanks to this, within the same process, from the python interface is possible to perform predictions of different networks in parallel on multiple threads, still mantaining complete thread safety. Before everything was locked by GIL (false sharing).

include/caffe/net.hpp
```diff
@@ -52,6 +54,19 @@ class Net {
...
    */
   Dtype ForwardFromTo(int start, int end);
+  /**
+   * Basically, a ForwardFromTo, that release Python GIL, so that pycaffe
+   * can make multithreaded predictions. This is exposed to python-boost interface
+   * instead of ForwardFromTo
+   */
+  Dtype PyForwardFromTo(int start, int end) {
+    // Release GIL
+    m_thread_state = PyEval_SaveThread();
```
ajtulloch: This should be an RAII wrapper around Save/Restore thread for exception safety.

alessandroferrari: RAII wrapper: Thank you for the suggestion. I will take a look to fix it.

Not right place: I agree. However, where would you suggest to place it? (consider that you have to reacquire GIL in case you forward/backward a python layer within the net). Even though this solution insert "Pythonish" code within the C++ logic, not releasing GIL for the pycaffe is a painful bug for python interface. If there is a cleaner alternative I would love to discuss about it.

ajtulloch:
Why not just simply:

a) Release the GIL in PyCaffe
b) Reaquire the GIL in PythonLayer before running the function in that layer?

alessandroferrari:
It would be cleaner and it was my first thought. However, I have been stopped by two reasons:
- PyEval_SaveThread return the thread_state of the current thread, you need somehow to pass from PyCaffe to the PythonLayer this state, without relying on Net (otherwise it would be the same);
- I suspect that in python layer, python-boost does some magic that require GIL for passing the python objects arguments to the C++ code, since I get segfault if I acquire GIL in PythonLayer before calling the inner forward. GIL needs to be acquired before calling forward method for Python layers.
Thus, even if releasing GIL within the Net is not clean, since pythonish stuffs are inserted in the C++ logic, looked like the only viable solution. But if you have any ideas on how to circumvent those problems, I will be happy to give them a try.

crowsonkb:
I use the multiprocessing package instead of threading both to work around the GIL and to parallelize PyCaffe over multiple GPUs. My main Python interpreter spawns several other interpreters each of which has one instance of PyCaffe.

alessandroferrari:
Of course multiprocessing does not have synchronization problems......
Still, multiprocessing come with a interprocess communication cost. There are many applications in production environment where multithreading may be beneficial, such as model ensembling and performance critical apps. Numpy, OpenCv, scikit all release GIL....

Makefile
```diff
@@ -593,19 +597,19 @@ $(TEST_ALL_BIN): $(TEST_MAIN_SRC) $(TEST_OBJS) $(GTEST_OBJ) \
 		| $(DYNAMIC_NAME) $(TEST_BIN_DIR)
 	@ echo CXX/LD -o $@ $<
 	$(Q)$(CXX) $(TEST_MAIN_SRC) $(TEST_OBJS) $(GTEST_OBJ) \
-		-o $@ $(LINKFLAGS) $(LDFLAGS) -l$(LIBRARY_NAME) -Wl,-rpath,$(ORIGIN)/../lib
+		-o $@ $(LINKFLAGS) -l$(LIBRARY_NAME)  $(LDFLAGS) -Wl,-rpath,$(ORIGIN)/../lib
```
seanbell: Why is this change necessary?

python/caffe/test/test_python_layer.py
```diff
@@ -97,8 +96,8 @@ def phase_net_file():
         return f.name
 
 
-@unittest.skipIf('Python' not in caffe.layer_type_list(),
-    'Caffe built without Python layer support')
+#@unittest.skipIf('Python' not in caffe.layer_type_list(),
+#    'Caffe built without Python layer support')
```
seanbell: Won't this break support for those building caffe without Python?

src/caffe/net.cpp
```diff
@@ -536,9 +539,16 @@ Dtype Net<Dtype>::ForwardFromTo(int start, int end) {
...
+  bool is_python;
   for (int i = start; i <= end; ++i) {
     // LOG(ERROR) << "Forwarding " << layer_names_[i];
+    is_python = strcmp(layers_[i]->type(), "Python") == 0;
+    if (is_python) {
+        scoped_gil_release.reset();
+    }
     Dtype layer_loss = layers_[i]->Forward(bottom_vecs_[i], top_vecs_[i]);
+    if (is_python)
+        scoped_gil_release = make_shared<ScopedGILRelease>();
```
seanbell: These changes are only relevant if `WITH_PYTHON_LAYER` are set.  That variable should probably be checked, so that it still works if `WITH_PYTHON_LAYER` is set to `0`.

src/caffe/net.cpp
```diff
@@ -581,10 +591,16 @@ template <typename Dtype>
...
+  bool is_python;
   for (int i = start; i >= end; --i) {
     if (layer_need_backward_[i]) {
+      is_python = strcmp(layers_[i]->type(), "Python") == 0;
+      if (is_python)
+          scoped_gil_release.reset();
       layers_[i]->Backward(
           top_vecs_[i], bottom_need_backward_[i], bottom_vecs_[i]);
+      if (is_python)
+          scoped_gil_release = make_shared<ScopedGILRelease>();
```
seanbell: Same as above.

src/caffe/net.cpp
```diff
@@ -536,9 +539,16 @@ Dtype Net<Dtype>::ForwardFromTo(int start, int end) {
   CHECK_GE(start, 0);
   CHECK_LT(end, layers_.size());
   Dtype loss = 0;
+  bool is_python;
   for (int i = start; i <= end; ++i) {
     // LOG(ERROR) << "Forwarding " << layer_names_[i];
+    is_python = strcmp(layers_[i]->type(), "Python") == 0;
+    if (is_python) {
```
seanbell: Style note: you use `{` inconsistently -- you use it here but not below.

src/caffe/net.cpp
```diff
@@ -19,6 +20,8 @@
 
 #include "caffe/test/test_caffe_main.hpp"
 
+#include "Python.h"
```
seanbell: This is not needed if `WITH_PYTHON_LAYER` is not set.

include/caffe/net.hpp
```diff
@@ -71,6 +85,17 @@ class Net {
    */
   void Backward();
   void BackwardFromTo(int start, int end);
+  /**
+   * Basically, a BackwardFromTo, that release Python GIL, so that pycaffe
```
seanbell: No need for the word "Basically"

python/caffe/_caffe.cpp
```diff
@@ -268,8 +268,11 @@ BOOST_PYTHON_MODULE(_caffe) {
     bp::no_init)
     .def("__init__", bp::make_constructor(&Net_Init))
     .def("__init__", bp::make_constructor(&Net_Init_Load))
-    .def("_forward", &Net<Dtype>::ForwardFromTo)
-    .def("_backward", &Net<Dtype>::BackwardFromTo)
+    // Instead of ForwardFromTo and BackwardFromTo,
+    // expose PyForwardFromTo and PyBackwardFromTo,
+    // basically the same thing but with GIL release
```
seanbell: The word "basically" is vague.  A better explaining comment would be:
```
// Instead of ForwardFromTo and BackwardFromTo,
// expose PyForwardFromTo and PyBackwardFromTo,
// which adds a GIL release around these functions
```

seanbell:
Thanks @alessandroferrari for the great PR, and I agree that the GIL should be released during heavy operations.  This enables pre-fetching with multi-threading (not multi-processing).  Can you please squash this into a single commit?

include/caffe/net.hpp
```diff
@@ -308,6 +333,9 @@ class Net {
   /// The root net that actually holds the shared layers in data parallelism
   const Net* const root_net_;
   DISABLE_COPY_AND_ASSIGN(Net);
+
+  // For releasing/reacquiring GIL with pycaffe
+  shared_ptr<ScopedGILRelease> scoped_gil_release;
```
seanbell: This should be `scoped_gil_release_` to be consistent with other naming.

python/caffe/test/test_python_layer.py
```diff
@@ -5,7 +5,6 @@
 
 import caffe
 
-
```
seanbell: Unnecessary change

Makefile
```diff
@@ -593,19 +597,19 @@ $(TEST_ALL_BIN): $(TEST_MAIN_SRC) $(TEST_OBJS) $(GTEST_OBJ) \
...
 	$(Q)$(CXX) $(TEST_MAIN_SRC) $< $(GTEST_OBJ) \
-		-o $@ $(LINKFLAGS) $(LDFLAGS) -l$(LIBRARY_NAME) -Wl,-rpath,$(ORIGIN)/../lib
+		-o $@ $(LINKFLAGS) -l$(LIBRARY_NAME) $(LDFLAGS) -Wl,-rpath,$(ORIGIN)/../lib
 
 $(TEST_CXX_BINS): $(TEST_BIN_DIR)/%.testbin: $(TEST_CXX_BUILD_DIR)/%.o \
 	$(GTEST_OBJ) | $(DYNAMIC_NAME) $(TEST_BIN_DIR)
 	@ echo LD $<
 	$(Q)$(CXX) $(TEST_MAIN_SRC) $< $(GTEST_OBJ) \
-		-o $@ $(LINKFLAGS) $(LDFLAGS) -l$(LIBRARY_NAME) -Wl,-rpath,$(ORIGIN)/../lib
+		-o $@ $(LINKFLAGS) -l$(LIBRARY_NAME) $(LDFLAGS) -Wl,-rpath,$(ORIGIN)/../lib
```
seanbell: Is this edit necessary?

include/caffe/common.hpp
```diff
@@ -181,7 +185,19 @@ class Caffe {
 
   DISABLE_COPY_AND_ASSIGN(Caffe);
 };
-
```
seanbell: Unnecessary edit

Makefile
alessandroferrari: It fixed an undefined reference to lpython2.7, that would be necessary even if not building with python this way. Thus, without it would not work.

cypof:
Each net can run in it's own fork now, there is an example in /python/train.py. #4563

willyd:
> Thanks @alessandroferrari for the great PR, and I agree that the GIL should be released during forward/backward. This enables pre-fetching with multi-threading (not multi-processing).

I think that even with new multiprocessing parallel training releasing the GIL is still relevant for the above use case. I can a get better throughput of data using the threading module and releasing the GIL in `Net::ForwardFromTo` and `Net::BackwardFromTo` that by using the multiprocessing module for single GPU training on windows. Of course we have to reacquire the GIL before calling any python layer functions like `setup`, `reshape`, etc.

willyd:
> PyEval_SaveThread return the thread_state of the current thread, you need somehow to pass from PyCaffe to the PythonLayer this state, without relying on Net (otherwise it would be the same);
> I suspect that in python layer, python-boost does some magic that require GIL for passing the python objects arguments to the C++ code, since I get segfault if I acquire GIL in PythonLayer before calling the inner forward. GIL needs to be acquired before calling forward method for Python layers.

> Thus, even if releasing GIL within the Net is not clean, since pythonish stuffs are inserted in the C++ logic, looked like the only viable solution. But if you have any ideas on how to circumvent those problems, I will be happy to give them a try.

@alessandroferrari  I am not sure I completely understand what you are saying here. Have you tried something like: 

Using:
```
class ReleasePyGIL {
 public:
  ReleasePyGIL() { state = PyEval_SaveThread(); }
  ~ReleasePyGIL() { PyEval_RestoreThread(state); }
 private:
  PyThreadState* state;
```
inside `ForwardFromTo` and `BackwardFromTo` and
```
class AcquirePyGIL {
 public:
  AcquirePyGIL() { state = PyGILState_Ensure(); }
  ~AcquirePyGIL() { PyGILState_Release(state); }
 private:
  PyGILState_STATE state;
};
```
inside `PythonLayer::LayerSetup`, `PythonLayer::Reshape`, etc.

This works flawlessly for me on Windows.

PR #4361, opened 2016-06-24
Examples correction

npanpaliya:
This pull request contains changes so that examples scripts can also run in distribution mode.  In build environment, there will not be any impact. However, if the scripts are to be run in distribution mode, one would need to set variables CAFFE_BIN and TOOLS to the path where caffe and test executable resides.  
Currently, symbolic links of tools binaries and some more directories like models and data are not shipped into distribution. This pull request will need that as well. A different PR for this change is on the way.

### DIFF
--- a/examples/cifar10/create_cifar10.sh
+++ b/examples/cifar10/create_cifar10.sh
@@ -5,15 +5,26 @@ EXAMPLE=examples/cifar10
 DATA=data/cifar10
 DBTYPE=lmdb
 
+# Check if CAFFE_ROOT is set
+if [ -z ${CAFFE_ROOT+x} ]; 
+# if unset
+then
+  EXAMPLES_BIN=./build/$EXAMPLE
+  TOOLS_BIN=./build/tools
+else
+  EXAMPLES_BIN=$CAFFE_BIN
+  TOOLS_BIN=$CAFFE_BIN
+fi  
+
 echo "Creating $DBTYPE..."
 
 rm -rf $EXAMPLE/cifar10_train_$DBTYPE $EXAMPLE/cifar10_test_$DBTYPE
 
-./build/examples/cifar10/convert_cifar_data.bin $DATA $EXAMPLE $DBTYPE
+$EXAMPLES_BIN/convert_cifar_data.bin $DATA $EXAMPLE $DBTYPE
 
 echo "Computing image mean..."
 
-./build/tools/compute_image_mean -backend=$DBTYPE \
+$TOOLS_BIN/compute_image_mean -backend=$DBTYPE \
   $EXAMPLE/cifar10_train_$DBTYPE $EXAMPLE/mean.binaryproto
 
 echo "Done."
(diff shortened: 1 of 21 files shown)


examples/cifar10/create_cifar10.sh
```diff
@@ -5,15 +5,26 @@ EXAMPLE=examples/cifar10
 DATA=data/cifar10
 DBTYPE=lmdb
 
+# Check if CAFFE_BIN is set
+if [ -z ${CAFFE_BIN+x} ]; 
+# if unset
```
seanbell: the `if unset` comment doesn't need its own line, and `then` can be on the same line:

`if [[ -z "$CAFFE_BIN" ]]; then`

examples/cifar10/train_full.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env sh
 
+# Check if TOOLS is unset
+if [ -z ${TOOLS+x} ];
+then
 TOOLS=./build/tools
+fi
```
seanbell: This could be shortened to:

```
if [[ -z "$TOOLS" ]]; then
    TOOLS=./build/tools
fi
```

examples/cifar10/create_cifar10.sh
npanpaliya: Thanks. Done.

examples/cifar10/train_full.sh
npanpaliya: Done. Changed as below -
if [ -z "$TOOLS" ]; then
    TOOLS=./build/tools
fi
Just instead of [[]], I used [] as the script has /bin/sh as a first line. [[]] is valid for bash.

npanpaliya:
Submitted a pull request to ship data, models and scripts dir. https://github.com/BVLC/caffe/pull/4400

hartb:
Howdy! Another member of npanpalya's team here; nice to meet you.

The "TOOLS" name is pretty general for an environment variable; it could conflict with other user settings. That's not a problem for several of these scripts because they always set TOOLS locally within the script. But a few of the scripts (e.g. train_full.sh) are checking for "TOOLS" in the environment and using that. I think those would be better switched to the "always set TOOLS to CAFFE_BIN or default" model.

npanpaliya:
Hello @seanbell ,
Could you please let me know if there are any more changes needed for this PR to get accepted?

seanbell:
Please squash your commits into a single commit.  Also, it still looks like the `if` statements are unnecessarily verbose with empty `if` branches, and `then` on a separate line.  To negate a condition, use `!`.

npanpaliya:
@seanbell - I'd corrected all these quite a long ago but somehow, the changes didn't get pushed to the branch. I just pushed them today. Please see them. I'll anyway generate a new pull request with just one commit for all the changes, and close this one.

npanpaliya:
Created a new pull request for this. https://github.com/BVLC/caffe/pull/4562

PR #4368, opened 2016-06-27
fix the bug of some gpus are blocked during a training iteration

tungld:
We were using multiple GPUs to train AlexNet and/or GoogleNet on ImageNet dataset. cuDNN was used for acceleration. However, during the backward (sometimes forward), one/some of GPUs are blocked and waiting for receiving gradients from the paired GPU before continuing its work. This phenomenon often happens to parent GPUs in the GPU tree. It dramatically reduces the performance of multi-GPUs training.

blankissue-multigpus

We tested with different configurations such as (POWER8 + 2 K40 GPU(s)/no P2P + CUDA 7.0 + cuDNN 3.0)/(POWER8 + up to 8 K80 GPUs/(mix of P2P and non-P2P) + CUDA 8.0/7.5 + cuDNNv5.0)/(x86 + 2 K40 GPUs/P2P + CUDA 7.5 + cuDNNv4.0), and the problem often appears.

We found that the bug is caused by the convolution layer that uses cuDNN (src/caffe/layers/cudnn_conv_layer.cu). In the convolution layer, there are multiple threads created, and Caffe uses an empty kernel to do a synchronization at the end of forward/backward. 

``` C++
// Synchronize the work across groups, each of which went into its own
// stream, by launching an empty kernel into the default (null) stream.
// NOLINT_NEXT_LINE(whitespace/operators)
sync_conv_groups<<<1, 1>>>();
```

In the case of multiple GPUs, after Forward and Backward, the root solver collects gradients from GPUs, and Caffe also uses the default stream to do another synchronization (see the callback function `on_gradients_ready()` in src/caffe/parallel.cpp).

``` C++
CUDA_CHECK(cudaMemcpyAsync(dst, src, size_ * sizeof(Dtype),  //
     cudaMemcpyDeviceToDevice, cudaStreamDefault));
CUDA_CHECK(cudaStreamSynchronize(cudaStreamDefault));
```

However, kernel calls are asynchronous with host as described [here](http://on-demand.gputechconf.com/gtc-express/2011/presentations/StreamsAndConcurrencyWebinar.pdf),
Hence, the `cudaMemcpyAsync` causes an implicit synchronization as described in the section “Implicit Synchronization” in the CUDA C Programming Guide, which blocks a parent GPU from executing threads in the convolution layer and causes it to wait until gradients are exchanged.

To fix this bug, we explicitly do a synchronization after the empty kernels in the file `src/caffe/layers/cudnn_conv_layer.cu`, by inserting a call `cudaStreamSynchronize(cudaStreamDefault)`. By this patch, there is no blocking on GPUs any more, and we can reduce on average about 20% of running time per iteration.

blankissue-multigpus-fixed


flx42:
I believe there is something similar in #3919, right @drnikolaev?

drnikolaev:
@flx42 No, this is different and we need to discuss this.

thatguymike:
Interesting.  The problem with doing the stream synchronize is the cost in the single GPU case, especially for inference.  Basically the issue here appears to be effectively needing a flush to get the kernels out of the way.  We'll look at coming up with something with less impact to the forward only single GPU case.  Thanks for finding this @tungld with the very nicely written description.

thatguymike:
@tungld - can you see if you just do a CUDA_CHECK(cudaStreamSynchronize(cudaStreamDefault)); at the top of on_gradients_ready() if this fixes your performance issue?  The other option perhaps is to only synchronize backwards_gpu and not foward_gpu (leaving the empty kernel trick here).

tungld:
@thatguymike - just inserting a CUDA_CHECK(cudaStreamSynchronize(cudaStreamDefault)); at the top of on_gradients_ready() can fix the issue also. I actually went in this way, but finally decided to change the forward_gpu() and backward_gpu() to clarify that the issue is caused by the convolution layer.

> 
The other option perhaps is to only synchronize backwards_gpu and not foward_gpu (leaving the empty kernel trick here).

This doesn't help, because the issue sometimes occurs during a forward too.

shelhamer:
This should be solved by #4563.

PR #4374, opened 2016-06-27
Intial implementation of loss layer which produce Confusion Matrix

sperry500:


seanbell:
Was this PR intentional?  If so, can you please add a description?

PR #4396, opened 2016-06-30
net_spec: handle enum values as argument

nitnelave:
This patch allows to specify enum values as strings when constructing a new layer. It looks up in the protobuf file the possible values, and gives the corresponding value or provides an error with the list of possible values.

### DIFF
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -61,6 +61,17 @@ def assign_proto(proto, name, val):
     lists; e.g., `my_repeated_int_field=3` is converted to
     `my_repeated_int_field=[3]`."""
 
+    enum_type = proto.DESCRIPTOR.fields_by_name[name].enum_type
+    if enum_type is not None and isinstance(val, str):
+        try:
+            val = enum_type.values_by_name[val].index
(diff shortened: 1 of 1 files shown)


cdoersch:
What's wrong with `caffe.layers.Eltwise(bottom,top,operation=caffe.params.Eltwise.PROD);`?

nitnelave:
The string is closer to the prototxt, I find, so easier for the users.

I have to admit that I didn't find `caffe.params.Eltwise.PROD` (is it documented anywhere?).

I'm not too attached to this PR, I understand it can make the code a bit more brittle.

nitnelave:
I'm working on a YAML to prototxt converter that allows more compact descriptions of networks, and I didn't find a good way to specify generically enums from the YAML, so a string is a good value, I feel.

ajtulloch:
Yeah, I'm not a fan of this. If you want to add this logic, it should belong in your YAML -> protobuf converter. You can access enum values from Protobuf easily with the tools in https://developers.google.com/protocol-buffers/docs/reference/python-generated#enum, you might find a lot of value in that page for your convertor.

PR #4397, opened 2016-06-30
net_spec: handle deconvolution layers

nitnelave:
Deconvolution shares parameters with Convolution, which wasn't properly handled up to now.


longjon:
See #3954.

PR #4237, opened 2016-05-30
Cmake using gnuinstalldirs

CDLuminate:
CMake module `GNUInstallDirs` fixes some install path, e.g. when users install prefix is `/usr` then it transfers path `/usr/lib/libcaffe.so` into `/usr/lib/x86_64-linux-gnu/libcaffe.so`, while no change will be applied if install prefix is `/home/somebody/caffe/` or something else.

This would benifit distribution packaging. And the same patch is applied downstream in Debian packaging.
http://anonscm.debian.org/cgit/debian-science/packages/caffe.git/tree/debian/patches/cmake-using-gnuinstalldirs

PR #4400, opened 2016-07-01
Caffe dist changes

npanpaliya:
This pull request contains the changes to ship/distribute data and models directories which are required to run examples scripts. It also adds a change to ship tool binaries link and scripts directory to distribution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -678,6 +678,8 @@ $(DISTRIBUTE_DIR): all py | $(DISTRIBUTE_SUBDIRS)
 	# add data and model
 	cp -r data $(DISTRIBUTE_DIR)/
 	cp -r models $(DISTRIBUTE_DIR)/
+	# add scripts
+	cp -r scripts $(DISTRIBUTE_DIR)/
 	mkdir -p $(DISTRIBUTE_DIR)/include/caffe/proto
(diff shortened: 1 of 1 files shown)


longjon:
Actually I'm not sure why examples are part of `distribute` at all, but happy to hear some opinions.

lukeyeager:
I'd like to see both buildsystems have a single `make install` target that follows standard conventions. I don't see the point of a `make distribute` that provides tests/docs/examples.

CMake is pretty close already. See https://github.com/BVLC/caffe/pull/4237 for a slight improvement.

hartb:
Member of @npanpaliya's team here...  Agree with @lukeyeager that a standard make install target for both build systems would be best. We've been using make distribute but will take a look at the CMake method.

To answer @longjon, we like that the make distribute target includes the examples. We want to provide our users a pre-built package (currently a .deb) for convenient install. And we want them to have access to all the self-tests, examples, etc. even if they've never built the source on the target machine.

So for us the more inclusive make distribute is, the better. (And if it converges with a standard/CMake make install behavior even better still.)

npanpaliya:
Hello,
Wanted to know further thoughts on this PR.

Issue #4365, opened 2016-06-25
draw net doesn't understand net phases


cdoersch:
For some networks, especially Variational Autoencoders, there can be significant differences between the training-time network and the testing-time network.  Hence, I propose that we add a "phase" flag to caffe.draw.draw_net which, if specified by the user, makes the function ignore any layers that belong to the wrong phase.  This would make the drawn network much more understandable.  My guess is that this change would only require about a dozen lines of code.  

If the Caffe devs agree, I can write a PR.  I'd appreciate a quick response since I'd like to include this with my tutorial on variational autoencoders.

lukeyeager:
While you're at it, why not add support for phase, level and stages? https://github.com/BVLC/caffe/pull/3863 https://github.com/BVLC/caffe/pull/4174

cdoersch:
Sounds like this would make the PR more complicated, although I don't think I would actually need to rely on #3863 or #4174 to make it work.  draw_net worked directly with the protocol buffer that defines the net structure, not with a caffe net.  I'm not familiar with levels and stages.  Would the interface just take in two extra parameters--stage and level--and if they're defined, do a simple equality test for any layers that have either of those two parameters defined?  Or is there more complexity in determining whether a layer is included in the net?

lukeyeager:
Yeah it's not simple. See the functions I added in #4174 to mirror [StateMeetsRule](https://github.com/BVLC/caffe/blob/rc3/src/caffe/net.cpp#L316-L380).

Honestly, I wouldn't worry about it for your PR. The @BVLC guys don't seem particularly eager to merge #3863, so I can't imagine they'd care about other netstate options for your "draw net" PR.

shelhamer:
Fixed in #4408.

PR #4408, opened 2016-07-03
Add phase support for draw net

cdoersch:
As mentioned in #4365, Caffe's draw.py doesn't understand net phases, which makes it difficult to draw networks like VAEs where training and testing networks tend to have very different structure.  This small PR adds support for it.  I have tested it with the networks in my VAE tutorial (there appear to be no unit tests for draw net, and it's not clear how they would be implemented), and it works.  It should be ready to merge.

### DIFF
--- a/python/draw_net.py
+++ b/python/draw_net.py
@@ -28,6 +28,11 @@ def parse_args():
                               'http://www.graphviz.org/doc/info/'
                               'attrs.html#k:rankdir'),
                         default='LR')
+    parser.add_argument('--phase',
+                        help=('Which network phase to draw: can be TRAIN, '
+                              'TEST, or ALL.  If ALL, then all layers are drawn '
+                              'regardless of phase.'),
+                        default="ALL")
 
     args = parser.parse_args()
     return args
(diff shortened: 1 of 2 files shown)


python/draw_net.py
```diff
@@ -38,7 +43,15 @@ def main():
...
     text_format.Merge(open(args.input_net_proto_file).read(), net)
     print('Drawing net to %s' % args.output_image_file)
-    caffe.draw.draw_net_to_file(net, args.output_image_file, args.rankdir)
+    phase=None;
+    if args.phase == "TRAIN":
+        phase = caffe.TRAIN
+    elif args.phase == "TEST":
+        phase = caffe.TEST
+    elif args.phase != "ALL":
+        raise Exception("Unknown phase: " + args.phase)
```
lukeyeager: Probably want a `ValueError` instead of a bare `Exception` here.

python/draw_net.py
```diff
@@ -38,7 +43,15 @@ def main():
...
+    caffe.draw.draw_net_to_file(net, args.output_image_file, args.rankdir,
+                                args.phase)
```
lukeyeager: Should be `phase` instead of `args.phase`

lukeyeager:
This doesn't actually work. Try running it on `examples/mnist/lenet_train_test.prototxt`. `ALL` and `TRAIN` come out looking equivalent.

Plus, you're not handling the `exclude` case.

python/draw_net.py
cdoersch: Good catch.  Is this the reason that it's failing on `lenet_train_test.prototxt`?  I'll admit that most of my testing was done using the library rather than the command line tool.

cdoersch: Nope, it worked on my test case completely by chance.

cdoersch:
Thanks @lukeyeager for your review, I think I have fixed all of the issues you mentioned.  I didn't even know about exclude!

lukeyeager:
LGTM. Works on LeNet and also some slightly tricker networks I made for testing https://github.com/BVLC/caffe/pull/3863.

shelhamer:
Thanks Carl!

PR #4410, opened 2016-07-04
[Python] Create solvers/nets without proto files.

BlGene:
This PR allows the creation of solver and net objects without requiring the creation of 
temporary files. 

1. The net interface was changes to allow initialization with a string protobuffer in text form. 
2. As the solver interface requires a network to be specified during creation, a net is needed when initializing the solver. To this end a helper function was added that copies a net specification into a solver specification and initializes the solver ( and thus the net ).

I originally wanted to pass a caffe_pb2.SolverParameter/NetParameter object directly to the c++ interface but I didn't find a way to convert between these and the boost c++ serialized SolverParameter/NetParameter, so I chose strings for everything.

This function can be called in the following way:
```
solver_proto = self.create_solver_proto() #from caffe_pb2.SolverParameter
netspec_proto = self.create_netspec().to_proto() #from netspec
self.solver = caffe.get_solver_net( str(solver_proto), str(netspec_proto))
```

### DIFF
--- a/include/caffe/net.hpp
+++ b/include/caffe/net.hpp
@@ -24,7 +24,7 @@ template <typename Dtype>
 class Net {
  public:
   explicit Net(const NetParameter& param, const Net* root_net = NULL);
-  explicit Net(const string& param_file, Phase phase,
+  explicit Net(const string& param_str_or_file, Phase phase,
       const Net* root_net = NULL);
   virtual ~Net() {}
 
--- a/include/caffe/util/io.hpp
+++ b/include/caffe/util/io.hpp
@@ -55,6 +55,8 @@ inline bool ReadProtoFromTextFile(const string& filename, Message* proto) {
   return ReadProtoFromTextFile(filename.c_str(), proto);
 }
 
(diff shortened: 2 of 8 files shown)


ajtulloch:
I think the implementation looks reasonably correct, but what is the higher-level goal of this (as opposed to the 2 lines in Python to create a tmpfile and write the spec into that tmpfile)? This adds complexity (e.g. the implementation of get_solver, Net::Net, etc), and I don't see how this improves the user experience? Why can't this just be solved in a 3 line Python function? 

```lang=py
def s(proto):
  with tempfile.NamedTemporaryFile(delete=False) as f:
    f.write(str(proto))
  return f.name

caffe.Solver(s(solver_proto), s(netspec_proto)
```

as opposed to this diff which adds some confusing semantics to Net/Solver construction?

BlGene:
I would like this to happen automatically. Maybe a better solution would be to wrap the caffe.Solver function in such a way that it does the conversion to protobuf and temporary files. This way it would only be a few lines of python code that should be easy to understand.

shelhamer:
See https://github.com/BVLC/caffe/pull/2801

PR #4418, opened 2016-07-05
Remove Python.h include since Boost Python already does it 

mlamarre:
This avoid adding the autolib for the debug version of Python.
This enables building in Debug without having the Python debug binaries.

See this file in Boost
boost\python\detail\wrap_python.hpp

specially the following line
`#undef _DEBUG // Don't let Python force the debug library just because we're debugging`.


seanbell:
LGTM!

longjon:
Hm, I'm not sure why that include was there in the first place. Merging this passing deletion, thanks!

PR #4419, opened 2016-07-05
net_spec: allow for dicts with repeated keys

nitnelave:
The DummyData layer parameters look like this:

```
 layer {
   name: "noise"
   type: "DummyData"
   top: "noise"
   dummy_data_param {
     shape {
       dim: 25
       dim: 32
       dim: 90
       dim: 45
     }
   }
 }
```

The shape is impossible to specify for now (a list of dicts is interpreted as a list of "shape" with only one "dim" per shape), so this commit makes it possible to specify it, by providing:
`shape=[("dim", 25), ("dim", 32), ("dim", 90), ("dim", 45)]`.

### DIFF
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -56,9 +56,10 @@ def to_proto(*tops):
 def assign_proto(proto, name, val):
     """Assign a Python object to a protobuf message, based on the Python
     type (in recursive fashion). Lists become repeated fields/messages, dicts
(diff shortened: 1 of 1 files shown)


cdoersch:
In a few-months-old caffe, `caffe.layers.DummyData(shape={"dim":[100,3,96,96]})` works fine for me.

nitnelave:
Ah, right, I didn't think of putting it this way. Thanks!

PR #4442, opened 2016-07-11
[windows] Fix missing %(DisableSpecificWarnings) in extract_features.vcxproj

nagadomi:
In a Japanese version of Visual Studio, we get lots of  `C4819` warnings from CUDA header files, so I added 
`<DisableSpecificWarnings>4819</DisableSpecificWarnings>` to `CommonSettings.props` but `extract_features.vcxproj` ignores it. `libcaffe.vcxproj` [does not ignore it.](https://github.com/BVLC/caffe/blob/5db30745fd2b47cd9762b9ef64cc33df781a2876/windows/libcaffe/libcaffe.vcxproj#L63)

### DIFF
--- a/windows/extract_features/extract_features.vcxproj
+++ b/windows/extract_features/extract_features.vcxproj
@@ -45,7 +45,7 @@
       <SubSystem>Console</SubSystem>
     </Link>
     <ClCompile>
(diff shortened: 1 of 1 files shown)


willyd:
Closing the VS based build is now deprecated. If you experience the same problem with the CMake build please reopen with changes to CMake build instead.

PR #4448, opened 2016-07-11
Fix comment error in parse_log.py

lalelale:
Aligned output description in docstring with actual output returned by parse_log function


seanbell:
LGTM

PR #4451, opened 2016-07-12
Support for spaces in directories when downloading cifar10

ShaggO:
The get_cifar10.sh script has been slightly modified to support caffe root directories with spaces.


longjon:
Looks right, thanks!

PR #4455, opened 2016-07-13
Support spaces in path when downloading ILSVRC12 and MNIST

ShaggO:
Similar changes to support spaces in the path when downloading ILSVRC12 and MNIST as CIFAR10.
I probably should've checked these too when I did the other one :P


longjon:
Yes, thanks!

Issue #3864, opened 2016-03-21
[Discussion] How to specify all-in-one networks


lukeyeager:
#3211 did most of the hard work to enable all-in-one networks (i.e. just a single `network.prototxt` instead of both `train_val.prototxt` and `deploy.prototxt`). I see several ways that all-in-one networks could be specified. The flexibility is nice for advanced users, but could be confusing to novices. I'd like to eventually update all the examples to use all-in-one networks in a standard way, and I think we should have a discussion first about how to do it.

The trick is to find a way to distinguish between layers used for _validation_ and those used for _deployment_.

## Option 1 - using not_phase

This is the least invasive option. The only interface that needs changing is the solver prototxt.

**Train state** - `phase: TRAIN`
**Val state** - `phase: TEST stage: "val"`
**Deploy state** - `phase: TEST`

**solver.prototxt**
```sh
test_state { stage: "val" }
```
**network.prototxt**
```sh
layer {
  name: "train_layer"
  include { phase: TRAIN }
}
layer {
  name: "val_layer"
  include { stage: "val" }
  # or (equivalent, but more explicit):
  include { phase: TEST stage: "val" }
}
layer {
  name: "deploy_layer"
  include { phase: TEST not_stage: "val" }
}
```

## Option 2 - using level

This is the least user-friendly option. But it is also non-invasive like Option 1.

**Train state** - `phase: TRAIN`
**Val state** - `phase: TEST level: 1`
**Deploy state** - `phase: TEST`

**solver.prototxt**
```sh
test_state { level: 1 }
```
**network.prototxt**
```sh
layer {
  name: "train_layer"
  include { phase: TRAIN }
}
layer {
  name: "val_layer"
  include { min_level: 1 }
  # or (equivalent, but more explicit):
  include { phase: TEST min_level: 1 }
}
layer {
  name: "deploy_layer"
  include { max_level: 0 }
  # or (equivalent, but more explicit):
  include { phase: TEST max_level: 0 }
}
```

## Option 3 - using phases

This is the most verbose option. It requires https://github.com/BVLC/caffe/pull/3863 (or something like it) in order to set the `stage` during deployment. I've already committed to it in DIGITS in hopes that this will be what Caffe chooses to go with (https://github.com/NVIDIA/DIGITS/pull/628).

**Train state** - `phase: TRAIN stage: "train"`
**Val state** - `phase: TEST stage: "val"`
**Deploy state** - `phase: TEST stage: "deploy"`

[...]

PR #4458, opened 2016-07-13
Convert examples to all-in-one networks

lukeyeager:
Implementation of https://github.com/BVLC/caffe/issues/3864.

**Pros**
- Less `.prototxt` to read since the number of files decreases
- Shows users how to build networks so that they only have to make changes to a single file

**Cons**
- Renaming files could break existing workflows and/or documentation

TODO:
- [x] Update all examples
- [ ] Update all documentation


lukeyeager:
Additional con for this implementation:

* Using the networks for timing/testing/deployment requires additional flags to `build/tools/caffe` or `pycaffe` (from https://github.com/BVLC/caffe/pull/3863)

See discussion at https://github.com/BVLC/caffe/issues/3864 - changing the network to require fewer flags makes networks less legible.

lukeyeager:
Some of the documentation isn't really updated yet. I'll wait for feedback on whether this type of change is desired before diving into the docs.

williford:
I hope these changes are merged.

I personally prefer the more explicit method of including the phase and the stage. However, I don't think that is a reason to hold back merging these changes.

lukeyeager:
@williford By "more explicit" do you mean this:
```
include { phase: TRAIN stage: "train" }
include { phase: TEST stage: "val" }
include { phase: TEST stage: "deploy" }
```
instead of this:
```
include { stage: "train" }
include { stage: "val" }
include { stage: "deploy" }
```
I'm happy with that if you think it's clearer. @shelhamer do you have an opinion on this? Is there any chance this gets merged?

williford:
@lukeyeager exactly. I wouldn't make the change unless others agree. I might just be biased because I'm used to having the phase.

lukeyeager:
Closing after a year of inactivity. This PR branch is backed up at https://github.com/lukeyeager/caffe/tree/backup/all-in-one-examples.

asfix:
Hello,

After spending more than one day, I have not still be able to run any model related to resnet or densenet on DIGITS6.

The only thing I can do is training google lenet (that is shipped with Digits6) I am taking any network from some sources (i.e) train_val.prototxt s. However, I can not train them. I ll drive crazy. There is no good documentation about how to make it possible to train those nets. Even the alexnet is problematic (shipped with digits6 model store)

I can not figure out why you did something like that. If I cannot run it, what is the purpose of Digits 6? Everybody who uses digits aims to gain time and play with the networks. With this version, it seems that it ll not be possible to train them with in 2 minutes.

Now, I want to ask @lukeyeager . Please convert the https://github.com/antingshen/resnet-protofiles/blob/master/ResNet_50_train_val.prototxt in order to be trained in Digits6. I have done every possible combination. It still says:

 "ERROR: Layer 'loss' references bottom 'label' at the TEST stage however this blob is not included at that stage. Please consider using an include directive to limit the scope of this layer."

I am excluding, including but nothing works. Either I am blind or your software is buggy. I even checked the caffe_train.py in order to understand whether or not it is out dated. But it is new. So please clarify us. I see lots of people who suffers from this error. Did we install it improperly or is there a bug? Or, must I convert every single layer to make it appropriate all-in-one network? Btw, as I stated, I have tried almost every possible combination including silly ones. But it does NOT work!

Looking forward to hear from you soon.

PR #4464, opened 2016-07-14
Improve performance by creating in_place versions of key vector functions

fenrus75:
Caffe implements a series of functions that operate on vectors, including
caffe_add, caffe_mul and caffe_div (there are more, but these are the
ones that seem to be used the most based on profiling).

These functions implement a vector  y = a op b  operation.

Now there is a slight inefficiency with this;
the compiler will nicely auto-vectorize these functions for the general
case, but because it can't prove that a, b and y are not overlapping aliases of eachother,
so it will also generate a non-SIMD version to fall back to.
(with aliases, the autovectorization in compilers, especially gcc, breaks
down so they fall back to the slow path).

this is all great, but there is a common pattern in the caffe code
that really wants to do

a = a op b

or in other words, an in-place add / mul / div.

This common pattern will ensure that "y" and "a" will alias, and this
then ends up executing the fallback, non-vectorized code, which is 4x or 8x
slower than the vectorized version.

This patch provides a solution to this problem by introducing "in_place" versions
of these 3 common operations; the next patch will introduce the users of these
new functions.

In the CPU profiles, the caffe_add/mul/div end up consuming several percent
CPU time (cpu-only configuration) and with this patch, that reduces drastically
giving a few percent (2 to 3% depending on hardware) performance improvement on
the alexnet benchmark.

### DIFF
--- a/include/caffe/util/math_functions.hpp
+++ b/include/caffe/util/math_functions.hpp
@@ -56,13 +56,22 @@ template <typename Dtype>
 void caffe_add(const int N, const Dtype* a, const Dtype* b, Dtype* y);
 
 template <typename Dtype>
+void caffe_add_in_place(const int N, Dtype* a, const Dtype* b);
+
+template <typename Dtype>
 void caffe_sub(const int N, const Dtype* a, const Dtype* b, Dtype* y);
 
 template <typename Dtype>
 void caffe_mul(const int N, const Dtype* a, const Dtype* b, Dtype* y);
 
 template <typename Dtype>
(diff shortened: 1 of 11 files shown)


fenrus75:
Hi,
I realize CPU performance is maybe not the most critical for the project team.. but there's something like an easy 20% total on the table with a series of patches, of which this is the first.
Do I need to do something more to move this forward?

fenrus75:
Obsoleted by the much simpler https://github.com/BVLC/caffe/pull/5130

PR #4466, opened 2016-07-14
Fixing relative import for caffe module

philkr:
Currently caffe can only be imported in python if there is a top level caffe directory, but not if caffe resides in a subdirectory. The reason for this is that several files in `python/caffe/` reference the caffe module (that only exists if caffe is a top level directory).
This PR fixes this and makes all import inside the `python/caffe/` directory relative.

I also removed the proto import warning in io.py as proto is imported in other places without warning.

### DIFF
--- a/python/caffe/classifier.py
+++ b/python/caffe/classifier.py
@@ -5,10 +5,9 @@ Classifier is an image classifier specialization of Net.
 
 import numpy as np
 
-import caffe
+from . import Net, io, TEST
 
-
-class Classifier(caffe.Net):
+class Classifier(Net):
     """
     Classifier extends Net for image class prediction
     by scaling, center cropping, or oversampling.
@@ -23,7 +22,7 @@ class Classifier(caffe.Net):
     def __init__(self, model_file, pretrained_file, image_dims=None,
                  mean=None, input_scale=None, raw_scale=None,
                  channel_swap=None):
-        caffe.Net.__init__(self, model_file, pretrained_file, caffe.TEST)
+        Net.__init__(self, model_file, pretrained_file, TEST)
 
         # configure pre-processing
         in_ = self.inputs[0]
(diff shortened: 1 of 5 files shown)


longjon:
Perhaps I am missing something, but I don't get this... what do you mean by "if caffe resides in a subdirectory"?

The `python/caffe` dir is the `caffe` module; you're not allowed to rearrange things inside it, or put it in another package.

philkr:
I probably could have been clearer. Let say caffe resides in `python/caffe` and there is a `python/__init__.py` file. You can now call `import python` which will work, but `import python.caffe` (or `from python import caffe`) will not in the current version of caffe. With this PR it will.

longjon:
Right, that is not allowed... you can't just take the `caffe` module (or any Python module) and decide it's part of another package, and I'm not sure why you would want to do that.

You can still argue for this patch inasmuch as it makes a certain kind of refactoring easier, but I think I prefer the explicit style (see, e.g., https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports; plus, it's more consistent with the usual user-level usage).

philkr:
We can argue if it's good style or not, but it is convenient. Also the changes here are fairly minor apart from `detector.py` and `classifier.py`.
The import `from .proto import caffe_pb2` is used in other places already, so changing to this would make is more consistent, so is the `from . import io`.

longjon:
> We can argue if it's good style or not, but it is convenient.

For what purpose? As above:

> I'm not sure why you would want to do that.

Part of style is making things you shouldn't do inconvenient.

> The import `from .proto import caffe_pb2` is used in other places already, so changing to this would make is more consistent, so is the `from . import io`.

I would argue (mildly) for the opposite change.

philkr:
without starting any religious argument, here is the concrete issue I'm facing right now:
I wrote some beautiful piece of code (a python module, let's call it `foo`) that uses caffe internally. The way things are setup is as follows `foo/caffe` is a simlink to the `caffe/python/caffe` directory. In that way foo can import caffe with the least amount if overhead (no `sys.path` stuff required).
Now I would like to use foo in another project et voila, things break.

longjon:
Well, then `foo` depends on `caffe`; if `caffe` is not in your `sys.path`, you don't really have `caffe`, e.g., Python layers will crash; allowing/encouraging use of the code this way is broken.

Note that if you `import caffe` from `foo/`, that works because `caffe` _is_ in your `sys.path`.

shelhamer:
Agreed with @longjon on this. Closing.

PR #4471, opened 2016-07-15
Refine intel spatial convolution engine

gongzg:
Fix one group related bugs for the IDLF kernel, and relax the image size restrication for kernels. Previous implementation, each kernel could only support one image size which is not good for real application which may need to handle many different images.

The size relax patch bring more register pressure which introduce some regression. This patchset use sub-buffer to eliminate the buffer offset for the IDLF kernel thus it could fix the regression, and finally could get about 5% to 10% overall performance improvement.


gongzg:

gongzg:
@naibaf7 ping for review. Thanks.

naibaf7:
@gongzg Scheduled.

PR #4474, opened 2016-07-15
Stop setting cache timeout in TravisCI

lukeyeager:
It refers to the caching command timeout, not how long before the caches expire as I had thought.

> The caches expire when they are 45 days old, and this value is not configurable.
 https://github.com/travis-ci/travis-ci/issues/6317#issuecomment-233065994

So this line is at best useless, and at worst will cause the build to hang for a week!?


jeffdonahue:
Makes sense, thanks for the fix!

PR #4484, opened 2016-07-19
Add random resize to Data Transformer

ssudholt:
Continuing on #1900, this PR adds the option to randomly resize an input datum/blob and extract a crop from the resized image. The other features of #1900 are not included.

Random resizing is controlled by the min_size and max_size parameters in the transform_param definition in the proto files. At this point, random resizing is only possible if a crop size is defined also.


ssudholt:
PR is ready for review

PR #4519, opened 2016-07-25
Mac build

hlzz:
The first commit deals with the 'Cannot find cblas.h header file' error on mac.
The second commit deals with cmake find cuDNN error on mac, because dynamic libs end with .dylib on mac.


shelhamer:
@hlzz thanks for the report and patches! I merged a fix for vecLib in #4604 and cherry-picked your cuDNN fix in #4605.

PR #4523, opened 2016-07-26
Fix search for Atlas on arch.

hgaiser:
On Arch, installing `atlas-lapack-base` results in a file `/usr/lib/liblapack.so`.  The cmake module for finding Atlas only looks for `alapack`, **not** for `lapack`. This commits adds the search for `liblapack.so`.


longjon:
Seems reasonable to me, thanks.

PR #4527, opened 2016-07-27
Extra interfaces to use caffe operator from mxnet

HrWangChengdu:
Hi,

I've added several interfaces in Blob & Layer to support caffe plugin in mxnet. Two prime purposes for those interfaces are:
- to set gpu_data/diff & cpu_diff in caffe::blob, currently it only supports cpu_data(). With those methods, mxnet could build a caffe::blob with its own data.
- to provide access to blobs_ in caffe::layer. In this way, mxnet could set its own weights into caffe layer.

With these interfaces, mxnet could call caffe op in its symbolic interface without doing memory copy.  

If you're interested in how caffe op is wrapped in mxnet, please check the https://github.com/dmlc/mxnet/pull/2746. The caffe-plugin feature would be merged to mxnet master in 1 or 2 days.

We deeply believe this plugin is beneficial to both caffe & mxnet communities, also for convenience of both side users.

Thank you, Haoran

### DIFF
--- a/include/caffe/blob.hpp
+++ b/include/caffe/blob.hpp
@@ -218,6 +218,9 @@ class Blob {
 
   const Dtype* cpu_data() const;
   void set_cpu_data(Dtype* data);
+  void set_cpu_diff(Dtype* data);
+  void set_gpu_data(Dtype* data);
+  void set_gpu_diff(Dtype* data);
   const int* gpu_shape() const;
   const Dtype* gpu_data() const;
   const Dtype* cpu_diff() const;
--- a/include/caffe/layer.hpp
+++ b/include/caffe/layer.hpp
@@ -177,6 +177,13 @@ class Layer {
       const vector<Blob<Dtype>*>& bottom);
 
   /**
+   * @brief Set phase: TRAIN or TEST
+   */
+  inline void SetPhase(Phase p) {
+    phase_ = p;
+  }
+
+  /**
    * @brief Returns the vector of learnable parameter blobs.
    */
   vector<shared_ptr<Blob<Dtype> > >& blobs() {
--- a/include/caffe/layer.hpp
+++ b/include/caffe/layer.hpp
@@ -184,25 +184,6 @@ class Layer {
   }
 
   /**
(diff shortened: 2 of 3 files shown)


seanbell:
Can you rebase this against the latest `BVLC/master` branch, and squash your edits into a single commit?  If you look at the commit history in the PR, you'll notice a bunch of extra commits.

include/caffe/layer.hpp
```diff
@@ -184,6 +184,25 @@ class Layer {
   }
 
   /**
+   * @brief Sets blobs
+   */
+  void SetBlobs(const vector<Blob<Dtype>*>& weights) {
```
ajtulloch: For both of these, Layer::blobs() is sufficient, there's no change needed to this file.

include/caffe/blob.hpp
```diff
@@ -218,6 +218,9 @@ class Blob {
 
   const Dtype* cpu_data() const;
   void set_cpu_data(Dtype* data);
+  void set_cpu_diff(Dtype* data);
```
ajtulloch: These are legit and should be added.

HrWangChengdu:
@seanbell @ajtulloch I've squashed commits and removed changes in blob file. please check

piiswrong:
@ajtulloch Is this good for merging?

ajtulloch:
This looks good to me - @shelhamer, @longjon, @jeffdonahue?

ajtulloch:
Maybe for completeness a unit test that instantiates a blob memory, sets various external pointers, and verifies that the pointer values are correct for before/after the external pointers are set.

mli:
ping, any progress on this PR?

jingpengwu:
this is interesting and useful, any update here?

shelhamer:
`phase` is immutable #1500 so `SetPhase()` should not be part of the interface. Adding the further setters/getters seems fine to me. A unit test as pointed out by @ajtulloch would be nice too.

HrWangChengdu:
Great. I'll modify the pr asap

PR #4528, opened 2016-07-27
Fix the cmake build for intel spatial convolution.

gongzg:


gongzg:
@naibaf7 please review this patchset, thanks.

PR #4533, opened 2016-07-27
Add SetMinLogLevel for setting glog log level.

hgaiser:
As per title, this allows one to set the log level of glog logging through a function in caffe.

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -97,6 +97,9 @@ using std::vector;
 // Currently it initializes google flags and google logging.
 void GlobalInit(int* pargc, char*** pargv);
 
+// Sets the Google logging min log level.
+void SetMinLogLevel(int level);
+
 // A singleton class to hold common caffe stuff, such as the handler that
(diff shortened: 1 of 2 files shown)


lukeyeager:
Why are you adding a function that's never called or exposed externally?

Also, why not just set the environment variable?
```sh
export GLOG_minloglevel=2
```

hgaiser:
What do you maan? It is exposed in caffe/common.hpp. 

I think it is much cleaner to expose such functionality through a function call than through an environment variable. I will leave it up to caffe developers to see if they agree or disagree with this perspective.

longjon:
We can debate the merits of glog's interface, but I don't think it's a good idea to create a custom interface to a library which substitutes one line with another one line.

hgaiser:
Fair enough, the reason I added it was because I dislike setting the environment variable and setting that flag in my code causes a linking error. My options were to link against glog or add this function in Caffe. I preferred the latter, but I will respect your opinion.

PR #4559, opened 2016-08-07
Fix Label Shape Check in LossLayer::Reshape

fyu:
The legacy shape is deprecated. The blobs are supposed to be created with new Reshape. This PL updates the blob shape check in LossLayer assuming the blobs are created without legacy shape.


jeffdonahue:
Thanks @fyu!

fyu:
No problem. Thanks for the speedy merge.

PR #4379, opened 2016-06-28
add default value to rms_decay and fix documentation

alicanb:
fixes #3472 and fixes #3930

PR #4567, opened 2016-08-09
Update rms_decay in caffe.proto to match docs

lukeyeager:
@smichalowski pointed out in https://github.com/NVIDIA/DIGITS/pull/956 that the docs say the default value for `rms_decay` is 0.02:

> The default value of $$\delta$$ (rms_decay) is set to $$\delta = 0.02$$.
https://github.com/BVLC/caffe/blob/rc3/docs/tutorial/solver.md#rmsprop

But it looks like the default is actually 0. Am I missing something?


jeffdonahue:
Thanks for fixing this to match the docs @lukeyeager, but merged #4379 instead as 0.99 is a much more reasonable default.

PR #4574, opened 2016-08-10
Dynamic masking capability for real-time data augmentation

chprasad:
This PR adds the capability to dynamically mask at a random position within the image, a 'mask_size' block, once in 'mask_freq' iterations.


shelhamer:
I appreciate the effort to contribute new data augmentations, but I think this is better handled by a Python data layer to avoid accumulating complexity in the core.

PR #4578, opened 2016-08-11
rc3 fix for mac-build

leoshine:
The first commit deals with the 'Cannot find cblas.h header file' error on mac.
The second commit deals with cmake find cuDNN error on mac, because dynamic libs end with .dylib on mac.

### DIFF
--- a/cmake/Cuda.cmake
+++ b/cmake/Cuda.cmake
@@ -174,11 +174,18 @@ function(detect_cuDNN)
             PATHS ${CUDNN_ROOT} $ENV{CUDNN_ROOT} ${CUDA_TOOLKIT_INCLUDE}
             DOC "Path to cuDNN include directory." )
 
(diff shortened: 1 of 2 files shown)


shelhamer:
@leoshine ah, thanks for pointing out the cuDNN issue in 3fc6792. Is NVIDIA still shipping it as a `dylib`? If so I'll cherry pick that commit.

leoshine:
Apologise for bothering!

The issue is correct: cuDNN lib on mac is still dylib.

But this is actually not my fix. I just copy from others, and accidental create a pull request.
Please see: https://github.com/BVLC/caffe/pull/4519 <https://github.com/BVLC/caffe/pull/4519>
The two fix is essential for build on mac.

shelhamer:
I see, thanks for the pointer. Closing.

PR #4589, opened 2016-08-14
Correct a minor mistake in statements

sungjunhong:
### Correct a statement from
The `fc6` inner product when carried out as convolution by `fc6-conv` turns into a 6 \times 6 filter with stride 1 on `pool5`.
### to
The `fc6` inner product when carried out as convolution by `fc6-conv` turns into a 6 x 6 filter with stride 1 on `pool5`.


longjon:
Okay, thanks!

Issue #3841, opened 2016-03-17
Caffe Makefile Command line tools, and paths in El Capitan


wcray:
Two intertwined issues

related to #3092

The caffe makefile produces

No receipt for 'com.apple.pkg.CLTools_Executables' found

when following the Atlas BLAS path on OS X 10.11

This can be fixed by changing : 

pkgutil --pkg-info=com.apple.pkg.CLTools_Executables

to 

pkgutil --pkg-info=com.apple.pkg.Xcode

Once that is taken care of, on machines with virgin installs of El Capitan, the /System/Library/Frameworks tree no-longer contains the Accelerate/veclib headers, and these have instead moved to a SDK tree under the Xcode app itself.

As a result, the line for BLAS_INCLUDE needs to be changes from:

BLAS_INCLUDE ?= /System/Library/Frameworks/Accelerate.framework/Versions/Current/Frameworks/vecLib.framework/Headers/

to

BLAS_INCLUDE ?= /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/Accelerate.framework//Vers\
ions/A/Frameworks/vecLib.framework/Versions/A/Headers/

shelhamer:
Thanks for reporting. The BLAS default dir is fixed in #4604, but the issue with the command line tools (`CLTools_Executables`) vs. Xcode is configured as desired if I recall correctly in that we prefer for the command line tools alone to suffice for compiling Caffe. If that is not working as expected, please comment.

PR #4604, opened 2016-08-18
Set default BLAS include for OS X 10.11

shelhamer:
The latest hunt for the ever-elusive vecLib/Accelerate. I just realized that I only ever set this in my `Makefile.config` and didn't contribute the fix upstream, sorry.

The need for this has been pointed out by others, including #3841.


shelhamer:
Credit to @hlzz for the report and fix in #4519.

PR #4609, opened 2016-08-20
Improvements to the build system

intelfx:
These are various improvements to the CMake buildsystem directed towards "modern" packaging with CMake IMPORTED targets and config-mode `find_package()`.

There are several bugfixes and a slight refactoring of the buildsystem which makes Caffe properly export its transitive dependencies ("usage requirements") to the library users.

Open questions:
- most of the dependencies are made PUBLIC (i. e. the client programs will link to them). This mimics the previous behavior, but it might be worth to revise configuration. E. g. the BLAS libraries — do we really need to make clients link to them?

Tests done:
- building as a static and shared library for GNU/Linux and Android, linking to a sample project.

### DIFF
--- a/src/caffe/CMakeLists.txt
+++ b/src/caffe/CMakeLists.txt
@@ -4,8 +4,11 @@ caffe_protobuf_generate_cpp_py(${proto_gen_folder} proto_srcs proto_hdrs proto_p
 
 # include python files either to force generation
 add_library(proto STATIC ${proto_hdrs} ${proto_srcs} ${proto_python})
-set(Caffe_LINKER_LIBS proto ${Caffe_LINKER_LIBS}) # note, crucial to prepend!
 caffe_default_properties(proto)
+target_link_libraries(proto PUBLIC ${PROTOBUF_LIBRARIES})
+target_include_directories(proto PUBLIC ${PROTOBUF_INCLUDE_DIR})
+
+list(INSERT Caffe_LINKER_LIBS 0 PUBLIC proto) # note, crucial to prepend!
 
 # --[ Caffe library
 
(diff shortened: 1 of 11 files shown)


shelhamer:
Thanks for the build polish.

src/caffe/CMakeLists.txt
```diff
@@ -18,8 +21,16 @@ if(HAVE_CUDA)
...
 
 add_library(caffe ${srcs})
-target_link_libraries(caffe proto ${Caffe_LINKER_LIBS})
 caffe_default_properties(caffe)
+target_link_libraries(caffe ${Caffe_LINKER_LIBS})
+target_include_directories(caffe ${Caffe_INCLUDE_DIRS}
+                                 PUBLIC
+                                 $<BUILD_INTERFACE:${Caffe_INCLUDE_DIR}>
+                                 $<INSTALL_INTERFACE:include>)
+target_compile_definitions(caffe ${Caffe_DEFINITIONS})
```
willyd: According to the CMake release notes: this command was added in 2.8.11. The root CMakeLists currently requires only 2.8.7.

src/caffe/CMakeLists.txt
```diff
@@ -18,8 +21,16 @@ if(HAVE_CUDA)
...
+if(Caffe_COMPILE_OPTIONS)
+  target_compile_options(caffe ${Caffe_COMPILE_OPTIONS})
```
willyd: According to the CMake [release notes](https://blog.kitware.com/cmake-2-8-12-available-for-download/) this command was added in 2.8.12. The root CMakeLists currently requires only 2.8.7.

willyd:
@intelfx Looks like we should also bump the minimum CMake version from 2.8.7 to 2.8.12.

severin-lemaignan:
@intelfx While moving toward 'modern' CMake is certainly laudable, IMPORTED targets are not yet very common and I'm sorely missing documentation/example on how to use them. Even the official CMake documentation does not help me much in that regard.

Could you provide examples (possibly in the documentation of `CaffeConfig.cmake`) showing how to integrate Caffe in a C++ CMake project?

Right now I'm here:
```cmake
# [...]
find_package(Caffe) 
add_library(Caffe SHARED IMPORTED)
# [...]
add_executable(myapp src/myapp.cpp)
target_link_libraries(myapp Caffe)
```

This fails because the Caffe includes are not found. Adding `target_include_directories(replay_with_poses PUBLIC Caffe)` before `add_executable` does not help as the target `myapp` does not exist yet; adding it after does not help either (headers not found at the `add_executable` step).

PR #4630, opened 2016-08-25
Made load_hd5 check blob dims by default, instead of reshaping.

BlGene:
### Reason:

Reshaping by default, as done by the `load_hd5` function can cause problems. I my case I was:
1. loading weights to a training net with `solver.net.load_hdf5(model_fn)`
2. Copying to solver with `solver.test_nets[0].share_with(solver.net)`
3. Getting the following error: 
`F0825 01:06:18.144207 21739 net.cpp:703] Check failed: target_blobs[j]->shape() == source_blob->shape() Cannot share param 0 weights from layer 'fc6'; shape mismatch.  
Source param shape is 4096 25088 (102760448); target param shape is 4096 512 7 7 (102760448)`

This was very confusing, because the actual train net was different from the version shown in the log files. I would expect an when calling `load_hdf5`. If I recall this would match the behavior of the protobuf serialization. 
### Open Questions:
1. Is the error message sufficiently clear or should `hdf5_load_nd_dataset_helper` return false and the error be throw where we know the layer for which this problem occurred.
2. An alternative implementation would be to call hdf5_load_nd_dataset with a temporary blob, and only copy this over to the current one if the shape is the same, this would move changes to net.cpp:846, but this might take more memory.

### DIFF
--- a/src/caffe/util/hdf5.cpp
+++ b/src/caffe/util/hdf5.cpp
@@ -9,7 +9,7 @@ namespace caffe {
 template <typename Dtype>
 void hdf5_load_nd_dataset_helper(
     hid_t file_id, const char* dataset_name_, int min_dim, int max_dim,
-    Blob<Dtype>* blob) {
+    Blob<Dtype>* blob, bool reshape) {
   // Verify that the dataset exists.
   CHECK(H5LTfind_dataset(file_id, dataset_name_))
       << "Failed to find HDF5 dataset " << dataset_name_;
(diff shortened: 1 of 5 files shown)


BlGene:
@erictzeng 
(haven't looked at what this means for unit tests.)

src/caffe/util/hdf5.cpp
```diff
@@ -56,26 +56,48 @@ void hdf5_load_nd_dataset_helper(
...
+      int count = 1;
+      for (int i = 0; i < blob_dims.size(); ++i) {
+        stream << blob_dims[i] << " ";
+        count = count * blob_dims[i];
+      }
+      stream << "(" << count << ")";
+      string source_shape_string = stream.str();
+
+      CHECK(blob_dims == blob->shape()) << "Cannot load blob from hdf5; shape "
+            << "missmatch. Source shape is " << source_shape_string
```
erictzeng: Minor nit: "missmatch" -> "mismatch"

BlGene:
Ok, I squashed. Pending tests its ready.

PR #4640, opened 2016-08-27
Copyright update

mvpatel2000:
Shouldn't the copyright be updated? Sorry if this is incorrect, just saw this when downloading and thought it might have been an oversight.


shelhamer:
Fixed in 9ab67099. Thanks for pointing out the old notice.

PR #4645, opened 2016-08-28
updates tense in docs

ClimbsRocks:
"could" seems to imply for some reason that something is blocking one from calling the registered layers. "can" lays out more directly that a user can choose to do this.


jeffdonahue:
Thanks for the fix @ClimbsRocks!

PR #4660, opened 2016-08-30
Opencl fixes for Intel OpenCL SDK

gongzg:
This PR fixes a bug for Intel OpenCL SDK and now could pass all the GPU device test cases with the official OpenCL SDK. Also updated the download link of Intel OpenCL SDK which could support SKL linux. The convolution kernel performance is good with the official driver. But the GEMM kernel from ISAAC has known performance issue with this driver. We already have a much better GEMM implementation, but need some time to integrate it to upstream ISAAC.


gongzg:
@naibaf7 please help to review. And you can switch to the latest OpenCL 2.0 SDK driver for your skylake machine to check the convolution performance.

naibaf7:
@gongzg 
Thanks! I will review and test today! Great work :)

naibaf7:
LGTM! Next up is driver testing on Skylake, I really hope it will work now.

PR #4677, opened 2016-09-02
Bilinear filler value corrected when matrix size is odd.

zhengwy888:
I believe there was an typo in BilinearFiller. 
E.g. if you set the kernel size to be 3. the filler produces the following result:

```
0.062500 0.187500 0.187500 
0.187500 0.562500 0.562500 
0.187500 0.562500 0.562500
```

Where it should have been

```
0.250000 0.500000 0.250000 
0.500000 1.000000 0.500000 
0.250000 0.500000 0.250000 
```

Looking at the code and this is because the matrix size condition (f%2) should have used the actual matrix size. 

Let me know what you think.


Noiredd:
Fixed in #5713.

PR #4699, opened 2016-09-09
added CrossEntropy loss layer

beniz:
When using an autoencoder, one of the losses of choice is `SigmoidCrossEntropyLoss`. However, at runtime, since the `Sigmoid` and the loss are merged in together, it is not possible to recover the decoded output easily, right after the `Sigmoid`.

Admittedly, this is a niche use case where we might want to look the decoded output up for debug or other assessment purposes (e.g. denoising, etc...).

This PR adds a `CrossEntropyLoss` layer to Caffe that can be used along a `Sigmoid` layer in `deploy.prototxt` so that at prediction time the output of the autoencoder can be easily retrieved. There's no implementation of `backward_cpu/_gpu()` since for training purposes the `SigmoidCrossEntropyLoss` should be preferred.

Of course Caffe already allows to do this by defining a `deploy.prototxt` with and without the `SigmoidCrossEntropyLoss` layer, so this PR really only allows to have only a single deploy file.

In summary, a niche PR just in case it proves useful to others one of these days.

### DIFF
--- a/include/caffe/layers/cross_entropy_loss.hpp
+++ b/include/caffe/layers/cross_entropy_loss.hpp
@@ -1,0 +1,44 @@
+#ifndef CAFFE_CROSS_ENTROPY_LAYER_HPP
+#define CAFFE_CROSS_ENTROPY_LAYER_HPP
+
+#include <vector>
+
+#include "caffe/blob.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/proto/caffe.pb.h"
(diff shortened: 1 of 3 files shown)


shaibagon:
You can have a Sigmoid layer in parallel to the loss layer in train prototxt.
Alternatively, you may implement a second top for the layer the same way it is implemented in the SoftmaxWithLoss layer

PR #4701, opened 2016-09-09
Work around for building on Ubuntu 16.04

cypof:
https://github.com/tensorflow/tensorflow/issues/1346
https://github.com/torch/cutorch/issues/401

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -406,7 +406,7 @@ CXXFLAGS += -MMD -MP
 # Complete build flags.
 COMMON_FLAGS += $(foreach includedir,$(INCLUDE_DIRS),-I$(includedir))
 CXXFLAGS += -pthread -fPIC $(COMMON_FLAGS) $(WARNINGS)
-NVCCFLAGS += -ccbin=$(CXX) -Xcompiler -fPIC $(COMMON_FLAGS)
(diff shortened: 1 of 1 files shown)


lukeyeager:
Please don't merge this. To build on Ubuntu 16.04 (or any machine with GCC >= 5.0), simply **install CUDA 8.0**. CUDA 7.5 does not support GCC < 5.0, and setting an internal compiler flag to trick CUDA into allowing the compilation is not safe.

Plus, you're only editing the raw Makefile build anyway.

shelhamer:
Agreed with @lukeyeager. The right resolution is to update the installation docs noting the CUDA 8 requirement for 16.04.

Thanks for reminding us about this difference across platform versions @cypof.

CDLuminate:
Now GCC-6 serves as the mainline of GCC. and CUDA 8.0~RC is not working with it. So in a long time the next Debian release cannot build cuda apps because there is only GCC-6. ArchLinux has already migrated to GCC-6 as default compiler and things related to CUDA just got borken and community can do nearly nothing about it except for dirty hacks on host_config or nvcc flags.  I don't know what gcc version will ubuntu maintain in the future releases but actually it won't diverge far from Debian because there is maintainer overlap.

In short, the CUDA compilation on new systems will be "interesting". Be prepared for more user complain about compiling failure on new systems ...

Issue #4453, opened 2016-07-12
Issue with Caffe's timing code


ethereon:
Caffe's `time` command includes two timers: a per layer timer and an overall timer. However, the overall timer currently also includes the overhead of the per-layer timers. For instance, for forward prop: https://github.com/BVLC/caffe/blob/master/tools/caffe.cpp#L383-L389

This overhead can be significant. For instance, the current version of Caffe reports the following for [ResNet-152](https://github.com/KaimingHe/deep-residual-networks/blob/master/prototxt/ResNet-152-deploy.prototxt) on an Nvidia Tegra TX1:

`Average Forward pass: 406.875 ms.`

After commenting out the nested timers:

`Average Forward pass: 193.922 ms.`

Similar caveats apply for the backward pass and total iteration timers.

flx42:
Right, the solution is to do 2 passes. Use a new timer for each individual layer, then use another timer for a full forward pass.
I'm doing something like that here:
https://github.com/flx42/caffe/commit/89064913221c57aeb0aabef1643afe4c931e12fc
(in this patch I also removed the backward pass because I was just benchmarking inference).

ujsyehao:
@flx42 Hi, I use your method, but I found the problem:
**caffe_net.Forward() consumes time unmatch sum of each layer consume time**
Code below:
![319](https://user-images.githubusercontent.com/17454232/51797926-bd911300-2245-11e9-8563-6e24d900af38.png)
![320](https://user-images.githubusercontent.com/17454232/51797927-c255c700-2245-11e9-83a6-9562a9ef7f0f.png)
net forward time: 5.1ms ~ 5.2ms
the sum of each layer forward time: 4.4 ~ 4.5ms

PR #4702, opened 2016-09-09
Benchmarking should not impact perf until timer is read

cypof:
For timing operations in a net, it should be possible to add events to a GPU stream without synchronizing, to have a lower impact on performance and more accuracy. Synchronizing is only required when reading the timer, e.g. once at the end of an iteration instead of at each layer.


shelhamer:
Looks good to me. Thanks Cyprien!

shelhamer:
@flx42 could you comment on whether this resolves #4453?

ujsyehao:
@shelhamer It is a big overhead.

PR #4703, opened 2016-09-09
Avoids missing return values during build.

cypof:


src/caffe/layer_factory.cpp
```diff
@@ -67,6 +67,7 @@ shared_ptr<Layer<Dtype> > GetConvolutionLayer(
 #endif
   } else {
     LOG(FATAL) << "Layer " << param.name() << " has unknown engine.";
+    throw;  // Avoids missing return warning
```
shelhamer: `return NULL` is more clear than `throw` at least in my reading, and consistent with #3362.

cypof:
I tried, but the methods return shared_ptr, for which NULL is not a valid value. We could return shared_ptr<Layer<Dtype> >(), but it's not very readable.

shelhamer:
Ok fine, seems like an inconsequential and perhaps necessary evil. If anybody sees a more stylish resolution of this, give a shout.

PR #4718, opened 2016-09-12
added apt command to install OpenBLAS

jgyllinsky:
I'm not sure about MKL, but this should at least install OpenBLAS using APT on Ubuntu 16.04.

Also note that `sudo apt install` may be sufficient on many devices, but for older versions the apt-get used here is likely better.

### DIFF
--- a/docs/install_apt.md
+++ b/docs/install_apt.md
@@ -14,7 +14,7 @@ The NVIDIA package tends to follow more recent library and driver versions, but 
 If installing from packages, install the library and latest driver separately; the driver bundled with the library is usually out-of-date.
 This can be skipped for CPU-only installation.
 
-**BLAS**: install ATLAS by `sudo apt-get install libatlas-base-dev` or install OpenBLAS (with `sudo apt-get install libopenblas-base libopenblas-dev`) or MKL for better CPU performance.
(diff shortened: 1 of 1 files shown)


docs/install_apt.md
```diff
@@ -14,7 +14,7 @@ The NVIDIA package tends to follow more recent library and driver versions, but
 If installing from packages, install the library and latest driver separately; the driver bundled with the library is usually out-of-date.
 This can be skipped for CPU-only installation.
 
-**BLAS**: install ATLAS by `sudo apt-get install libatlas-base-dev` or install OpenBLAS or MKL for better CPU performance.
+**BLAS**: install ATLAS by `sudo apt-get install libatlas-base-dev` or install OpenBLAS (with `sudo apt-get install libopenblas-base libopenblas-dev`) or MKL for better CPU performance.
```
lukeyeager: No need to include both packages. Just install the `-dev` one and the other gets installed automatically.
```
$ apt-cache depends libopenblas-dev
libopenblas-dev
  Depends: libopenblas-base
  Depends: libblas-dev
```

jgyllinsky: ok, thanks

jgyllinsky: Updated the push / branch

PR #4720, opened 2016-09-13
OpenCL: Add missing macro definition for new spatial convolution kernels.

gongzg:
LOOP0 is required for the new GEMM like convolution kernels.


gongzg:

PR #4736, opened 2016-09-18
Fix missing proto header path in CMake

zhang-xin:
The generated protobuf header path is not included in Caffe target,
a consumer links to Caffe will not find the protobuf header.

The generated header is at CMake build directory but only source include direcotry is added in CMake target Caffe.

For example a minimal CMake project:
```
find_package(Caffe REQUIRED)
add_executable(mytest main.cpp)
target_link_libraries(mytest ${Caffe_LIBRARIES})
```

Compile error:
> In file included from /home/xzhang84/projects/caffe/include/caffe/caffe.hpp:7:0,
>                  from /home/xzhang84/projects/test/main.cpp:2:
> /home/xzhang84/projects/caffe/include/caffe/blob.hpp:10:34: fatal error: caffe/proto/caffe.pb.h: No such file or directory

### DIFF
--- a/src/caffe/CMakeLists.txt
+++ b/src/caffe/CMakeLists.txt
@@ -26,6 +26,7 @@ target_link_libraries(caffe ${Caffe_LINKER_LIBS})
 target_include_directories(caffe ${Caffe_INCLUDE_DIRS}
                                  PUBLIC
                                  $<BUILD_INTERFACE:${Caffe_INCLUDE_DIR}>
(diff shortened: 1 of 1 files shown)


naibaf7:
@shelhamer 
What do you think of this? Also an issue in mainline Caffe? Not sure if handling this as OpenCL-only issue is correct...

zhang-xin:
@naibaf7 I think it's an OpenCL branch only issue.

PR #4749, opened 2016-09-21
Add Visual Studio Code files to .gitignore.

kyu-sz:
Using Visual Studio Code to organize the code, I find it creating a directory named ".vscode" containing files such as ".browse.VC.db", which should be ignored by Git.

PR #4779, opened 2016-09-27
Ignore Visual Studio Metadata

shelhamer:
Replacement for #4749. Thanks @kyu-sz


shelhamer:
For the record I'm of the opinion all of these editor ignores should be handled in the user's `~/.gitignore`, but since we currently have others included we might as well have this editor ignored too since users have requested it.

PR #4785, opened 2016-09-29
slightly relax batch norm check

bwilbertz:
As already raised in #4704, the current LayerSetup method of the BatchNormLayer expects to have **no** LayerParamter configured at all, but also adds 3 LayerParameter itself (cf. [batch_norm_layer.cpp](https://github.com/BVLC/caffe/blob/master/src/caffe/layers/batch_norm_layer.cpp#L39-L44)).
This all is done to avoid any unintenional backward modification of the global stats in the BatchNormLayer and therefore forcing `lr_mult = 0.f`.

As a consequence, a valid (upgraded) net, which was passed once into `Net<Dtype>(net_param)` and read back via `ToProto(...)` cannot be passed a second time into `Net` without being upgraded again.

This PR slightly relaxes the checks in LayerSetUp, to just ensure `lr_mult == 0.f` or adding it, if the LayerParameter is missing (as it was introduced in #4704).

### DIFF
--- a/src/caffe/layers/batch_norm_layer.cpp
+++ b/src/caffe/layers/batch_norm_layer.cpp
@@ -36,11 +36,15 @@ void BatchNormLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
   }
   // Mask statistics from optimization by setting local learning rates
   // for mean, variance, and the bias correction to zero.
-  CHECK_EQ(this->layer_param_.param_size(), 0)
-      << "Cannot configure batch normalization statistics as layer parameters.";
   for (int i = 0; i < this->blobs_.size(); ++i) {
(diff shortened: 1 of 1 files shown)


shelhamer:
Thanks for issuing a patch for this @bwilbertz and sorry I couldn't handle it immediately. I'll try to check and merge today.

Note that the CI failures seem to be some bogus issue with CUDA packages on the Travis machines: https://travis-ci.org/BVLC/caffe/jobs/163831672 We can hopefully resolve that shortly. @lukeyeager have you seen this issue w/ Travis?

lukeyeager:
> @lukeyeager have you seen this issue w/ Travis?

Yeah we're working on it - sorry.

shelhamer:
Thanks again for this @bwilbertz -- merged after checking it myself since the CI issue is unrelated.

@lukeyeager thanks for looking into the CI problem!

PR #4792, opened 2016-09-30
Make sure memory is allocated to only one device, enforce if debug

cypof:
I added checks while debugging the new multi-GPU code. It's simpler to always enforce a single device instead of trying to switch on the fly. If we merge that I can rebase the nccl branch on it.


cypof:
In the nccl branch, WorkerSolver goes away so it makes this simpler too.

shelhamer:
Closed as handled by #4563

PR #4799, opened 2016-10-03
update the module of the scikit-learn (in example/brewing-logreg)

wit543:
Newer version of scikit-learn doesn't have sklearn.cross_validation.train_test_split anymore, they use sklearn.model_selection.train_test_split instread.

### DIFF
--- a/examples/brewing-logreg.ipynb
+++ b/examples/brewing-logreg.ipynb
@@ -73,7 +73,7 @@
     ")\n",
     "\n",
     "# Split into train and test\n",
-    "X, Xt, y, yt = sklearn.cross_validation.train_test_split(X, y)\n",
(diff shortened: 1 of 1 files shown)


yghlc:
I had this problem today with sklearn 0.18.1., It seems that the example codes haven't been updated.
The error message was below, 
```

----> 7 X, Xt, y, yt = sklearn.cross_validation.train_test_split(X, y)
      8 #X, Xt, y, yt = sklearn.model_selection.train_test_split(X, y)
AttributeError: 'module' object has no attribute 'cross_validation'
```

then I changed the code like below,

```
# Split into train and test
#X, Xt, y, yt = sklearn.cross_validation.train_test_split(X, y)
X, Xt, y, yt = sklearn.model_selection.train_test_split(X, y)
```

Noiredd:
Done in #5770.

PR #4805, opened 2016-10-04
Update caffe.cpp

yuanzhihang1:
If there is a GPU in your computer and you enable the CPU_ONLY, you cannot train a net using the build/tools/caffe.bin command becuase the gpus.size() !=0 if you have a GPU and you will run the function Caffe::SetDevice(gpus[0]); then error is reported:
`F1004 13:59:57.352250 4383 common.cpp:66] Cannot use GPU in CPU-only Caffe: check mode.`

### DIFF
--- a/tools/caffe.cpp
+++ b/tools/caffe.cpp
@@ -203,9 +203,9 @@ int train() {
           FLAGS_gpu = "" + boost::lexical_cast<string>(0);
       }
   }
-
   vector<int> gpus;
   get_gpus(&gpus);
+#ifndef CPU_ONLY
   if (gpus.size() == 0) {
     LOG(INFO) << "Use CPU.";
     Caffe::set_mode(Caffe::CPU);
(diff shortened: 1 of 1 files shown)


cypof:
Doesn't it only happen if you specify --gpu on the command line?

yuanzhihang1:
I checked my prototxt and shell script and found It will always happen no matter you specify --gpu or not. Only if you specipy solver_mode: CPU in the solver.prototxt it will work.

cypof:
I will try to look at it in more detail. Would you be OK to search for why does the gpus vector is not empty as it should be, instead of defensively ignoring its content? Also this way of doing it leads to duplicate CPU setup code.

cypof:
If the user specifies solver_mode: GPU, you want Caffe to stop if it cannot satisfy the setting, not silently ignore it and continue. So the current behavior is correct.

cypof:
Ah, but it seems not specifying anything in the solver params also causes the issue, let me see...

cypof:
OK, so we need to test that the setting is given by the user, not the default (GPU), it can be done with the line
```
&& solver_param.has_solver_mode()
```
at caffe.cpp:198, do you want to update your PR?

PR #4806, opened 2016-10-04
Move root_net_ check in net constructor

junshi15:
This PR extends recurrent_layer to the multi-gpu settings.

Currently, the recurrent_layer constructs a net internally (the unrolled net), https://github.com/BVLC/caffe/blob/master/src/caffe/layers/recurrent_layer.cpp#L108.
No root_net_ is set in that constructor.   In a multi-gpu setting, worker solver will create an unrolled net too, however it will fail at the line below, https://github.com/BVLC/caffe/blob/master/src/caffe/net.cpp#L50-L51, because the worker solver is not a root solver and the net does not have a root_net.

Since root_net is only used in share_from_root, this PR moves the check to where share_from_root is defined.

### DIFF
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -47,8 +47,6 @@ Net<Dtype>::Net(const string& param_file, Phase phase,
 
 template <typename Dtype>
 void Net<Dtype>::Init(const NetParameter& in_param) {
-  CHECK(Caffe::root_solver() || root_net_)
-      << "root_net_ needs to be set for all non-root solvers";
   // Set phase from the state.
   phase_ = in_param.state().phase();
   // Filter layers based on their include/exclude rules and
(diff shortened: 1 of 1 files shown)


shelhamer:
Heads-up: the `root_net_` is going away entirely as part of the simplifications in #4563.

junshi15:
That's a large PR, may take a while. It may make sense merging this one first. 

Similar complaints here: https://github.com/BVLC/caffe/issues/4851

shelhamer:
This should be unnecessary with the merge of #4563, but thank you for proposing a fix. @junshi15 and @SIshijima could you check that #4563 in fact fixes the issue?

junshi15:
@shelhamer I won't have time to verify the recurrent layer at this moment as I am quite busy at work. Let me close this PR for now. If we see problems in the future, we will re-visit it.

PR #4813, opened 2016-10-05
Fix: docs/yum_install.md glog broken link

Shamanoid:
fixes the broken glog link in yum_install.md which is currently returning a 404.


shelhamer:
Thanks for checking the links.

PR #4840, opened 2016-10-13
Checks inside Xcode for latest OSX SDK

Zylphrex:
Checks for latest OSX SDK as in Xcode 8, MacOSX10.11.sdk no available by default


soldier828:
oh, I recently updated to osx10.12, which also had this problem

PR #4842, opened 2016-10-14
pytest fix: Files created with NamedTemporary files cannot be opened on Windows

willyd:
As the title says. This will fix the python tests when master is merged back in the windows branch. It should not impact the UNIX builds.

### DIFF
--- a/python/caffe/test/test_net.py
+++ b/python/caffe/test/test_net.py
@@ -173,12 +173,12 @@ layer {
 """
 
     def setUp(self):
-        self.f = tempfile.NamedTemporaryFile(mode='w+')
+        self.f = tempfile.NamedTemporaryFile(mode='w+', delete=False)
         self.f.write(self.TEST_NET)
-        self.f.flush()
+        self.f.close()
 
     def tearDown(self):
-        self.f.close()
+        os.remove(self.f.name)
 
     def check_net(self, net, blobs):
         net_blobs = [b for b in net.blobs.keys() if 'data' not in b]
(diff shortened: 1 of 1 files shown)


lukeyeager:
Won't this leave a bunch of temporary files lying around?

willyd:
@lukeyeager Yes you are right. It was late when I submitted this. I did not noticed that the other test were actually removing the files by hand. I actually don't understand why the file is removed in the first place, it should be remove only when file.close() is called not when file.flush() is. I will do some more investigation.

willyd:
I reviewed the python docs and I think that the answer is here: https://docs.python.org/2/library/tempfile.html.

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later)

A simple test at reopening the file created with `NamedTemporaryFile` on Windows fails with `IOError: [Errno 13] Permission denied:`. 

@lukeyeager Do you think it is acceptable (if it works) to keep the `delete=False`, replace the `self.f.flush()` with `self.f.close()` and put and `os.remove(f.name)` in the `tearDown` method?

lukeyeager:
Well I can't speak for what the @BVLC team will want, but that's pretty much what we do in DIGITS.

willyd:
Thanks for your feedback. I will update the PR as described and see what the core devs think of this.

willyd:
@shelhamer If you have some time to check this out, I will soon merge the CMake build into the windows branch and after merge the latest master into it. All tests pass on appveyor except for the python tests which require this change.

willyd:
Bump. Anyone from @BVLC can review this? This will be one less change between UNIX and Windows.

shelhamer:
Thanks for the compatibility fix @willyd !

PR #4849, opened 2016-10-17
Add support for vs2015

EnjoyLeole:
most dependencies are base on vc120(vs2013), while some of them doesn't have an vs2015 compatible version at this moment.


willyd:
@EnjoyLeole Thanks for submitting this. VS 2015 is planned with the CMake build. There seems to be a lot more than just VS 2015 support in this PR though. If you really want VS 2015 support with the current VS build please cherry-pick only the needed changes. 

> most dependencies are base on vc120(vs2013), while some of them doesn't have an vs2015 compatible version at this moment.

Do you mean that not all nuget packages were updated with VS 2015 versions?

EnjoyLeole:
@willyd  Yes, several pkgs  don't have a version for vs2015, LevelDB\boost-python2.7,etc.

willyd:
@EnjoyLeole You can find a working (at least for CPU and Python) branch based on the current windows CMake build that works with VS 2015 at https://github.com/willyd/caffe/tree/vs2015.

willyd:
VS 2015 support was just merged in the windows branch. The CMake build must be used.

PR #4892, opened 2016-10-25
Fix wrong conditional after merge branch 'master'

GlueCrow:


willyd:
I have disabled the checks on appveyor for the opencl branch for now since they fail because of too many merge conflicts. @naibaf7 the CMake build on the windows branch is getting pretty stable. If you want my help in any way to get the opencl branch working on windows let me know.

naibaf7:
@willyd Sure, just link me to what I need to know to do it!

willyd:
@naibaf7 I haven't had time to check how different is the OpenCL branch from master but the windows branch is up-to-date with master. The main differences are changes made to the CMake build system and a few changes related to MSVC (`#if defined _MSC_VER`) here and there to support differences between MSVC and gcc or clang.

If the opencl is up-to-date with master it might just boil down to merging the two branches together and solving any conflicts, which will most probably be mainly in the CMake files. As far as the OpenCL code and changes are concerned I have no experience with OpenCL so I can't tell if there will be changes required to the code itself to support windows. The current conflicts can be seen [here](https://ci.appveyor.com/project/BVLC/caffe/build/1.0.4).

Are the OpenCL conv layers much faster on Intel CPUs (including Atom) when compared to standard caffe with OpenBLAS or MKL?

naibaf7:
@willyd 
Not on CPUs, but on the onboard GPUs of those CPUs; which often can beat the CPU part of the chip!
OpenCL is up-to-date with master, but I had to change the CMake scripts to accomodate detecting ViennaCL, OpenCL and clBLAS headers.

willyd:
@naibaf7 Ok. Thanks for the clarification. 

> I'll try if I can merge on Friday/Saturday.

Great! If you have questions don't hesitate to ask.

PR #4909, opened 2016-10-27
Update install_apt.html

hmybmny:
The requested url https://google-glog.googlecode.com/files/glog-0.3.3.tar.gz was not found and the download link is https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/google-glog/glog-0.3.3.tar.gz


williford:
I wonder if the URL should be:
https://github.com/google/glog/archive/v0.3.3.tar.gz ?

Issue #4886, opened 2016-10-24
Add MATLAB support with the CMake build on windows


willyd:
Current CMakeLists.txt disables the MATLAB support. Need to change:

```
caffe_option(BUILD_matlab "Build Matlab wrapper" OFF IF UNIX OR APPLE OR WIN32)
```

FindMatlabMex.cmake fails to find mex.bat on Windows. CMake added the matlab_add_mex in the [FindMatlab.cmake](https://cmake.org/cmake/help/v3.6/module/FindMatlab.html) maybe we can use this Find module instead on windows?

lunzueta:
Hi @willyd. I already have experience in compiling matcaffe in previous Caffe versions for Windows. I'm currently trying to solve this issue with CMake in my local repo. What I'm trying to do is a custom finder and linker of the required Matlab files (I don't think that using FindMatlab.cmake is a good idea in this case, which I've also tried). Hopefully, during this week I'll be able to give you an update on this (I mean, a PR).

willyd:
@lunzueta In one of my previous PRs I actually had MATLAB working using the FindMatlab.cmake module you can check it out there: https://github.com/willyd/caffe/tree/msvc

I believe all the MATLAB related changes were done in this commit: https://github.com/willyd/caffe/tree/cbbd3a7d00d04bc7a536a969d9e1c3730f4c3fe6.

The most relevant files I believe were:

- https://github.com/willyd/caffe/blob/msvc/matlab/CMakeLists.txt
- https://github.com/willyd/caffe/blob/msvc/cmake/Dependencies.cmake
- https://github.com/willyd/caffe/blob/msvc/cmake/Modules/FindMatlab.cmake

> Hopefully, during this week I'll be able to give you an update on this (I mean, a PR).

I will wait for your feedback then.

lunzueta:
OK @willyd, I didn't try that one, so I'll compare and test both approaches and if in the end what I'm trying to do has something that can benefit the community, I'll send the PR, otherwise, it wouldn't make too much sense to do so :-P

lunzueta:
@willyd, your approach is clearer and simpler than what I was trying to do, so I'd definitively go for it. I've already submitted a PR (https://github.com/BVLC/caffe/pull/4928) where I handpicked some lines from some cmake files of your msvc branch. I also updated the README.md file. I was able to run `classification_demo.m` correctly in Matlab 2014b with the generated `.mexw64` file.

willyd:
[...]

PR #4928, opened 2016-10-30
Update CMake files to build Matcaffe on Windows

lunzueta:
This PR is related with this issue: https://github.com/BVLC/caffe/issues/4886


willyd:
I believe we should required CMake 3.3 since the matlab_add_mex was added in this version. However, the C++ build itself requires 3.4 so I guess a comment would be sufficient. 

We could also use matlab_add_unit_test to add the tests.

willyd:
> @willyd you mean to add a comment in the Matlab section of README.md? But, if in the Windows Setup section there is already a comment saying that CMake 3.4+ is required, I'm not sure if it makes sense to emphasize that CMake 3.3+ is required because of matlab_add_mex. Do you think it is really necessary to add this comment? or it might add some confusion about the minimum required CMake version? (I think this could lead to confusion)

You are right I don't change anything.

> Regarding matlab_add_unit_test, I agree that it could be useful for running the run_tests.m script included in the MATLAB package, but this should be considered for the non-Windows versions of Matcaffe as well. At this moment I can't test the non-Windows Matcaffe versions, so that's why I maintained the non-Windows CMake commands for Matcaffe untouched. Maybe the possible inclusion of this CMake command should be considered separately from this PR? 

Let's do this in a second pass as you suggested.

willyd:
Closing in favor of #5180. Sorry for the delay @lunzueta.

PR #4938, opened 2016-11-01
Domain confusion layers for the ICCV2015 paper "Simultaneous Deep

mahfuj1093:
Transfer Across Domains and Tasks"


shelhamer:
Most papers keep their own fork and include a submodule of Caffe. Closing since these layers are particular to the domain confusion paper.

PR #4948, opened 2016-11-03
add instance norm as an option of BatchNormLayer

lzx1413:
instance normalization has a good effect in [Instance Normalization: The Missing Ingredient for Fast Stylization](https://arxiv.org/abs/1607.08022),so I add it as an option of BatchNormLayer,and the option is false as default

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -509,6 +509,9 @@ message BatchNormParameter {
   // Small value to add to the variance estimate so that we don't divide by
   // zero.
   optional float eps = 3 [default = 1e-5];
+  //if false,mean and variance will be calculated across instances. If true ,mean and variance will be 
+  //calculated in single instance
+  optional bool use_instance_norm = 4 [default = false];
 }
 
 message BiasParameter {
--- a/include/caffe/layers/batch_norm_layer.hpp
+++ b/include/caffe/layers/batch_norm_layer.hpp
@@ -62,8 +62,11 @@ class BatchNormLayer : public Layer<Dtype> {
 
   Blob<Dtype> mean_, variance_, temp_, x_norm_;
   bool use_global_stats_;
+  bool use_instance_norm_;
   Dtype moving_average_fraction_;
   int channels_;
+  int pre_batch_size_;
+  int batch_size_;
   Dtype eps_;
 
   // extra temporarary variables is used to carry out sums/broadcasting
--- a/src/caffe/layers/batch_norm_layer.cpp
+++ b/src/caffe/layers/batch_norm_layer.cpp
@@ -11,6 +11,7 @@ void BatchNormLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
   BatchNormParameter param = this->layer_param_.batch_norm_param();
   moving_average_fraction_ = param.moving_average_fraction();
+  use_instance_norm_ = param.use_instance_norm();
   use_global_stats_ = this->phase_ == TEST;
   if (param.has_use_global_stats())
     use_global_stats_ = param.use_global_stats();
@@ -18,13 +19,18 @@ void BatchNormLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
     channels_ = 1;
   else
     channels_ = bottom[0]->shape(1);
+  batch_size_ = bottom[0]->shape(0);
   eps_ = param.eps();
   if (this->blobs_.size() > 0) {
     LOG(INFO) << "Skipping parameter initialization";
+    pre_batch_size_ = this->blobs_[0]->shape(0);
   } else {
     this->blobs_.resize(3);
     vector<int> sz;
-    sz.push_back(channels_);
+    if (use_instance_norm_)
+        sz.push_back(channels_*batch_size_);
+    else
+        sz.push_back(channels_);
     this->blobs_[0].reset(new Blob<Dtype>(sz));
     this->blobs_[1].reset(new Blob<Dtype>(sz));
     sz[0] = 1;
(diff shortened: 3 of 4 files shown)


x10000year:
I have experimented this implementation in my own project, and I found something strange: even if I always use minibatch size 1 and set use_global_stats=false in both training and testing phase, the model trained with instance norm always outperform the one trained with batch norm. I have repeated the experiment several times. Aren't the implementation for instance norm and batch norm equivalent when minibatch size is 1 and use_global_stats is set to false?

src/caffe/proto/caffe.proto
```diff
@@ -306,7 +306,7 @@ message ParamSpec {
 // NOTE
 // Update the next available ID when you add a new LayerParameter field.
 //
-// LayerParameter next available layer-specific ID: 147 (last added: recurrent_param)
+// LayerParameter next available layer-specific ID: 148 (last added: recurrent_param)
```
williford: Was this modified by accident?

williford:
The commits would need to be squashed. See: https://github.com/BVLC/caffe/blob/master/CONTRIBUTING.md

Would it be better for InstanceNorm and BatchNorm to be separate layers?

Is it possible to do instance normalization with existing layers, for example the ReductionLayer using mean and then dividing the original values with the mean?

src/caffe/proto/caffe.proto
lzx1413: I do't think too much about that.If I can't modified it,you can rewrite it.sorry

lzx1413:
@x10000year they should be same while the bathsize is one,I will test it again.
@williford If it is better to be separate,I'd like to  make a new version and thanks for your opinion

williford:
Does the MVN (Mean Variance Normalization) layer do the same thing? It looks like it can perform channel-wise normalization, although it lacks much documentation and seems to silently fail if the blobs are not at least 4D.

http://caffe.berkeleyvision.org/doxygen/classcaffe_1_1MVNLayer.html
https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto#L869 (MVNParameter)

lzx1413:
@williford yes，MVNLayer can also do this,but it lacks some parameters such as average_mean.Besides it can't handle the condition where the batchsize changes in the same model.So should I fix that layer or continue refining this one?

williford:
@lzx1413 Can you elaborate on lacking `average_mean` and not being able to handle changes in batch size in the same model?

I don't see any reference to `average_mean`, in your changes or in caffe.proto. According to Caffe's philosophy, the normalization should set the mean to 0 and then the Bias layer should be used to make the mean something else.

My opinion doesn't hold much (any) weight - so be cautious on implementing any of my suggestions :-). I do think the commits would need to be cleaned up (squashed), which is fairly easy to do. Also it generally takes a long time for PRs to be requested.

lzx1413:
@williford while training the model, I want to save the moving_average_mean and moving_average_var just like BNLayer
```
// compute and save moving average 
this->blobs_[2]->mutable_cpu_data()[0] *= moving_average_fraction_; 
this->blobs_[2]->mutable_cpu_data()[0] += 1; 
caffe_cpu_axpby(mean_.count(), Dtype(1), mean_.cpu_data(), 
moving_average_fraction_, this->blobs_[0]->mutable_cpu_data()); 
int m = 0; 
if (use_instance_norm_) 
    bottom[0]->count()/(channels_*batch_size_);
else 
    bottom[0]->count()/(channels_); 
type bias_correction_factor = m > 1 ? Dtype(m)/(m-1) : 1; 
caffe_cpu_axpby(variance_.count(), bias_correction_factor, 
variance_.cpu_data(), moving_average_fraction_, 
this->blobs_[1]->mutable_cpu_data()); 
```
so we can use the FLAG `use_global_stats` to use them in test process later. but rather than the size of BNLayer's mean is the number of channels,  the size of instance norm is batchsize*channels, So if the batch size change, we should reshape the origin moving_average_mean to new shape,just like that:
```
if (pre_batch_size_ != num && use_instance_norm_) {

        vector<int> sz;

        sz.push_back(num*channels_);

        vector<int> tmp_size;

        tmp_size.push_back(channels_);

        Blob<Dtype> tmp_blob;

        tmp_blob.Reshape(tmp_size);

        caffe_cpu_gemv<Dtype>(CblasTrans, pre_batch_size_, channels_, 1.,

            this->blobs_[0]->cpu_data(), batch_sum_multiplier_.cpu_data(), 0.,

            tmp_blob.mutable_cpu_data());

        this->blobs_[0]->Reshape(sz);

        caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, num, channels_, 1, 1,

            batch_sum_multiplier_.cpu_data(), tmp_blob.cpu_data(), 0.,

            this->blobs_[0]->mutable_cpu_data());



        caffe_cpu_gemv<Dtype>(CblasTrans, pre_batch_size_, channels_, 1.,

            this->blobs_[1]->cpu_data(), batch_sum_multiplier_.cpu_data(), 0.,

            tmp_blob.mutable_cpu_data());

        this->blobs_[1]->Reshape(sz);

        caffe_cpu_gemm<Dtype>(CblasNoTrans, CblasNoTrans, num, channels_, 1, 1,

            batch_sum_multiplier_.cpu_data(), tmp_blob.cpu_data(), 0.,

            this->blobs_[1]->mutable_cpu_data());

        mean_.Reshape(sz);

        variance_.Reshape(sz);

        pre_batch_size_ = num;

    }
```

williford:
@lzx1413 Thanks for the clarification that you meant `moving_average_mean`.

Since the size of the moving average mean and variance is now batchsize*channels, does that mean there is a separate running mean and variance for each example in the batch? For example, the first example of each batch would have a separate mean than the second example of each batch? If so, that seems a bit strange?

If you use instance normalization during training, it would make sense to me to use instance normalization during testing. I don't think it makes sense to directly switch from instance normalization to "global" normalization, since I think of the operation of instance normalization fundamentally different than global normalization, whereas batch normalization approximates global normalization. It might be better to switch from instance normalization to training with mini-batch normalization and then using global normalization. Does this make sense?

lzx1413:
@williford I implete the functions based on [this one in Torch](https://github.com/DmitryUlyanov/texture_nets/blob/master/InstanceNormalization.lua#L15),and I don't know the MVNLayer before,So if the running mean is useless in instance normlization,MVNLayer can do the same work,I'll cancel this PR.thank for reviewing it.

williford:
@lzx1413 Thanks for answering my questions. Yes, the documentation is lacking and it can be hard not to duplicate effort. If the original paper or another paper indicate that tracking the moving average mean is useful, it might be of interest to pursue this more.

PR #4954, opened 2016-11-06
fix build on Jetson TK1

howard0su:
Jetson TK1 only support cuDNN v2. Skip cuDNN support but continue building with cuda.

### DIFF
--- a/cmake/Cuda.cmake
+++ b/cmake/Cuda.cmake
@@ -216,12 +216,13 @@ function(detect_cuDNN)
 
     string(COMPARE LESS "${CUDNN_VERSION_MAJOR}" 3 cuDNNVersionIncompatible)
     if(cuDNNVersionIncompatible)
-      message(FATAL_ERROR "cuDNN version >3 is required.")
+      message(STATUS "cuDNN version >3 is supported. Build without cuDNN.")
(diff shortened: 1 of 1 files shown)


naibaf7:
@shelhamer 
@howard0su 
wouldn't it be better to introduce a cuDNNv2 convolution layer again, because I imagine Jetson TK1 performance with native convolutions ought to be terrible.

howard0su:
TK1 only has cuda 6.5 and cuDNN R2 since it is 32bit platform. I just reimage my board with latest jetpack. end up with this.

support cudnn r2 is great and I think we don't need too much code change.

howard0su:
https://developer.nvidia.com/embedded/jetpack-notes

there is official doc on this. so what's the call here?

shelhamer:
Hmm, this is awkward. So far we have followed cuDNN releases in lockstep and deprecated past versions as to not leave a trail of legacy code.

@thatguymike is there an NVcaffe or the like targeted for TK1 compatibility that we could point users too instead?

mkaraki48:
@naibaf7 i have a TK1 as well. They literally dropped 32Bit support from all their libraries starting CUDA 7.0 and cuDNN 3.0 the moment the 64Bit TX1 came out... I didn't have time to port a new Caffe master to work with cuDNN 2.0, but if u did, please let me know.

PR #4955, opened 2016-11-06
Clarified CUDNN install instructions

elhigu:
Fixed `CUDNN_ROOT` to be written correctly.
Added missing pieces to the last cmake install command
Added an example showing which part of the unpacked cudnn.zip CUDNN_ROOT variable should point

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,16 +58,16 @@ In case one of the steps in the above procedure is not working please refer to t
 
 ### Use cuDNN
 
-To use cuDNN you need to define the CUDNN_ROOT cache variable to point to where you unpacked the cuDNN files. For example, the build command above would become:
+To use cuDNN you need to define the CUDNN_ROOT cache variable to point to where you unpacked the cuDNN files e.g. `C:/Users/myuser/Projects/machine-learning/cudnn-8.0-windows10-x64-v5.1/cuda`. For example, the build command above would become:
 
 ```
 > set CMAKE_GENERATOR=Ninja
 > set CMAKE_CONFIGURATION=Release
 > mkdir build
 > cd build
-> cmake -G%CMAKE_GENERATOR% -DBLAS=Open -DCMAKE_BUILD_TYPE=%CMAKE_CONFIGURATION% -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=<install_path> -DCUDNNROOT=<path_to_cudnn> -C %CAFFE_DEPENDENCIES%\caffe-builder-config.cmake  ..\
+> cmake -G%CMAKE_GENERATOR% -DBLAS=Open -DCMAKE_BUILD_TYPE=%CMAKE_CONFIGURATION% -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=<install_path> -DCUDNN_ROOT=C:/Users/myuser/Projects/machine-learning/cudnn-8.0-windows10-x64-v5.1/cuda -C %CAFFE_DEPENDENCIES%\caffe-builder-config.cmake  ..\
 > cmake --build . --config %CMAKE_CONFIGURATION%
-> cmake
+> cmake --build . --config %CMAKE_CONFIGURATION% --target install
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,16 +58,16 @@ In case one of the steps in the above procedure is not working please refer to t
...
-To use cuDNN you need to define the CUDNN_ROOT cache variable to point to where you unpacked the cuDNN files. For example, the build command above would become:
+To use cuDNN you need to define the CUDNN_ROOT cache variable to point to where you unpacked the cuDNN files e.g. `C:\Users\myuser\Projects\machine-learning\cudnn-8.0-windows10-x64-v5.1\cuda`. For example, the build command above would become:
 
 ```
 > set CMAKE_GENERATOR=Ninja
 > set CMAKE_CONFIGURATION=Release
 > mkdir build
 > cd build
-> cmake -G%CMAKE_GENERATOR% -DBLAS=Open -DCMAKE_BUILD_TYPE=%CMAKE_CONFIGURATION% -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=<install_path> -DCUDNNROOT=<path_to_cudnn> -C %CAFFE_DEPENDENCIES%\caffe-builder-config.cmake  ..\
+> cmake -G%CMAKE_GENERATOR% -DBLAS=Open -DCMAKE_BUILD_TYPE=%CMAKE_CONFIGURATION% -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=<install_path> -DCUDNN_ROOT=<path_to_cudnn> -C %CAFFE_DEPENDENCIES%\caffe-builder-config.cmake  ..\
```
willyd: Why not make it `-DCUDNN_ROOT=C:/Users/myuser/Projects/machine-learning/cudnn-8.0-windows10-x64-v5.1/cuda`

willyd:
@elhigu Can you apply the requested change then I will merge the change. Thanks.

elhigu:
@willyd thanks for the review. I changed the example path to be directly to be part of cmake command.

willyd:
@elhigu Unless I am mistaken it should be  `-DCUDNN_ROOT=C:/Users/myuser/Projects/machine-learning/cudnn-8.0-windows10-x64-v5.1/cuda` with forward slashes and not backslashes. Can you update and squash all the commits together and force push?

elhigu:
@willyd I hadn't noticed those slashes at all, I have no idea how I got path that way, since I used "copy addess" feature from file explorer.

IIRC Cmake works with both slashes in most of cases, but I'll try it out in a moment with my windows machine (just to make it sure that it still works with backslashes) and fix this if it still works. 

Btw. github has an option to select squash merge instead of normal when merging pull requests, but I can do squashing too of course.

github-merge-options

elhigu:
Looks like I indeed had compiled it already using path with backslashes. No idea where I copied the path to docs... fixing..

willyd:
> IIRC Cmake works with both slashes in most of cases, but I'll try it out in a moment with my windows machine (just to make it sure that it still works with backslashes) and fix this if it still works.

My experience is somewhat different but maybe this has improved in newer CMake version or the errors where on my part and not CMake's. I guess that if you specify the option like `CUDNN_ROOT:PATH=` now CMake has everything it needs to handle the slash or backslash correctly.

> Btw. github has an option to select squash merge instead of normal when merging pull requests, but I can do squashing too of course.

Thanks I did not know that.

elhigu:
Now I think that everything should be in order, I also tested that forward slashes did compile ok.

willyd:
@elhigu Great. One last thing. In the first version you had something that looked like:

You downloaded and extracted cuDNN to `C:\Users\myuser\Projects\machine-learning\cudnn-8.0-windows10-x64-v5.1\cuda`

could you add that back? Since it might confuse some people that will now need to figure out where this path comes from. 

Also please replace `elhigu` with `myuser` like before.

Sorry for being so picky...

After that I will wait for CI to pass and merge. Thanks.

elhigu:
@willyd sure will do and no worries (sorry for being sloppy with fixes), I know its not fun either to have to ask making changes to PR many times until it gets right :D

elhigu:
Now everything should be ok.

willyd:
Thanks.

PR #4958, opened 2016-11-08
Opencl - add padding support to Intel spatial convolution kernel

gongzg:
This patchset is to eliminate the seprate image padding for Intel's direct spatial convolution kernels. Also clean up the corresponding kernel code.


gongzg:

PR #4969, opened 2016-11-11
Put quotes around titles in YAML front matter.

williford:
When generating the documents, the colon produces errors unless the title is in quotes. This commit just puts the titles of the installation pages in quotes.

This commit makes the HTML title appear for the corresponding 3 pages. You can see that it is currently missing:
http://caffe.berkeleyvision.org/install_apt.html

See https://github.com/jekyll/jekyll/issues/549

### DIFF
--- a/docs/install_apt.md
+++ b/docs/install_apt.md
@@ -1,5 +1,5 @@
 ---
-title: Installation: Ubuntu
+title: "Installation: Ubuntu"
 ---
 
 # Ubuntu Installation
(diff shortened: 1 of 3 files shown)


williford:
The only difference this commit makes is that it sets the HTML title for the 3 installation pa.

And perhaps more importantly, no annoying errors :-).

williford:
Closing to create PR from a different branch.

PR #4971, opened 2016-11-11
Put quotes around titles in YAML front matter of installation docs to remove errors

williford:
When generating the installation docs, the colon in the titles of the YAML front matter produces errors unless the title is in quotes. This commit just puts the titles of the installation pages in quotes.

This commit makes the HTML title appear for the corresponding 3 pages. You can see that it is currently missing:
http://caffe.berkeleyvision.org/install_apt.html

See jekyll/jekyll#549

### DIFF
--- a/docs/install_apt.md
+++ b/docs/install_apt.md
@@ -1,5 +1,5 @@
 ---
-title: Installation: Ubuntu
+title: "Installation: Ubuntu"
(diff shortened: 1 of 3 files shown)


shelhamer:
Thanks for catching the formatting error.

PR #4979, opened 2016-11-14
Support solver resumes in parse_log.py

davidbrai:
Currently parse_log.py skips all non timestamped lines
only once. When resuming a solver and appending to the
same log file, it creates more non timestamped log lines.
This change allows the script to silently skip those lines.

### DIFF
--- a/tools/extra/parse_log.py
+++ b/tools/extra/parse_log.py
@@ -48,8 +48,13 @@ def parse_log(path_to_log):
                 # iteration
                 continue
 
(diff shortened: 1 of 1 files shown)


shelhamer:
Seems a little risky in case formatting breaks in the future, but I suppose that will be obvious enough when there aren't enough iterations in the plot. Thanks.

davidbrai:
@shelhamer thanks for merging!

Issue #2353, opened 2015-04-23
undefined reference to `cblas_*"


wbtxd2004:
When I make all to compile caffe, I got this error:
.build_debug/lib/libcaffe.so: undefined reference to `cblas_dasum'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_dcopy'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_sdot'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_dscal'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_sgemm'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_dgemm'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_sgemv'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_sscal'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_scopy'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_saxpy'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_dgemv'
.build_debug/lib/libcaffe.so: undefined reference to `cblas_sasum'

I use mkl as BLAS. How to solve this problem? 
Thanks!

StevenLOL:
Do you have a build folder, once I got same problem when I  build caffe in its own directory.


mkdir build
cd build
cmake ..
make

wbtxd2004:
Thank you! I use your method and successful installed caffe.

wincle:
I also have this problem, check your Makefile.config , Is there leave one space after BLAS := open  ?delete the space after "open",then recompile it.

vishwa15:
Thank you Wincle. It worked!!

noldorj:
Hi. Were you able to solve this issue with Intel MKL ? I'm facing the same error.

northeastsquare:
@noldorj , iset mkl not MKL, solve
BLAS := mkl
# Custom (MKL/ATLAS/OpenBLAS) include and lib directories.
# Leave commented to accept the defaults for your choice of BLAS
# (which should work)!
BLAS_INCLUDE := /opt/intel/mkl/include/
BLAS_LIB := /opt/intel/mkl/lib/intel64/

srbnghosh99:
I am finding this error message when I am trying to make. 


CXX/LD -o .build_release/test/test_all.testbin src/caffe/test/test_caffe_main.cpp
[...]

PR #4981, opened 2016-11-15
add note about  BLAS

qiaohaijun:
https://github.com/BVLC/caffe/issues/2353


shelhamer:
Thanks for highlighting the sensitivity to spaces, but this is true of all values in make and not specific to just this variable. I'd rather rely on general make knowledge and our example not including trailing spaces.

PR #4983, opened 2016-11-15
Cosine Loss Layer

ssudholt:
This PR implements the Cosine Loss Layer as used in "Information-theoretical label embeddings for large-scale image classification".


shaibagon:
Thank you for this contribution!

1. Is there a particular reason you did not add a testing for this layer? 
2. Can you please add a link to the paper describing the loss you implemented?
3. Can you please add documentation for the loss layer you added? Please see other layers and the documentation/comments they have in their hpp files. See, for example, [`euclidean_loss_layer`](https://github.com/BVLC/caffe/blob/master/include/caffe/layers/euclidean_loss_layer.hpp).
4. You also added functions to math functions. Can you please add comments describing them?
4. Please consider squashing all the commits into a single one.

Thanks!

ssudholt:
I added a set of unit tests including the gradient checker. The gradient checker fails for all data types. However, I witnessed that the computed and estimated gradient are offset by a factor of 512. I'm not too sure though that I'm using the gradient checker tool correctly.
In practice, this constant factor can simply be dealt with by adapting the learning rate. I'm already using the Cosine Loss for a couple of experiments.

I'll get to the documentation part soon. I'll squash everything once the gradient checking problem is solved.

src/caffe/layers/cosine_loss_layer.cpp
```diff
@@ -0,0 +1,90 @@
...
+    len_label_data[i] = caffe_cpu_nrm2(dim, label_data + (i*dim));
+    dots_data[i] = caffe_cpu_dot(dim, inp_data + (i*dim),
+                                 label_data + (i*dim));
+  }
+
+  Dtype loss(0.0);
+  for (int i = 0; i < num; ++i) {
+    loss += 1 - dots_data[i] / (len_inp_data[i]*len_label_data[i]);
+  }
+  top[0]->mutable_cpu_data()[0] = loss / num;
```
shaibagon: You scale your loss here by `num`. Where is this scale factor in backward computation? Could this be the reason why your gradient check fails?

ssudholt: The scale factor is extracted in line 65 and then applied to the gradient in lines 77 and 78. I'm a bit confused also as changing the factor factor manually in the code has no effect on the test what so ever. All numbers stay exactly the same even if I set the scale to 512 by hand.

src/caffe/layers/cosine_loss_layer.cpp
```diff
@@ -0,0 +1,90 @@
...
+    int dim = bottom[0]->count()/bottom[0]->num();
+    const Dtype scale = top[0]->cpu_diff()[0];
+
+    // set bottom to negative label
+    caffe_cpu_axpby(bottom[0]->count(),
+                    Dtype(-1), label_data,
+                    Dtype(0), bottom_diff);
+    for (int i = 0; i < num; ++i) {
+        Dtype s = Dtype(1) / (len_inp_data[i] * len_label_data[i]);
+        Dtype c = dots_.cpu_data()[i] / (len_inp_data[i] * len_inp_data[i]);
```
shaibagon: you are calling `cpu_data()` in a loop. why not `const Dtype* dots_data = dots_.cpu_data();` before the loop?

src/caffe/layers/cosine_loss_layer.cpp
shaibagon: the scale factor in line 65 is the top gradient, not the size of the batch. You are scaling the loss by the batch size `num` in the forward pass, but there is no equivalent scale in the backward pass. In the test, you are using batch size of `512` and this is the error you observe. Try changing the test batch size and see if this affect the gradient error.

src/caffe/layers/cosine_loss_layer.cpp
ssudholt: Good point, will change it

shaibagon:
Some more comments:
1. You are using [deprecated][1] `blob` method `num()` in your code, please use `shape(0)` instead.
2. Your code assume the input blob is 2D: you have `num()` vectors with `count(1)` dim each. What if someone wants to compute the loss in a "pixelwise" fashion? That is, I have 4D input blob, with `num()x height()x width()` independent vectors of dim `channels()`?  
Consider `"SoftmaxWithLoss"` layer for example, it has `axis` parameter that allows for such functionality.

 [1]: https://github.com/BVLC/caffe/blob/master/include/caffe/blob.hpp#L131-L132
 [2]: https://github.com/BVLC/caffe/blob/master/include/caffe/blob.hpp#L71

src/caffe/layers/cosine_loss_layer.cpp
ssudholt: That's what I'm saying, I changed line 65 to `const Dtype scale = Dtype(1.0) / num;` just to see what happens and the results are exactly the same for the computed gradients in the gradient checker.

src/caffe/layers/cosine_loss_layer.cpp
```diff
@@ -0,0 +1,90 @@
...
+  }
+  top[0]->mutable_cpu_data()[0] = loss / num;
+}
+
+template <typename Dtype>
+void CosineLossLayer<Dtype>::Backward_cpu(
+    const vector<Blob<Dtype>*>& top, const vector<bool>& propagate_down,
+    const vector<Blob<Dtype>*>& bottom) {
+  if (propagate_down[1]) {
+    LOG(FATAL) << this->type()
```
shaibagon: why don't you allow to propagate down to labels? this loss is symmetric. what if I want to use it in a "Siamese" network to compare two inputs? You should propagate to both inputs unless `propagate_down[i]` is false.

src/caffe/layers/cosine_loss_layer.cpp
shaibagon: and what happens if you change the [batch size][1] in the test?

BTW, why are you testing on 2D vectors? I suggest make the batch size smaller and the dimension larger for test. 
 [1]: https://github.com/BVLC/caffe/pull/4983/files#diff-aa111b9ec2b96938b1a4f6fccc3aa005R23

ssudholt: I was finally able to figure out the error (my build system was obviously not configured correctly). You were right, the error came from the absence of scaling during backward. I'm correcting the unit tests and will push everything once I'm done.

src/caffe/layers/cosine_loss_layer.cpp
ssudholt: It'd be better to use the [CosineEmbeddingLoss](https://github.com/torch/nn/blob/master/doc/criterion.md#nn.CosineEmbeddingCriterion) for stability for Siamese networks. This is simply a safety to ensure this.

shaibagon:
Adding a link to the paper describing this loss:
[François Chollet, **Information-theoretical label embeddings for large-scale image classification**, arXiv 2016][1].

 [1]: https://arxiv.org/abs/1607.05691

ssudholt:
@shaibagon thanks for all your help, reviewing and suggestions

shaibagon:
@ssudholt you are most welcome. Thank you for making this contribution. I hope it will not take forever for this PR to be accepted and merged.

src/caffe/layers/cosine_loss_layer.cpp
shaibagon: Indeed, the `CosineEmbeddingLoss` seems more robust. Nevertheless, since this layer is symmetric, you should allow gradients to be backprop to both inputs. See, e.g., [`"EuclideanLoss"`][1] layer: the layer is symmetric w.r.t its inputs and thus it allows for backprop to **both** `bottom`s according to `propagate_down` indicator.

 [1]: https://github.com/BVLC/caffe/blob/master/src/caffe/layers/euclidean_loss_layer.cpp#L35

ssudholt:
@shaibagon Ok, I addressed all of your points and squashed everything into one commit. From my side, this PR is ready for review

PR #4998, opened 2016-11-18
Add missing spaces besides equal signs in batch_norm_layer.cpp

chenzeyuczy:
Though not affect the function of codes, it makes sense to improve the readability of codes.


jeffdonahue:
Thanks for the cleanup!

PR #4999, opened 2016-11-18
Add doxygen comments for the batchnorm params.

williford:
I'm adding documentation for the BatchNormLayer parameters (use_global_stats, moving_average_fraction, eps).


williford:
I will reopen when I have updated the caffe.proto also.

PR #5009, opened 2016-11-21
Solver: check and set type to reconcile class and proto type

shelhamer:
The solver checks its proto type (SolverParameter.type) on instantiation:

- if the proto type is unspecified it's set according to the class type `Solver::type()` 
- if the proto type and class type conflict, the solver dies loudly

This helps avoid accidental instantiation of a different solver type than intended when the solver def and class differ. 
For instance if the solver def specifies `type: "Nesterov"` but one instantiates an `SGDSolver` with that def which type the solver is could be a surprise (it's an `SGDSolver`).

Guaranteed type information in the SolverParameter will simplify multi-solver coordination too; in particular it's needed for correctness of #4563. 

@cypof


shelhamer:
Merging—CI failed due to transient networking issue.

PR #5014, opened 2016-11-22
Logging from python, e.g. for lower log level on multi-GPU workers

cypof:


cypof:
I will just do another commit in the NCCL pr instead.

shelhamer:
👍

wadefelix:
init_log should be called in the python/caffe/__init__.py

PR #5022, opened 2016-11-24
Fixed Point Caffe

JiannhuiWang:
The project train fixed point parameters for Caffe


Noiredd:
My opinion (#6052):  
>I don't think it is a good idea to merge any of the specialized forks back into Caffe core - both for the Caffe team (as it shifts the burden of maintaining this code onto us) and the forks' developers (as their code now exists in two places).

PR #5025, opened 2016-11-24
fix error link

hmybmny:
fix error link of glog


shelhamer:
Thanks for updating the link

PR #5035, opened 2016-11-27
fix many typos by using codespell

CDLuminate:
Thanks to https://github.com/lucasdemarchi/codespell/

PR #5038, opened 2016-11-28
Make lint happy (> 80 characters)

CDLuminate:
To fix https://github.com/BVLC/caffe/pull/5035#pullrequestreview-10242924

### DIFF
--- a/src/caffe/layers/crop_layer.cu
+++ b/src/caffe/layers/crop_layer.cu
@@ -39,10 +39,10 @@ void CropLayer<Dtype>::crop_copy_gpu(const vector<Blob<Dtype>*>& bottom,
                 src_data, dest_data, is_forward);
     }
   } else {
-    // We are at the last two dimensions, which are stored continuously in memory
-    // With (N,C,H,W)
-    //      (0,1,2,3) cur_dim   -> H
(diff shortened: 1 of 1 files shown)


src/caffe/layers/crop_layer.cu
```diff
@@ -39,10 +39,10 @@ void CropLayer<Dtype>::crop_copy_gpu(const vector<Blob<Dtype>*>& bottom,
                 src_data, dest_data, is_forward);
     }
   } else {
-    // We are at the last two dimensions, which are stored continuously in memory
-    // With (N,C,H,W)
-    //      (0,1,2,3) cur_dim   -> H
-    //                cur_dim+1 -> W
+    // We are at the last two dimensions, which are stored continuously in
+    // memory With (N,C,H,W)
```
jeffdonahue: Thanks for the fix @CDLuminate.  Sorry for the nitpick but could you change this to `memory.` (add period `.`) so that the capitalized `With` immediately following doesn't look odd?

CDLuminate: Oops sorry for that. Fixed.

jeffdonahue:
Thanks @CDLuminate!

PR #5047, opened 2016-11-30
fix typo when including cmake files of opencv

twmht:
fix typo which may cause build failures


willyd:
Thanks for the fix @twmht.

PR #5057, opened 2016-12-02
Add Pascal to all cuda architectures

Queuecumber:
The known gpu architectures were missing the Pascal sm_60 and sm_61 compute capabilities. When building for this GPU, but on a separate machine, like a CI server or inside a docker image, caffe would be built for at most capability sm_50 and crash when run on the Pascal GPU.


shelhamer:
Thanks for including Pascal.

PR #5060, opened 2016-12-04
make web_demo work offline

aospan:
in some cases client doesn't have internet access (only access to
web-server with caffe web_demo running). Keep css/js in our web-server
for this case.

Signed-off-by: Abylay Ospan <aospan@netup.ru>


shelhamer:
Thanks for bringing up this use case, but I'd rather not bundle web resources in master. This is a simple change to make downstream, as you've made it here, and this PR can point out how to do so for others.

PR #5063, opened 2016-12-05
Filler compatibility for ND convolution and inner product

### LINKED ISSUE #3129: Does caffe support Nd Convolution?
I tried 3d convolution referring to #2049, and got an error 
F0625 16:20:14.583122  4782 blob.hpp:141] Check failed: num_axes() <= 4 (5 vs. 4) Cannot use legacy accessors on Blobs with > 4 axes.

As @Tgaaly said, I uncomment a sentence "CHECK_LE(num_axes(), 4)<< "Cannot use legacy accessors on Blobs with > 4 axes.";", and the program seems to run.

Will things go right?

### LINKED ISSUE #4456: ND convolution not working, again?
Hi,

I am trying to use FCN model. The model is giving error that it can not handle more than 4 axes.

F0713 14:12:04.008546 32691 blob.hpp:140] Check failed: num_axes() <= 4 (5 vs. 4) Cannot use legacy accessors on Blobs with > 4 axes.

does the current latest caffe work with Nd data? If not how to fix this?

Tried with and without CuDNN (v4.0). 

Best,

### LINKED ISSUE #4485: Filler in ND Convolutions use legacy shape
Hi, 
I am not sure if Xavier initialization does not make sense for 3D data,
but if it does, 
There will be this error:
Check failed: num_axes() <= 4 (5 vs. 4) Cannot use legacy accessors on Blobs with > 4 axes.

Which happens because in the filler we use the legacy shape (but we only use num and channels which would be shape[0] san shape[1])

Is xavier initiallization only for 2d data, or can it just use the new shape accessors?

Thanks!

knsong:
hi, I notice that `caffe` will always throw an error like
>F0713 14:12:04.008546 32691 blob.hpp:140] Check failed: num_axes() <= 4 (5 vs. 4) Cannot use legacy accessors on Blobs with > 4 axes.

when in the ND(>2) convolution case, e.g. [#4456](https://github.com/BVLC/caffe/issues/4456), [#4485](https://github.com/BVLC/caffe/issues/4485), [#3129](https://github.com/BVLC/caffe/issues/3129) and [this SO post](http://stackoverflow.com/q/40935127/6281477). 

Besides the `Xavier` filler and `MSRA` filler are not well compatible for both `Convolution` and `InnerProduct` layer. 

So I make some small modification in the `filler.hpp` to fix them.

Fixes #4456
Fixes #4485
Fixes #3129


Noiredd:
Nice catch, @knsong - could you please rebase and squash your commits so I can merge them?  
Also, would you mind explaining why did the logic for `fan_out` change?

shelhamer:
@Noiredd the github merge interface now lets you combine merging with squashing and I find it can help in cases like this—just in case you haven't tried that before.

Noiredd:
@shelhamer Thanks I did not notice that option!

PR #5069, opened 2016-12-06
Update inner_product_layer.cu

IdiosyncraticDragon:
a simple mistake, gpu data should be used.


shelhamer:
`alpha` can be on host or device memory: http://docs.nvidia.com/cuda/cublas/#cublas-lt-t-gt-axpy
If this is changed it should be changed everywhere for consistency. I doubt there'll be any real difference, since these scalars will stay in host (or device, with this change) memory and not change.

Noiredd:
I agree. Note that `Layer::Forward_gpu` and `caffe_gpu_axpy` are host functions, executed by the CPU. We're not passing a pointer to memory on either host or device, but dereference a pointer to retrieve a value, so I would even say that `gpu_memory()[0]` is incorrect.

PR #5074, opened 2016-12-07
Fix plot_training_log.py.example

garion9013:
The wrong marker hash access incurs infinite loop errors.
This PR fixed it and removed the unnecessary try / except statement.


shelhamer:
Thanks for the fix

PR #5084, opened 2016-12-12
Update installation.md

sbrugman:
Fixed the installation of requirements using pip.


shelhamer:
This is not generally correct; closing.

See https://github.com/BVLC/caffe/pull/1950#issuecomment-75712421

PR #5117, opened 2016-12-23
use cv2 to load_image, resize_image

ZhangXinNan:
scikit-image is too slow, so I use cv2 to load/resize image

I made a test with an image (image_size = (3120,4160)), 
run 100 times
| load image function | load image time(s) | resize image function | resize image(s) | 
|---|---|---|---|
| cv2.imread | 14.90 | cv2.resize | 0.087 |
| skimage.io.imread | 18.18 | skimage.transform.resize | 29.609 |
| caffe.io.load_image | 38.01 | caffe.io.resize | 26.35 |
| caffe.io.load_image(cv2) | 22.01 | caffe.io.resize(cv2) | 7.085 |

### DIFF
--- a/python/caffe/io.py
+++ b/python/caffe/io.py
@@ -295,7 +295,7 @@ def load_image(filename, color=True):
         of size (H x W x 1) in grayscale.
     """
     # img = skimage.img_as_float(skimage.io.imread(filename, as_grey=not color)).astype(np.float32)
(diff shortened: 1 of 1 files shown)


shelhamer:
OpenCV isn't a required dependency, but switching to Pillow makes sense to me. `scikit-image` is a poor choice for this use.

PR #5119, opened 2016-12-23
pycaffe fixes and googlenet oversample method

koelscha:
At the moment, testing a network with 'caffe.bin test' gives different results than testing with pycaffe, which is very confusing. This commit fixes pycaffe and the example notebooks, so the classification results of the python interface and the cpp interface are equal.


Noiredd:
I think we should be careful about adding new dependencies, and this PR adds `opencv-python` to Python req's. Besides, the changes to `python/caffe/io.py` seem rather drastic - do we really need to alter that much to provide full compatibility between Pycaffe and CLI? I do realize that it is somewhat confusing, especially for new users (one of the frequently asked questions on caffe-users), but modifying the entire interface (which users have a bunch of scripts built around) seems risky.

The second contribution (oversampling) looks interesting - however I would recommend splitting this PR into two distinct ones for easier review.

PR #5131, opened 2016-12-27
Use the __restrict__ keyword for the low level vector functions

fenrus75:
The C++ restrict keyword gives the hint to the compiler that pointers
don't do partial overlaps, which in turn allows the compiler to apply
a set of optimizations that it otherwise would struggle with.

Specifically, for the vector functions, this results in SIMD getting
used for all the vector add/sub/div/sqrt operations, giving a huge
speedup on a micro-scale (4x to 8x) while giving a 5%+ speedup
on a full Alexnet


gustavopinto:
Dear @fenrus75,

I'm a researcher studying pull-requests non-acceptance, and I found this one. Could you give some insights about why this PR was opened and closed in the same day, with no comments?

Thanks

Gustavo Pinto
www.gustavopinto.org

PR #5136, opened 2016-12-29
Fix batchnorm layer numerics by replacing `powx()`

pfollmann:
I had problems to train ResNet on ImageNet, by using the solver-setting debug_info: true, I saw that there were nan's produced by the BatchNorm layer. 
The problem was that the square inside the variance-term was computed via the caffe_gpu_powx function. This is slow on the one hand, but also unstable, especially for negative numbers. Therefore it was replaced by a caffe_gpu_mul call.
In the same way, I added a caffe_gpu_sqrt function in order to replace the caffe_gpu_powx(..., ..., 0.5, ...), that could lead to the same problems.
Additionally I added Dtype() castings to make sure, where appropriate.
Note: I didn't change the CPU version as it seams to be working. Nevertheless, using caffe_sqr instead of caffe_powx (..., 2) could speed up the layer... of course, also here a sqrt-function could be better than a caffe_powx (..., 0.5).

### DIFF
--- a/include/caffe/util/math_functions.hpp
+++ b/include/caffe/util/math_functions.hpp
@@ -209,6 +209,9 @@ void caffe_gpu_log(const int n, const Dtype* a, Dtype* y);
 template <typename Dtype>
 void caffe_gpu_powx(const int n, const Dtype* a, const Dtype b, Dtype* y);
 
+template <typename Dtype>
+void caffe_gpu_sqrt(const int n, const Dtype* a, Dtype* y);
+
 // caffe_gpu_rng_uniform with two arguments generates integers in the range
 // [0, UINT_MAX].
 void caffe_gpu_rng_uniform(const int n, unsigned int* r);
--- a/src/caffe/layers/batch_norm_layer.cu
+++ b/src/caffe/layers/batch_norm_layer.cu
@@ -38,17 +38,18 @@ void BatchNormLayer<Dtype>::Forward_gpu(const vector<Blob<Dtype>*>& bottom,
   }
 
   // subtract mean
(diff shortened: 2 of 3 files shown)


jeffdonahue:
Thanks for this fix @pfollmann!  Merged in c560658 with the corresponding change for CPU batch norm.  I left off the explicit casts and comments as I don't think these are needed for this stabilization.

John1231983:
I am merging my CAFFE to the pull request (for Batch Norm issue). I almost have done without the remaining lines 200-209 in `src/caffe/util/math_functions.cpp`
```
    template <>
     void caffe_sqrt<float>(const int n, const float* a, float* y) {
       vsSqrt(n, a, y);
     }     
     template <>
     void caffe_sqrt<double>(const int n, const double* a, double* y) {
       vdSqrt(n, a, y);
     }
```    
After adding the above line and recompile, I got the error
```
CXX src/caffe/util/math_functions.cpp
    src/caffe/util/math_functions.cpp: In function ‘void caffe::caffe_sqrt(int, const Dtype*, Dtype*) [with Dtype = float]’:
    src/caffe/util/math_functions.cpp:281:17: error: ‘vsSqrt’ was not declared in this scope
       vsSqrt(n, a, y);
                     ^
    src/caffe/util/math_functions.cpp: In function ‘void caffe::caffe_sqrt(int, const Dtype*, Dtype*) [with Dtype = double]’:
    src/caffe/util/math_functions.cpp:286:17: error: ‘vdSqrt’ was not declared in this scope
       vdSqrt(n, a, y);
                     ^
    make: *** [.build_release/src/caffe/util/math_functions.o] Error 1
```

How can I solve it? I found that vdSqrt() is a standard function. I am using the caffe version https://github.com/BVLC/caffe/tree/ae5343d72d04b16cc1f395538012febe75749f65

PR #5138, opened 2016-12-30
Crop layer remove redundance

xmyqsh:
remove the redundancy for loop in crop.cpp
remove the redundancy compute in crop.cu

### DIFF
--- a/src/caffe/layers/crop_layer.cpp
+++ b/src/caffe/layers/crop_layer.cpp
@@ -86,27 +86,25 @@ void CropLayer<Dtype>::crop_copy(const vector<Blob<Dtype>*>& bottom,
     }
   } else {
     // We are at the last dimensions, which is stored continuously in memory
-    for (int i = 0; i < top[0]->shape(cur_dim); ++i) {
-      // prepare index vector reduced(red) and with offsets(off)
-      std::vector<int> ind_red(cur_dim, 0);
-      std::vector<int> ind_off(cur_dim+1, 0);
-      for (int j = 0; j < cur_dim; ++j) {
-          ind_red[j] = indices[j];
-          ind_off[j] = indices[j] + offsets[j];
-      }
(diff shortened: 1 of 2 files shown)


williford:
The change in the crop_layer.cpp file looks good to me. The code was in a for-loop when it does the same thing every loop (index is used and nothing is incremented).

I haven't figured out how the crop_layer.cu changes are equivalent yet. Can you give some explanation?

xmyqsh:
@williford   glad to see you interesting my PR
case 
`    const int lines = top[0]->shape(cur_dim);`
`    const int height = top[0]->shape(cur_dim);`
so 
`lines` is equal to `height`
from 
`__global__ void copy_kernel(const int n, const int height, const int width,`
`copy_kernel<<<CAFFE_GET_BLOCKS(lines), CAFFE_CUDA_NUM_THREADS>>>(
            lines, height, width,`
we know `n` is equal to `lines`
from 
`CUDA_KERNEL_LOOP(index, n) {`
we know `index` is less than `n` 
so `index` is also less than `height`
so `index / height` is always zero
and `index % height` is always `index`

so 
`int src_start = index / height * src_outer_stride
                   + index % height * src_inner_stride;`
is equivalent to
`int src_start = index * src_inner_stride;`

these simplicity can remove `N*C*H*2` times  division, multiply, mod and add operations in one forward or backwark process

shelhamer:
@BlGene could you take a look when you have the chance?

BlGene:
It makes sense that we don't need to iterate over the size of the last spatial dimension because we are copying it in one go, as it is stored in memory continuously. 

Sorry for overlooking this before.


@xmyqsh: Thanks for finding this, can you please squash the PR.
@shelhamer: This looks correct to me.

shelhamer:
Thanks @xmyqsh for the simplification and removal of the redundant operations.

PR #5148, opened 2017-01-01
Overhaul layer catalogue documentation.

williford:
I have been overhauling the layer catalogue documentation, so that it is comprehensive and relies more on the documentation in the code (including caffe.proto).

You can find an online version of these changes here:
http://git.caffe.help/tutorial/layers.html

An example layer page is:
http://git.caffe.help/tutorial/layers/batchnorm.html

I created the links to the headers and caffe.proto in order to avoid the need to duplicate information, or at least make is easier to make sure that the more user-friendly version is up-to-date with the code.
Some of the layers had quite a lot of more user-friendly documentation added, in this case I have left it in with some reformatting, for example: http://git.caffe.help/tutorial/layers/convolution.html

Changes include:

- Creating a separate page for every layer.
- Linking every current layer type from the layers.html page.
- Each layer page has links to their header files, cpp files, etc.
- I've added a script that splits up the caffe.proto file, which allows the layer pages to include the corresponding sections.

I am also making an alternative version available, but requires a separate git repository (which I will also make available):
http://caffe.help/manual/layers.html

For those with the permissions: Please feel free to make direct modifications to this PR, without asking for permission.

### DIFF
--- a/docs/tutorial/layers.md
+++ b/docs/tutorial/layers.md
@@ -1,186 +1,77 @@
 ---
 title: Layer Catalogue
 ---
+
 # Layers
 
 To create a Caffe model you need to define the model architecture in a protocol buffer definition file (prototxt).
 
 Caffe layers and their parameters are defined in the protocol buffer definitions for the project in [caffe.proto](https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto).
 
-### Vision Layers
+## Data Layers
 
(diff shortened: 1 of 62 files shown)


shelhamer:
Thanks @williford for the definite improvement. Follow-up PRs can add more detail and we should consider promoting the layer catalogue to the homepage instead of stashing it in tutorial documentation.

PR #5165, opened 2017-01-09
OpenCL: some optimizations for Intel Gen Graphics

gongzg:
This PR added some new convolution kernels into the spatial convolution engine and also optimized some other layers such as LRN layer/batchnorm layer/softmax layer and softmax loss layer. We also create an intel isaac project to optimize those blas functions via OpenCL. Please use github.com/intel/isaac with Intel platform.

### DIFF
--- a/include/caffe/layers/conv_spatial_layer.hpp
+++ b/include/caffe/layers/conv_spatial_layer.hpp
@@ -199,7 +199,8 @@ class ConvolutionLayerSpatial : public BaseConvolutionLayer<Dtype> {
   Dtype* top_data;
   Dtype* col_data;
   const Dtype* weight;
-  Dtype* swizzled_weights;
+  const Dtype* weight_cpu;
+  Dtype* swizzled_weights_;
   int_tp weight_offset;
   int_tp col_offset;
   int_tp top_offset;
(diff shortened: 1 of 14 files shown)


naibaf7:
Ok cool, I will review this in the next few days :)

atlury:
@naibaf7 @gongzg Does this mean after this PR merge, we will need to use isaac with say clbas to get better performance?

gongzg:
@atlury Exactly. clBLAS is much slower than the isaac.

naibaf7:
@gongzg The point is also that we have windows support now, but it's not so easy to get all the external libraries working there (yet). Also the Intel spatial kernels are not fully working yet, so that's also something I'm currently investigating. :)

gongzg:
@naibaf7 oh. I have not been aware of the windows support so far. The Intel spatial convolution should only work on Linux now. May be we can start to enable windows support in next step. Thanks.

naibaf7:
@gongzg I thought I wrote you an email about it 2 weeks ago... :)
However yes, that would be cool to get working as well.

gongzg:
@naibaf7 Don't know why I missed your previous email. Too busy these days :).  But it's cool to support both Linux and Windows. Will work on that latter.

atlury:
@naibaf7 @gongzg I have tested these merge commits on an Intel tv stick, webcam using Intel Spatial kernels and using 19-layer model vgg model. I am able to get real time classification. 

Amazing feat!! And all under 3.5 watts. This blows out nvidia tx1. Congrats to you guys.

Next would be to test object detection. (one Warning I encountered is "reading dangerously large protocol message, if the message turns out to be larger, parsing will be halted for security reasons" nevertheless this should be advertised more)

curtain

naibaf7:
@atlury Thanks a lot :) It has been a lot of work to get here by @gongzg, various community contributions and me.

naibaf7:
@gongzg Merged. There have been a few minor issues but I think it should be good now.

PR #5088, opened 2016-12-13
Adding support for python 3.5 in windows

willyd:
Adding support for building with python 3.5. This is a work in progress

**TODOs**
- [x] Add preliminary support for python 3.5 with MSVC 2015
- [x] ~Add support for MSVC 2013 (create protobuf package vc12 with python 3.5)~ Will only support 3.5 with MSVC 2015
- [x] Update README.md
- [x] Test dependencies download script more thoroughly
- [x] ~Fix linting on appveyor~

PR #5189, opened 2017-01-16
Fix a link error.

taeguk:


willyd:
Thanks for submitting. Since #5088 was merge CMake takes care of downloading the dependencies so this change is not required anymore. If you find any other typos other PRs are welcome!

PR #5204, opened 2017-01-20
Fix broken build on windows when not defining WITH_CUDA

forderud:
Fix the following error when running the script without first defining "WITH_CUDA": `( was unexpected at this time.`

The problem was introduced by 372e920eb5c0ccdf70a7caaa344847d1485919c6 (part of https://github.com/BVLC/caffe/pull/4895)


willyd:
Thanks for the fix @forderud!

PR #5205, opened 2017-01-20
Promote the URLs of individual layer doc pages.

williford:
This PR is in response to @shelhamer 's [comment](https://github.com/BVLC/caffe/pull/5148#issuecomment-273589521) in my last PR that "we should consider promoting the layer catalogue to the homepage instead of stashing it in tutorial documentation."

I'm not sure whether this should be done or not, but if it will be done, we should move the URLs of the new layer doc pages up one level before many external websites link to the pages at the current location (within `/tutorial/layers/*.md`).

Move the individual layer pages up one level, to make it easier to
promote the layers documentation in the future. This will prevent
breaking links from external websites, as long as we don't decide they
should actually be located somewhere else.

The layers.md is being left in tutorial for now, because it already has
a lot of external links.


williford:
I also fixed some links that were broken in the layer documentation. You can see the changes here:
http://git.caffe.help/tutorial/layers.html

shelhamer:
It seems simpler to leave all the existing pages where they are, but add link on the homepage directly to the layer catalogue. What do you think? I still need to mull this over more b.c. there might be confusion between tutorial documentation, API documentation, and layer catalogue.

williford:
I think that is a good idea. Perhaps we should also leave the existing links from the tutorial to the layer catalogue, because it does seem to fit in with some of the pages there. I find that the layer catalogue stands out as it is in Google searches.
 
Thanks for your quick response!

PR #5206, opened 2017-01-20
Windows build configuration changes without editing build_win.cmd

forderud:
Make it possible to change Windows build configuration without having to edit build_win.cmd.

This capability seem to have been unintentionally removed in 82c087ca05795d23c939e55b1b12f9046140ed5c (part of #5088).

### DIFF
--- a/scripts/build_win.cmd
+++ b/scripts/build_win.cmd
@@ -66,29 +66,29 @@ if DEFINED APPVEYOR (
 ) else (
     :: Change the settings here to match your setup
     :: Change MSVC_VERSION to 12 to use VS 2013
-    set MSVC_VERSION=14
+    if NOT DEFINED MSVC_VERSION set MSVC_VERSION=14
     :: Change to 1 to use Ninja generator (builds much faster)
-    set WITH_NINJA=1
+    if NOT DEFINED WITH_NINJA set WITH_NINJA=1
     :: Change to 1 to build caffe without CUDA support
-    set CPU_ONLY=0
+    if NOT DEFINED CPU_ONLY set CPU_ONLY=0
     :: Change to Debug to build Debug. This is only relevant for the Ninja generator the Visual Studio generator will generate both Debug and Release configs
-    set CMAKE_CONFIG=Release
+    if NOT DEFINED CMAKE_CONFIG set CMAKE_CONFIG=Release
     :: Change to 1 to build a caffe.dll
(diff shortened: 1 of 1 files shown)


willyd:
Thanks @forderud. My original intention was that one could set global environment variables to override the ones in the batch file. Like:
```
$ set WITH_NINJA=1
$ set WITH_PYTHON=1
$ scripts/build_win.cmd
```
But after some time I realized that somehow some variables defined in the batch file were leaking to my current cmd.exe session even with the `@setlocal` . I could not figure out why this was happening at the time so I decided to remove the `IF NOT DEFINED`. I will revisit this to see if the issue is still there.

Is your use case similar to my origin intent?

forderud:
Yes, my use case exactly what you're mentioning above. I want to build Caffe with WITH_NINJA=0, CPU_ONLY=1, BUILD_PYTHON=0 & BUILD_PYTHON_LAYER=0 without having to modify any versioned files. 

You might be right that variables leak out to the cmd.exe session. However, this is not a concern for me, since I'm using new cmd.exe sessions for each build. What is important to me is to be able to automatically build Caffe "as is" without any modification.

willyd:
Ok. I will wait for CI to pass and check if I the variables still leak into the cmd.exe session and report back. Thanks again.

forderud:
@willyd I've now tested a bit, but have not been able to reproduce your problem of variables leaking out to the calling cmd.exe session.

forderud:
Gentle reminder.

willyd:
> @willyd I've now tested a bit, but have not been able to reproduce your problem of variables leaking out to the calling cmd.exe session.

Thanks for validating this.

> Gentle reminder.

Sorry, I don't have much time to check this now. It will have to wait a few days (or a week) more. I get a lot of requests from the community but not a lot of help... 

I am looking to get help on integrating the latest merge of NCCL (see PR #4563) in master. If you would like to help let me know.

willyd:
I created about the nccl related changes at #5216.

forderud:
I'm sorry, but I don't know much about CUDA and have never heard about NCCL. It will therefore be difficult for me to contribute on that specific problem.

My main focus here is to ease usage of Caffe on Windows.

forderud:
Gentle reminder. This PR will make life easier for anyone wanting to build Caffe on windows with an automatic script. I would really appreciate if this PR or something similar could be merged in the near future.

forderud:
Thanks a lot for making my life easier @willyd.

PR #5207, opened 2017-01-21
cmake: bump soversion to rc4

CDLuminate:


shelhamer:
Oh right, thanks for catching that!

PR #5210, opened 2017-01-21
Obsolete reference to `bool solver` in caffe.proto

ftokarev:
`NetState` field `bool solver` was introduced in https://github.com/BVLC/caffe/pull/734, but it has never made it to the master, @jeffdonahue decided against it (https://github.com/BVLC/caffe/pull/734#issuecomment-49568384).


shelhamer:
Thanks for cleaning up the comment.

PR #5213, opened 2017-01-23
Added Cmake Support to VS 2017 (and its RC)

PENGUINLIONG:
This pull request added CMake support to recently released Visual Studio 2017 RC (and VS 2017 in the future).

### DIFF
--- a/scripts/build_win.cmd
+++ b/scripts/build_win.cmd
@@ -66,6 +66,7 @@ if DEFINED APPVEYOR (
 ) else (
     :: Change the settings here to match your setup
     :: Change MSVC_VERSION to 12 to use VS 2013
+    :: Change MSVC_VERSION to 15 to use VS 2017 or VS 2017 RC
     set MSVC_VERSION=14
     :: Change to 1 to use Ninja generator (builds much faster)
     set WITH_NINJA=1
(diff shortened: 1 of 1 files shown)


willyd:
We would need to download to make sure that CMake downloads the right precompiled libraries too. Since VS 2017 and VS 2015 are supposed to be binary compatible no need to have a different set of libraries just download the VS 2015 ones. See https://github.com/BVLC/caffe/blob/windows/cmake/WindowsDownloadPrebuiltDependencies.cmake.

PENGUINLIONG:
Noticed that. CMake was unable to find Boost libs and OpenCV because of their naming convention. But all the dependencies are, of course, binary compatible. Renaming is an acceptable solution.

However, another thing should be paid attention is that the MSVC executable is moved from `/path/to/vs/VC/bin` to `/path/to/vs/VC/Tools/MSVC/14.10.24728/bin/HostX64/x64`. The generation of cuda objects will also be affected. Manual replacement is required again because the `FindCUDA.cmake` provide by CMake 3.7.2 is buggy, check it out at line 487.

willyd:
> Noticed that. CMake was unable to find Boost libs and OpenCV because of their naming convention. But all the dependencies are, of course, binary compatible. Renaming is an acceptable solution.

Good catch. I did not think about that. 

> However, another thing should be paid attention is that the MSVC executable is moved from /path/to/vs/VC/bin to /path/to/vs/VC/Tools/MSVC/14.10.24728/bin/HostX64/x64. The generation of cuda objects will also be affected. Manual replacement is required again because the FindCUDA.cmake provide by CMake 3.7.2 is buggy, check it out at line 487.

This is an upstream CMake issue. I guess we should wait until CMake fixes this before integrating this change then. You will hit some problems with CUDA and MSVC 2017 compatibility as well. VS 2015 is now supported by CUDA 8 but was not in 7.5. I guess we will have to wait for 8.5?

PENGUINLIONG:
Sure. It depends on a lot of things. Let this closed first.

PR #5215, opened 2017-01-23
Restore can be invoked on rank > 0

cypof:


shelhamer:
@cypof is there a test you could add to catch this in the future?

cypof:
OK, it might not be too bad to write. There is one for net apparently, but nothing yet for solvers if I did not miss anything.

PR #5220, opened 2017-01-24
parse_log.py was not using --verbose argument

kts:


shelhamer:
Thanks for catching that.

Issue #5209, opened 2017-01-21
[doc] problematic markdown rendering at http://caffe.berkeleyvision.org/install_apt_debian.html


CDLuminate:
It seems that I have to make a newline before the verbatim part in the markdown source ...

http://caffe.berkeleyvision.org/install_apt_debian.html

PR #5242, opened 2017-02-01
docs: update install_apt_debian guide

CDLuminate:
Closes: https://github.com/BVLC/caffe/issues/5209

It's time to update the Debian guide again. We rescued the CUDA package for Stretch. That means both `caffe-cpu` and `caffe-cuda` will be available for Debian/Stretch.


shelhamer:
Thanks for the packaging help!

PR #4920, opened 2016-10-29
Fail when loss_param is used with EuclidanLoss layer 

BlGene:
...as this layer does not implemente the functionality specified by the parameter,
although this might plausibly be expected.

PR #5250, opened 2017-02-03
Feature/ignore euclideanloss

matthill:
This PR allows me to use an "ignore_label" value in the EuclideanLoss layer.  For example:

```
layer {
  name: "loss_match"
  type: "EuclideanLoss"
  bottom: "convb"
  bottom: "regression_label"
  top: "loss_match"
  loss_param {
    ignore_label: 0
  }
}
```

The issue I was having was that I am training against multiple loss functions with different data.  Part of my dataset is missing.  I still want to use the data when I have it, but I do not want to back-propagate when those values are missing.  Simply setting the values to zero causes the regression to lean towards the zero values, rather than learning to ignore them.

SoftmaxWithLoss already has this feature, this PR adds parity for the EuclideanLoss layer.

99% of this PR is pulled from this PR: https://github.com/BVLC/caffe/pull/3677

However, that PR was not ideal, since it simply casts the value to an int and ignores it if it matches.  My values are normalized between -1 and 1 so that wouldn't work.  Instead, I am simply checking that the float value is within a narrow range around the configured integer.  After making this change, my network's accuracy was much higher.

### DIFF
--- a/include/caffe/layers/euclidean_loss_layer.hpp
+++ b/include/caffe/layers/euclidean_loss_layer.hpp
@@ -99,6 +99,10 @@ class EuclideanLossLayer : public LossLayer<Dtype> {
   virtual void Backward_gpu(const vector<Blob<Dtype>*>& top,
       const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
 
+  /// Whether to ignore instances with a certain label.
+  bool has_ignore_label_;
+  /// The label indicating that an instance should be ignored.
+  int ignore_label_;
   Blob<Dtype> diff_;
 };
 
--- a/src/caffe/layers/euclidean_loss_layer.cpp
+++ b/src/caffe/layers/euclidean_loss_layer.cpp
@@ -33,8 +33,9 @@ void EuclideanLossLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
     const Dtype* label_data = bottom[1]->cpu_data();
     Dtype* diff__data = diff_.mutable_cpu_data();
     for (int i = 0; i < count; ++i) {
-      const int label_value = static_cast<int>(label_data[i]);
-      if (label_value == ignore_label_) {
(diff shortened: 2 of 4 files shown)


happyharrycn:
I do not think this is the best way of ignoring samples for EuclideanLoss. You are also throwing away all samples with a target value close to 0. A more principled way is to implement some sort of masking function. For example, allow the third bottom blob to take a (binary) mask. This mask can be used in backward path to reset/block the gradients.

matthill:
That seems like a more complex solution (both in terms of implementation and use) without much benefit.  Realistically, how many spurious data points will fall within such an extremely narrow range?  I think probably zero or very close to it.

BlGene:
@matthill: Can you please `run make` lint on your PR and fix the mentioned items.

Just to cross reference, this would make #4920 redundant.

Noiredd:
Closing as a better approach to this has been [suggested](https://github.com/BVLC/caffe/pull/3677#issuecomment-185527051).

nnop:
However, the suggested approach didn't consider the normalization issue. @Noiredd

Noiredd:
I see. But wouldn't it be a better way to go if we added a mask input to the EuclideanLossLayer? Loss label in the case of an L2 loss not only feels confusing (we don't even have actual labels here), but also pointless - effectively this means: "never learn values of x", which isn't generally useful. Taking a mask input (i.e. "do not learn from these examples") would be more universal - and it still fulfills @matthill's original reason for implementing this.

I'll reopen this for further conversation.

nnop:
I had muddled the way `EuclideanLoss` layer doing normalization. Actually, it behaves differently from `SoftmaxWithLoss` layer.  Euclidean loss propagates it gradients **separately** to each position (or dimension) of its input. Whereas softmax loss treat each position **equally**, and need to consider the ignored label when doing normalization. So, for Euclidean loss we only need to use the batch size (N) as the normalization count which has nothing to do with the ignored (or masked) elements.

Correct me if I'm wrong. @Noiredd

PR #5272, opened 2017-02-09
Add Pascal CUDA architectures to Makefile.config.example

crowsonkb:
The new list of -gencode options is available at http://docs.nvidia.com/cuda/pascal-compatibility-guide/#building-pascal-compatible-apps-using-cuda-8-0.


shelhamer:
Thanks for the more complete listing of archs.

Issue #4637, opened 2016-08-26
cmake caffe vecLib not found  OS X El Capitan


JulianBlin:
I've been trying to compile caffe on a mac with El Capitan.

`git clone https://github.com/BVLC/caffe`
`mkdir build`
`cd build`
`cmake .. `

The result was an error `vecLib not found`

The issue to solve this is to modify the file in `cmake/Modules/FindvecLib.cmake` 

Line 19, replace 
`          /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk/System/Library/Frameworks/Accelerate.framework/Versions/Current/Frameworks/vecLib.framework/Headers/`

by
`                /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/Accelerate.framework/Versions/Current/Frameworks/vecLib.framework/Headers/`

Don(t know if you can modify this in the master branch ?

Thanks

shelhamer:
It sounds like a more general expression is needed that includes the OS X version number in the search path, be it `10.9`, `10.11`, or whatever.

Pull requests for build improvements are always welcome. Thanks for noting the issue.

zhenpingfeng:
open the CMakeCache.txt file,
find this line: 
//vecLib include directory
vecLib_INCLUDE_DIR:PATH=/System/Library/Frameworks/vecLib.framework/Headers

change to: 
//vecLib include directory
vecLib_INCLUDE_DIR:PATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/vecLib.framework/Headers

work fine for me.

ledjorge:
change to: 
The general and correct expression is:

//vecLib include directory
vecLib_INCLUDE_DIR:PATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Accelerate.framework/Versions/Current/Frameworks/vecLib.framework/Headers/

It works if your system is not broken ;)

Laknath1996:
I have the same error popping up. And I using 'make all' command to build instead of 'came'. It gives me the following error.

clang: warning: argument unused during compilation: '-pthread'
ld: framework not found vecLib
[...]

PR #5278, opened 2017-02-12
Find veclib on MacOS automatically.

ethanluoyc:
Fix https://github.com/BVLC/caffe/issues/4637.


shelhamer:
The versionless path in this comment https://github.com/BVLC/caffe/issues/4637#issuecomment-257249351 seems to be right too. Is there a reason not to use it?

shelhamer:
Merged versionless path in #5236. Please follow-up if that doesn't work for you.

PR #5297, opened 2017-02-17
fixed typo: GPU->CPU

fxrcode:


cypof:
It's GPU by default, the page was correct

Issue #5262, opened 2017-02-07
Using c++11 by default


hgaiser:
It's been 6 years since the release of c++11, I think it's time Caffe started using it ;)

I noticed in the source code there are [some comments ](https://github.com/BVLC/caffe/blob/master/include/caffe/common.hpp#L77-L78) which suggest C++11 would preferably be enabled, but it cannot be due to NVCC incompatibilities with C++11? I tested the compilation (using CMake) of Caffe using `-std=c++14` and indeed it gave issues regarding NVCC. It turns out that CMake by default [propagates host flags](https://github.com/Kitware/CMake/blob/v2.8.7/Modules/FindCUDA.cmake#L76-L85) to `nvcc`. If this is turned off, compilation works fine on my system and everything works as expected.

For completeness, I compiled with the following command:
```
cmake .. \
	-DCMAKE_BUILD_TYPE=Release \
	-DBLAS="Open" \
	-Dpython_version=3 \
	-DBoost_PYTHON_LIBRARY_RELEASE=/usr/lib/libboost_python3.so \
	-DCUDA_HOST_COMPILER=/usr/bin/g++-5 \
	-DCUDA_PROPAGATE_HOST_FLAGS=off \
	-DCMAKE_CXX_FLAGS="-std=c++14"
make -j8
```

Note that I compile with OpenBLAS, python3, and since I am using gcc6 I have to tell NVCC to use g++5. It is the last two options (`CUDA_PROPAGATE_HOST_FLAGS` and `CMAKE_CXX_FLAGS`) that are of importance here, but I thought I'd share the entire build process. In addition, I compiled using c++14 which worked fine. Using c++14 if available would in my opinion also be a valid option, but I figured lets discuss to enable c++11 by default first.

Also as a related side note:
1. I think it'd be better for Caffe to start using the CMake build system by default, I feel it is far superior to the Makefile.config method.
2. The current minimum required version according to CMakeLists.txt is 2.8.7, if this is bumped to 3.x then useful features can be used to determine the presence of c++11 etc.
3. Features from c++11 would make the design of the feature discussed in https://github.com/BVLC/caffe/issues/5243 more clean, as it allows move semantics and other features to be used.

I would love to hear thoughts on enabling c++11 by default, but thoughts regarding a move to CMake or bumping the CMake version are also welcome (though perhaps more suitable for a separate discussion?).

de-vri-es:
[...]

PR #5325, opened 2017-02-24
Enable C++11 in cmake

ssudholt:
This PR enables C++11 for compilation of both C++ and CUDA code (see https://github.com/BVLC/caffe/issues/5262).

As C++11 has been supported by CUDA since 7.0, the minimally required version is bumped as well in the cmake file. In order to not raise warnings, host compiler flag propagation to nvcc is disabled (see https://cmake.org/Bug/view.php?id=15240).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,12 @@ if(USE_libstdcpp)
   message("-- Warning: forcing libstdc++ (controlled by USE_libstdcpp option in cmake)")
 endif()
 
+if (CMAKE_VERSION VERSION_LESS "3.1")
+  set(CMAKE_CXX_FLAGS "--std=c++11 ${CMAKE_CXX_FLAGS}")
+else ()
+  set(CMAKE_CXX_STANDARD 11)
+endif ()
+
 # ---[ Warnings
 caffe_warnings_disable(CMAKE_CXX_FLAGS -Wno-sign-compare -Wno-uninitialized)
 
--- a/cmake/Cuda.cmake
+++ b/cmake/Cuda.cmake
@@ -231,7 +231,7 @@ endfunction()
 ###  Non macro section
 ################################################################################################
 
-find_package(CUDA 5.5 QUIET)
+find_package(CUDA 7.0 QUIET)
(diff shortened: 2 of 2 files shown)


hgaiser:
I agree that c++11 should be enabled, however it is also included in https://github.com/BVLC/caffe/pull/5294 because it was necessary for move semantics and other c++11 features. Be aware that this PR also includes enabling c++11 but in a slightly different manner. It bumps required cmake version to 3.1, since this works on Ubuntu 14.04 I don't see an issue there.

ssudholt:
@hgaiser cmake < 3.1 would allow compilation on Debian 8 with cmake from stable sources. That was pretty much the only reason for me to include it. Wouldn't be opposed to changing it though if this feature is not required. Wasn't aware that #5294 enables C++11 as well.

de-vri-es:
Note that #5294 doesn't bump the required version for CUDA to 7.0 and doesn't enable C++11 for `nvcc`. #5294 can be rebased on this PR if it is merged, but before that C++11 should also be enabled in the old non-cmake build system. Otherwise anything using C++11 features will still fail the CI tests.

It does seem that just dropping the old build system would save everyone a lot of trouble. But I suppose that is a different discussion.

pkdogcom:
This commit resolves my issue of adding C++11 dependent layers. Hopefully it will be merged into master soon so that people won't return into the same issue.

cooperpellaton:
@de-vri-es Is there any particular benefit to not, by default, enabling "C++11 for `nvcc`"? I don't see any reason why this would negatively affect the build system or procedure.

de-vri-es:
> @de-vri-es Is there any particular benefit to not, by default, enabling "C++11 for nvcc"? I don't see any reason why this would negatively affect the build system or procedure.

The main reason would be to support CUDA < 7.0. There doesn't seem to be a pressing need to enable C++11 support in CUDA code right now. However, I'm not against bumping the minimum CUDA version and enabling C++11.

Though the cmake 3.1 path should also have this then:
```
set(CMAKE_CXX_STANDARD_REQUIRED on)
```

PENGUINLIONG:
`constexpr` is really helpful making the codes clearer. Although I would love C++11 to be enabled, I still couldn't assert it's a necessity since too much has to be changed, and boost is already there.

PR #5333, opened 2017-02-28
Add support for netlib reference cblas in the CMake build system.

de-vri-es:
The reference `cblas` from http://www.netlib.org/lapack/ seems to work fine with caffe, and is widely available in a large number of distros, so it makes sense to support it. This PR adds it as supported BLAS provider for the cmake build.

The reference `cblas` ships with a `cblas-config.cmake` file, so no additional cmake module is needed.


de-vri-es:
Ping? There should be no risk of breaking anything that currently works, the user has to specify the CBLAS provider manually.

de-vri-es:
Could this be reviewed? It's pretty low risk.

de-vri-es:
There is absolutely no risk of breaking anything that currently works, this only adds an extra blas provider. I'd think it would be easy to review and merge this.

de-vri-es:
Seriously, this is an extremely low risk change. Can somebody please review this?

de-vri-es:
Rebased again...

de-vri-es:
ping

de-vri-es:
Well, this is never happening...

PR #5337, opened 2017-03-01
Remove not used variable in base_conv_layer.cpp

zhuyuanhao:
Variable "bottom_dim_blob_shape" is not used in LayerSetUp().


shelhamer:
Thanks for trimming!

PR #5370, opened 2017-03-08
[experimental] Halide Layer

BlGene:
This is an extension of caffe that enables the calling of Halide functions as layers.

To test it please.
1. Install Halide
2. build caffe using cmake with the `BUILD_halide` option
3. run the short python script below

TODOs:

- [ ]  Improve build / find bugs

The process of building layers is still quite elaborate and a lot of cmake code went into masking this complexity. That said I'm not a cmake expert, so I would greatly appreciate if others had a look at his too.

- [ ] Make layer zoo?

Once useful Halide layers start cropping up it would be nice to have a designated place to put them, together with some information on usage.

Some additional things worth considering:

1. Halide Python Bindings
Halide has python bindings which are a nicer user interface. Whats the cleanest way to provide access to these?

2. Automatic Backward pass
Automatic backward functions for Halide would make it much easier to use. There has been some chatter about implementing this, but I don't know the current state of this.

This is an extension of @longjon prior work.

----
Simple Test script
```
import time
import numpy as np
import caffe
from caffe import layers as L

INSTALL_DIR = "."
HEIGHT, WIDTH = 200, 200

def create_neural_net(batch_size=1):
    library = INSTALL_DIR + "/lib/halide/libdevfunc_wrapper.so"
    netspec = caffe.NetSpec()
    netspec.data = L.DummyData(shape=[dict(dim=[batch_size, 1, WIDTH, HEIGHT])], ntop=1)
    netspec.halide  = L.DevFuncLayer(netspec.data)

    with open('net_halide.proto','w') as f:
        f.write(str(netspec.to_proto()))
        f.write('external_layers : "{0}"'.format(library))
        f.flush()
        net = caffe.Net(f.name, caffe.TEST)
    return net

if __name__=='__main__':
    caffe.set_mode_gpu()

    m = 3
    tmp = np.zeros( (1,1,HEIGHT,WIDTH))
    h_c = int(HEIGHT/2)
    w_c = int(WIDTH/2)
    tmp[0,0,h_c,w_c] = 1

    net = create_neural_net()

    net.blobs['data'].data[...]  = tmp
    startTime = int(round(time.time() * 1000))
    net.forward()
    endTime = int(round(time.time() * 1000))
    print(endTime - startTime,'ms')
    print(net.blobs["halide"].data[0,0,h_c-m:h_c+m,w_c-m:w_c+m])
```


shelhamer:
> Once useful Halide layers start cropping up it would be nice to have a designated place to put them, together with some information on usage.

This is true of Python layers too. Let's try to work out an arrangement for both.

BlGene:
Note: github is sorting commits by time-stamp not order, so commits `5c8e354` and `1e02d62` should be reviewed first. If they are merged (or rejected) I can rebase.

halide/generator/register_gen.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+    Func inp_c("inp_c");
+    inp_c(x,y, c, b) = clamped(x,y,c, b);
+
+
+    // The algorithm - no storage or order
+    blur_x(x, y, c, b) = (inp_c(x-1, y, c, b) + inp_c(x, y, c, b) + inp_c(x+1, y, c, b))/3;
+    blur_y(x, y, c, b) = (blur_x(x, y-1, c, b) + blur_x(x, y, c, b) + blur_x(x, y+1,c ,b))/3;
+
+    // The schedule - defines order, locality; implies storage
```
BlGene: This should be removed.

halide/sync_halide.hpp
```diff
@@ -0,0 +1,88 @@
...
+namespace caffe {
+
+// the data/diff param is ugly
+template <typename Dtype>
+buffer_t HalideWrapBlob(const Blob<Dtype>& blob, bool data = true) {
+  CHECK_LE(blob.num_axes(), 4) << "Halide does not support blobs with more "
+    "than four axes";
+  buffer_t buf = {};
+  for (int i = 0; i < blob.num_axes(); ++i) {
+    // XXX reversing dimensions here -- should I?
```
BlGene: All XXX should be removed from comments

halide/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
+# This file differs from ../CMakeLists.txt only in the target_name_mod
+# var and lack of add_subdirectory call.
+
+if(BUILD_halide)
+
+    #add_subdirectory(test)
```
BlGene: Should be removed.

halide/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+
+    # Target name modification
```
BlGene: The argument for keeping this is that I easily allows defining a separate CMake script in a subfolder, e.g. for tests.

halide/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+    # ---[ Run generator executable to generate halide objects
+    set(Halide_OBJECTS "")
+    file(GLOB gens generator/*.gen)
+    foreach(infilePath ${gens})
+        # Generate output file name
+        get_filename_component(in_obj ${infilePath} NAME_WE)
+        message(STATUS "Halide build: found generator ${in_obj} at ${infilePath}")
+        SET(outfile "${PROJECT_BINARY_DIR}/halide/${in_obj}.o")
+        set( in_tgt "${in_obj}${target_name_mod}")
+        #message(STATUS halide_generator_run_${in_tgt} "creates" ${outfile} )
```
BlGene: remove.

PR #5376, opened 2017-03-09
Updated column awk parses from to obtain the correct value for loss

clancyian:
This updates ensures that the loss parameter is parsed correctly from the caffe log output


shelhamer:
Merged #5514 for it's slightly more robust log parsing by setting the field separator. Thanks for contributing a fix!

Issue #5369, opened 2017-03-08
Stage not passed to TEST state during training


JonBoyleCoding:
When I run caffe in training mode, if I give it a stage, it only appears to pass it through to the TRAIN state, and not the TEST state. This seems like a bug to me, hence why I'm posting here rather than the user group.

> caffe train --solver=autoencoder.solver.prototxt -stage="train-stage-0"

With the error being:
> F0308 11:18:16.666119  5201 insert_splits.cpp:29] Unknown bottom blob 'finalinner' (layer 'finalactivation', bottom index 0)

I note that the states in the console output are...
```
state {
  phase: TRAIN
  level: 0
  stage: "train-stage-0"
}
```

and...

```
state {
  phase: TEST
}
```

The relevant extract from the model prototxt is (which the finalinner_0 is included in the TRAIN state, but not the TEST state):
```
layer {
  name: "finalinner_0"
  type: "InnerProduct"
  bottom: "inner1activation"
  top: "finalinner"
  include {
    stage: "train-stage-0"
  }
  inner_product_param {
    num_output: 3212
    weight_filler {
      type: "gaussian"
      std: 1
      sparse: 15
    }
  }
}
layer {
  name: "finalactivation"
  type: "Sigmoid"
  bottom: "finalinner"
  top: "finalactivation"
}
```

PR #5383, opened 2017-03-10
Modification to Solver to get Stages in Testing

JonBoyleCoding:
As mentioned in #5369, when specifying stages to the caffe executable, the created test nets in the solver do not include the stages.

The simplest way I found to deal with this was to extract the stages from the train_state. While I don't think this is the most ideal solution, the alternative as best as I could see (modifying one of the constructors to take the stages) would involve a much larger change to the code base.

I specified the parameter to InitTestNets to be a pointer with a default NULL in order to not break any existing code that may rely on InitTestNets without any arguments. I do note that the only place in the code-base that calls InitTestNets is in the Init method; however, as it's a public method, I decided it best to try and not break the interface.

### DIFF
--- a/include/caffe/solver.hpp
+++ b/include/caffe/solver.hpp
@@ -45,7 +45,7 @@ class Solver {
   explicit Solver(const string& param_file);
   void Init(const SolverParameter& param);
   void InitTrainNet();
-  void InitTestNets();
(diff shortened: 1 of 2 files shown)


JonBoyleCoding:
Modified to not fail lint test.

JonBoyleCoding:
Closing to swap branch in my repository, look at #5395.

PR #5387, opened 2017-03-11
Reorder BLAS configuration

lababidi:
On a Mac a user cannot force Caffe CMake to choose which BLAS to use. This reorders the APPLE option as last if a user does not choose a proper BLAS package.


hgaiser:
How would someone trigger the `vecLib` option with this PR? By passing an argument which is not `"Atlas;Open;MKL"` ?

I understand the issue, systems like OpenBLAS are as far as I know also available on OSX but they are currently not optional in Caffe's CMakeLists.txt. Perhaps it is better to add an option called `vecLib` which checks `APPLE` and fails if it is not?

PR #5393, opened 2017-03-13
Multi GPU training from Python can use any solver

jfolz:
During multi GPU training from Python, using anything other than SGDSolver results in a crash because of method signature mismatch. Modifying the signature of Solver_add_nccl fixes this and networks learn as expected with other solvers in my tests


lolongcovas:
@jfolz how to configure solver to use multi gpu?

Issue #5347, opened 2017-03-02
caffe.NCCL.new_uid() - UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa9 in position 17: invalid start byte


GaryKT:
How can we work around this error? If we set uid to a fixed number it doesn't work and with caffe.NCCL.new_uid() we get the following error:

uid = caffe.NCCL.new_uid()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa9 in position 17: invalid start byte

Any help highly appreciated!

### Steps to reproduce
Run train.py

uid = caffe.NCCL.new_uid()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa9 in position 17: invalid start byte

### Your system configuration
Operating system: Windows
Compiler: MS Visual Studio 2015
CUDA version (if applicable): 8
CUDNN version (if applicable): 5.1
BLAS: ?
Python

willyd:
Thanks for reporting. I haven't tried the training script yet, I will get back to you on this ASAP. What python version are you using?

GaryKT:
Python 3.5.3 (v3.5.3:1880cb95a742, Jan 16 2017, 16:02:32) [MSC v.1900 64 bit (AMD64)] on win32

Let me know if there is some work around and/or if you are able to reproduce. Thanks.

willyd:
I can reproduce the issue. I think the problem is that boost python maps std::string to a *standard* string (i.e. not unicode) in python 2 and maps it to a unicode string in python 3. 

Can you try to see if the issue exists as well with python 2.7?

I think it would not be to hard write wrappers around the constructor or NCCL and the new_uid method returning and accepting bp::objects and manually converting from bytes objects to string objects in those wrappers.

See this https://github.com/boostorg/python/issues/85#issuecomment-242864985 for some ideas on how to implement the wrappers.

@cypof Any comments? Has any @BVLC member tried to use multi-GPU training with python 3?

GaryKT:
How would the wrapper work? Can we do a wrapper in python?

willyd:
No. The wrapper needs to be C++. 

Try to replace your CAFFE_ROOT/python/caffe/_caffe.cpp with

https://gist.github.com/willyd/0dbd1fabb06eeedc3289e656be03a022

Let me know if that works for you.

GaryKT:
@willyd 
We did a quick compile now there is a different error:
Traceback (most recent call last):
[...]

PR #5400, opened 2017-03-14
Added support to use NCCL with python3 and fixed nvml.dll related error under windows

willyd:
WIP. This may be of interest for the master branch since the python 3 error is not OS specific. I may split this PR in two.


willyd:
@GaryKT can you confirm that this PR fixes #5401 and #5347 for you as well?

shelhamer:
Closing as presumed fixed until hearing otherwise.

willyd:
@shelhamer I will keep this one open since I would like that fix to go in on windows. There are two fixes in this PR:

1. Getting nccl to correctly load the nvml.dll since it is not on `PATH` by default. This one is Windows specific.
2. Getting the uid's to work on python3 since python 3 and python 2 handle strings differently. Although I have not tested this change on a UNIX platform I believe the changes could also benefit the master branch.

cypof:
I tested the first commit about uid and it works, can you remove the second commit so that we can merge?

willyd:
@cypof I wrote the code in a rush so let me clean it up (I think I can come up with less changes) and then I will resubmit the first commit to master.

willyd:
@cypof @shelhamer I could simplify the first commit if we use a relatively recent version of boost. 

According to the docs we require 1.55 but according to the [CMake](https://github.com/BVLC/caffe/blob/master/cmake/Dependencies.cmake#L8) build we require 1.46, I assume that the docs is more up to date than the CMake build. Am I correct?

I am asking since 1.53 handles the conversion from python strings and bytes to C++ std::string automatically. If we can assume this we only have to convert manually from C++ to python.

See this: https://github.com/boostorg/python/blob/boost-1.53.0/src/converter/builtin_converters.cpp#L381

willyd:
Part of this was merge in #5527.

PR #5402, opened 2017-03-14
Opencl - replace deprecated method call in prelu layer backwards_gpu()

vpa1977:
While developing a 3d convnet I've noticed that prelu layer implementation uses a deprecated offset() method. Replaced it with a "modern" vector-based one.


naibaf7:
Ok, needs reviewing. Thanks.

naibaf7:
Does not work, runtests failed.

vpa1977:
Thanks, I'll look into it.

naibaf7:
It is rolled back. Please make sure these tests return without issue before the next PR:
`make lint -j4`
`make runtest -j4`
or if runtest has issues and aborts: `/build/test/test_all.testbin --gtest_filter=*LU* 1`

PR #5407, opened 2017-03-15
Pyloss backward bug

kli-nlpr:
```
template <typename Dtype>
void EuclideanLossLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
    const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom) {
  for (int i = 0; i < 2; ++i) {
    if (propagate_down[i]) {
      const Dtype sign = (i == 0) ? 1 : -1;
      const Dtype alpha = sign * top[0]->cpu_diff()[0] / bottom[i]->num();
      caffe_cpu_axpby(
          bottom[i]->count(),              // count
          alpha,                              // alpha
          diff_.cpu_data(),                   // a
          Dtype(0),                           // beta
          bottom[i]->mutable_cpu_diff());  // b
    }
  }
}
```

### DIFF
--- a/examples/pycaffe/layers/pyloss.py
+++ b/examples/pycaffe/layers/pyloss.py
@@ -34,4 +34,4 @@ class EuclideanLossLayer(caffe.Layer):
                 sign = 1
             else:
                 sign = -1
-            bottom[i].diff[...] = sign * self.diff / bottom[i].num
+            bottom[i].diff[...] = sign * top[0].diff[0] * self.diff / bottom[i].num


shaibagon:
If I'm not mistaken, `top[0].diff` holds the `loss_weight`.
You can see this term in the [cpp implementation of the loss][1]:

>     const Dtype alpha = sign * top[0]->cpu_diff()[0] / bottom[i]->num();

 [1]: https://github.com/BVLC/caffe/blob/master/src/caffe/layers/euclidean_loss_layer.cpp#L37

Issue #1726, opened 2015-01-15
HDF5 crashed when using layer HDF5_OUTPUT


tartavull:
I0114 19:41:35.961170 18670 hdf5_output_layer.cpp:33] Saving HDF5 file/usr/people/me/caffe/examples/test/test_prediction.h5
    I0114 19:41:35.962885 18670 hdf5_output_layer.cpp:38] Successfully saved 100 rows
    I0114 19:41:35.986912 18670 hdf5_output_layer.cpp:33] Saving HDF5 file/usr/people/me/caffe/examples/test/test_prediction.h5
    HDF5-DIAG: Error detected in HDF5 (1.8.11) thread 140309532535360:
      #000: ../../../src/H5D.c line 170 in H5Dcreate2(): unable to create dataset
        major: Dataset
        minor: Unable to initialize object
      #001: ../../../src/H5Dint.c line 439 in H5D__create_named(): unable to create and link to dataset
        major: Dataset
        minor: Unable to initialize object
      #002: ../../../src/H5L.c line 1638 in H5L_link_object(): unable to create new link to object
        major: Links
        minor: Unable to initialize object
      #003: ../../../src/H5L.c line 1882 in H5L_create_real(): can't insert link
        major: Symbol table
        minor: Unable to insert object
      #004: ../../../src/H5Gtraverse.c line 861 in H5G_traverse(): internal path traversal failed
        major: Symbol table
        minor: Object not found
      #005: ../../../src/H5Gtraverse.c line 641 in H5G_traverse_real(): traversal operator failed
        major: Symbol table
        minor: Callback failed
      #006: ../../../src/H5L.c line 1674 in H5L_link_cb(): name already exists
        major: Symbol table
        minor: Object already exists
    F0114 19:41:35.987098 18670 io.cpp:178] Check failed: status >= 0 (-1 vs. 0) Failed to make float dataset data
    *** Check failure stack trace: ***
        @     0x7f9c5635bdaa  (unknown)
        @     0x7f9c5635bce4  (unknown)
        @     0x7f9c5635b6e6  (unknown)
        @     0x7f9c5635e687  (unknown)
        @           0x4a0a41  caffe::hdf5_save_nd_dataset<>()
        @           0x4feee8  caffe::HDF5OutputLayer<>::SaveBlobs()
        @           0x5242d9  caffe::HDF5OutputLayer<>::Forward_gpu()
        @           0x45a6ab  caffe::Net<>::ForwardFromTo()
[...]

PR #5418, opened 2017-03-17
Fix a bug when using hdf5_output_layer 

77695:
The PR is mainly for bug fixing when using hdf5_output_layer saving data of the bottom layer. In the latest hdf5_outout_laye, all data is saved in the same file when iteration. This will cause hdf5 crashed(https://github.com/BVLC/caffe/issues/1726). In this code, index of the iteration is added in the output filename. 
e.g.
layer {
  name: "pool5_out"
  type: "HDF5Output"
  bottom: "pool5"
  bottom: "label"
  hdf5_output_param {
    file_name: "abc"
  }
}
output of HDF5Output layer is "abc_0", "abc_1", ...
Then there will be no error.

### DIFF
--- a/src/caffe/layers/hdf5_output_layer.cpp
+++ b/src/caffe/layers/hdf5_output_layer.cpp
@@ -1,5 +1,8 @@
-#include <vector>
 
+
+#include <vector>
+#include <string>
+#include <sstream>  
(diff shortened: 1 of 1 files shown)


77695:
Edit code to be better

PR #5419, opened 2017-03-17
Fix a bug when using hdf5_output_layer

77695:
The PR is mainly for bug fixing when using hdf5_output_layer saving data of the bottom layer. In the latest hdf5_outout_laye, all data is saved in the same file when iteration. This will cause hdf5 crashed(https://github.com/BVLC/caffe/issues/1726). In this code, index of the iteration is added in the output filename. 
e.g.
layer {
  name: "pool5_out"
  type: "HDF5Output"
  bottom: "pool5"
  bottom: "label"
  hdf5_output_param {
    file_name: "abc"
  }
}
output of HDF5Output layer is "abc_0", "abc_1", ...
Then there will be no error.

### DIFF
--- a/src/caffe/layers/hdf5_output_layer.cpp
+++ b/src/caffe/layers/hdf5_output_layer.cpp
@@ -1,8 +1,6 @@
-
-
-#include <vector>
+#include <sstream>
 #include <string>
-#include <sstream>  
+#include <vector>
 #include "hdf5.h"
 #include "hdf5_hl.h"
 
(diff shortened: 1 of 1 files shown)


williford:
I would suggest that `cnt_file` be initialized for each hdf5 output layer. For example, there could be multiple hdf5 output layers and you would not want `cnt_file` to be shared between them.

You might want to add a `ext`, so that you could have `abc_0.h5` instead of `abc.h5_0`. You might also want to consider allowing the file_name to be pattern, such as `abc_%05d.h5`.

77695:
@williford 

1. It seems true....I will fixed it later
2. Usually we dont add ext after fillename in linux. Thus I fixed this bug in a simplest way.

PR #5420, opened 2017-03-17
Added python 3 compatibility to cpp_lint.py

willyd:
This PR modifies cpp_lint.py so that it can be executed with both python 2 or 3.

### DIFF
--- a/scripts/cpp_lint.py
+++ b/scripts/cpp_lint.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/env python
 #
 # Copyright (c) 2009 Google Inc. All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


scripts/cpp_lint.py
```diff
@@ -52,6 +52,7 @@
 import sys
 import unicodedata
 
+PY_3K = (sys.version_info[0] == 3)
```
williford: Perhaps this should be:

`PY_3K = (sys.version_info[0] >= 3)`

It sounds like Python 4 will probably be mostly backwards compatible with Python 3 (Python 4 will be the version after Python 3.9).

scripts/cpp_lint.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/python
```
williford: According to http://stackoverflow.com/a/19305076/249226, the following would be recommended:

`#!/usr/bin/env python`

I'm not aware of any system where either shebang would fail. It looks like either should also work in Windows.

willyd:
@williford I applied the requested changes or used the `six` module and made use of `past.builtins` and `future.utils` to avoid impeding performance in python 2.

shelhamer:
Thanks for the py3 compatibility.

PR #5437, opened 2017-03-23
testdir path fix

BlGene:
added Atlas lapack Library name atllapack
added ABS_TEST_DATA_DIR var (only actually absolute for cmake)


shelhamer:
Thanks for adjusting this.

PR #5444, opened 2017-03-24
libdnn: add spatial convolution implmentation

wzw-intel:
A new class in libdnn LibDNNConvSpatial which is porting
from clcaffe's ConvolutionLayerSpatial class but with following
changes:
 - Use basic kernel as benchmark kernel
 - Add LibDNNConvConfig::phase_test to indicate using in TEST phase

### DIFF
--- a/include/caffe/greentea/libdnn.hpp
+++ b/include/caffe/greentea/libdnn.hpp
@@ -2,6 +2,7 @@
 #define CAFFE_GREENTEA_LIBDNN_HPP_
 
 #include <iomanip>
+#include <map>
 #include <memory>
 #include <string>
 #include <vector>
@@ -86,6 +87,7 @@ struct LibDNNConvConfig {
   bool fast_unsafe_math = false;
   bool weights_backward = true;
   bool bias_backward = true;
+  bool phase_test = true;
   libdnnConvolutionWeightAlgo_t wgalgo =
       LIBDNN_CONVOLUTION_WG_ALGO_ATOMIC;
   libdnnConvolutionBackwardAlgo_t bwalgo =
@@ -147,7 +149,6 @@ class LibDNN {
 #endif  // USE_CUDA
 
   std::string kernel_;
-
   bool fast_unsafe_math_;
 };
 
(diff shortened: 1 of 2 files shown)


wzw-intel:
@naibaf7 @gongzg
Please help to review this pull request for libdnn

naibaf7:
@wzw-intel Scheduled for review as discussed.

naibaf7:
So ok, this doesn't fully adhere to the LibDNN format yet, i.e. kernel generation is not embedded. Consequently, because the standalone LibDNN does not do kernel resource embedding, this won't work in the standalone.

wzw-intel:
@naibaf7 Thanks for your comments. I will refine the kernel source and make it more reusable and tuneable. Considering that the spatial convolution kernel is heavy dependent on the cl_intel_subgroups extension, it may not reproduce the same performance on another hardware, so we may not put a lot of efforts on supporting other hardware for now.

naibaf7:
Yes, but if we write the kernel source like the original LibDNNConv class does, then it might be possible to reuse the code modularly also on other hardware (skipping cl_intel_subgroup).
However I do know this is not strongly in your interest. But at least otherwise, the Intel spatial kernel implementation inside LibDNN needs to be independent of other Caffe components (such as kernel code embedding) so that I am able to pull it into the standalone library.

wzw-intel:
Sure, I will make an embedded version that has no dependence on Caffe components.

naibaf7:
Great :) after that it should only take about 1-2 weeks until it's available in the standalone.

wzw-intel:
@naibaf7  @gongzg
New version for LibDNNConvSpatial. This version make the kernel source embedded into host code and some refine for kernel source. Please help review, thanks.

naibaf7:
Ok, I'll check the implementation.

wzw-intel:
There is no backward implementation in this version, since currently we just focused on inference.

naibaf7:
Should I adapt it to work with the LibDNN fused convolution backward pass? Like that, inference would be done with LibDNN-IntelSpatial and backward pass would still work.

wzw-intel:
Great:) nice to have it.

naibaf7:
Ok, it will be ready and merged on approx. the 27th of April.

naibaf7:
@wzw-intel Postponed after NIPS deadline, sorry.

naibaf7:
@wzw-intel @gongzg please review the merge. I cleaned up some unnecessary variables and fixed a few other things. Next step should be replacing the normal Caffe Intel Spatial layer with the new implementation.

wzw-intel:
@naibaf7 Reviewed, very nice refining. 

For "replacing the normal Caffe Intel Spatial layer with the new implementation.", do you mean that we need to add a new subclass (i.e. LibDNNConvolutionSpatialLayer) of class ConvolutionLayer by using LibDNN? 

Another question is when we use "ConvolutionParameter_Engine_LIBDNN", which one do we choose as default convolution layer? LibDNNConvolutionLayer or LibDNNConvolutionSpatialLayer?

naibaf7:
@wzw-intel We could, for now, just pick LibDNNConvolutionSpatialLayer with the LIBDNN engine if we see that it is an Intel GPU, and otherwise use the LibDNNConvolutionLayer.

And you are correct, it needs a LibDNNConvolutionSpatialLayer. When that works we can remove the old spatial convolution layer.

Glad that you like the refining.

PR #5449, opened 2017-03-25
fix a bug when data_ or diff_ is shared

constroy:
When data_ or diff_ is shared, capacity_ should be updated to the smaller one.
And the shared memory should be reallocated in Reshape().

Otherwise, the capacity_ will be invalid, and memory will be still shared after Reshape(). That's not the right behavior.


Noiredd:
Please try to rebase your PR onto the most recent `master`. Tests failed for seemingly unrelated reasons and we've had several updates to Travis since your original submission (e.g. #5973).  
Consider [squashing](https://github.com/BVLC/caffe/wiki/Contributing#2-squashing-commits) your commits while you're at it.

PR #5460, opened 2017-03-28
Dependencies URLs updated for newer version

AhmedOS:


willyd:
Thanks for the consistency fix.

PR #5468, opened 2017-03-30
Fix typos in filename extensions in docs

zarak:
The links to the platform specific installation files are broken because they are pointing to `.html` instead of `.md`.


williford:
Where is it broken? The links are made to work after they are transformed by jekyll to html. See:
http://caffe.berkeleyvision.org/installation.html

PR #5470, opened 2017-03-30
Efficient GPU implementation of ReductionLayer forward pass

Callidior:
This PR fixes a performance issue with `ReductionLayer` that I came across when I tried to reduce a blob of size `10 x 512 x 17 x 23` to `10 x 512`.

On the CPU, this is done in 1 to 2 ms. On the GPU however, a simple `SUM` reduction took 1458 ms, dominating the forward pass of the entire network. This is due to the expensive loop over all outer axes used in the original code, which involved 5120 CUDA kernel instantiations in my case.

In this PR, I propose to get rid of that loop in GPU code entirely by replacing it with a single matrix-vector multiplication.

The following table shows a comparison of average forward pass runtime of the `ReductionLayer` on a blob of the aforementioned size using the old and the new code:

|  OP  |   old   |   new   |
|------|---------|---------|
| `SUM`  | 1458 ms | 0.17 ms |
| `ASUM`  | 1377 ms | 0.49 ms |
| `SUMSQ` | 1866 ms | 0.27 ms |

This corresponds to a speed-up by 3 orders of magnitude.


Callidior:
The PR has been rebased on top of c0597b1593685353ed806a05ca83a7fdb16cfb6d.

Callidior:
@shelhamer Is rebasing this PR required before code review can take place?

dragonQian:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #5474, opened 2017-03-31
Fixed memory leaks in cudnn conv and relu

willcrichton:


cypof:
activ_desc_ seems fine, but workspace should be a regular delete, not an array one right? It only contains aliases.

willcrichton:
Not sure what you mean by it only contains aliases. My understanding is that `delete []` is used if the deleted things was allocated with `new []`, which here is the case for `workspace`.

https://github.com/willcrichton/caffe/blob/a32114e6b2e098e2fdef47e397542b105eb58b66/src/caffe/layers/cudnn_conv_layer.cpp#L38

cypof:
OK yes it's fine. By aliases I mean the pointed data is owned by workspaceData.

PR #5010, opened 2016-11-21
Fix Python net drawing code

ngaloppo:
Fixes #3201

Issue #3201, opened 2015-10-15
Draw Net Problem (%d format: a number is required)


FreakTheMighty:
This seem similar to #1709. In the master branch, when running `./python/draw_net.py`, I get this traceback. Seems like this is an area that regresses frequently, what would a good test look like to prevent this from happening again?

```
root@0e28b3340f95:/opt/caffe# python ./python/draw_net.py ./foo.proto /data/net.png
libdc1394 error: Failed to initialize libdc1394
Couldn't import dot_parser, loading of dot files will not be possible.
Drawing net to /data/net.png
Traceback (most recent call last):
  File "./python/draw_net.py", line 45, in <module>
    main()
  File "./python/draw_net.py", line 41, in main
    caffe.draw.draw_net_to_file(net, args.output_image_file, args.rankdir)
  File "/opt/caffe/python/caffe/draw.py", line 213, in draw_net_to_file
    fid.write(draw_net(caffe_net, rankdir, ext))
  File "/opt/caffe/python/caffe/draw.py", line 195, in draw_net
    return get_pydot_graph(caffe_net, rankdir).create(format=ext)
  File "/opt/caffe/python/caffe/draw.py", line 142, in get_pydot_graph
    node_label = get_layer_label(layer, rankdir)
  File "/opt/caffe/python/caffe/draw.py", line 89, in get_layer_label
    layer.convolution_param.pad)
TypeError: %d format: a number is required, not RepeatedScalarFieldContainer
```

erogol:
change the corresponding function as following 
```
def get_layer_label(layer, rankdir):
    """Define node label based on layer type.

    Parameters
    ----------
    layer : ?
    rankdir : {'LR', 'TB', 'BT'}
        Direction of graph layout.

    Returns
    -------
    string :
        A label for the current layer
    """

    if rankdir in ('TB', 'BT'):
        # If graph orientation is vertical, horizontal space is free and
        # vertical space is not; separate words with spaces
        separator = ' '
    else:
        # If graph orientation is horizontal, vertical space is free and
        # horizontal space is not; separate words with newlines
        separator = '\\n'

    if layer.type == 'Convolution' or layer.type == 'Deconvolution':
        # Outer double quotes needed or else colon characters don't parse
        # properly
        param = layer.convolution_param
        node_label = '"%s%s(%s)%skernel size: %d%sstride: %d%spad: %d"' %\
                    (layer.name,
                     separator,
                     layer.type,
                     separator,
[...]

PR #5477, opened 2017-03-31
Fix draw_net() and add a test

lukeyeager:
Reverts https://github.com/BVLC/caffe/pull/5010

> A minimal but helpful improvement would be to add a Python test that imports caffe.draw and calls draw_net() to check that it at least runs without exception. That could be a good PR to encourage further testing.
@shelhamer https://github.com/BVLC/caffe/issues/3201#issuecomment-148872889

Done.

### DIFF
--- a/python/caffe/draw.py
+++ b/python/caffe/draw.py
@@ -104,11 +104,11 @@ def get_layer_label(layer, rankdir):
                       pooling_types_dict[layer.pooling_param.pool],
                       layer.type,
                       separator,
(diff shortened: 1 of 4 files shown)


shelhamer:
Thanks Luke for fixing the fix (sorry about that everybody!) and finally adding a test to make sure this doesn't happen again.

StevenPuttemans:
Thanks! :+1:

ascll82:
it really works，thanks a lot！

PR #5478, opened 2017-03-31
Add support for cuDNN v6

flx42:
Support for cuDNN v4 and v5 is preserved.

The Dockerfile will have to be modified when a new released is tagged.


shelhamer:
Thanks Felix for integrating the latest version of cuDNN!

PR #5480, opened 2017-04-01
Improved matlab interfaces for robustness

zhangyuanW:
**Major improvements**
1. Prevent crushing by installing failure function and throw-catch exception
in caffe_.cpp
2. Add LogSink to log error and warning to matlab, in caffe_.cpp
3. Add check for the return value from mex-calls in matlab code, to handle some of the common exceptions

**Reasons for the previous crushing**
google::glog call abort() by default when a Fatal message occurs. When loaded as a mex function, this causes matlab to crush.

**Implementation notes**
1. add first-call flag in failurefunction and matlabsink, since the error message may be sent more than once before it got handled.
2. use mexWarnMsgTxt to print (error) messages in matlab, since
mexErrMsgTxt results in crushing
3. remove reference to data_reader.cpp from libcaffe project

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -170,6 +170,12 @@ class Caffe {
   inline static void set_multiprocess(bool val) { Get().multiprocess_ = val; }
   inline static bool root_solver() { return Get().solver_rank_ == 0; }
 
+#ifdef MATLAB_MEX_FILE
+  // Whether is called from matlab
+  inline static bool using_Matlab() { return Get().using_Matlab_; }
+  inline static void set_using_Matlab(bool val) { Get().using_Matlab_ = val; }
+#endif
+
  protected:
 #ifndef CPU_ONLY
   cublasHandle_t cublas_handle_;
@@ -183,6 +189,11 @@ class Caffe {
   int solver_count_;
   int solver_rank_;
   bool multiprocess_;
+
+#ifdef MATLAB_MEX_FILE
+  // Use matlab
+  bool using_Matlab_ = false;
+#endif
 
  private:
   // The private constructor to avoid duplicate instantiation.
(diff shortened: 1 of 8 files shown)


include/caffe/common.hpp
```diff
@@ -170,6 +170,12 @@ class Caffe {
   inline static void set_multiprocess(bool val) { Get().multiprocess_ = val; }
   inline static bool root_solver() { return Get().solver_rank_ == 0; }
 
+#ifdef MATLAB_MEX_FILE
+  // Whether is called from matlab
+  inline static bool using_Matlab() { return Get().using_Matlab_; }
```
willyd: Why are those necessary?

matlab/+caffe/private/caffe_.cpp
```diff
@@ -586,6 +586,41 @@ static handler_registry handlers[] = {
...
+
+// Failure function to prevent Matlab from crushing
+// by throwing an exception
+// Failurefunction might be called multiple times, therefore set a first-call variable
+bool firstFail = true;
+void MyFailureFunction(){
+	if (Caffe::using_Matlab() && firstFail){
+		firstFail = false;
+		mexPrintf("failure function\n");
+		throw ("Fatal Error");
```
willyd: I think using the matlab error funcitons would make more sense here since the google C++ style guide forbids exceptions. Also the caffe code based was not really designed to be recovered from when a `CHECK` fails so this might have some unexpected side effects.

willyd:
Please rebase on the latest windows. And submit only the windows specific changes to the windows branch and the other changes to the master branch. Thanks for you contribution!

matlab/+caffe/private/caffe_.cpp
zhangyuanW: The idea here is that the error message will be printed to matlab from matlabsink, and the failure function (called at last) aims to jump to caffe_.cpp by throwing an error. Note that previously the failure function is simply abort(), I guess a normal exit in caffe_ might also work (?).

PR #5487, opened 2017-04-04
Additional fixes for draw_net

nitheeshas:
The previous fix #5477  for draw_net had some issues
 - `draw.py` still wasn't working for most of the nets.
 - The unittest didn't get executed:
   - unittest didn't have call for `unittest.main()` (not sure if there is any way to run unittest without this).
   - import statement for `text_format` should be `from google.protobuf import text_format`.

These are fixed in this PR.

### DIFF
--- a/python/caffe/draw.py
+++ b/python/caffe/draw.py
@@ -91,11 +91,11 @@ def get_layer_label(layer, rankdir):
                       separator,
                       layer.type,
                       separator,
-                      layer.convolution_param.kernel_size[0] if len(layer.convolution_param.kernel_size._values) else 1,
+                      layer.convolution_param.kernel_size[0] if len(layer.convolution_param.kernel_size) else 1,
                       separator,
-                      layer.convolution_param.stride[0] if len(layer.convolution_param.stride._values) else 1,
+                      layer.convolution_param.stride[0] if len(layer.convolution_param.stride) else 1,
                       separator,
-                      layer.convolution_param.pad[0] if len(layer.convolution_param.pad._values) else 0)
+                      layer.convolution_param.pad[0] if len(layer.convolution_param.pad) else 0)
     elif layer.type == 'Pooling':
         pooling_types_dict = get_pooling_types_dict()
         node_label = '"%s%s(%s %s)%skernel size: %d%sstride: %d%spad: %d"' %\
--- a/python/caffe/test/test_draw.py
+++ b/python/caffe/test/test_draw.py
@@ -1,7 +1,7 @@
 import os
 import unittest
 
-from google import protobuf
+from google.protobuf import text_format
 
 import caffe.draw
 from caffe.proto import caffe_pb2
@@ -29,5 +29,9 @@ class TestDraw(unittest.TestCase):
         for filename in getFilenames():
             net = caffe_pb2.NetParameter()
             with open(filename) as infile:
-                protobuf.text_format.Merge(infile.read(), net)
+                text_format.Merge(infile.read(), net)
             caffe.draw.draw_net(net, 'LR')
+
+
+if __name__ == "__main__":
+    unittest.main()


lukeyeager:
> unittest didn't have call for unittest.main() (not sure if there is any way to run unittest without this).

Ah, I usually use `nose` or `pytest` instead of unittest directly. My bad.

lukeyeager:
> draw.py still wasn't working for most of the nets.

What's the error you're seeing?

> The unittest didn't get executed. 

Actually, yes it does. Turn on the `-v` flag to check if you like:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 3056d75..d53be93 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ add_custom_target(lint COMMAND ${CMAKE_COMMAND} -P ${PROJECT_SOURCE_DIR}/cmake/l
 
 # ---[ pytest target
 if(BUILD_python)
-  add_custom_target(pytest COMMAND python${python_version} -m unittest discover -s caffe/test WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/python )
+  add_custom_target(pytest COMMAND python${python_version} -m unittest discover -v -s caffe/test WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/python )
   add_dependencies(pytest pycaffe)
 endif()
 ```

File: python/caffe/draw.py [hunk 1]
```diff
@@ -91,11 +91,11 @@ def get_layer_label(layer, rankdir):
...
                       layer.type,
                       separator,
-                      layer.convolution_param.kernel_size[0] if len(layer.convolution_param.kernel_size._values) else 1,
+                      layer.convolution_param.kernel_size[0] if len(layer.convolution_param.kernel_size) else 1,
                       separator,
-                      layer.convolution_param.stride[0] if len(layer.convolution_param.stride._values) else 1,
+                      layer.convolution_param.stride[0] if len(layer.convolution_param.stride) else 1,
                       separator,
-                      layer.convolution_param.pad[0] if len(layer.convolution_param.pad._values) else 0)
+                      layer.convolution_param.pad[0] if len(layer.convolution_param.pad) else 0)
```
lukeyeager: Why do you think this is necessary?
```py
>>> from caffe.proto import caffe_pb2
>>> net = caffe_pb2.NetParameter()
>>> layer = net.layer.add()
>>> layer.convolution_param.kernel_size
[]
>>> layer.convolution_param.kernel_size.append(1)
>>> layer.convolution_param.kernel_size
[1]
>>> len(layer.convolution_param.kernel_size)
1
>>> len(layer.convolution_param.kernel_size._values)
1
```
Looks like it should work as-is.

python/caffe/test/test_draw.py
```diff
@@ -1,7 +1,7 @@
 import os
 import unittest
 
-from google import protobuf
+from google.protobuf import text_format
```
lukeyeager: Personally, I feel like the existing syntax is slightly clearer.

python/caffe/test/test_draw.py
```diff
@@ -29,5 +29,9 @@ def test_draw_net(self):
         for filename in getFilenames():
             net = caffe_pb2.NetParameter()
             with open(filename) as infile:
-                protobuf.text_format.Merge(infile.read(), net)
+                text_format.Merge(infile.read(), net)
             caffe.draw.draw_net(net, 'LR')
+
+
+if __name__ == "__main__":
+    unittest.main()
```
lukeyeager: This isn't necessary when the `pytest` target is `python -m unittest discover`

python/caffe/draw.py
nitheeshas: I'm getting `AttributeError: 'google.protobuf.pyext._message.RepeatedScalarConta' object has no attribute '_values'`  when trying with the included example nets. 

This happens when using `protobuf>2.6.1` (NVIDIA#303). I have tested the PR with `protobuf==2.5.0` and it works fine, just in case.

python/caffe/test/test_draw.py
nitheeshas: The existing syntax is more clear, but its returning error:
`AttributeError: 'module' object has no attribute 'text_format'`

Replacing the import statement as mentioned in PR fixed it. `draw_net.py` also uses the same syntax, probably because of this issue.

python/caffe/test/test_draw.py
nitheeshas: Yes this isn't necessary when running the unittest as you mentioned. My bad!

lukeyeager:
Aha! So all the fixes are for `protobuf >= 3.0`. That makes sense. Sounds good to me.

shelhamer:
Thanks for the further fixes @nitheeshas !

PR #5491, opened 2017-04-05
Add example and small blurb about Sigmoid layer

williford:


shelhamer:
Thanks for the further detail!

PR #5503, opened 2017-04-08
Eltwise layer shape mismatch => log dimensions

brunobowden:
When layer shapes mismatch for the eltwise layer, caffe will fail a
check but doesn't give any information on how the shapes mismatch.
This logging information will make it easier to debug. Additionally
this reorders the variables to CHECK(expected == actual), matching
the JUnit convention.

```
BEFORE: Check failed: bottom[i]->shape() == bottom[0]->shape()
```
```
AFTER:  Check failed: bottom[0]->shape() == bottom[i]->shape()
        bottom[0]: 1 3 (3), bottom[1]: 1 5 (5)
```

NOTE: This removes use of CHECK_EQ in an earlier version of this PR,
which caused a build warning due to include of glog/stl_logging.h.

### DIFF
--- a/src/caffe/layers/eltwise_layer.cpp
+++ b/src/caffe/layers/eltwise_layer.cpp
@@ -31,7 +31,9 @@ template <typename Dtype>
 void EltwiseLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
   for (int i = 1; i < bottom.size(); ++i) {
-    CHECK(bottom[i]->shape() == bottom[0]->shape());
+    CHECK(bottom[0]->shape() == bottom[i]->shape())
+        << "bottom[0]: " << bottom[0]->shape_string()
+        << ", bottom[" << i << "]: " << bottom[i]->shape_string();
   }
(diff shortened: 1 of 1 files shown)


brunobowden:
The build produces the same build warning I mentioned in the original comment. I'll presume that glog won't be updated and modify the PR to drop the dependency on `glog/stl_logging.h`

_Warning:
/usr/include/c++/4.8/backward/backward_warning.h:32:2: warning: #warning This file includes at least one deprecated or antiquated header which may be removed without further notice at a future date. Please use a non-deprecated interface with equivalent functionality instead. For a listing of replacement headers and interfaces, consult the file backward_warning.h. To disable this warning use -Wno-deprecated. [-Wcpp]_

brunobowden:
I'd like feedback before I progress on this...

brunobowden:
Please review and merge this. It's a safe 3 line change that only changes logging. All lint and build checks have passed and it's been rebased on the latest master.

shelhamer:
Thanks for making the error more clear!

brunobowden:
@shelhamer - much obliged and thanks for the speedy review

PR #5506, opened 2017-04-09
Removed repeated imports Layer, get_solver

willyd:


shelhamer:
Thanks for the clean-up!

PR #5510, opened 2017-04-11
opencl: Fix segfault in spatial convolution engine when VIENNACL_CACHE_PATH is not set

gongzg:
It's a regression caused by the preivous patch.


gongzg:
@naibaf7 ping for review, thanks.

PR #5514, opened 2017-04-12
Fix log parsing

cypof:


shelhamer:
Isn't the python plot_training_log.py still broken too? By the way I just noticed #5376 fixes the sh issue, so perhaps that should be merged and then this PR can fix the py issue? Sorry I didn't see that before.

PR #5515, opened 2017-04-12
Allow using env vars for glog init from python

cypof:


shelhamer:
Thanks for the more flexible logging.

Issue #5518, opened 2017-04-12
Pycaffe partial inference/backprop fails for certain networks


Noiredd:
### Background
Network prototxt examples commonly name layers and their respective output blobs identically. Yet we often change layer names - if only to prevent loading their weights from a pretrained model, allowing partial transfer. Renaming their top blobs with them is tedious and error-prone, since in general case it is not straightforward to track which of the further layers will be using those blobs. Therefore, I often end up renaming only the layer, while its blob is left intact.

Another feature I make extensive use of is partial inference/backprop, that is call similar to `Net.forward(start='data', end='fc7')` - I can alter the activations in the middle of inference and analyze the subsequent change in prediction, for example. It turns out, both forward() and backward() passes result in a KeyError if the `end` layer's name does not match the name of the blob it produces.

### Issue summary
Both `caffe.Net()` and `caffe.Classifier()` work just fine for a standard setup where a layer's top blob name equals its own. But try to use the following [deploy.prototxt](https://gist.github.com/Noiredd/4b3692245501c989e368e6c7ed581675) - it is just an AlexNet setup whose `fc` layers were renamed, but their top blobs were not. Run the following code in your python terminal:
`import caffe`
`model = 'deploy.prototxt'`
`network = caffe.Net(model, caffe.TRAIN)`
`network.forward(end='fc6_renamed')`  
You should get an error similar to this:
`Traceback (most recent call last):`
`  File "<stdin>", line 1, in <module>`
`  File "/usr/local/caffe/python/caffe/pycaffe.py", line 183, in _Net_backward`
`    return {out: self.blobs[out].diff for out in outputs}`
`  File "/usr/local/caffe/python/caffe/pycaffe.py", line 183, in <dictcomp>`
`    return {out: self.blobs[out].diff for out in outputs}`
`KeyError: 'fc6_renamed'`

### Explanation and fix
Functions `_Net_forward` and `_Net_backward` in [pycaffe.py](../blob/master/python/caffe/pycaffe.py#L88) take a named, string type argument `end` to decide the extent of partial execution. The bug is caused by using the same string to decide the blob whose data is to be output. Therefore, if a layer's name does not match its top blob's name, two errors happen:  
KeyError is thrown if the correct layer name is passed, since a blob of this name does not exist  
ValueError is thrown if the correct blob name is passed, since a layer of this name does not exist.  
I have solved this issue (#5519) by retrieving the name of a blob the given layer produces, before attempting to extract its data.

### My system configuration
[...]

PR #5519, opened 2017-04-12
fixes pycaffe forward() and backward() behavior for issue #5518

Noiredd:
Fixes issue #5518.

### DIFF
--- a/python/caffe/pycaffe.py
+++ b/python/caffe/pycaffe.py
@@ -113,7 +113,7 @@ def _Net_forward(self, blobs=None, start=None, end=None, **kwargs):
 
     if end is not None:
         end_ind = list(self._layer_names).index(end)
-        outputs = set([end] + blobs)
+        outputs = set(self.top_names[end] + blobs)
     else:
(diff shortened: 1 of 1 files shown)


shelhamer:
Thanks for fixing this awkward issue first noted in #2829. See #2975 for an earlier fix that had to alter more at the time.

Noiredd:
The only thing I still ponder here: wouldn't it be better if `Net.backward()` returned diffs on the given layer's _bottom_ blobs instead of tops? I know code from before this merge would output diffs for the `end` layer's top, but this seems somewhat more logical. Just as top blobs are layers' output when traversing forward, bottoms are the outputs when traversing back. Even the [comment](../blob/master/python/caffe/pycaffe.py#L143) hints at this.
This would make life easier when analysing gradient flow in branched networks - backpropagating only to merge point currently only returns one blob, while it would be of more use to see gradients for each of the branch inputs. They are already calculated, just accessing them needs to be done manually. Let me know what you think, @shelhamer.

PR #5526, opened 2017-04-13
Bump boost version to 1.55 in CMake build

willyd:
As mentioned in #5400

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -5,7 +5,7 @@ set(Caffe_DEFINITIONS "")
 set(Caffe_COMPILE_OPTIONS "")
 
 # ---[ Boost
-find_package(Boost 1.46 REQUIRED COMPONENTS system thread filesystem)
+find_package(Boost 1.55 REQUIRED COMPONENTS system thread filesystem)
 list(APPEND Caffe_INCLUDE_DIRS PUBLIC ${Boost_INCLUDE_DIRS})
 list(APPEND Caffe_LINKER_LIBS PUBLIC ${Boost_LIBRARIES})
 
--- a/scripts/travis/install-deps.sh
+++ b/scripts/travis/install-deps.sh
@@ -9,10 +9,10 @@ apt-get -y update
 apt-get install -y --no-install-recommends \
   build-essential \
   graphviz \
-  libboost-filesystem-dev \
-  libboost-python-dev \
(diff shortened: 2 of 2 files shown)


willyd:
CMake currently requires 1.46
Travis installs 1.54
Documentation requires 1.55

Should settle for 1.54 everywhere or update travis build to match the requirement?

shelhamer:
I'm perfectly happy with requiring 1.54 or 1.55 everywhere. Let's pick 1.55?

willyd:
@shelhamer Ok great. Can you get the Travis build to install 1.55 or point me to what I should change and where?

shelhamer:
The dependencies for travis are handled by install-deps.sh and boost can be found here: https://github.com/BVLC/caffe/blob/master/scripts/travis/install-deps.sh#L12-L15

For trusty, I believe there are packages for boost 1.55 in particular but I don't know the names offhand. Could you have a look?

willyd:
Ok thanks. I'll see what I can find.

willyd:
@shelhamer I updated the Travis dependencies to 1.55 so this should be good to go.

shelhamer:
Thanks @willyd! 👍

pjreed:
For reference, this breaks compatibility with [ROS Indigo](http://wiki.ros.org/indigo) on Ubuntu 14.04 because it's linked against boost 1.54.  If I manually edit `cmake/Dependencies.cmake` to require 1.54, Caffe seems to compile without any problems.

willyd:
I wouldn't mind requiring 1.54 instead of 1.55 since it seems that this is what was used by default on the Travis build before this change went in. As mentioned the referenced issue. For me, 1.53+ is sufficient. I will let it to @shelhamer to decide if we should switch back to 1.54.

lukeyeager:
Please switch back to 1.54. That is the version [packaged for](http://packages.ubuntu.com/source/trusty/boost-defaults) Ubuntu 14.04. This change broke our DIGITS builds: https://travis-ci.org/NVIDIA/DIGITS/jobs/231102340.

shelhamer:
I'm completely fine with switching back to 1.54—I didn't realize compatibility with Ubuntu 14.04 was lost with this switch.

wilsonge:
For what it's worth Centos 7 includes Boost 1.53 - is it "OK" to drop compatibility with that too?

PR #5527, opened 2017-04-13
Added support for python 3 and NCCL

willyd:
@cypof I reworked the NCCL and python3 compatibility of #5400. Can you try this and let me know.


cypof:
OK, everything seems to work, thanks

willyd:
Thanks

lvzhaoyang:
Hi,

I have compiling issue which is 

```
/home/zlv30/framework/caffe/python/caffe/_caffe.cpp: In function ‘boost::python::api::object caffe::NCCL_New_Uid()’:
/home/zlv30/framework/caffe/python/caffe/_caffe.cpp:374:10: error: could not convert ‘uid’ from ‘std::__cxx11::string {aka std::__cxx11::basic_string<char>}’ to ‘boost::python::api::object’
   return uid;
```

which I think it is caused by this pull request. Can you take a look at it? 

I am using python 2 on ubuntu 16.04.

willyd:
This was supposed to be fixed in https://github.com/BVLC/caffe/commit/eeebdab16155d34ff8f5f42137da7df4d1c7eab0

@lvzhaoyang Can you confirm that you are using the latest master?

PR #5530, opened 2017-04-14
Explicit std::string to bp::object conversion

willyd:
Follow up on #5527, some compilers will require explicit conversion since boost::python::object is marked as `explicit`.


antran89:
:+1:

OUCliuxiang:
I am sorry that I am using the latest caffe-master, yet the similar problem like #5541 and #5629  still occurs on my desktop:

> Policy CMP0022 is not set: INTERFACE_LINK_LIBRARIES defines the link
  interface

I use ubuntu16.04, with anaconda4.1(python2.7), CUDA8.0, cuDNN5.1 all running normally.
And are anything could I do ?
(Also sorry for my English.)

PR #5531, opened 2017-04-14
Test for python forward and backward with start and end layer

shelhamer:
From #2975


shelhamer:
(Merged in spite of spurious CI failures due to fixed lint issue.)

PR #5545, opened 2017-04-17
Shape mismatch CHECK logging improvements

brunobowden:
- More logs when shapes mismatch
- Manually reviewed all matches of “grep -r CHECK.*shape src”
- Follow JUnit convention of:
```
  CHECK(expected == actual)
      << “expected: “ << expected
      << “ vs. actual: ” << actual;
```
Layers:
- Recurrent
- BaseConvolutionLayer
- Slice (improves logging for incorrect parameter)

CHECK_EQ
- Would prefer to use CHECK_EQ(expected, actual):
https://github.com/google/glog/blob/master/src/glog/stl_logging.h.in#L36
- However this requires include of stl_logging.h which introduces a
build warning. I believe this would work if caffe was updated to the
latest version of glog.


brunobowden:
Seeing flaky build failures in `test_backward_start_end`, logged as issue #5546. The cause is rather a mystery but might be connected to this PR.

Noiredd:
Merging. Thank you for the definite improvement!

Also, the randomly failing build has been solved in #5973.

PR #5547, opened 2017-04-18
Suppress skimage warning "The default mode, 'constant', will be changed to 'reflect'"

Banus:
A warning has been introduced in `skimage.transform.resize` due to a change in the default value for the padding mode parameter.
This change suppresses the warning by setting the old default value explicitly.


Banus:
Already fixed in #5969. Closing.

PR #5548, opened 2017-04-19
Rewrite crop layer GPU implementation

erictzeng:
The crop layer currently in Caffe is really slow on GPU. For example, in [fcn8s](https://github.com/shelhamer/fcn.berkeleyvision.org/tree/master/voc-fcn8s-atonce) on a 500x375 image, the final crop layer alone takes 8.3ms out of a 65.1ms forward pass (12.7%)!

This seems to be a result of the fact that the original GPU implementation is a fairly faithful reproduction of the CPU version. The CPU version is a series of recursive calls that eventually delegates to `caffe_copy` to copy a contiguous portion of the crop. The original GPU version is thus a similar series of recursive calls that eventually delegates to a CUDA kernel. This ends up being horribly inefficient in practice, since we are forced to sync after each copy, and we do a large number of copies (one for each leaf of the recursion tree).

This PR rewrites the GPU implementation to do the entire operation in a single kernel call. Under the same conditions as before, the new implementation takes 0.3ms for a forward pass, which is roughly a 28x speedup. In practice, the speedup depends on the size of the input, with the largest gains on the largest input blobs.

I think this should be good to go. Let me know if anything seems off.

### DIFF
--- a/include/caffe/layers/crop_layer.hpp
+++ b/include/caffe/layers/crop_layer.hpp
@@ -41,13 +41,15 @@ class CropLayer : public Layer<Dtype> {
   virtual void Backward_gpu(const vector<Blob<Dtype>*>& top,
       const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom);
 
-  vector<int> offsets;
+  Blob<int> offsets;
+  Blob<int> src_strides_;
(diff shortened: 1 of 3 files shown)


shelhamer:
Thanks for the speed-up Eric!

PR #5549, opened 2017-04-19
Added const getters and fixed constness of existing getters for vario…

ensafi:
Added const getters and fixed constness of existing getters for various parameters that applications may wish to track via the C++ API during training.  Added member variables to make temporary variables persistent and accessible via getters, including learning rate and mean scores (accuracy).

Also, undefined preprocessor macros that get redefined and sometimes cause conflicts.  Perhaps CAFFE_STRINGIFY and CAFFE_AS_STRING are better choices?

### DIFF
--- a/include/caffe/common.hpp
+++ b/include/caffe/common.hpp
@@ -19,7 +19,9 @@
 #include "caffe/util/device_alternate.hpp"
 
 // Convert macro to string
+#undef STRINGIFY
 #define STRINGIFY(m) #m
+#undef AS_STRING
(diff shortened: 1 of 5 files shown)


ensafi:
1. Added protocol buffer snapshotting support and a custom snapshot callback to facilitate user-defined behavior without necessarily relying on disk files for saving/restoring snapshots.
2. Renamed global macros with include CAFFE prefix: CAFFE_STRINGIFY and CAFFE_AS_STRING.

PR #5562, opened 2017-04-25
Support finding MKL on Windows

GlueCrow:


cmake/Modules/FindMKL.cmake
```diff
@@ -19,7 +19,13 @@ caffe_option(MKL_USE_STATIC_LIBS "Use static libraries" OFF IF NOT MKL_USE_SINGL
 caffe_option(MKL_MULTI_THREADED  "Use multi-threading"   ON IF NOT MKL_USE_SINGLE_DYNAMIC_LIBRARY)
 
 # ---[ Root folders
-set(INTEL_ROOT "/opt/intel" CACHE PATH "Folder contains intel libs")
+if(WIN32)
+  set(ProgramFilesx86 "ProgramFiles(x86)")
+  set(INTEL_ROOT $ENV{${ProgramFilesx86}}/IntelSWTools/compilers_and_libraries/windows CACHE PATH "Folder contains intel libs")
```
GlueCrow: Just preserve `INTEL_ROOT` in the original file.
But it's OK to replace it by `${MKL_ROOT}/..` I think.

willyd: Right. I missed that, sorry my bad.

GlueCrow: Is the OpenMP part still working?

willyd: If you are taking about https://github.com/BVLC/caffe/blob/master/cmake/Dependencies.cmake#L17-L30 

The default windows build links to openblas which does not need OpenMP to link properly. Maybe MKL needs it I have never tested caffe with MKL.

cmake/Modules/FindMKL.cmake
```diff
@@ -42,10 +47,12 @@ else()
     if(WIN32)
       list(APPEND __mkl_libs intel_c)
     else()
-      list(APPEND __mkl_libs intel gf)
+      list(APPEND __mkl_libs intel)
+      #list(APPEND __mkl_libs intel gf)
```
GlueCrow: Intel doesn't support mkl_gf* on windows

cmake/Modules/FindMKL.cmake
```diff
@@ -54,16 +61,25 @@ else()
      list(APPEND __mkl_libs sequential)
   endif()
 
-  list(APPEND __mkl_libs core cdft_core)
+  list(APPEND __mkl_libs core)
+  #list(APPEND __mkl_libs core cdft_core)
```
GlueCrow: CDFT is for cluster use, and it won't work if only link `mkl_cdft_core`.

willyd:
@GlueCrow LGTM. I'll give it a try in the next few days and get back to you. Have you tried if it still works on UNIX?

GlueCrow:
@willyd Haven't tried. I'll try on Ubuntu 16.04 later.
On Windows, when use `MKL_USE_SINGLE_DYNAMIC_LIBRARY=OFF,MKL_MULTI_THREADED=ON`,
it gives the following error:
`LINK : fatal error LNK1181: cannot open input file 'libiomp5md.dll'`
Still try to figure out what the problem it is.

GlueCrow:
Found the problem causes LNK1181 error:
Some of MKL's soft links are missing.
In my case, `libiomp5md.dll` can be found in `C:\Program Files (x86)\IntelSWTools\compilers_and_libraries_2017.2.187\windows\redist\intel64(_win)\compiler`,
but can't be found in `C:\Program Files (x86)\IntelSWTools\compilers_and_libraries\windows\redist\intel64(_win)\compiler`.
Now I have three solutions:
1. Add  `C:\Program Files (x86)\IntelSWTools\compilers_and_libraries_201X.X.XXX\windows\redist\intel64(_win)\compiler` to `PATH`
2. Execute  `C:\Program Files (x86)\IntelSWTools\compilers_and_libraries_201X.X.XXX\windows\mkl\bin\mklvars.bat intel64`,
not `C:\Program Files (x86)\IntelSWTools\compilers_and_libraries\windows\mkl\bin\mklvars.bat intel64`
3. Complete the missing soft links

willyd:
@GlueCrow I think that none of your solutions will work. My guess is that CMake is finding and tries to link to the `.dll` when it should find and link to the `.lib`. In any case you never need a `.dll` at build or link time.

In my install I have `"C:\Program Files (x86)\IntelSWTools\compilers_and_libraries\windows\compiler\lib\intel64\libiomp5md.lib"` but by looking at https://github.com/GlueCrow/caffe/blob/63bd20a796c2f183ced0b641365623c3b949fee7/cmake/Modules/FindMKL.cmake#L100-L103 CMake should be able to find it. Maybe you need to delete some of your cache variables?

GlueCrow:
@willyd I don't think it is because CMake tries to "link" to the `.dll`.
In https://github.com/BVLC/caffe/blob/windows/cmake/TargetResolvePrerequesites.cmake#L179 ,
it will try to find all shared libraries required by the target and hard-link them after build.
I guess that is why VS drops some errors like LNK1181 which doesn't mean that it tries to link.

> 4>EXEC : warning : target 'libiomp5md.dll' is not absolute...
> 4>EXEC : warning : target 'libiomp5md.dll' does not exist...
> 4>  CMake Error at C:/Program Files/CMake/share/cmake-3.7/Modules/GetPrerequisites.cmake:816 (message):
> 4>    C:/Program Files (x86)/Microsoft Visual Studio
> 4>    14.0/VC/bin/x86_amd64/dumpbin.exe failed: 1181
> 4>
> 4>
> 4>
> 4>    Microsoft (R) COFF/PE Dumper Version 14.00.24215.1
> 4>
> 4>    Copyright (C) Microsoft Corporation.  All rights reserved.
Sorry for late to paste the output log

willyd:
I believe these are two different errors. 

The copying/hardlinking of .dlls is done after a successful link so the LNK1181 error must be a *real* link error. 

That being said, the dumpbin error above is definitely related to the copying/hardlinking and most probably due to the fact that the MKL dlls are not on your `PATH`.  You can try to fix the hardlinking error by modifying the TargetResolvePrerequesites.cmake file in a similar manner to the cuDNN https://github.com/BVLC/caffe/blob/windows/cmake/TargetResolvePrerequesites.cmake#L29-L35. That is, append the directory containing the mkl dlls to the `_directories` list.

PR #5569, opened 2017-04-26
typo: `-gpu` -> `--gpu`

ajschumacher:


PENGUINLIONG:
`-gpu` and `--gpu` are both accepted by Caffe. Not quite sure whether it should be considered as a problem.

ajschumacher:
Okay!

PR #5584, opened 2017-05-03
Removing the duplicated Reshape in Layer->Forward()

cyberfire:
Reshape is called already in Layer->SetUp().
The other call in Layer->Forward() is not necessary and will result in Reshape be called twice, while this behavior is never documented.

### DIFF
--- a/include/caffe/layer.hpp
+++ b/include/caffe/layer.hpp
@@ -413,7 +413,7 @@ template <typename Dtype>
 inline Dtype Layer<Dtype>::Forward(const vector<Blob<Dtype>*>& bottom,
     const vector<Blob<Dtype>*>& top) {
   Dtype loss = 0;
-  //Reshape(bottom, top);
+  // Reshape(bottom, top);
   switch (Caffe::mode()) {
   case Caffe::CPU:
     Forward_cpu(bottom, top);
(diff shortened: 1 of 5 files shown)


include/caffe/layer.hpp
```diff
@@ -413,7 +413,7 @@ template <typename Dtype>
 inline Dtype Layer<Dtype>::Forward(const vector<Blob<Dtype>*>& bottom,
     const vector<Blob<Dtype>*>& top) {
   Dtype loss = 0;
-  Reshape(bottom, top);
+  // Reshape(bottom, top);
```
antran89: I think the Reshape in Layer is necessary for those cases with inputs with different size in a batch. Also, the Reshape will take almost zeros cost if the blob is in correct shape. Hence, I do not think it is necessary to remove this line.

antran89:
I think the Reshape in Layer is necessary for those cases with inputs with different size in a batch. Also, the Reshape will take almost zeros cost if the blob is in correct shape. Hence, I do not think it is necessary to remove this line.

include/caffe/layer.hpp
cyberfire: The different size inputs of a batch should call net->reshape() explicitly to update all networking layers, instead of doing this magically by Forward(). There are two shortages to have a call reshape just before forward(): 1. bring overhead to forward for cases that have no shape change. 2. If reshape have allocated some resource, it will have to handle two different situations for the second call in Forward(): the shape changes and it is necessary to release and reallocate resource, or the shape does not change and it should not do resource reclaim and allocation.

It is really confusing to have a Reshape() in Forward, based on Reshape()'s semantic.

Thanks,

Cyber

antran89: @cyberfire You should you leave your PR open. The core developers will take a look at it.

PR #5588, opened 2017-05-04
Handling destruction of empty Net objects

### LINKED ISSUE #5346: matlab warning
make all ok
make test ok
make matcaffe ok
make mattest 7 passed
but when I ran "net = caffe.Net(model, weights, 'test');" in matlab,I got this
Warning: The following error was caught while executing 'caffe.Net' class destructor:
Error using caffe_
Usage: caffe_('delete_solver', hNet)

Error in caffe.Net/delete (line 72)
      caffe_('delete_net', self.hNet_self);

Error in caffe.Net (line 31)
        self = caffe.get_net(varargin{:}); 
> In caffe.Net (line 31) 

system configuration
Operating system:Ubuntu16.10
Compiler:gcc 5.4
CUDA version (if applicable):8.0
CUDNN version (if applicable):5.1
BLAS:openblas
Python or MATLAB version (for pycaffe and matcaffe respectively):matlab2016b

ShaggO:
When constructing a network, the internal object initialization of Matlab actually creates a new network and replaces the already instantiated object, causing an empty network to be deleted. This PR checks if the caffe net pointer is defined before calling the net-deletion-function in caffe_.cpp.
Fixes #5346

### DIFF
--- a/matlab/+caffe/Net.m
+++ b/matlab/+caffe/Net.m
@@ -69,7 +69,9 @@ classdef Net < handle
       self.blob_names = self.attributes.blob_names;
     end
     function delete (self)
-      caffe_('delete_net', self.hNet_self);
+      if ~isempty(self.hNet_self)
+        caffe_('delete_net', self.hNet_self);
+      end
     end
     function layer = layers(self, layer_name)
       CHECK(ischar(layer_name), 'layer_name must be a string');


shelhamer:
Thanks for the fix!

meixiaofeng:
This fix may still lead to error,I change the code to this, and everything gets all right:
        if self.isvalid
           caffe_('delete_net', self.hNet_self);
        end

meixiaofeng:
This fix may still lead to error,I change the code to this, and everything gets all right:
`if self.isvalid
caffe_('delete_net', self.hNet_self);
end`

mingzizongyouchongtu:
The similar issue that I have encountered as follow:
Error using caffe_ 
Usage: caffe_('delete_solver', hSolver)

Error in caffe.Solver/delete (line 40)
caffe_('delete_solver', self.hSolver_self);

The solution to this issue as the same to @meixiaofeng way. I change the code to as follow:
      if self.isvalid
        caffe_('delete_solver', self.hSolver_self);
      end


Finally, thx to @meixiaofeng.

sfarkya:
@GuoxiaWang I changed the code as you suggested, but I ran into this error as give in [link](https://github.com/BVLC/caffe/issues/6270)

PR #5595, opened 2017-05-06
Handling destruction of empty Solver objects

Johnqczhang:
When running 'make mattest', I found a similar issue with #5346 solved by #5588 that causes an empty solver to be deleted. This PR checks if the caffe solver pointer is defined before calling the delete_solver function in caffe_.cpp.


Noiredd:
I restarted the tests, this looks good now. Any comment from someone more experienced with MATLAB interface?

PR #5598, opened 2017-05-08
Update euclidean_loss_layer.hpp with corrected test

ZoroDerVonCodier:
previous test name doesn't appear to exist.


Noiredd:
Thanks for the catch, @ZoroDerVonCodier!

PR #5617, opened 2017-05-12
Downgrade boost requirement from 1.55 to 1.54

lukeyeager:
Address concern mentioned at https://github.com/BVLC/caffe/pull/5526#issuecomment-300934771.


shelhamer:
Thanks for the compatibility adjustment @lukeyeager!

PR #5619, opened 2017-05-12
Docker conda

yangcha:
Dockerfiles for building images of Caffe with Anaconda (Python 3.6 version).


zym1010:
looks like it's actually 3.5?

yangcha:
Yes. You are right. It is Python 3.5.

yangcha:
Yes. You are right. It is Python 3.5.

zym1010:
@yangcha  I think actually Caffe now works with Python 3.6, based on my experience...

PR #5633, opened 2017-05-19
Added pydot as a requirement

bhargav-patel:
I get an `ImportError: No module named pydot` error when running `./python/draw_net.py`.

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -15,3 +15,4 @@ python-gflags>=2.0
 pyyaml>=3.10
 Pillow>=2.3.0
 six>=1.1.0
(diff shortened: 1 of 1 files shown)


Noiredd:
Isn't `pydot` outside the requirements on purpose? This comment in [draw.py](https://github.com/BVLC/caffe/blob/691febcb83d6a3147be8e9583c77aefaac9945f8/python/caffe/draw.py#L7-L9) would indicate so. Maybe instead this should be documented more clearly, eg. in [installation tutorial](http://caffe.berkeleyvision.org/installation.html) among optional dependencies?

PR #5644, opened 2017-05-24
using of center_loss

busrakb:
how can i use center_loss? it is not working. i am taking this error;

[libprotobuf ERROR google/protobuf/text_format.cc:245] Error parsing text-format caffe.NetParameter: 1142:21: Message type "caffe.LayerParameter" has no field named "center_loss_param".


PENGUINLIONG:
Have you modified `caffe.proto` and define center_loss_param? If you don't need other additional functionality you can simply copy and replace `/src/caffe/proto/caffe.proto`.

It seems you misused the pull-request feature. If you want to ask a question you should consider to use caffe user group (which could be found in README.md). You send a pull-request when and only when you added features or improved Caffe.

Noiredd:
Please post issues related to specific forks of Caffe to respective repositories' issue trackers - in this case I suppose this is [ydwen/caffe-face](https://github.com/ydwen/caffe-face). Additionally, you can try asking in the [caffe-users list](https://groups.google.com/forum/#!forum/caffe-users).  
Closing this as [not a valid PR](https://github.com/BVLC/caffe/blob/master/CONTRIBUTING.md).

PR #5665, opened 2017-06-02
add ConvolutionDepthwise layer

sp2823:
https://arxiv.org/pdf/1704.04861v1.pdf
MobileNets: Efficient Convolutional Neural Networks for Mobile Vision Applications

convolution depthwise layer
faster and less memory than the "convolution layer with group" (with CuDNN and without CuDNN)

### DIFF
--- a/src/caffe/layers/conv_dw_layer.cpp
+++ b/src/caffe/layers/conv_dw_layer.cpp
@@ -1,0 +1,315 @@
+#include <algorithm>
+#include <vector>
+#include "caffe/filler.hpp"
+#include "caffe/layers/conv_dw_layer.hpp"
+
+namespace caffe {
+
+template <typename Dtype>
+void ConvolutionDepthwiseLayer<Dtype>::LayerSetUp(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  ConvolutionParameter conv_param = this->layer_param_.convolution_param();
+  if (conv_param.has_kernel_h() && conv_param.has_kernel_w()) {
+    kernel_h_ = conv_param.kernel_h();
+    kernel_w_ = conv_param.kernel_w();
+  } else {
+    if (conv_param.kernel_size_size() == 1)
+    {
+      kernel_h_ = conv_param.kernel_size(0);
+      kernel_w_ = conv_param.kernel_size(0);
+    }
+    else
+    {
+      kernel_h_ = conv_param.kernel_size(0);
+      kernel_w_ = conv_param.kernel_size(1);
+    }
+  }
+  if (conv_param.has_stride_h() && conv_param.has_stride_w()) {
+    stride_h_ = conv_param.stride_h();
+    stride_w_ = conv_param.stride_w();
+  } else {
+    if (conv_param.stride_size() == 1)
+    {
+      stride_h_ = conv_param.stride(0);
+      stride_w_ = conv_param.stride(0);
+    }
+    else
+    {
+      stride_h_ = conv_param.stride(0);
+      stride_w_ = conv_param.stride(1);
+    }
+  }
+  if (conv_param.has_pad_h() && conv_param.has_pad_w()) {
+    pad_h_ = conv_param.pad_h();
+    pad_w_ = conv_param.pad_w();
+  } else {
+    if (conv_param.pad_size() == 1)
+    {
+      pad_h_ = conv_param.pad(0);
+      pad_w_ = conv_param.pad(0);
+    }
+    else
+    {
+      pad_h_ = conv_param.pad(0);
+      pad_w_ = conv_param.pad(1);
+    }
+  }
+  if (conv_param.dilation_size() > 0)
+  {
+    if (conv_param.dilation_size() == 1)
+    {
+      dilation_h_ = conv_param.dilation(0);
+      dilation_w_ = conv_param.dilation(0);
+    }
+    else
+    {
+      dilation_h_ = conv_param.dilation(0);
+      dilation_w_ = conv_param.dilation(1);
+    }
+  }
+  else
+  {
+    dilation_h_ = 1;
+    dilation_w_ = 1;
+  }
+  vector<int> weight_shape(4);
+  weight_shape[0] = bottom[0]->channels();
+  weight_shape[1] = 1;
+  weight_shape[2] = kernel_h_;
+  weight_shape[3] = kernel_w_;
+  vector<int> bias_shape;
+  if (conv_param.bias_term())
+  {
+    bias_shape.push_back(bottom[0]->channels());
+  }
+  if (this->blobs_.size() == 0) {
+    if (conv_param.bias_term()) {
+      this->blobs_.resize(2);
+    } else {
+      this->blobs_.resize(1);
+    }
+    this->blobs_[0].reset(new Blob<Dtype>(weight_shape));
+    shared_ptr<Filler<Dtype> > weight_filler(GetFiller<Dtype>(conv_param.weight_filler()));
+    weight_filler->Fill(this->blobs_[0].get());
+    if (conv_param.bias_term()) {
+      this->blobs_[1].reset(new Blob<Dtype>(bias_shape));
+      shared_ptr<Filler<Dtype> > bias_filler(GetFiller<Dtype>(conv_param.bias_filler()));
+      bias_filler->Fill(this->blobs_[1].get());
+    }
+  }
+  this->param_propagate_down_.resize(this->blobs_.size(), true);
+}
+
+template <typename Dtype>
+void ConvolutionDepthwiseLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top) {
+  vector<int> top_shape;
+  top_shape.push_back(bottom[0]->num());
+  top_shape.push_back(bottom[0]->channels());
+  top_shape.push_back((bottom[0]->height() + 2 * pad_h_ - (dilation_h_ * (kernel_h_ - 1) + 1)) / stride_h_ + 1);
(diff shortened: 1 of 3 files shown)


src/caffe/layers/conv_dw_layer.cpp
```diff
@@ -0,0 +1,280 @@
...
+  weight_buffer_shape.push_back(bottom[0]->num());
+  weight_buffer_shape.push_back(top[0]->height());
+  weight_buffer_shape.push_back(top[0]->width());
+  weight_buffer_.Reshape(weight_buffer_shape);
+  vector<int> weight_multiplier_shape;
+  weight_multiplier_shape.push_back(bottom[0]->num());
+  weight_multiplier_shape.push_back(top[0]->height());
+  weight_multiplier_shape.push_back(top[0]->width());
+  weight_multiplier_.Reshape(weight_multiplier_shape);
+  caffe_set(weight_multiplier_.count(), Dtype(1),
```
fengziyong: caffe_set just for cpu_data @sp2823

sp2823: We only need to set mutable_cpu_data or mutable_gpu_data once.
There is a similar implementation of batch_sum_multiplier_ in BatchNormLayer.
If it is necessary, we should use caffe_set in Forward_cpu and caffe_gpu_set in Forward_gpu.

fengziyong: I mean caffe_set is just for pointer of cpu_data, and set data to pointer of gpu_data would crash.

Bardo91:
I am trying to merge the PR with caffe_ssd repository (https://github.com/weiliu89/caffe/tree/ssd). Now when I try to compile it I get `Field number 41 has already been used in "caffe.SolverParameter" by field "eval_type".` Is it possible to change the field number so they are compatible?

zj19921221:
请问，两个问题请教下：
1、您这个实现和caffe中用group实现有什么优点吗？
2、模糊理解了group不能够并行的，并不是很理解为什么加了for循环就不能并行？

NHZlX:
cpu下有待于优化

zjchuyp:
@sp2823
Is it faster by using forloop than using gemm in CPU mode?

mathmanu:
Great to see this work - I hope it gets merged soon. The correct name for this should be "DepthwiseSeparable". Just "Depthwise" gives almost the opposite meaning.

sp2823:
I didn't optimize the CPU mode because the Convolution layer with group is slow in GPU mode. You can use this code for training and use Convolution layer for prediction.

youngwanLEE:
Could you share your .prototxt which show how to set parameters? or test_examples?

mathmanu:
I have attached the files required to train the popular mobilenet model:

[imagenet_mobilenet1.0_2017-07-04_10-44-00.zip](https://github.com/BVLC/caffe/files/1123829/imagenet_mobilenet1.0_2017-07-04_10-44-00.zip)

I added the following code in layer_factory.cpp, GetConvolutionLayer() so that this layer will be called whever its appropriate to use:
  if(conv_param.num_output() == conv_param.group()) {
    return CreateLayerBase<ConvolutionDepthwiseLayer>(param, ftype, btype);
  }

There is a speedup when using the proposed ConvolutionDepthwise layer instead of Convolution layer. But it is not as much as I expected. 

If fact if I just comment the group parameter in all convolution layers in both train.prototxt and test.prototxt, so that the 3x3 convolution becomes are traditional 3x3 convolution instead of DepthWise seperable, it becomes slightly faster! This was not what I was expecting.

Is there something that I am missing? Please try the files that I shared.

sp2823:
You only need to edit the .prototxt file like this.
type: "Convolution"
type: "ConvolutionDepthwise"

ryusaeba:
@sp2823  How do I merge your implementation into my CAFFE? Just download the hpp/cpp is OK? Thanks :)

sp2823:
download the .hpp/.cpp/.cu file and compile

leolee96:
Hi @sp2823 ,
I am new to caffe. I got an error saying like "class caffe::ConvolutionParameter’ has no member named ‘kernel_size_size’" in conv_dw_layer.cpp when I'm trying to compile.
Any idea of this error?

SophieZhou:
Hi, @sp2823 
I have trained mobilenet using your code for 20 epoches, and the top1  about 52%, and about 76% for top5.  Do you have any experiments results? 
By the way, the code is very well. The speed for training is much faster than group convolution way.   And I hope the results are well too.

zj19921221:
@SophieZhou 你好，请问下在cpu下训练速度有变很快吗？性能达到什么程度。

birdwcp:
up

birdwcp:
you did not implement CuDNNConvolutionDepthWiseLayer.  Isn't it necessary?

src/caffe/layers/conv_dw_layer.cu
```diff
@@ -0,0 +1,223 @@
...
+    ConvolutionDepthwiseBiasForward<Dtype>
+          // NOLINT_NEXT_LINE(whitespace/operators)
+          <<<CAFFE_GET_BLOCKS(count), CAFFE_CUDA_NUM_THREADS>>>(
+        count, bias_data, num, channels,
+        top_height, top_width, top_data);
+  }
+}
+
+template <typename Dtype>
+__global__ void ConvolutionDepthwiseWeightBackward(const int nthreads,
```
violet17: It seems that it is different from the result of CPU implement.

7oud:
The new implement runs faster than "convolution layer with group" (with CuDNN and without CuDNN), but it seems no fast enough. e.g. AlexNet has more FLOPS than MobileNet, but runs faster. 
Cudnn v7 has already released and brings grouped convolution feature, maybe it will be faster, Did you try it?

violet17:
The implement depthwise convolution of tensorflow https://github.com/tensorflow/tensorflow/blob/2d23960e29c3efb1b30c5cb6f8f3368635b97f29/tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
which one is faster?

twmht:
@sp2823 

Do you compare the FPS with current caffe implementation?

libra7:
@sp2823  how to optimize the CPU mode?

dawuchen:
@mathmanu The reason  traditional 3x3 convolution   faster  than Depthwise separable conv is that caffe use forloop to execute the conv operation that with multiple group , it is very slow . But DepthWise separable conv has much less parms.

lijuan123:
hello, I added ConvolutionDepthwise layer cpp and cu in caffe_ssd in TX2 and recompile caffe, however, when I test the model for mobilenet, it shows the error:  cudnn_softmax_layer.cpp:15] Check failed: status == CUDNN_STATUS_SUCCESS (4 vs. 0)  CUDNN_STATUS_INTERNAL_ERROR  just after conv3/dw -> conv3/dw.   Should I make some changes for ConvolutionDepthwise layer cu file?

src/caffe/layers/conv_dw_layer.cpp
```diff
@@ -0,0 +1,280 @@
...
+        - (dilation_w_ * (kernel_w_ - 1) + 1)) / stride_w_ + 1);
+  top[0]->Reshape(top_shape);
+  vector<int> weight_buffer_shape;
+  weight_buffer_shape.push_back(bottom[0]->channels());
+  weight_buffer_shape.push_back(kernel_h_);
+  weight_buffer_shape.push_back(kernel_w_);
+  weight_buffer_shape.push_back(bottom[0]->num());
+  weight_buffer_shape.push_back(top[0]->height());
+  weight_buffer_shape.push_back(top[0]->width());
+  weight_buffer_.Reshape(weight_buffer_shape);
```
Noiredd: Do we seriously need a 6-dimensional buffer for weights? If I have a batch of 64 feature maps, let's say 256 channels, 32x32, and want to convolve with a 3x3 filter, this line would allocate 256*3*3*64*32*32 = 150,994,994 floats, so almost 600 MB - that sounds like a significant overkill.

Noiredd:
I have made some tests, both on a raw convolution-oriented benchmark and some actually useful network. This is indeed faster than the default Caffe convolution engine for grouped conv, but the RAM requirements are higher, and I'm not so sure if this is a good thing.  
On a synthetic 1x256x256x256 blob with a 5x5 convolution in group of 256 this PR gives a 33% speedup (312 ms -> 217 ms for a forward-backward pass) at a barely noticeable RAM increase (1924 MB -> 1987 MB). But on a more realistic network with several conv layers, some of them grouped some not, results varied more drastically: iteration time went down from 119 ms to 46 ms, but RAM usage jumped from 425 MB to 590 MB.

So the question is: can the same speedup be achieved while allocating less memory? Like I said in the comment, do we really need to have such a large buffer for weights?

chibai:
No offense, I'm new in github. I just want to know whether  ConvolutionDepthwise add to master branch or not?? 
If not, what's the major problem??

Phil-Lin:
Excuse me，I had download the core files and compile successfully. But it's wrong when training Net.
Was I miss some step??
*** Aborted at 1563169270 (unix time) try "date -d @1563169270" if you are using GNU date ***
PC: @     0x7fa004053d3e caffe::ConvolutionDepthwiseLayer<>::LayerSetUp()
*** SIGSEGV (@0x0) received by PID 25146 (TID 0x7fa00473bb00) from PID 0; stack trace: ***
    @     0x7fa0020d74b0 (unknown)
    @     0x7fa004053d3e caffe::ConvolutionDepthwiseLayer<>::LayerSetUp()
    @     0x7fa004150157 caffe::Net<>::Init()
    @     0x7fa00415289e caffe::Net<>::Net()
    @     0x7fa00410eb3a caffe::Solver<>::InitTrainNet()
    @     0x7fa004110005 caffe::Solver<>::Init()
    @     0x7fa00411031f caffe::Solver<>::Solver()
    @     0x7fa00412e4b1 caffe::Creator_SGDSolver<>()
    @           0x40a788 train()
    @           0x407578 main
    @     0x7fa0020c2830 __libc_start_main
    @           0x407e49 _start
    @                0x0 (unknown)

PR #5678, opened 2017-06-09
Add try/catch around python function calls.

hgaiser:
Every Python call from c++ should be guarded with a try/catch to print its Python error. Previously a Python exception would only print :

```
terminate called after throwing an instance of 'boost::python::error_already_set'
```
With this PR, caffe will print the entire stack trace where the exception occurred, making debugging 100x easier ;)

### DIFF
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -8,6 +8,14 @@
 #ifdef WITH_PYTHON_LAYER
 #include <boost/python.hpp>
 namespace bp = boost::python;
+
+void PrintPyError() {
+  PyObject * ptype, * pvalue, * ptraceback;
+  PyErr_Fetch(&ptype, &pvalue, &ptraceback);
+  PyErr_Print();
+  PyErr_Restore(ptype, pvalue, ptraceback);
+}
+
 #endif
 
 #include "hdf5.h"
@@ -128,8 +136,8 @@ void Net<Dtype>::Init(const NetParameter& in_param) {
 #endif
       layers_[layer_id]->SetUp(bottom_vecs_[layer_id], top_vecs_[layer_id]);
 #ifdef WITH_PYTHON_LAYER
-    } catch (bp::error_already_set) {
-      PyErr_Print();
+    } catch (bp::error_already_set const &) {
+      PrintPyError();
       throw;
     }
 #endif
@@ -542,7 +550,7 @@ Dtype Net<Dtype>::ForwardFromTo(int start, int end) {
       layer_loss = layers_[i]->Forward(bottom_vecs_[i], top_vecs_[i]);
 #ifdef WITH_PYTHON_LAYER
     } catch (bp::error_already_set) {
-      PyErr_Print();
+      PrintPyError();
       throw;
     }
 #endif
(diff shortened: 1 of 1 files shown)


hgaiser:
Kindly requesting a review.

shelhamer:
Hey @hgaiser, thanks for trying to make pycaffe friendlier but it seems the change is failing the exception test for continuous integration. Please have a look at that first.

hgaiser:
@shelhamer Apparently PyErr_Print clears the [error indicator](https://docs.python.org/2/c-api/exceptions.html#c.PyErr_PrintEx), so I changed it to fetch and restore the error indicator. It's a bit ugly if you ask me (PyErr_Print shouldn't clear the error indicator IMO) but apparently this is how it is intended by the Python people.

That was the cause of the RuntimeError not being thrown in Python, so the test case works now. We'll see what Travis has to say about it :)

hgaiser:
Seems two of the Travis jobs failed, but they appear unrelated to my changes. @shelhamer can you verify if these are something for me to be concerned about ?

shelhamer:
Sorry, on second look I'm confused by the need for this PR. When I raise an exception from a Python layer it is properly reported. Under what conditions do you encounter the error in this issue? What version of Python and boost::python are you using?

> Seems two of the Travis jobs failed, but they appear unrelated to my changes.

Those assertion failures are very strange. I have not seen them elsewhere, but they do not seem related to this PR either. I'll try to have a closer look soon.

hgaiser:
Try with the following prototxt:

```
name: 'ExceptionNet'

input: "x"
input_dim: 1
input_dim: 1
input_dim: 1
input_dim: 1

layer {
  type: 'Python'
  name: 'exception'
  top: 'exception'
  bottom: 'x'
  python_param {
    module: 'exception_layer'
    layer: 'ExceptionLayer'
  }
}
```

And corresponding python layer:

```
import caffe

class ExceptionLayer(caffe.Layer):
        def setup(self, bottom, top):
                pass
                pass
                raise RuntimeError
                pass
```

Place them in the same dir and run something like this:

```
PYTHONPATH=. caffe time --model net.prototxt
```

The error I'm seeing with master is:

```
I0630 00:22:24.016942  3414 caffe.cpp:360] Performing Forward
terminate called after throwing an instance of 'boost::python::error_already_set'
```

The error I am seeing with this PR is:

```
caffe.cpp:364] Performing Forward
  File "/home/hgaiser/dev/caffe-exception/exception_layer.py", line 9, in forward
    raise RuntimeError
RuntimeError
terminate called after throwing an instance of 'boost::python::error_already_set'
```

For the record, I'm using python 3.6 and boost 1.64.

hgaiser:
@shelhamer any update?

hgaiser:
Ping?

Noiredd:
I have restarted the tests that failed prior to #5973, they failed for unrelated reasons but pass now. We could resume review now, @shelhamer?

MikalaiDrabovich:
This could be very helpful for debugging - any updates on merging?

hgaiser:
Closing this PR due to inactivity.

PR #5697, opened 2017-06-15
parse_log.py modified for multiple test-nets

romangrothausmann:
parse_log.py can now distinguish between multiple test nets
this adds an extra column 'NetNum' to the output
which can be used for e.g. plotting each test net loss and accuracy separately:

caffe info_ps_lrinc_zoom log


Noiredd:

PR #5704, opened 2017-06-21
Fixed bug where make distribute duplicates python files

ArneSuppe:
To trigger bug:
make distribute
make distribute

The second call to make will copy the python files to distribute/python/python, not distribute/python


Noiredd:
I have restarted the tests that failed prior to #5973, it passes now. Thanks for catching this!

Issue #5718, opened 2017-06-29
ValueError: Mean shape incompatible with input shape in case of "./examples/images/cat.jpg"


leemgs:
### Issue summary
We can not run `$ python python/classify.py  examples/images/cat.jpg foo` command. 

Below is an error messages. 

```
 . . . . Upper Omission . . . . 
09:30:47.006155  5678 net.cpp:744] Ignoring source layer loss
Traceback (most recent call last):
  File "python/classify.py", line 138, in <module>
    main(sys.argv)
  File "python/classify.py", line 110, in main
    channel_swap=channel_swap)
  File "/work/sdc/caffe/python/caffe/classifier.py", line 34, in __init__
    self.transformer.set_mean(in_, mean)
  File "/work/sdc/caffe/python/caffe/io.py", line 259, in set_mean
    raise ValueError('Mean shape incompatible with input shape.')
ValueError: Mean shape incompatible with input shape.
```

### Steps to reproduce
 It is a procedure to train & test imagenet data set with ImageNet model

** Train **
We have to run below command to get model and label data. (It tooks long time. Approximately, 10m, 232MB, 405KB/s ). We can download the data files that was used at 2012 ILSVRC (ImageNet Large-Scale Visual Recognition Challenge) with the below commands.

```
$ time python scripts/download_model_binary.py models/bvlc_reference_caffenet
$ sh data/ilsvrc12/get_ilsvrc_aux.sh
```

** Test **
`$ python python/classify.py  examples/images/cat.jpg foo`

 

### Your system configuration
Operating system: Ubuntu 16.04.1 LT64bit
Compiler: GCC 5.4.0
CUDA version (if applicable):  8.0
CUDNN version (if applicable): 8.0
BLAS: 3.6.0
Python or MATLAB version (for pycaffe and matcaffe respectively): Python 2.7.12

Noiredd:
I assume #5719 resolved this? Please reopen if the issue persists.

PR #5719, opened 2017-06-29
Fix: mean shape in compatible with input shape

leemgs:
This commit is to fix issue #5718.

* reference:
1. https://groups.google.com/forum/#!topic/caffe-users/nBpWJCcJoCU
2. https://stackoverflow.com/questions/28692209/using-gpu-despite-setting-cpu-only-yielding-unexpected-keyword-argument

Signed-off-by: Geunsik Lim <geunsik.lim@samsung.com>

### DIFF
--- a/python/caffe/io.py
+++ b/python/caffe/io.py
@@ -256,7 +256,12 @@ class Transformer:
             if len(ms) != 3:
                 raise ValueError('Mean shape invalid')
             if ms != self.inputs[in_][1:]:
-                raise ValueError('Mean shape incompatible with input shape.')
+                in_shape = self.inputs[in_][1:]
+                m_min, m_max = mean.min(), mean.max()
+                normal_mean = (mean - m_min) / (m_max - m_min)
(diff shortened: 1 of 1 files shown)


Noiredd:
:+1:, I remember having to hack a fix like that into every caffe installation in order to run my models. Would be really great to see it in stock.

goodloop:
any problem with this pull request? why it have not been merged to master?

Noiredd:
The only thing I don't like is the `print` statement - when bulk-testing multiple nets with their own transformers it generates a lot of doubtfully useful output, which tended to be rather annoying. But other than that, do we have anything this PR?

leemgs:
@Yangqing, @jeffdonahue, @shelhamer, Please, have a look at this.

Noiredd:
@leemgs Please remove the `print` line and I will merge this.

leemgs:
@Noiredd I have removed the `print` line. Then resubmitted PR with `squash`.

Noiredd:
Thank you for the fix @leemgs!

Issue #5386, opened 2017-03-11
Caffe freezes when running in a child process


thertweck:
Caffe seems to freeze when multiprocessing is used to do predictions in a child process.

Following code may be used to reproduce:

```
import caffe
import multiprocessing as mp
import numpy as np

def main():
    pool = mp.Pool(processes=1)
    pool.apply_async(predict)
    pool.close()
    pool.join()

def predict():
    net = caffe.Net(network_file, caffe.TEST, weights=weights_file)
    arr = np.random.rand(1, 3, 227, 227)
    print 'before_forward'
    pred = net.forward(data=arr) # This call will never return
    print 'after_forward'

if __name__ == "__main__":
    main()
```

It seems to happen in CPU mode only, though.

My machine runs macOS 10.12.3 with a [recent](https://github.com/BVLC/caffe/commit/85ab6100a122042c7dfd4adaf06f4c0b2e71148d) Caffe.

PR #5734, opened 2017-07-03
Optimized accuracy calculation

Noiredd:
Accuracy calculation, particularly for segmentation tasks with `top_k: 1`, can be unnecessarily slow: note the [two nested for loops](https://github.com/BVLC/caffe/blob/f16b5f2eb96cbb97d9a4b2b7312a23cb16f43dac/src/caffe/layers/accuracy_layer.cpp#L62) (over the batch and over image pixels) with `partial_sort` inside. The main culprit here is the need to copy all data to a new container so we can sort it.

My proposal is to replace that with a dynamically updated `priority_queue`. Instead of copying every prediction vector and sorting it later, it is much faster to iterate over it just once and only copy an element when it's larger than the smallest of `k` currently best ones (provided by an automatically sorted container, priority queue being the fastest).

Benchmark settings:
* net: FCN_AlexNet, batch size 1,
* dataset: PASCAL_VOC (21 classes, 1464 images in training set, 1449 for validation, images are about 360x550) - about 250M top_k searches per epoch,
* solver setup: `max_iter: 14640`, `test_interval: 1464`, `test_iter: 1449`,
* hardware: Titan Z, 2x Xeon E5620 (2.4 GHz),
* software: Ubuntu 14.04, CUDA 8.0, cuDNN 5.1, ATLAS, DIGITS 5.1-dev.

Each build was ran 4 times (Titan Z has two cores so I ran two nets in parallel, making it effectively 2 times 2 runs); times were measured from job initialization to completion (as reported by DIGITS).

Results:

top_k | current master | priority_queue | optimal search*
--- | --- | --- | ---
**1** | 90m 00s | 58m 57s | 58m 34s
**5** | 98m 49s | 88m 42s | N/A

(*) for `top_k==1` we don't need any container, just remember the best element's value and index - as we see, this is only marginally faster so for the sake of code clarity I opted against having a separate case for that.

For me this is about 10% faster for top-5 accuracy, and over 30% faster for top-1.

### DIFF
--- a/src/caffe/layers/accuracy_layer.cpp
+++ b/src/caffe/layers/accuracy_layer.cpp
@@ -1,4 +1,5 @@
 #include <functional>
+#include <queue>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


shaibagon:
I think what can really help here is a GPU implementation for this layer...

shaibagon:
Look at PR#5836. 
In order to estimate `top_k` accuracy one does not need to sort / partially sort / priority_queue the scores AT ALL! 
We know **in advance** what is the predicted score for the "true label", we only need to go once over all the other predictions and make sure there are no more than `top_k` better predictions.
This way, we can estimate accuracy with O(L) (number of labels) per pixels without any fancy data structures.

Noiredd:
@shaibagon Indeed, we don't need to know _what did the network answer_, only whether the correct label was among the predictions - and without sorts/queues it's so much easier to do on GPU. This is pretty clever ;)
Closing this in favor of #5386

PR #5745, opened 2017-07-06
Opencl - half floating point support and introduce layer fusion for inference

gongzg:
This PR's major contributions are the following two features:
1. FP16 support for OCL device and CPU device.  As it depends on half clBLAS, it requires to use ISAAC to enable FP16 feature.
2. Layer fusion support. For inference optimization, some layer could be fused into one, for example: CONV + RELU, CONV + BN + SCALE + RELU etc.

The FP16 version currently could get about 91% pass rate. Most of the failures are for the gradient check and solver test cases. The inference function works fine and I already verified it on my tree with both SSD and yolo2. The FP16 version get very good performance gain compare the FP32 for yolo2 model. I will fix those failures eventually.


naibaf7:
Ok, I will review this in parallel while developing the improved Caffe for MPN+MOE as discussed. This may take a while.

naibaf7:
This has been reviewed, tested, merged and improved to the point where it should work on most OpenCL GPUs and on CUDA. Tested on GTX 1080, RX 480, Vega FE, Mali T760.

PR #5753, opened 2017-07-08
docs: update apt installation guide for Debian and Ubuntu

CDLuminate:
This PR changes not only the Debian guide but also the Ubuntu guide.


shelhamer:
Thanks for the installation update!

PR #5762, opened 2017-07-11
Add "append" mode for convert_imageset.

tonykcn:
Sometimes, we need to append additional images into lmdb/leveldb, especiallly in fine-tuning stage.

If we construct all data with convert_imageset from ZERO, it's inefficient when the image count is large.

If "append" is set explicitly, new image data are appended after existing rows, no need to touch existing constructed data.


tonykcn:
Sometimes, we need to append additional images into lmdb/leveldb, especiallly in fine-tuning stage.
If we construct all data with convert_imageset from ZERO, it's inefficient when the image count is large.
If "append" is set explicitly, new image data are appended after existing rows, no need to touch existing constructed data.

@shelhamer , could you help review and merge this change?

tools/convert_imageset.cpp
```diff
@@ -133,22 +152,23 @@ int main(int argc, char** argv) {
...
+    string key_str = caffe::format_int(line_id + start_line_id, 8)
+            + "_" + lines[line_id].first;
 
     // Put in db
     string out;
     CHECK(datum.SerializeToString(&out));
     txn->Put(key_str, out);
 
-    if (++count % 1000 == 0) {
+    if (++count % 10000 == 0) {
```
shaibagon: why did you change the number of items processed between prints?

shaibagon:
Thank you for making a contribution to caffe.

(1) why did you add `convert_imageset_ex.cpp`? 
(2) Can you please squash your commits into a single one?

Thank you.

tonykcn:
@shaibagon , thanks for your review. 
I just clean the commit, please take a look.

Thanks

shaibagon:
Is there an automatic test for this utility? I suppose you should add an automatic test for this functionality as well.

WormCoder:
I think you don't have to make a loop to get the number of items in the lmdb/leveldb. 
For lmdb, you can use this function  mdb_env_stat() to get the number of items.
And how do you re-shuffle the images in db?

Noiredd:
@shaibagon Ideally yes, we would like to cover with tests as much code as reasonably possible. But for now, is anything inside `convert_imageset.cpp` tested?

PR #5770, opened 2017-07-12
API updates in Logistic Regression example

skylarjhdownes:


shelhamer:
Thanks for the API update.

PR #5780, opened 2017-07-18
May be a bug in net.cpp?

gauss-clb:
line 94 may be a bug, blob_need_backward_[blob_id] is always false because line 386.  And from line 411 to 416 shows that need_backward equals layer_param.propagate_down(bottom_id), so I think it needs `blob_need_backward_[blob_id] = need_backward;`.

### DIFF
--- a/src/caffe/net.cpp
+++ b/src/caffe/net.cpp
@@ -413,6 +413,7 @@ int Net<Dtype>::AppendBottom(const NetParameter& param, const int layer_id,
   if (layer_param.propagate_down_size() > 0) {
     need_backward = layer_param.propagate_down(bottom_id);
   }
+  blob_need_backward_[blob_id] = need_backward;
   bottom_need_backward_[layer_id].push_back(need_backward);
   return blob_id;
 }


PENGUINLIONG:
Seems not, it's been set [here](https://github.com/BVLC/caffe/blob/367a6e2c7f6be1abbba6253f7e8da5bccaf51fc2/src/caffe/net.cpp#L159).

gauss-clb:
But it is set if `need_backward=true`, see [here](https://github.com/BVLC/caffe/blob/367a6e2c7f6be1abbba6253f7e8da5bccaf51fc2/src/caffe/net.cpp#L157). So I think it should be set previously.

PENGUINLIONG:
See [the loop](https://github.com/BVLC/caffe/blob/367a6e2c7f6be1abbba6253f7e8da5bccaf51fc2/src/caffe/net.cpp#L144-L151). And, by [default](https://github.com/BVLC/caffe/blob/367a6e2c7f6be1abbba6253f7e8da5bccaf51fc2/src/caffe/proto/caffe.proto#L302) it will be 1.0.

gauss-clb:
But user can set learning rate 0, and `need_backward` also can't work and equals to `false`.

PENGUINLIONG:
If the learning rate is 0, nothing would be changed in BP, which means there is no need to BP. If you want to change something inside a layer, you can do it at the end of forward process.

Noiredd:
> But `blob_need_backward_[blob_id]=true` and `param_spec->lr_mult()==0`, the `need_backward=true`. (...) It's not our intention.

But it *is* our intention. `blob_need_backward_[blob_id]` might have been set before due to `force_backward` being set for the entire network, which should have precedence over stuff like `lr_mult` being zero. If I explicitly forced the network to compute the backward pass, then I don't want to see this disabled simply because I forgot to check learning rate and left it at zero. And if I understand your change correctly, this is what would happen.

If I misunderstood this, please point it out. Until then I'm closing.

PR #5801, opened 2017-07-26
Upgrading SoftmaxWithLoss layer to accept also spatially varying weights

shaibagon:
adding "WeightedSoftmaxWithloss" layer. 
This new loss layer upgrades "SoftmaxWithLoss" layer to accept spatially varying non-negative weights for the loss.
Usage:

    layer {
      name: "weighted_loss"
      type: "WeightedSoftmaxWithLoss"
      bottom: "predictions"  # raw predictions, e.g., B-C-H-W
      bottom: "labels"  # per-pixel label, e.g., B-1-H-W
      bottom: "weights"  # per pixel loss weight, e.g., B-1-H-W
      top: "loss"
      softmax_param { axis: 1 }
      loss_param { ignore_label: -1 normalization: VALID } # normalize by the SUM of the valid weights
    }

This PR includes GPU implementation and tests. 

This PR is **SELF CONTAINED**: it has NO changes to other code in caffe (not even to caffe.proto!).

### DIFF
--- a/include/caffe/layers/weighted_softmax_loss_layer.hpp
+++ b/include/caffe/layers/weighted_softmax_loss_layer.hpp
@@ -1,0 +1,153 @@
+#ifndef CAFFE_WEIGHTED_SOFTMAX_WITH_LOSS_LAYER_HPP_
+#define CAFFE_WEIGHTED_SOFTMAX_WITH_LOSS_LAYER_HPP_
+
+#include <vector>
+
+#include "caffe/blob.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+#include "caffe/layers/loss_layer.hpp"
+#include "caffe/layers/softmax_layer.hpp"
+
+namespace caffe {
+
+/**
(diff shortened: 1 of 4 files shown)


shaibagon:
@shelhamer for some reason **one** of the automatic Travis tests fails on some python test. It is very odd as this PR did not touch the python interface, and it seems like the same test passed the autotest for all other configurations. Could it be an issue with the auto  test?

shelhamer:
@shaibagon looks to be the fault of the test, although on inspection the test does seem to be fine. Let's leave that to another investigation.

Ideally we could add weighting across all losses, unified with ignoring the loss (by setting the corresponding weight to zero instead), but this is useful so extending only the softmax loss in this way would be alright for now. However, could this be done by inheriting from the existing `SoftmaxWithLossLayer` instead of coding it all again?

shaibagon:
@shelhamer you are right. I should make this loss a derivative of class softmaxWithLoss. I'll make the change and update. 

Thanks!

shaibagon:
PR #5828 replaces this PR. @shelhamer The new PR implements this loss as a derived class of `"SoftmaxWithLoss"` layer.

PR #5813, opened 2017-08-01
[DOC][FIX] fix web demo install instruction link

jqueguiner:


Noiredd:
This is only a problem when the demo is browsed from GitHub - when read on berkeleyvision.org, it did and does work properly. An ideal solution would redirect a GitHub visitor to the [GitHub version](https://github.com/BVLC/caffe/blob/master/docs/installation.md) of installation instructions, and work as is for a caffe.berkeleyvision.org visitor.  
Since I doubt that is possible without some deeper changes to the mechanism that pulls GH docs to Berkeley pages, I'm merging this as the best compromise.

I have restarted the tests that failed prior to #5973.

PR #5824, opened 2017-08-04
Fix hardcode xcode path

ek9852:
User may not install xcoder into default directory
especially if there are two different versions of xcoder installed.


shelhamer:
Thanks for the build polish.

Issue #5615, opened 2017-05-12
can't use Ctrl+C to stop training process


liyangliu:
I update my caffe to the latest master branch and installed NCCL to support multi-GPU support. But when I hit Ctrl+C in the training process, it would not exit from the program and would stuck at saving caffemodel and solverstate. When killing the process using "kill pid", it won't be killed totally. I use nvidia-smi to check the GPU status and it said
"|    0     27148    C   Unknown Error                                  642MiB |
The process still occupied some GPU memory and I was not able to run the training again util I reboot my computer.

Operating system: Ubuntu 14.04
Compiler: g++
CUDA version (if applicable): 8.0
CUDNN version (if applicable): v5
BLAS: atlas
Python or MATLAB version (for pycaffe and matcaffe respectively):

bfdream:
I also met similar issue, but I can stop the process by kill -9.
The solver state seems also saved, but thread.join() meets deadlock.
I just change join() to try_join_for() for a temp solving.

svobora:
I have the same problem.

weiliu89:
Hope #5825 would help.

talrav:
Same problem here.

noirmist:
still not fixed..

soulslicer:
ctrl + \ will kill it.

PR #5825, opened 2017-08-05
Fix MultiGPU hangs when interrupting training

weiliu89:
I met the same problem as mentioned in #5615 for months. Inspired by @bfdream's comment and after some digging, I think it could solve the hanging issue after canceling the job during MultiGPU training.

### DIFF
--- a/src/caffe/parallel.cpp
+++ b/src/caffe/parallel.cpp
@@ -354,11 +354,12 @@ void NCCL<Dtype>::Run(const vector<int>& gpus, const char* restore) {
   // Run first solver on current thread
   Broadcast();
   solver_->Solve();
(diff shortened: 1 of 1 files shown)


cypof:
Sorry I don't think this will work. It might fix interrupted runs, but in the normal case when training to the end, if the barrier is after join, worker threads will wait indefinitely. Isn't that what you see?

This issue is not easy to fix because the signal mechanism in Caffe is not thread safe. I tried to make worker threads check on the main solver requested_early_exit_ flag. It seems to work, please try it and use something like this for your PR if it works for you:

https://github.com/cypof/caffe/commit/e995da3b26ce419f98e9689169ea5436b5aebafc

Issue #2968, opened 2015-08-24
SoftmaxWithLoss creation fails with with two top blobs


leifblaese:
Hi,

I ran into a problem with the SoftmaxWithLoss layer recently that I wanted to share because it seems it has not been fixed yet.
The softmax loss layer can have two outputs, the first holds the loss, the second one the probabilities computed by the attached softmax layer.

However, the attached softmax layer can only have one top blob, that is why when creating the loss layer, it goes like this: 
 - Create the loss layer
 - Duplicate the LayerParameter object called softmax_param used to build the softmax loss layer
 - create a new top and bottom vector where you push_back() the first bottom blob of the loss layer (the blob holding the data) and the first top blob of the loss layer. 
- Create the softmax layer using softmax_param, reshape it to have the bottom/top vector

The problem is, when you have more than two top blobs in the loss layer, you also have two loss_weight parameters in the layer_param_ class that is used. This creates a mess when setting up the softmax layer because it expects only one of these parameters and checks for it using CHECK_EQ(), in the function SetLossWeights(). That means in the end the layer creation fails.

So, a quick fix would be to change  SetLossWeights() to check for at least as many top blobs as there are loss_weights, instead of an equal amount of blobs and loss_weights. However, this is of course messy as well, so a better way would be to alter the softmax_param and only retain the first loss_weight parameter.

I hope that description helps and the issue has not been reported before; I at least did not find it.

zimenglan-sysu-512:
@leifblaese had you ever solved this problem?

siebenkopf:
It would be great if you could come back to this issue. It still seems to be unsolved, and it is more that one year old now.

shaibagon:
I think PR #5828 resolves this issue.

PR #5828, opened 2017-08-07
Upgrading SoftmaxWithLoss layer to accept also spatially varying weights

shaibagon:
adding "WeightedSoftmaxWithloss" layer.
This new loss layer upgrades "SoftmaxWithLoss" layer (and is derived from it) to accept spatially varying non-negative weights for the loss.

(replaces PR #5801 - due to change in implementation)

Usage:

```
layer {
  name: "weighted_loss"
  type: "WeightedSoftmaxWithLoss"
  bottom: "predictions"  # raw predictions, e.g., B-C-H-W
  bottom: "labels"  # per-pixel label, e.g., B-1-H-W
  bottom: "weights"  # per pixel loss weight, e.g., B-1-H-W
  top: "loss"
  softmax_param { axis: 1 }
  loss_param { ignore_label: -1 normalization: VALID } # normalize by the SUM of the valid weights
}
```
This PR includes GPU implementation and tests.

This PR 
- adds a new layer
- made minor changes to `"SoftmaxWithLoss"` layer 
   (a) changed type of `get_normalizer` argument ot `Dtype`
   (b) ignore `loss_weight` for the internal `"Softmax"` layer

### DIFF
--- a/src/caffe/layers/softmax_loss_layer.cpp
+++ b/src/caffe/layers/softmax_loss_layer.cpp
@@ -13,6 +13,8 @@ void SoftmaxWithLossLayer<Dtype>::LayerSetUp(
   LossLayer<Dtype>::LayerSetUp(bottom, top);
   LayerParameter softmax_param(this->layer_param_);
   softmax_param.set_type("Softmax");
+  // no loss weight for the Softmax internal layer.
+  softmax_param.clear_loss_weight();
   softmax_layer_ = LayerRegistry<Dtype>::CreateLayer(softmax_param);
   softmax_bottom_vec_.clear();
   softmax_bottom_vec_.push_back(bottom[0]);
@@ -57,14 +59,14 @@ void SoftmaxWithLossLayer<Dtype>::Reshape(
 
 template <typename Dtype>
 Dtype SoftmaxWithLossLayer<Dtype>::get_normalizer(
-    LossParameter_NormalizationMode normalization_mode, int valid_count) {
+    LossParameter_NormalizationMode normalization_mode, Dtype valid_count) {
   Dtype normalizer;
   switch (normalization_mode) {
     case LossParameter_NormalizationMode_FULL:
       normalizer = Dtype(outer_num_ * inner_num_);
       break;
     case LossParameter_NormalizationMode_VALID:
-      if (valid_count == -1) {
+      if (valid_count <= 0) {  // valid_count is Dtype, better use <= than ==
         normalizer = Dtype(outer_num_ * inner_num_);
       } else {
         normalizer = Dtype(valid_count);
--- a/src/caffe/layers/weighted_softmax_loss_layer.cpp
+++ b/src/caffe/layers/weighted_softmax_loss_layer.cpp
@@ -1,0 +1,156 @@
+#include <algorithm>
+#include <cfloat>
+#include <vector>
+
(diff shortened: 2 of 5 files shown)


shaibagon:
@shelhamer please see this updated PR. I made the weighted loss layer inherit from `"SoftmaxWithLoss"` as you recommended.

Thanks!

src/caffe/layers/softmax_loss_layer.cpp
```diff
@@ -13,6 +13,8 @@ void SoftmaxWithLossLayer<Dtype>::LayerSetUp(
   LossLayer<Dtype>::LayerSetUp(bottom, top);
   LayerParameter softmax_param(this->layer_param_);
   softmax_param.set_type("Softmax");
+  // no loss weight for the Softmax internal layer.
+  softmax_param.clear_loss_weight();
```
shaibagon: This change suppose to resolve issue #2968

shaibagon:
Note that one of the changes proposed in this PR to `"SoftmaxWithLoss"` layer resolves issue #2968.

src/caffe/layers/weighted_softmax_loss_layer.cpp
```diff
@@ -0,0 +1,156 @@
...
+      DCHECK_LT(label_value, this->prob_.shape(this->softmax_axis_));
+      DCHECK_GE(weight_value, 0);  // do not allow negative weights...
+      loss -= weight_value
+        * log(std::max(prob_data[i * dim + label_value * this->inner_num_ + j],
+                           Dtype(FLT_MIN)));
+      agg_weight += weight_value;
+    }
+  }
+  top[0]->mutable_cpu_data()[0] = loss
+    / this->get_normalizer(this->normalization_, agg_weight);
```
weiliu89: Have you tried to normalize by count instead of agg_weight? For example, [this](https://github.com/shicai/Weighted_Softmax_Loss/blob/master/weighted_softmax_loss_layer.cpp) weighted softmax loss implementation is normalized by count. Although I think yours makes a bit more sense, otherwise one might need to tune the loss weight a bit.

shaibagon: This is a good point. It seems like the right way to normalize. However this is not set in stone.

shaibagon:
@shelhamer thank you for "focus"ing on this PR. Is there anything I can do to ease the process of accepting this PR?

src/caffe/layers/weighted_softmax_loss_layer.cpp
weiliu89: I tried a bit on my dataset. I set different weight for different classes. But it looks like normalizing by `count` is better than `agg_weight`. Maybe I should tune the learning late or loss_weight a bit to achieve good results if I choose to use 'agg_weight'. Have you tried the two different ways of normalization?

shaibagon: @weiliu89 I use it mainly for semantic segmentation where each sample has several labels. 
How about changing caffe.proto and `get_normalizer` to have an additional normalizing method? @shelhamer what do **you** think about it?

giihyun:
@shaibagon Thank you for the post. I want to use this layer on my network. How can I add this layer to existing caffe? I installed the caffe with visual studio 2013 and my operating system is Windows.

PR #5836, opened 2017-08-10
upgrading Accuracy layer

shaibagon:
As noted by @Noiredd in PR #5734, current implementation of `"Accuracy"` layer is inefficient.
This PR makes **two** contributions:
**(1)** efficient CPU implementation, takes only O(L) for top_k per pixel prediction. 
Unlike PR#5734 no need for fancy priority_queue or other complex data structures.  
**(2)** GPU implementation: no need to sync host/device memory.

### DIFF
--- a/src/caffe/layers/accuracy_layer.cu
+++ b/src/caffe/layers/accuracy_layer.cu
@@ -1,0 +1,147 @@
+#include <vector>
+
+#include "caffe/layers/accuracy_layer.hpp"
+#include "caffe/util/math_functions.hpp"
+
+
+namespace caffe {
+
+template <typename Dtype>
+__global__ void AccuracyForwardGPU(const int nthreads,
+          const Dtype* bottom_data, const Dtype* label, Dtype* acc,
+          const int num, const int dim, const int spatial_dim,
+          const int num_labels, const int top_k,
+          const bool has_ignore_label_, const int ignore_label_,
+          Dtype* counts) {
+  CUDA_KERNEL_LOOP(index, nthreads) {
+    const int n = index / spatial_dim;
+    const int s = index % spatial_dim;
+    const int label_value = static_cast<int>(label[n * spatial_dim + s]);
+    const Dtype prob_of_true_class = bottom_data[n * dim
+                                                 + label_value * spatial_dim
+                                                 + s];
+    int num_better_predictions = -1;  // true_class also counts as "better"
+    if (has_ignore_label_ && label_value == ignore_label_) {
+      acc[index] = 0;
+      counts[index] = 0;
+    } else {
+      for (int k = 0; k < num_labels & num_better_predictions < top_k; k++) {
+        num_better_predictions +=
+          (bottom_data[n * dim + k * spatial_dim + s] >= prob_of_true_class);
+      }
+      acc[index] = (num_better_predictions < top_k);
+      counts[index] = 1;
+    }
+  }
+}
+
+template <typename Dtype>
+__global__ void AccuracyForwardWithPerClassGPU(const int nthreads,
+          const Dtype* bottom_data, const Dtype* label,
+          Dtype* acc, Dtype* counts,
+          const int num, const int dim, const int spatial_dim,
+          const int num_labels, const int top_k,
+          const bool has_ignore_label_, const int ignore_label_) {
+  CUDA_KERNEL_LOOP(index, nthreads) {
+    const int n = index / spatial_dim;
+    const int s = index % spatial_dim;
+    const int label_value = static_cast<int>(label[n * spatial_dim + s]);
+    const Dtype prob_of_true_class = bottom_data[n * dim
+                                                 + label_value * spatial_dim
+                                                 + s];
+    if (has_ignore_label_ && label_value == ignore_label_) {
+      // nothing to be done.
+    } else {
+      int num_better_predictions = -1;  // true_class also counts as "better"
+      for (int k = 0; k < num_labels & num_better_predictions < top_k; k++) {
+        num_better_predictions +=
+          (bottom_data[n * dim + k * spatial_dim + s] >= prob_of_true_class);
+      }
+      acc[label_value*nthreads + index] += (num_better_predictions < top_k);
+      counts[label_value*nthreads + index] = 1;
+    }
+  }
+}
+
+template <typename Dtype>
+void AccuracyLayer<Dtype>::Forward_gpu(
+    const vector<Blob<Dtype>*>& bottom, const vector<Blob<Dtype>*>& top) {
+  const Dtype* bottom_data = bottom[0]->gpu_data();
+  const Dtype* bottom_label = bottom[1]->gpu_data();
+  const int dim = bottom[0]->count() / outer_num_;
+  const int num_labels = bottom[0]->shape(label_axis_);
+  const int nthreads = outer_num_ * inner_num_;
+  // Since this memory is not used for anything,
+  // we use it here to avoid having to allocate new GPU
+  // memory to accumulate intermediate results in the kernel.
+  Dtype* acc_data = bottom[0]->mutable_gpu_diff();
+  if (top.size() == 1) {
+    // simple case - report only global accuracy.
+
+    // Similarly, this memory is never used elsewhere, and thus we can use it
+    // to avoid having to allocate additional GPU memory.
+    Dtype* counts = bottom[1]->mutable_gpu_diff();
+    // NOLINT_NEXT_LINE(whitespace/operators)
+    AccuracyForwardGPU<Dtype><<<CAFFE_GET_BLOCKS(nthreads),
+        CAFFE_CUDA_NUM_THREADS>>>(nthreads, bottom_data, bottom_label,
+        acc_data, outer_num_, dim, inner_num_, num_labels, top_k_,
+        has_ignore_label_, ignore_label_, counts);
+    Dtype acc;
+    caffe_gpu_asum(nthreads, acc_data, &acc);
+    Dtype valid_count;
+    caffe_gpu_asum(nthreads, counts, &valid_count);
+    if (valid_count > 0) {
+      top[0]->mutable_cpu_data()[0] = acc / valid_count;
+    } else {
+      top[0]->mutable_cpu_data()[0] = 0;
+    }
+  } else {
+    // need to report per-class accuracy as well
+
+    // allocate space for more detailed "counts"
+    nums_buffer_.ReshapeLike(*bottom[0]);
+    Dtype* counts = nums_buffer_.mutable_gpu_data();
+
+    caffe_gpu_set(bottom[0]->count(), Dtype(0), acc_data);
+    caffe_gpu_set(nums_buffer_.count(), Dtype(0), counts);
+
+    // NOLINT_NEXT_LINE(whitespace/operators)
+    AccuracyForwardWithPerClassGPU<Dtype><<<CAFFE_GET_BLOCKS(nthreads),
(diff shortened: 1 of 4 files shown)


shelhamer:
Thanks for boosting the efficiency!

Sorry for the total nitpick, but can you fix the file permissions changes that switch 100644 → 100755 for merge?

shaibagon:
@shelhamer Thankyou for "focus"ing on this PR. 
I just `chomd` the files and committed. 
Auto-tests seems to pass smoothly.

Is there anything else I can do to ease the process of accepting this PR?

Noiredd:
I just gave this PR a spin but the results don't look right... at first I noticed my DIGITS plots broken, like accuracy going out of scale. Then I dug a little and found the weird thing: it tends to stably increase even when the network is already at a loss minimum. It increases by about (accuracy for one minibatch)*(number of minibatches in a test epoch) at every epoch. It's just something I observed after 2 runs on 2 different nets and datasets (no time to investigate it deeper at the moment), so this may be a bit far-fetched theory, but something is definitely off. Maybe it's just for me though? Has anyone tested it _in action_ beside doing `make runtest`?

src/caffe/layers/accuracy_layer.cu
```diff
@@ -0,0 +1,147 @@
...
+  // Since this memory is not used for anything,
+  // we use it here to avoid having to allocate new GPU
+  // memory to accumulate intermediate results in the kernel.
+  Dtype* acc_data = bottom[0]->mutable_gpu_diff();
+  if (top.size() == 1) {
+    // simple case - report only global accuracy.
+
+    // Similarly, this memory is never used elsewhere, and thus we can use it
+    // to avoid having to allocate additional GPU memory.
+    Dtype* counts = bottom[1]->mutable_gpu_diff();
```
Noiredd: `acc_data` and `counts` need to be reset between iterations, like so:
```c++
caffe_gpu_set(bottom[0]->count(), Dtype(0), acc_data);
caffe_gpu_set(bottom[1]->count(), Dtype(0), counts);
```

Noiredd:
I ran a quick benchmark on a TITAN Z using FCN AlexNet on PASCAL VOC 2012 data (1464 training images, 1449 validation, 21 classes; test on entire val set every training epoch) at top 3 accuracy. Results look very good:
* 56m 1s on current master (`partial_sort`),
* 47m 9s on PR #5734 (`priority_queue`),
* 28m 35s on this PR.

One thing I noticed is I keep getting 100% result at the first test iteration - current master (and #5734) output much smaller (and more believable) numbers. Over the next epochs it looks normal again. This seems to only happen for segmentation networks, regardless of params (`top_k`, `ignore_label`).

shaibagon:
@Noiredd Thank you for spotting the bug! I updated `test_accuracy_layer.cpp` so that it run each forward test for several iterations: this way the test reproduce the error and the fix can be confirmed by the automatic testing.

src/caffe/layers/accuracy_layer.cu
shaibagon: @Noiredd this is a bit of an overkill to go over all the arrays. 
The bug is much simpler to fix. In line 32, there is a `+=` for the accumulator:
```
acc[index] += (num_better_predictions < top_k);
```
Replacing it with a simple `=` assignment is enough to fix the bug.

shaibagon:
@shelhamer After the very thorough review of @Noiredd, I made these changes:
(1) Updating `test_accuracy_layer.cpp` - run all tests for 3 iterations. This change catches the bug in the auto tests.  
(2) Fixed the bug in `accuracy_layer.cu` and verified the new tests are passing.

Thank you @Noiredd for your work to make this PR very high quality!

src/caffe/layers/accuracy_layer.cu
Noiredd: @shaibagon Yep this works too and is clearer. Regarding speed, it shaves about 30 seconds more (though bear in mind I didn't do any serious benchmarks with clean test environment, repeated runs, averaged times etc).

shaibagon: @Noiredd  Thank you!! 
You really upgraded this PR with all your tests and benchmarks.

Noiredd:
@shaibagon I can confirm that accuracies look good now. What I'm still worried about is that first iteration (before the training starts) when training FCNs. I load a "convolutionized" AlexNet and initialize it with converted pretrained model except the last layer, which is initialized with Gaussian noise. Results at the first iteration should be bad and that's what the current master build says:
```
I0824 10:35:16.934343 29462 solver.cpp:397]     Test net output #0: accuracy = 0.0374341
I0824 10:35:16.934394 29462 solver.cpp:397]     Test net output #1: accuracy_perclass = 0
I0824 10:35:16.934404 29462 solver.cpp:397]     Test net output #2: accuracy_perclass = 0
I0824 10:35:16.934412 29462 solver.cpp:397]     Test net output #3: accuracy_perclass = 0
I0824 10:35:16.934418 29462 solver.cpp:397]     Test net output #4: accuracy_perclass = 0
I0824 10:35:16.934425 29462 solver.cpp:397]     Test net output #5: accuracy_perclass = 0
I0824 10:35:16.934432 29462 solver.cpp:397]     Test net output #6: accuracy_perclass = 0
I0824 10:35:16.934438 29462 solver.cpp:397]     Test net output #7: accuracy_perclass = 0
I0824 10:35:16.934445 29462 solver.cpp:397]     Test net output #8: accuracy_perclass = 0
I0824 10:35:16.934453 29462 solver.cpp:397]     Test net output #9: accuracy_perclass = 0
I0824 10:35:16.934459 29462 solver.cpp:397]     Test net output #10: accuracy_perclass = 0
I0824 10:35:16.934466 29462 solver.cpp:397]     Test net output #11: accuracy_perclass = 0
I0824 10:35:16.934473 29462 solver.cpp:397]     Test net output #12: accuracy_perclass = 0
I0824 10:35:16.934479 29462 solver.cpp:397]     Test net output #13: accuracy_perclass = 0
I0824 10:35:16.934486 29462 solver.cpp:397]     Test net output #14: accuracy_perclass = 0
I0824 10:35:16.934494 29462 solver.cpp:397]     Test net output #15: accuracy_perclass = 0
I0824 10:35:16.934499 29462 solver.cpp:397]     Test net output #16: accuracy_perclass = 0
I0824 10:35:16.934506 29462 solver.cpp:397]     Test net output #17: accuracy_perclass = 0
I0824 10:35:16.934514 29462 solver.cpp:397]     Test net output #18: accuracy_perclass = 0
I0824 10:35:16.934520 29462 solver.cpp:397]     Test net output #19: accuracy_perclass = 0.0621118
I0824 10:35:16.934526 29462 solver.cpp:397]     Test net output #20: accuracy_perclass = 0.057971
I0824 10:35:16.934533 29462 solver.cpp:397]     Test net output #21: accuracy_perclass = 0.0510697
```
When I switch to this PR, this is what I get instead:
```
I0824 10:27:48.350045 29098 solver.cpp:397]     Test net output #0: accuracy = 1
I0824 10:27:48.350124 29098 solver.cpp:397]     Test net output #1: accuracy_perclass = 0.995859
I0824 10:27:48.350136 29098 solver.cpp:397]     Test net output #2: accuracy_perclass = 0.0621118
I0824 10:27:48.350144 29098 solver.cpp:397]     Test net output #3: accuracy_perclass = 0.0545204
I0824 10:27:48.350152 29098 solver.cpp:397]     Test net output #4: accuracy_perclass = 0.0710835
I0824 10:27:48.350158 29098 solver.cpp:397]     Test net output #5: accuracy_perclass = 0.0496894
I0824 10:27:48.350164 29098 solver.cpp:397]     Test net output #6: accuracy_perclass = 0.0662526
I0824 10:27:48.350172 29098 solver.cpp:397]     Test net output #7: accuracy_perclass = 0.0510697
I0824 10:27:48.350178 29098 solver.cpp:397]     Test net output #8: accuracy_perclass = 0.0876467
I0824 10:27:48.350185 29098 solver.cpp:397]     Test net output #9: accuracy_perclass = 0.0821256
I0824 10:27:48.350191 29098 solver.cpp:397]     Test net output #10: accuracy_perclass = 0.0848861
I0824 10:27:48.350198 29098 solver.cpp:397]     Test net output #11: accuracy_perclass = 0.0489993
I0824 10:27:48.350205 29098 solver.cpp:397]     Test net output #12: accuracy_perclass = 0.0517598
I0824 10:27:48.350213 29098 solver.cpp:397]     Test net output #13: accuracy_perclass = 0.0883368
I0824 10:27:48.350219 29098 solver.cpp:397]     Test net output #14: accuracy_perclass = 0.0545204
I0824 10:27:48.350225 29098 solver.cpp:397]     Test net output #15: accuracy_perclass = 0.05245
I0824 10:27:48.350232 29098 solver.cpp:397]     Test net output #16: accuracy_perclass = 0.307798
I0824 10:27:48.350239 29098 solver.cpp:397]     Test net output #17: accuracy_perclass = 0.0586611
I0824 10:27:48.350245 29098 solver.cpp:397]     Test net output #18: accuracy_perclass = 0.0393375
I0824 10:27:48.350252 29098 solver.cpp:397]     Test net output #19: accuracy_perclass = 0.0621118
I0824 10:27:48.350260 29098 solver.cpp:397]     Test net output #20: accuracy_perclass = 0.057971
I0824 10:27:48.350265 29098 solver.cpp:397]     Test net output #21: accuracy_perclass = 0.0510697
```
Accuracy = 1 should imply each per-class accuracy of 1 - am I not right?

In further iterations it drops to about 0.87 which is consistent with master. But this first one just doesn't make sense.

PS: @shelhamer will probably remind you about the `chmod` - you're setting 100755 again :P

shaibagon:
(1) Regarding the high accuracy at first iteration:
It might be the case that all predictions has same probability. This implementation checks top_k **stricktly greater** probabilities than this label, thus for same probabilities it will count as "accurate".
(2) Regarding accuracy=1 and per-class is much lower: that's odd. Does it happen in GPU? CPU? Both?
(3) file permissions :( I'll fix. again.

Noiredd:
Took me a while to get down to the bottom of this; turned out it was a problem of my model. Due to bad initialization, it outputted all zeros on the first iteration - so your answer (1) explains it. My fault of course. 

On the other hand though, if P out of N elements of the output probability vector have equal, high values, and the real answer is also among them, should that be treated as a hit even if P > `top_k`? Thinking from the point of view of a deployed model, such answer is not very informative.

shaibagon:
@Noiredd You are making a good point here: Accuracy layer should only count predictions that are **strictly** within top_k. If there are other predictions with equal probability, they should be counted as part of top_k.

I'll make the necessary changes

Noiredd:
@shaibagon Just modifying the inequality in the `num_better_predictions` increment from strict to `>=` does it.

shaibagon:
@Noiredd does that solves also the issue with Accuracy=1 while per-class accuracy is less than 1?

Noiredd:
@shaibagon Yes. Accuracy=1 and per-class accuracy<1 only happened when the network produced an all-zero output. With those changes, both total and per-class accuracy will be 0.

shaibagon:
@shelhamer @Noiredd 
(1) Issue with many predictions with same probability fixed.
(2) File permissions fixed (again). Hopefully this time it will last.

Thank you for your efforts on improving this PR!

Noiredd:
I have no further objections ;) Great contribution by @shaibagon!

vlomonaco:
Hello, I've just made a new issue (#5981) which may be related to this merge!

vlomonaco:
#5981

PR #5841, opened 2017-08-12
relu_layer speed up

HoboChen:


shaibagon:
Thank you for your contribution.
Can you please describe your proposed PR? what issue are you addressing and how?
Moreover, note that your PR did not pass the automatic travis tests. Look at the logs and correct the errors.
Make sure your code does not only compile, but also adheres to the coding standards (try `make lint` to see style errors).

Noiredd:
I vote to close this. Introducing a conditional statement does not seem to be a great idea in this case - the speed-up due to one less multiply-add and `std::min` is marginal, and I'm convinced that the slow-down due to a ton of failed branch predictions will make ReLU actually slower overall.

shaibagon:
@Noiredd agree.

Noiredd:
Sorry @shaibagon, I have deleted my comment almost immediately after writing it, and before noticing your reply, as I noticed that it was not exactly right.

In the comment I speculated that the branch prediction will slow this down overally, but this is not exactly true, since neither `eps` and `negative_slope` change between iterations - so this might actually breeze through the conditional at zero cost.

HoboChen:
hello there
so sorry for not responding to this thread
when i was a intern of microsoft, i worked to speed up caffe network, and in vs debug mode i noticed that relu_layer is very slow ( but fast in release mode ); 
the merge may speed up the relu layer for about 50%, and in my workstation (e5 1630 v4), single picture forward of googlenet/resnet is more than 1ms faster (release mode, just as for production use)
so are there anybody tell me how to fix the failed checks or just simply open another pull requests including my code, thanks

shaibagon:
@HoboChen Thank you for contributing to caffe.
Any changes/fixes to this PR can be done on the local branch you created for this feature. Once you push them to your fork, this PR will be updated accordingly.  
In order for this PR to move forward:

- [ ] Please **edit** your first (currently empty) message in this discussion, and provide a comprehensive description of this PR. Including the effect you observed on runtime.  
- [ ] what happens is `negative_slope` is negative? why does the code discard it in this case?
- [ ] verify that `backward` function does not require similar change: as the code currently stands there is a difference between `forward` and `backward`.
- [ ] Make sure your branch passes **all** automatic tests including code style check (`make lint`)

PR #5852, opened 2017-08-17
Adding Weighted Euclidean loss layer

AlexanderSavochkin:
Hi,

I propose weighted euclidean loss layer. I think it might be very useful, for example it can naturally handle missing labeling data in training of regression models.

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -1235,6 +1235,7 @@ message V1LayerParameter {
     DROPOUT = 6;
     DUMMY_DATA = 32;
     EUCLIDEAN_LOSS = 7;
+    WEIGHTED_EUCLIDEAN_LOSS = 40;
     ELTWISE = 25;
     EXP = 38;
     FLATTEN = 8;
--- a/src/caffe/test/test_weighted_euclidean_loss_layer.cpp
+++ b/src/caffe/test/test_weighted_euclidean_loss_layer.cpp
@@ -24,14 +24,16 @@ class WeightedEuclideanLossLayerTest : public CPUDeviceTest<TypeParam> {
         blob_top_loss_(new Blob<Dtype>()) {
     // fill the values
     FillerParameter filler_param;
-    GaussianFiller<Dtype> filler(filler_param);
-    filler.Fill(this->blob_bottom_data_);
+    filler_param.set_min(0);
+    filler_param.set_max(1);
+    GaussianFiller<Dtype> gaussian_filler(filler_param);
+    gaussian_filler.Fill(this->blob_bottom_data_);
     blob_bottom_vec_.push_back(blob_bottom_data_);
-    filler.Fill(this->blob_bottom_label_);
+    gaussian_filler.Fill(this->blob_bottom_label_);
     blob_bottom_vec_.push_back(blob_bottom_label_);
-    filler.Fill(this->blob_bottom_certainty_);
+    UniformFiller<Dtype> uniform_filler(filler_param);
+    uniform_filler.Fill(this->blob_bottom_certainty_);
     blob_bottom_vec_.push_back(blob_bottom_certainty_);
-
     blob_top_vec_.push_back(blob_top_loss_);
   }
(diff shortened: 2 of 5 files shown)


src/caffe/proto/caffe.proto
```diff
@@ -1235,6 +1235,7 @@ message V1LayerParameter {
     DROPOUT = 6;
     DUMMY_DATA = 32;
     EUCLIDEAN_LOSS = 7;
+    WEIGHTED_EUCLIDEAN_LOSS = 40;
```
shaibagon: (1) I think ordering here is done alphabetically; `WEIGHTED_EUCLIDEAN_LOSS` should be roughly at the bottom.
(2) `V1LayerParameter` is only for backward compatibility, I don't think you need to add your layer here anymore.

src/caffe/test/test_weighted_euclidean_loss_layer.cpp
```diff
@@ -0,0 +1,91 @@
...
+        blob_top_loss_(new Blob<Dtype>()) {
+    // fill the values
+    FillerParameter filler_param;
+    GaussianFiller<Dtype> filler(filler_param);
+    filler.Fill(this->blob_bottom_data_);
+    blob_bottom_vec_.push_back(blob_bottom_data_);
+    filler.Fill(this->blob_bottom_label_);
+    blob_bottom_vec_.push_back(blob_bottom_label_);
+    filler.Fill(this->blob_bottom_certainty_);
+    blob_bottom_vec_.push_back(blob_bottom_certainty_);
```
shaibagon: filling the weights with `Gaussian` filler will results with negative weights. Does it make any sense?

src/caffe/test/test_weighted_euclidean_loss_layer.cpp
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  void TestForward() {
+    // Get the loss without a specified objective weight -- should be
+    // equivalent to explicitly specifying a weight of 1.
+    LayerParameter layer_param;
+    WeightedEuclideanLossLayer<Dtype> layer_weight_1(layer_param);
+    layer_weight_1.SetUp(this->blob_bottom_vec_, this->blob_top_vec_);
+    const Dtype loss_weight_1 =
+        layer_weight_1.Forward(this->blob_bottom_vec_, this->blob_top_vec_);
```
shaibagon: Doesn't `Forward` return `void`? What are you testing here exactly? I would expect "forward test" to include an explicit manual computation of the loss and comparing the manual calculation to the value computed by the layer...

shaibagon:
Thank you for making this PR.
Please make sure your code passes the tests and follows this repository's coding standards: run `make lint` to run the automatic test.

AlexanderSavochkin:
Thanks @shaibagon for reviewing the pull request. I will reiterate after necessary fixes

src/caffe/proto/caffe.proto
AlexanderSavochkin: I will check if I can remove it. Thanks.

src/caffe/test/test_weighted_euclidean_loss_layer.cpp
AlexanderSavochkin: Mathematically speaking it should be correct result. Meaningless in terms of machine learning though. Will fix.

src/caffe/test/test_weighted_euclidean_loss_layer.cpp
AlexanderSavochkin: Forward returns DType (not void). It calculates loss based on Forward_cpu or Forward_gpu "top" output if the layer has any non-zero loss_weight. I saw similar code in test_infogain_loss_layer.cpp and test_euclidean_loss_layer.cpp. The test checks if loss function scales accordingly with loss_weight. I borrowed this from test_euclidean_loss_layer.cpp . Probably comparing with explicit calculation is better idea, I will change it in the next iteration.

AlexanderSavochkin:
Doing another iteartion

Issue #5771, opened 2017-07-13
Missing gflags, glog, lmdb


Zumbalamambo:
Im unable to install it in my machine. Im getting the following error

```
CMake Warning at C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:769 (message):
  Imported targets not available for Boost version
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:873 (_Boost_COMPONENT_DEPENDENCIES)
  C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:1501 (_Boost_MISSING_DEPENDENCIES)
  cmake/Dependencies.cmake:8 (find_package)
  CMakeLists.txt:46 (include)


CMake Warning at C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:769 (message):
  Imported targets not available for Boost version
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:873 (_Boost_COMPONENT_DEPENDENCIES)
  C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:1501 (_Boost_MISSING_DEPENDENCIES)
  cmake/Dependencies.cmake:8 (find_package)
  CMakeLists.txt:46 (include)


CMake Warning at C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:769 (message):
  Imported targets not available for Boost version
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:873 (_Boost_COMPONENT_DEPENDENCIES)
  C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:1501 (_Boost_MISSING_DEPENDENCIES)
  cmake/Dependencies.cmake:8 (find_package)
  CMakeLists.txt:46 (include)


CMake Error at C:/Program Files/CMake/share/cmake-3.9/Modules/FindBoost.cmake:1877 (message):
  Unable to find the requested Boost libraries.

  Unable to find the Boost header files.  Please set BOOST_ROOT to the root
  directory containing Boost or BOOST_INCLUDEDIR to the directory containing
  Boost's headers.
Call Stack (most recent call first):
  cmake/Dependencies.cmake:8 (find_package)
  CMakeLists.txt:46 (include)


[...]

PR #5854, opened 2017-08-18
add instructions regarding compiler choice

skylarjhdownes:
I ran into issues compiling Caffe in Windows, see issue #5771
I've infodumped what I thought might help someone who had been in my position here.  Please advise if there's a more suitable way to document.  

Cheers!


willyd:
Thanks.

PR #5865, opened 2017-08-21
Fix caffe rpath

cheshirekow:
The logic for setting the library RPATH checks whether or not
${CMAKE_INSTALL_PREFIX}/lib is a system directory, and if not
adds it to the library RPATH. However, caffe does not install
to ${CMAKE_INSTALL_PREFIX}/lib, it installs to
${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR} (from
GNUInstallDirs). CMAKE_INSTALL_LIBDIR may be something like
"lib/x86_64-linux-gnu"


Noiredd:

PR #5866, opened 2017-08-21
Update link to google style guide.

cijianzy:


Noiredd:
I have restarted the tests that failed prior to #5973, it passed (as it should have). Thank you for the fix!

PR #5872, opened 2017-08-24
Some batch scripts for facilitating running examples under Windows

Coderx7:
These are actually the scripts I created for myself more than a year ago but never got the chance to submit them as a pr.
so here they are, I also added a CIFAR100 example which I guess would be nice to have.


Noiredd:
I would recommend caution, as this PR introduces several third-party tools with their own licenses.

Other than that, it's hard to tell what this PR really does, due to its scarce documentation.

PR #5879, opened 2017-08-29
CuDNN 7 grouped convolution

kouyoumin:
Uses CuDNN 7's new grouped convolution instead of for loop.

### DIFF
--- a/include/caffe/util/cudnn.hpp
+++ b/include/caffe/util/cudnn.hpp
@@ -45,6 +45,12 @@ inline const char* cudnnGetErrorString(cudnnStatus_t status) {
     case CUDNN_STATUS_RUNTIME_PREREQUISITE_MISSING:
       return "CUDNN_STATUS_RUNTIME_PREREQUISITE_MISSING";
 #endif
+#if CUDNN_VERSION_MIN(7, 0, 0)
+    case CUDNN_STATUS_RUNTIME_IN_PROGRESS:
+      return "CUDNN_STATUS_RUNTIME_IN_PROGRESS";
+    case CUDNN_STATUS_RUNTIME_FP_OVERFLOW:
+      return "CUDNN_STATUS_RUNTIME_FP_OVERFLOW";
+#endif
   }
   return "Unknown cudnn status";
 }
--- a/src/caffe/layers/cudnn_conv_layer.cu
+++ b/src/caffe/layers/cudnn_conv_layer.cu
@@ -16,7 +16,11 @@ void CuDNNConvolutionLayer<Dtype>::Forward_gpu(
     Dtype* top_data = top[i]->mutable_gpu_data();
 
     // Forward through cuDNN in parallel over groups.
+#if CUDNN_VERSION_MIN(7,0,0)
+    for (int g = 0; g < 1; g++) {
+#else
     for (int g = 0; g < this->group_; g++) {
+#endif
       // Filters.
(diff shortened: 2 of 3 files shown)


fengziyong:
@kouyoumin Have you test the forward time?
I have try to use cuDNNv7 before, but I found it was slower than the original implementation in group mode.

Noiredd:
I have restarted the tests that failed prior to #5973, it passes now. It still needs a [squash](https://github.com/BVLC/caffe/wiki/Contributing#2-squashing-commits) before merging, however I think that for a review it's good as is.

One question @kouyoumin: is Travis even able to test cuDNN 7 right now? Compare with #5972, where the dependencies script was also modified.

twmht:
not fast enough.

Anyone have tested this on MobileNet?

Noiredd:
@kouyoumin Please note before rebasing that #5972 modified the `cudnn.hpp`, adding the v7 cases.

Noiredd:
I have made several tests using a simple network: a [1,256,256,256] input and a convolution with varying kernel size (3 and 5) and group parameter (16, 64, 256). cuDNN 7 installed, made sure layer uses the cuDNN engine, compared times for only-forward and complete forward-backward pass. I cannot see any difference with this PR.

Can it be due to the fact that it was done on an old Titan Z (compute cap 3.5)?

src/caffe/layers/cudnn_conv_layer.cu
```diff
@@ -61,19 +65,43 @@ void CuDNNConvolutionLayer<Dtype>::Backward_gpu(const vector<Blob<Dtype>*>& top,
...
       if (this->bias_term_ && this->param_propagate_down_[1]) {
-        CUDNN_CHECK(cudnnConvolutionBackwardBias(handle_[0*this->group_ + g],
+#if CUDNN_VERSION_MIN(7, 0, 0)
+        CUDNN_CHECK(cudnnConvolutionBackwardBias(handle_[g],
               cudnn::dataType<Dtype>::one,
               top_descs_[i],  top_diff + top_offset_ * g,
               cudnn::dataType<Dtype>::one,
               bias_desc_, bias_diff + bias_offset_ * g));
+#else
+        CUDNN_CHECK(cudnnConvolutionBackwardBias(handle_[g],
```
Noiredd: Shouldn't we leave `CUDNN_CHECK(cudnnConvolutionBackwardBias(handle_[0*this->group_ + g],` here?

include/caffe/util/cudnn.hpp
```diff
@@ -44,6 +44,12 @@ inline const char* cudnnGetErrorString(cudnnStatus_t status) {
 #if CUDNN_VERSION_MIN(6, 0, 0)
     case CUDNN_STATUS_RUNTIME_PREREQUISITE_MISSING:
       return "CUDNN_STATUS_RUNTIME_PREREQUISITE_MISSING";
+#endif
+#if CUDNN_VERSION_MIN(7, 0, 0)
+    case CUDNN_STATUS_RUNTIME_IN_PROGRESS:
+      return "CUDNN_STATUS_RUNTIME_IN_PROGRESS";
+    case CUDNN_STATUS_RUNTIME_FP_OVERFLOW:
+      return "CUDNN_STATUS_RUNTIME_FP_OVERFLOW";
```
Noiredd: Adding this fragement is no longer necessary (since #5972), please remove this while rebasing.

PR #5893, opened 2017-09-05
Adding weighted euclidean loss layer

AlexanderSavochkin:
Hi,

I propose weighted euclidean loss layer. I think it might be very useful, for example it can naturally handle missing labeling data in training of multivariable regression models. It is possible to assign low weight to missing labels while leverage information from existing labels by assigning them higher weight. 


AlexanderSavochkin:
Need to remove extra whitespace

PR #5894, opened 2017-09-05
Adding weighted euclidean loss layer

AlexanderSavochkin:
Hi,

I propose weighted euclidean loss layer. I think it might be very useful, for example it can naturally handle missing labeling data in training of multivariable regression models. It is possible to assign low weight to missing labels while leverage information from existing labels by assigning them higher weight.

### DIFF
--- a/include/caffe/layers/weighted_euclidean_loss_layer.hpp
+++ b/include/caffe/layers/weighted_euclidean_loss_layer.hpp
@@ -1,0 +1,102 @@
+#ifndef CAFFE_EUCLIDEAN_LOSS_LAYER_HPP_
+#define CAFFE_EUCLIDEAN_LOSS_LAYER_HPP_
+
+#include <vector>
+
+#include "caffe/blob.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+#include "caffe/layers/loss_layer.hpp"
+
+namespace caffe {
+
+/**
+ * @brief Computes weighted Euclidean (L2) loss @f$
+ *          E = \frac{1}{2N} \sum\limits_{n=1}^N w_n \left| \left| \hat{y}_n - y_n
+ *        \right| \right|_2^2 @f$ for real-valued regression tasks.
+ *
+ * @param bottom input Blob vector (length 3)
+ *   -# @f$ (N \times C \times H \times W) @f$
+ *      the predictions @f$ \hat{y} \in [-\infty, +\infty]@f$
+ *   -# @f$ (N \times C \times H \times W) @f$
+ *      the targets @f$ y \in [-\infty, +\infty]@f$
+ *   -# @f$ (N \times C \times H \times W) @f$
+ *      the weights @f$ w \in [0, +\infty] @f$
+ * @param top output Blob vector (length 1)
+ *   -# @f$ (1 \times 1 \times 1 \times 1) @f$
+ *      the computed Euclidean loss: @f$ E =
+ *          \frac{1}{2n} \sum\limits_{n=1}^N \w_n \left| \left| \hat{y}_n - y_n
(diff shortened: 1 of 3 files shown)


PENGUINLIONG:
You can keep committing to the same branch in your fork, those commits will be updated in your PRs.

include/caffe/layers/weighted_euclidean_loss_layer.hpp
```diff
@@ -0,0 +1,102 @@
...
+class WeightedEuclideanLossLayer : public LossLayer<Dtype> {
+ public:
+  explicit WeightedEuclideanLossLayer(const LayerParameter& param)
+      : LossLayer<Dtype>(param), diff_() {}
+  virtual void Reshape(const vector<Blob<Dtype>*>& bottom,
+      const vector<Blob<Dtype>*>& top);
+
+  virtual inline const char* type() const { return "WeightedEuclideanLoss"; }
+  /**
+   * Unlike most loss layers, in the EuclideanLossLayer we can backpropagate
```
shaibagon: Indeed your layer can propagate to `bottom[0]` and `bottom[1]`, but what about `bottom[2]`?
If this function was copied from `"EuclideanLoss"` layer, you need to consider if it is still appropriate here as well...

src/caffe/layers/weighted_euclidean_loss_layer.cpp
```diff
@@ -0,0 +1,67 @@
...
+    if (propagate_down[i]) {
+      const Dtype sign = (i == 0) ? 1 : -1;
+      const Dtype alpha = sign * top[0]->cpu_diff()[0] / bottom[i]->num();
+      caffe_cpu_axpby(
+          bottom[i]->count(),              // count
+          alpha,                              // alpha
+          diff_.cpu_data(),                   // a
+          Dtype(0),                           // beta
+          bottom[i]->mutable_cpu_diff());  // b
+      for (int j = 0; j < bottom[i]->count(); ++j) {
```
shaibagon: this loop of element-wise multiplication by the weights can be replaced with [`void caffe_mul(const int N, const Dtype* a, const Dtype* b, Dtype* y);`](https://github.com/BVLC/caffe/blob/master/include/caffe/util/math_functions.hpp#L65)

src/caffe/layers/weighted_euclidean_loss_layer.cpp
```diff
@@ -0,0 +1,67 @@
...
+    const vector<Blob<Dtype>*>& top) {
+  int count = bottom[0]->count();
+  caffe_sub(
+      count,
+      bottom[0]->cpu_data(),
+      bottom[1]->cpu_data(),
+      diff_.mutable_cpu_data());
+
+  Dtype wdot(0.0);
+  for (int i = 0; i < count; ++i) {
```
shaibagon: You can compute the loss in a slightly different way (assuming your weights are non-negative):
1. compute diff <- (x1-x2)
2. scale the diff by **sqrt** of the weights: sdiff <- sqrt(x).*diff
3. compute the dot product of <sdiff,sdiff>

These three steps can be easily carried out using off-the-shelf math functions that can be found in `math_functions.cpp`.
Once your forward/backward functions are implemented using caffe "math functions" it should be fairly easy to implement a GPU version of the layer - this can be really important for successful usage of the layer.

include/caffe/layers/weighted_euclidean_loss_layer.hpp
AlexanderSavochkin: From mathematical point of view it is doable. But what is the point to differentiate loss function by sample weight from machine learning point of view?

AlexanderSavochkin: Hi @shaibagon. I have implemented back propagation for weights (bottom[2]). Could you take a look? Thanks

PR #5900, opened 2017-09-06
Fix division operator for Compatibility of python 3 in classifier.py

wasnot:
On python 3, `Classifier`'s `predict ()` method is failed on image `reshape`, because using float division(/) on the reshape of the image at oversample.
On python 2, division operator `/` is integer division. But on python 3, `/` is float division and `//` is integer division(floored division).
And `//` division operator is compatible with python 2.

So, I replaced to integer division (//) which is compatible with both python 2/3.

### DIFF
--- a/python/caffe/classifier.py
+++ b/python/caffe/classifier.py
@@ -92,7 +92,7 @@ class Classifier(caffe.Net):
 
         # For oversampling, average predictions across crops.
         if oversample:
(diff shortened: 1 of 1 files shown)


longjon:
This is the correct thing to do, thanks!

(Unrelatedly, at least one of our tests seems to be failing nondeterministically ಠ_ಠ)

PR #5904, opened 2017-09-07
Expose GPU pointers to Python

longjon:
This is a rebased and cleaned-up version of a commit from @tnarihi, https://github.com/BVLC/caffe/pull/2102/commits/7b99436379141de709666d3910357cfd23988979.

The purpose of this is to expose the GPU blob pointers, permitting use of libraries like PyCUDA without needless and expensive blob roundtripping.

This commit has been modified from the original in the following ways:
* Wrapper functions are removed; instead the accessors are cast directly. This reduces the amount of code and also makes the low-level nature of the call a bit more evident at the place where such things are normally specified.
* `reinterpret_cast` is used instead of C-style casting.
* `uintptr_t` is used instead of `size_t`.
* An underscore is prepended to the property names, emphasizing that these are low-level, at-your-own-risk properties. Since these are raw pointers, it's up to the user to decide when they remain valid.

I am making this PR now to have minimal mergeable access to blob GPU memory. On the Python side, it's easy to write code to convert `Blob`s to (e.g.) `GPUArray`s. Such code could also be brought in as desired (see @tnarihi's PRs).

### DIFF
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -464,6 +464,14 @@ BOOST_PYTHON_MODULE(_caffe) {
     .add_property("count",    static_cast<int (Blob<Dtype>::*)() const>(
         &Blob<Dtype>::count))
     .def("reshape",           bp::raw_function(&Blob_Reshape))
+#ifndef CPU_ONLY
+    .add_property("_gpu_data_ptr",
+        reinterpret_cast<uintptr_t (Blob<Dtype>::*)()>(
+          &Blob<Dtype>::mutable_gpu_data))
+    .add_property("_gpu_diff_ptr",
+        reinterpret_cast<uintptr_t (Blob<Dtype>::*)()>(
+          &Blob<Dtype>::mutable_gpu_diff))
+#endif
     .add_property("data",     bp::make_function(&Blob<Dtype>::mutable_cpu_data,
           NdarrayCallPolicies()))
     .add_property("diff",     bp::make_function(&Blob<Dtype>::mutable_cpu_diff,


shaibagon:
Is there any automatic test for this feature?

vimalthilak:
would be a very nice feature! ಠ_ಠ (yes, I lifted this from @longjon's comment )

longjon:
@shaibagon No test is included here; since this code just exposes an existing call, it's not clear what to test, especially without bringing in additional Python dependencies. The idea here is to have a small mergeable patch which just makes it possible to access GPU memory from Python. If you are a regular old user, the public interface is unchanged and there is no new functionality. If you are a hacker, you can dig into these functions and interface to PyCUDA &c. free of round-trips (this is what I am doing). If you are a developer, you can think about adding these interfaces to pycaffe.

@vimalthilak I'm flattered but I'm not sure you've grokked the expression (https://www.google.com/search?q=ಠ_ಠ).

vimalthilak:
@longjon You are right! My usage is more along the lines of I'm keeping my eyes on this thread!

shaibagon:
@longjon This is indeed a very simple (yet important) feature, but I suppose a very simple test verifying that indeed the exposed GPU pointer can be used in python.

longjon:
Well, I don't think there is such a thing as a simple test here: in order to check that these functions return something valid, we would need to read GPU memory from Python. We would need a new Python library to do that, which we would not want to make a hard dependency of pycaffe. That means we would need a way to disable some tests, or to acknowledge that not all tests should run and pass all the time. Then we would need to make sure that the new test was sometimes run, even though it wouldn't be runnable on Travis.

That's all possible (after all, Caffe has a CPU-only mode and GPU tests on the C++ side), but I don't think I'm going to do any of that in this PR, and I don't really feel bad about it, because there is no change to the public interface, only a private exposure of calls that are already tested in Caffe.

Future code which adds a public interface wrapping these calls probably merits testing. If such code interfaced to PyCUDA, e.g., it would already have to deal with the issue of dependencies.

naibaf7:
I'm currently working on something to allow writing GPU layers for both CUDA and OpenCL from Python on the OpenCL branch (single source and runtime compile for both CUDA and OpenCL).

shaibagon:
@longjon sounds reasonable. no need to over comicate things. What I like most in caffe is the automatic tests: I can be certain that everything does what it is expected to do.

shaibagon:
@naibaf7 GPU pyhon layeris a huge step forward for caffe. Thanks!

shelhamer:
Thanks for bringing this commit back so we can sip from the GPU through Python @longjon !

naibaf7:
@shaibagon @shelhamer Since this PR is usable for CUDA only: I'll release an initial version of the "device abstracted" Caffe soon, Python GPU will follow in 1-2 months approximately.

nmerrill67:
To test one could set up a simple net like learning the XOR function, and make sure it works correctly.

PR #5906, opened 2017-09-07
use ffmpeg_qsv to decode and process video

FangleixLiu:
use ffmpeg_qsv to decode and process video


naibaf7:
This PR seems to be misplaced against the master branch. Did you intend to file it against the OpenCL branch?

PR #5924, opened 2017-09-14
Implement CuDNN-based deconvolution layer and test

### LINKED ISSUE #5093: need cudnn deconvolution layer
I found that the current implementation of DeconvolutionLayer in caffe is not as fast as cudnn version, and it also costs more gpu memory. Can anyone provide a cudnn implementation of deconvolution layer in caffe?

bowang:
This PR implemented the CuDNN-based deconvolution layer and testcase. It runs 7-9x faster than the current implementation and more scalable on multiple GPUs.

The layer-wise runtime for the deconvolution layer is measured as below.
```
+------------+----+----+-----+-----+
|   #GPUs    | 1  | 2  |  4  |  8  |
+------------+----+----+-----+-----+
| Caffe (ms) | 80 | 96 | 120 | 225 |
| CuDNN (ms) | 12 | 12 | 14  | 33  |
| Speedup    | 7x | 8x | 9x  | 7x  |
+------------+----+----+-----+-----+
```

Fixes #5093

### DIFF
--- a/include/caffe/layers/cudnn_deconv_layer.hpp
+++ b/include/caffe/layers/cudnn_deconv_layer.hpp
@@ -1,0 +1,68 @@
+#ifndef CAFFE_CUDNN_DECONV_LAYER_HPP_
+#define CAFFE_CUDNN_DECONV_LAYER_HPP_
+
+#include <vector>
+
+#include "caffe/blob.hpp"
+#include "caffe/layer.hpp"
+#include "caffe/proto/caffe.pb.h"
+
+#include "caffe/layers/deconv_layer.hpp"
+
+namespace caffe {
+
+#ifdef USE_CUDNN
+/*
+ * @brief cuDNN implementation of DeConvolutionLayer.
+ *        Fallback to DeConvolutionLayer for CPU mode.
+ *
+ * cuDNN accelerates deconvolution through forward kernels for filtering and
+ * bias plus backward kernels for the gradient w.r.t. the filters, biases, and
+ * inputs. Caffe + cuDNN further speeds up the computation through forward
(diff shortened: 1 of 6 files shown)


Noiredd:
I've done a few runs of my old models with this PR and the results are amazing. An autoencoder with multiple deconvolution layers with kernels 3, 4 and 5 (including grouped deconv for unpooling) accelerated from 117 minutes to 14. Some FCN-like segmentation networks with a single large deconvolution for upsampling, which previously took 36 (kernel 16) and 43 (kernel 63) minutes to train, now only needed 34 and 16 respectively. Times are the better, the larger kernel is in use (and obviously the more deconvs we do).

I don't know cuDNN so I will refrain from doing a thorough code review, but basing on the experiments this *looks* good to me.

shelhamer:
Thanks for the speed-up @bowang and review @Noiredd!

Note that the cuDNN conv + deconv layers could be refactored to extend a base cuDNN conv/deconv layer as the vanilla Caffe layers do. They only differ in the top/bottom and a few dimension calculations so this would reduce code duplication. However, this can be left to a follow-up PR.

3DdeepAI:
@bowang  I also implemented a deconvolution layer using cuDNN; but I'm not sure, whether I did it correctly. Then I saw your code and I don't understand it... However, do you know of any simple train/test dataset I could use to test the results ?

deconv layer:
([https://github.com/3DdeepAI/3DdeepAI/blob/3DdeepAI/NNcontext.h](https://github.com/3DdeepAI/3DdeepAI/blob/3DdeepAI/NNcontext.h))
functions "ForwardDeConvolution" + "BackwardDeConvolution"

PR #5929, opened 2017-09-18
Update `include/caffe/blob.hpp` to remove `-Wsign-compare` warning

gineshidalgo99:
I've just reformulated `include/caffe/blob.hpp` to remove a `-Wsign-compare` warning. No other difference.

I know Caffe does uses `-W-nosign-compare` in their Makefile to remove these kind of warnings, but I must use Caffe with that warning, so adding this change the warning disappears.

EDITED: 2nd commit to change `auto` keyword by `unsigned int` so that non-C++11 compilers can also compile it.

Highly tested and it's being currently used in [OpenPose](https://github.com/CMU-Perceptual-Computing-Lab/openpose)

### DIFF
--- a/include/caffe/blob.hpp
+++ b/include/caffe/blob.hpp
@@ -53,7 +53,7 @@ class Blob {
   void ReshapeLike(const Blob& other);
   inline string shape_string() const {
     ostringstream stream;
-    for (auto i = 0u; i < shape_.size(); ++i) {
+    for (unsigned int i = 0u; i < shape_.size(); ++i) {
       stream << shape_[i] << " ";
     }
     stream << "(" << count_ << ")";
--- a/include/caffe/blob.hpp
+++ b/include/caffe/blob.hpp
@@ -53,7 +53,7 @@ class Blob {
   void ReshapeLike(const Blob& other);
   inline string shape_string() const {
     ostringstream stream;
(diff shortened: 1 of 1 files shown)


longjon:
We follow the Google style guide here (see https://google.github.io/styleguide/cppguide.html#Integer_Types), which is why we don't use `unsigned int` in loops.

We've debated this before and could debate it again, but we can't make one-off changes to this decision.

gineshidalgo99:
@longjon  Thank you for the explanation! But then would you admit the following change instead?
```
for (int i = 0; i < (int)shape_.size(); ++i) {
```

I just wanna get rid of the warning, both ways are equivalent in functionality (I changed it in my forked version, but it is not shown here because the pull request was closed).

longjon:
Well, that's part of the debate... we could change our style to add those kinds of casts (they should be `static_cast`s, btw), but that looks pretty sketchy and makes bad things silently happen if the bound actually exceeds `INT_MAX`. (Of course, bad things also silently happen now in that case, but we were warned!)

I think you're best either:
* using a modified version of this header file,
* using `-Wno-sign-compare` when you build with this header, or
* accepting this warning when you include this header, and isolating that from that your normal build process.

gineshidalgo99:
OK, thanks for the elaborated answer! I've learned from it!

PR #5939, opened 2017-09-22
Fixing broken links to installation

jameshamm:
It seems like installation.html was renamed to installation.md. This update fixes the two links in the page to compilation that were broken.


shelhamer:
Thanks for checking, but this is coded as intended: the docs dir is compiled into the site served at caffe.berkeleyvision.org by jekyll and so all of the links are to the compiled site. See http://caffe.berkeleyvision.org/development.html#documentation for more detail.

PR #5941, opened 2017-09-22
Fix summarize.py for protobuf 3.x

ssahu:
Same issue as mentioned here:
https://github.com/BVLC/caffe/pull/5487#discussion_r109833026


Noiredd:
I'm getting `AttributeError: 'module' object has no attribute 'Parse'` on protobuf 2.5 - replacing `Parse` with `Merge` seems to fix this. Would you care to investigate this, @ssahu?

PR #5969, opened 2017-10-10
Fix default mode warning in io.resize_image

### LINKED ISSUE #5602: warning on `resize_image` from pycaffe with new version of skimage
### Issue summary

when calling `io.resize_image` with relatively new skimage installed (0.13 in this case), there's deprecation warning.

### Steps to reproduce

```
from __future__ import division, print_function, absolute_import
import caffe
import numpy as np

if __name__ == '__main__':
    caffe.io.resize_image(np.random.rand(100,100,3), (200, 200))
```

result:

```
.../python2.7/site-packages/skimage/transform/_warps.py:84: UserWarning: The default mode, 'constant', will be changed to 'reflect' in skimage 0.15.
```

an easy fix is to explicitly state the padding mode when calling `resize`. This would suppress the warning, as well as keeping the behavior of `resize_image` consistent against future skimage.

### Your system configuration
Operating system: CentOS 6
Compiler:
CUDA version (if applicable):
CUDNN version (if applicable):
BLAS:
Python or MATLAB version (for pycaffe and matcaffe respectively): 2.7

developius:
Signed-off-by: Finnian Anderson <get@finnian.io>

This fixes a warning in `io.resize_image` which originates in `skimage.transform.resize`.
The default mode for `resize` will change from `constant` to `reflect`, so this commit changes the call to `resize` so that it passes `mode` manually.
Warning:
```
.../python2.7/site-packages/skimage/transform/_warps.py:84: UserWarning: The default mode, 'constant', will be changed to 'reflect' in skimage 0.15.
```

Fixes #5602

### DIFF
--- a/python/caffe/io.py
+++ b/python/caffe/io.py
@@ -323,7 +323,7 @@ def resize_image(im, new_dims, interp_order=1):
             # skimage is fast but only understands {1,3} channel images
             # in [0, 1].
             im_std = (im - im_min) / (im_max - im_min)
-            resized_std = resize(im_std, new_dims, order=interp_order)
+            resized_std = resize(im_std, new_dims, order=interp_order, mode='constant')
             resized_im = resized_std * (im_max - im_min) + im_min
         else:
             # the image is a constant -- avoid divide by 0


developius:
Not sure why this test is failing, the code change shouldn't have been a breaking change.

Noiredd:

Indeed, `skimage.transform.resize` used to have a default argument `mode='constant'` which will change to `None` in future versions. Thank you for the fix!

alexellis:
Nice one @developius 👍 Caffe is awesome..

PR #5972, opened 2017-10-11
add support for cuDNN v7

icyblade:
This PR will try to add support for cuDNN v7.
Tested on Tesla K80/Ubuntu 16.04/CUDA 9.0/cuDNN v7.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -42,14 +42,14 @@ Optional dependencies:
 
 * [OpenCV](http://opencv.org/) >= 2.4 including 3.0
 * IO libraries: `lmdb`, `leveldb` (note: leveldb requires `snappy`)
-* cuDNN for GPU acceleration (v6)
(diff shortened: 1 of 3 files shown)


wjcskqygj2015:
Perfectly, I originally try to pull this request

wjcskqygj2015:
However, in order to fit cuda 9.0, we might need to anotate  
line 64 in cuda/include/crt/common_functions.h since the preprocessor is not worked well
#define __CUDACC_VER__ "__CUDACC_VER__ is no longer supported.  Use __CUDACC_VER_MAJOR__, __CUDACC_VER_MINOR__, and __CUDACC_VER_BUILD__ instead."

Noiredd:
From a few tests I ran myself (before and after an update to cuDNN 7), this works well. Thank you for the contribution, @icyblade!

Issue #5960, opened 2017-10-03
Travis builds randomly fail - proposed solution


Noiredd:
I suppose I'm not the only one who noticed that some Travis builds fail due to seemingly random causes - even if no related code has been modified, and even if no actual code has been modified at all. This leads to weird situations where a [PR](https://travis-ci.org/BVLC/caffe/builds/282264423) passes all tests (see #5713), but the same commit merged to [master](https://travis-ci.org/BVLC/caffe/builds/282530789) fails. This has troubled me for a while so I decided to dig a little and I finally got to the root cause, which seems to be a badly designed unit test.

Having analysed logs from several events (eg. [1](https://travis-ci.org/BVLC/caffe/jobs/264708550#L9719), [2](https://travis-ci.org/BVLC/caffe/jobs/282530796#L8831), [3](https://travis-ci.org/BVLC/caffe/jobs/282530790#L9509)), the source of the problems turns out to be `test_forward_start_end` and `test_backward_start_end` in `class TestNet` of `python/caffe/test/test_net.py`. It uses `np.testing.assert_allclose` with *relative tolerance* parameter of `1e-3`. What happens depends on results of two randomization rolls: `np.random.uniform()` that [initializes](https://github.com/BVLC/caffe/blob/master/python/caffe/test/test_net.py#L77) the test data, and weight fillers defined in the `simple_net_file()`.  
If the rolls are "lucky", the elements of `forward_blob` (in `test_forward_start_end`, and respectively `backward_blob` in `test_backward_start_end`) are large (above `1e-3`), and the distance to `manual_forward` isn't relatively large. But when they turn out "bad", very small elements appear (less than `1e-4`), and errors calculated with respect to their values end up large - larger than currently set `rtol` of `1e-3` - and the test fails.

This proves that the test results are partially random, which I think should be fixed. The question remains how. I found that adding a tiny *absolute tolerance* level (`atol=1e-5`) solves the problem - but does it still allow the test to weed out bugs in the operations that it's supposed to verify?  
I'm ready to open a PR but would like to discuss this first. [Here](https://gist.github.com/Noiredd/4cc5feda4ff46ceb2480f5e5b50c9478) is how I got into test internals (modify `test_net.py`), and [here's](https://gist.github.com/Noiredd/79af15afe594472173658cf8dbbf5899) how I analyzed them. You will need to iterate over  
`python -m unittest discover -s caffe/test -p "test_net.py"`  
and the attached gist  
`python test_analyze.py`  
[...]

PR #5973, opened 2017-10-11
Add absolute tolerance to test_net.py to prevent random Travis fails

Noiredd:
Follow up on issue #5960 - fix by adding absolute tolerance of `1e-5` to `test_forward_start_end()` and `test_backward_start_end()` in `test_net.py`. I iterated `python -m unittest discover -s caffe/test -p "test_net.py"` over the current master 500 times and no random fails occurred.

Additionally I removed semicolons that ended each line in those two functions - it looked weird.

### DIFF
--- a/python/caffe/test/test_net.py
+++ b/python/caffe/test/test_net.py
@@ -72,41 +72,41 @@ class TestNet(unittest.TestCase):
         self.net.backward()
 
     def test_forward_start_end(self):
-        conv_blob=self.net.blobs['conv'];
-        ip_blob=self.net.blobs['ip_blob'];
-        sample_data=np.random.uniform(size=conv_blob.data.shape);
-        sample_data=sample_data.astype(np.float32);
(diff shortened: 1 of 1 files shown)


shaibagon:
Thanks!
Why did you opt for adding `atol` instead of increasing slightly the random input numbers?

shaibagon:
LGTM

Noiredd:
> Why did you opt for adding atol instead of increasing slightly the random input numbers?

I explained that in the [comment](https://github.com/BVLC/caffe/issues/5960#issuecomment-335738664) under the issue itself. This test does not only concern randomly generated data, but also weights in the InnerProduct layer of the test net. We can increase the data slightly but the weights (drawn from a normal distribution) can still roll such that some output will be small.

shaibagon:
@Noiredd  what holds you from merging this PR?

shelhamer:
Thanks for addressing these spurious flaky tests!

Issue #5981, opened 2017-10-15
New Accuracy Layer on GPU interferes with training


vlomonaco:
### Issue summary
Using the "Accuracy" layer in the "Training net" on GPU breaks the training. The layer somehow interferes with the gradient. Loss explodes quickly and Train/Test Accuracies stall to 1.

### Steps to reproduce
1. Download the the latest version of Caffe (commit 691febcb83d6a3147be8e9583c77aefaac9945f8)
2. Compile it with this [Makefile.config](https://github.com/BVLC/caffe/files/1385875/Makefile.txt)
3. Run a network with Accuracy layer in the TRAIN phase.

### My system configuration
Operating system: Ubuntu 14.04.5 LTS
CUDA version (if applicable): 7.0
CUDNN version (if applicable): 4.7
BLAS: libblas.so.3
Python version: 3.5

### How to fix it
All these three solutions can fix it:
- Remove the Accuracy Layer from the "Training net", not problems with the phase: TEST.
- Change back-end to CPU.
- Rollback before commit 62e0c8559045cb2b5a12e0d6c41acd25d412263 (which I suspect caused the issue).

Noiredd:
I can reproduce this behavior for the GPU implementation. CPU does not seem to be affected. I'll be looking into this today; In the meantime, could you take a look too, @shaibagon?

shaibagon:
@Noiredd Is it possible [this](https://github.com/BVLC/caffe/blob/62e0c8559045cb2b5a12e0d6c41acd25d4122630/src/caffe/layers/accuracy_layer.cu#L143) causes the issue? I will look into it.

Noiredd:
Removing the `if` and replacing with an unconditional `NOT_IMPLEMENTED;` did not change anything.

However, forcing:  
`caffe_gpu_set(bottom[0]->count(), Dtype(0), acc_data);`
`caffe_gpu_set(bottom[0]->count(), Dtype(0), counts);`
at the end of `Forward_gpu()` fixes the problem, supporting my guess that Caffe indeed propagates from there.  
This is not counter-intuitive. Think of intermediate classifiers - if we have a prediction blob A that is a bottom to one Accuracy layer but it's also a bottom to, let's say, InnerProduct, we want the IP to propagate. By reusing the blob's gradient memory, we effectively overrode the other gradients.

shaibagon:
[...]

PR #5987, opened 2017-10-16
Hotfix for accuracy interfering with training

Noiredd:
Instead of reusing "unused" memory of the bottom blob (`diff`), it's safer to allocate some internal blob (`gpu_buffer_`) to make sure the layer does not interact with external memory that might be potentially used elsewhere (as was demonstrated in #5981). Bug was initially introduced in 62e0c85.

In the long run, we might want to optimize this to eg. only allocate GPU memory (instead of an entire blob), or even take a closer look at what is exactly going on in that buffer (i.e. how much memory do we actually need). On the other hand, using Accuracy layer during training is not very common usage - so we might leave this PR pending for now (for the affected users to pull) and only accept after said optimizations.

### DIFF
--- a/include/caffe/layers/accuracy_layer.hpp
+++ b/include/caffe/layers/accuracy_layer.hpp
@@ -92,6 +92,8 @@ class AccuracyLayer : public Layer<Dtype> {
   int ignore_label_;
   /// Keeps counts of the number of samples per class.
   Blob<Dtype> nums_buffer_;
+  /// Intermediate results for the GPU implementation
+  Blob<Dtype> gpu_buffer_;
 };
 
 }  // namespace caffe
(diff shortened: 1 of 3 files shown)


Noiredd:
@shaibagon Can we optimize this memory-wise in any way? I.e. do we need the intermediate blob to be as large as the whole `bottom[0]`? If not, I say we merge. @shelhamer?

eimikls:
Hi,
what the progress on merging this fix? It is already almost a month for caffe not being able to calculate/display accuracy statistics in training phase.

duygusar:
I have made the changes and made a clean Caffe build but my problem still persists. I did not have this problem when working with a dataset of about 60K training images, now I oversample them and have 10*60K images in total (not sure why this should be a problem because the way the batches are handled are still same), and for some reason after few iterations Loss output t quickly jumps to and stays at 87.3365 . It is a fixed number.

shelhamer:
@Noiredd see #6202 for a combined fix to the scratch usage of bottoms diffs. I decided to clear the diffs in that fix instead of needing separate memory. Please let me know what you think of that choice.

shelhamer:
Closed in favor of #6202.

PR #5989, opened 2017-10-17
spatial conv layer for Xception

yihui-he:
From discussion in #778 : when group == num_output (spatial conv), the memory consuming is unbearable. This PR implements spaital conv in conv_layer.cu. This implementation could be used Xception, a prototxt can be found here: https://github.com/yihui-he/Xception-caffe/blob/master/xception-50.prototxt . 
An example of usage:
```
layer {
  name: "interstellar2a_branch1_s"
  type: "Convolution"
  bottom: "pool1"
  top: "interstellar2a_branch1_s"
  param {
    lr_mult: 1
    decay_mult: 1
  }
  convolution_param {
    num_output: 64
    bias_term: false
    pad: 1
    kernel_size: 3
    group: 64
    stride: 1
    weight_filler {
      type: "msra"
    }
    engine: CAFFE
  }
}
```

### DIFF
--- a/src/caffe/layers/conv_layer.cu
+++ b/src/caffe/layers/conv_layer.cu
@@ -5,21 +5,167 @@
 namespace caffe {
 
 template <typename Dtype>
+__global__ void spatial_conv_fwd_kernel(const int count, const int channels, const int output_height, const int output_width,
+            const int input_height, const int input_width, 
+            const int kernel_height, const int kernel_width,
+            const int stride_y, const int stride_x, const int pad_y, const int pad_x,
+            const Dtype *input_data, const Dtype *weight_data, Dtype *output_data) { //, const Dtype *bias_data
+  CUDA_KERNEL_LOOP(index, count) {
+      output_data += index;
+      int ow = index % output_width;
+      index /= output_width;
+      int oh = index % output_height;
+      index /= output_height;
+      int c = index % channels;
+      int n = index / channels;
+
+      int iw = ow * stride_x - pad_x;
+      int ih = oh * stride_y - pad_y;
+
+      input_data += ((n * channels + c) * input_height + ih) * input_width + iw;
+      weight_data += c * kernel_width * kernel_height;
+
+      Dtype v = 0;
+      for (int kh = 0; kh < kernel_height; kh++) {
+          if (ih + kh >= 0 && ih + kh < input_height) {
+              for (int kw = 0; kw < kernel_width; kw++) {
+                  if (iw + kw >= 0 && iw + kw < input_width) {
+                      v += input_data[kw] * weight_data[kw];
+                  }
+              }
+          }
+          input_data += input_width;
+          weight_data += kernel_width;
+      }
+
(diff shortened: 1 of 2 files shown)


Noiredd:
I compared performance of this implementation with the latest master on my GTX TITAN Z, taking a 1x256x256x256 input and a conv layer with `num_output: 256`, `group: 256`, `kernel_size: 5`, `stride: 1`, `pad: 1`. Below I show a forward-backward iteration time and VRAM allocated as shown by `nvidia-smi`:

implementation | iteration time | VRAM taken
--- | --- | ---
master | 38.7 ms | 324 MB
this PR | 20.2 ms | 324 MB

For a 1x32x256x256 input and a conv with `num_output: 32`, `group: 32`, `kernel_size: 5`, `stride: 1`, `pad: 1`:

implementation | iteration time | VRAM taken
--- | --- | ---
master | 297.5 ms | 1924 MB
this PR | 143.5 ms | 1924 MB

So it was faster, but I was unable to observe any RAM savings.

---

Additional issues with this PR:
* it does not implement biases (though there is some commented-out code for a forward pass that could be easily made to work), failing the forward tests (`ConvolutionLayerTest/2.TestSimpleConvolutionGroup,`),
* even with this fixed, it fails the backward tests (`ConvolutionLayerTest/2.TestGradientGroup`),
* from what I see, it does not implement dilation (though we do not seem to have a test for group+dilation),
* numerous lint issues.

Are you willing to continue working on this, @yihui-he? RAM savings and speedups for grouped conv would be **very** useful, but this needs some more attention.

src/caffe/layers/conv_layer.cu
```diff
@@ -4,24 +4,170 @@
...
   for (int i = 0; i < bottom.size(); ++i) {
     const Dtype* bottom_data = bottom[i]->gpu_data();
     Dtype* top_data = top[i]->mutable_gpu_data();
-    for (int n = 0; n < this->num_; ++n) {
-      this->forward_gpu_gemm(bottom_data + n * this->bottom_dim_, weight,
-          top_data + n * this->top_dim_);
-      if (this->bias_term_) {
-        const Dtype* bias = this->blobs_[1]->gpu_data();
-        this->forward_gpu_bias(top_data + n * this->top_dim_, bias);
+    if (this->group_ == this->channels_){
```
Noiredd: In the description, you mention a case when group==num_output, yet here you compare it to channels instead. Is this correct?  
Also, you don't really need `this->` when accessing attributes.

naibaf7:
There is also an implementation that does not consume any extra memory for convolution here:
https://github.com/naibaf7/caffe/tree/master/src/caffe/libdnn (CUDA and OpenCL)
and here:
https://github.com/BVLC/caffe/blob/opencl/src/caffe/greentea/libdnn_conv.cpp (CUDA and OpenCL)
and here:
https://github.com/BVLC/caffe/blob/opencl/src/caffe/greentea/libdnn_conv_spatial.cpp (OpenCL)

PR #3855, opened 2016-03-21
upgrading InfogainLoss layer 

shaibagon:
Making InfogainLoss layer more similar to SoftmaxWithLoss in term of functionality:
(1) incorporating Softmax layer to make the gradient computation robust, much like SoftmaxWithLoss layer (see: [this SO thread](http://stackoverflow.com/a/34917052/1714410) for more information). 
(2) supporting loss along axis (see [this SO thread for motivation](http://stackoverflow.com/q/36115026/1714410)).

Updated `test_infogain_loss_layer.cpp` accordingly.

This PR is self-contained: it only affects infogain loss layer.

PR #5990, opened 2017-10-17
Infogain loss: fix bottom blobs description

anmikh:
It seems that the description of the bottoms blobs is out of date for the Infogain loss layer. At [this PR](https://github.com/BVLC/caffe/pull/3855) SoftmaxLoss layer was integrated to the InfogainLoss for numerical stability, and after that InfogainLoss takes just scores as one of the the bottoms blobs rather than probabilities. Thereafter it's equivalent to SoftmaxWithLoss layer rather than MultinominalLogisticLoss layer when Infogain matrix is the identity.

### DIFF
--- a/include/caffe/layers/infogain_loss_layer.hpp
+++ b/include/caffe/layers/infogain_loss_layer.hpp
@@ -13,20 +13,21 @@
 namespace caffe {
 
 /**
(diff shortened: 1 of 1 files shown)


shaibagon:
LGTM

PR #5999, opened 2017-10-18
add confusion matrix layer for classification task.

melody-rain:
Signed-off-by: Jin Ma <xiah_sunny@hotmail.com>


shaibagon:
What about a test for this layer?

PR #6017, opened 2017-10-31
Make Caffe buildable on OSX 10.11.6 with cmake and python3

mprat:
In building Caffe with CMake on OSX 10.11.6 in a virtualenv with python3 I had problems finding the appropriate Boost libraries (in this case it was 1.65.1 installed with Homebrew), but adding these two search statements for Boost (and removing the version restriction) allowed me to build it.

There was also a missing dependency in the requirements.txt.


Noiredd:
`pydot` is not in requirements on [purpose](https://github.com/BVLC/caffe/blob/b4ffad848e7a54e2f22081f07c8c7e49f31ca5f4/python/caffe/draw.py#L7-L9).

PS: Travis failed to build your PR for some reason, but I manually restarted it and it looks fine now.

mprat:
Re: pydot - that's interesting. If you do `make pytest` without having pydot installed it fails, so I assumed it was mandatory. Also, I installed Caffe with python3 (and `pip3 install pydot`) and it passed all the tests as well. It might be that newer versions of pydot fixed the problem.

Issue #5964, opened 2017-10-04
CHECK_LE(n, num()) should  be CHECK_LT


hkxIron:
https://github.com/BVLC/caffe/blob/fac74347106ea16140ace828e8d278716e4d4742/include/caffe/blob.hpp#L156

I think CHECK_LE(n, num()) should  be CHECK_LT , same as CHECK_LE in the following lines of this function,
which can be in accordance with the other fuction  "offset(const vector<int>& indices)"

hkxIron:
CHECK_LE in offset()n,c,h,w) should be replace with CHECK_LT,
when n==num(), it out of ranges actually!
which can be equal with offset(indices),

shaibagon:
Good point.

in fact [all `CHECK_LE`](https://github.com/BVLC/caffe/blob/fac74347106ea16140ace828e8d278716e4d4742/include/caffe/blob.hpp#L156-L162) in `inline int offset(const int n, const int c = 0, const int h = 0, const int w = 0) const` should be converted to `CHECK_LT` in accordance with [`inline int offset(const vector<int>& indices) const`](https://github.com/BVLC/caffe/blob/fac74347106ea16140ace828e8d278716e4d4742/include/caffe/blob.hpp#L173).

Can you please open a pull-request fixing this issue? See [*Developing and Contributing*](http://caffe.berkeleyvision.org/development.html) how to do this.

Thank you very much,
-Shai

hkxIron:
ok

hkxIron:
when I run the command " git push origin feature_issue_5964"
I get the following error:
remote: Permission to BVLC/caffe.git denied to hkxIron.
fatal: unable to access 'https://github.com/BVLC/caffe.git/': The requested URL returned error: 403

dashesy:
@hkxIron you should push to your own clone. Clone BVLC Caffe (through github), then change origin to point to your Caffe and set upstream to track the upstream Caffe at BVLC: 

    git remote set-url origin git@github.com:hkxIron/caffe.git
    git remote add upstream git@github.com:BVLC/caffe.git

Then 

   git push -u origin feature_issue_5964

Then go to BVLC Caffe and click on the button to make that a PR (github knows it is cloned and suggests the easy way to PR)

hkxIron:
[...]

PR #6021, opened 2017-11-01
Update blob.hpp for bug issue #5964

hkxIron:
Fix bug issue #5964 : change CHECK_LE to CHECK_LT in function offset.

### DIFF
--- a/include/caffe/blob.hpp
+++ b/include/caffe/blob.hpp
@@ -153,13 +153,13 @@ class Blob {
   inline int offset(const int n, const int c = 0, const int h = 0,
       const int w = 0) const {
     CHECK_GE(n, 0);
-    CHECK_LE(n, num());
+    CHECK_LT(n, num());
     CHECK_GE(channels(), 0);
-    CHECK_LE(c, channels());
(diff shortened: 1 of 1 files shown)


hkxIron:
Thanks.

hkxIron:
@shaibagon ,  run unit test failed.
  F1101 15:51:51.257525 12917 blob.hpp:158] Check failed: c < channels() (1 vs. 1)

shaibagon:
@hkxIron  It seems like pooling layer's CPU code uses `offset()` to compute a step size inside the blob, This step is taken at the end of a loop, so it might be the case that the step is computed, but when it falls outside the Blob it's the end of the loop and the memory is not accessed.  
I guess this PR will have to deal/fix this as well.
I would try and pre-compute these steps in advance (i.e., calling `offset(0,1)` once outside the loop and saving the value and then re-using it instead of calling `offset(0,1)` multiple times.

Noiredd:
I would also add #792 for reference - I think that PR was more comprehensive - and #3420 for the conversation.

PR #6023, opened 2017-11-01
Ssd

buff4life123:


shaibagon:
This PR seems like an accidental push of SSD branch of caffe.

PR #6026, opened 2017-11-02
Fix build warnings with CUDA9. 

lkm1321:
Hi, not sure if this has been fixed by someone else, but I fixed switch-case build warnings for the latest CUDA 9.0 and cudnn 7.0. There are just a couple of more enums that need to be cased. Please review.


Noiredd:
This has already been done - see #5972. Thank you for the submission though!

PR #6048, opened 2017-11-12
Fix Makefile parallel builds missing protobuf header

sclarkson:
`make -j##` can fail seemingly randomly if one of the source files including the caffe.pb.h header starts to compile before `protoc` finishes.


Noiredd:
Thank you for the fix!

PR #6052, opened 2017-11-12
Mechanism of Quantization

RudyChin:
Hi all,

I found [Ristretto](http://lepsucd.com/?page_id=630) pretty useful when we want to play with the precision of weights and activations of a neural network in Caffe.

Hence, I cherry-pick and fixed merge conflicts resulting in this PR.

Thanks,
Rudy


RudyChin:
It seems there are some lint errors with Travis-CI, I'll fix it and update it again

Noiredd:
Personally, I don't think it is a good idea to merge any of the specialized forks back into Caffe core - both for the Caffe team (as it shifts the burden of maintaining this code onto us) and the forks' developers (as their code now exists in two places).

PR #6053, opened 2017-11-12
OpenCL - Improved reshape performance

artyom-beilis:
When using caffe-opencl with dynamic batch size (for obvious performance improvements) the reshape procedure for libdnn based layers conv/deconv/pool creates a new instance and thus rebuilds the kernels. It is suboptimal since libdnn itself already supports batch size.

So provided pool requests fixes this issue allowing to use dynamic batch size without penalty.


naibaf7:
Ok cool, I will review it.

PR #6066, opened 2017-11-21
Fixed a bug in AccuracyLayer.

ryanflower:
gpu_diff of bottom[0] used as temporary buffer needs to be set back to
zeros. Otherwise the AccurarcyLayer may propagate back wrong gradient
values in training.

### DIFF
--- a/src/caffe/layers/accuracy_layer.cu
+++ b/src/caffe/layers/accuracy_layer.cu
@@ -134,6 +134,9 @@ void AccuracyLayer<Dtype>::Forward_gpu(
       }
     }
   }
+  // Set gpu_diff of bottom[0] back to zeros to avoid propagating down any
+  // gradient from the accuracy layer during training.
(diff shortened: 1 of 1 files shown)


shaibagon:
This is not a good fix: if the prediction layer is shared with other layers (e.g., "SoftmaxWithLoss") resetting `diff` might reset the gradients coming from the loss layer. 

Please see PR #5987 for a fix for this issue.

ryanflower:
I see your point, though I think split layers are automatically added to avoid such cases.

A better fix is to use bottom[0]->diff().use_count() to check whether the diff memory is shared with other layers. If so, initialize and use an internal buffer; otherwise reuse bottom[0]->diff() and reset it  to zeros after usage.

shaibagon:
@ryanflower this is an interesting point you are making. Can you please move this discussion to the open PR #5987?

PR #6092, opened 2017-12-03
fix above to below

sonack:


Noiredd:
Could you explain what does this PR do?

sonack:
@Noiredd  I correct some spell mistake in the code comment and remove some redundant code piece in batch_norm layer, it's trivial, just look at the file diff :）

Noiredd:
This is exactly the problem - someone trying to review your code has no idea what's going on until they read into it and analyze it. For someone like me, who's never read the BN layer code, this means the necessity to go through not just a few lines that you changed, but also the whole surrounding infrastructure. Not even to mention some motivation - why is the change needed, what improvements does it bring etc.

PR #6093, opened 2017-12-04
reinit global_instance_ when it has been deleted

dzyong:
when global_instance_ has been deleted, we have to set 'first' to true to trigger Caffe::Get() to create a new global_instance_, otherwise we will use deleted global_instance_ to initialize other Caffe object and this will cause segment fault.

### DIFF
--- a/python/caffe/draw.py
+++ b/python/caffe/draw.py
@@ -244,7 +244,7 @@ def get_pydot_graph(caffe_net, rankdir, margin, page, pagesize, size, label_edge
     return pydot_graph
 
 
-def draw_net(caffe_net, rankdir, margin, page, pagesize, size, ext='png', phase=None):
+def draw_net(caffe_net, rankdir, margin='', page='', pagesize='', size='', ext='png', phase=None):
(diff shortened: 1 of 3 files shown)


dzyong:
There seems some error in original source code:
<code>
_caffe.obj : error LNK2019: unresolved external symbol "public: static void __cdecl caffe::Caffe::SetDevices(class std::vector<int,class std::allocator<int> >)" (?SetDevices@Caffe@caffe@@SAXV?$vector@HV?$allocator@H@std@@@std@@@Z) referenced in function "void __cdecl caffe::set_devices(class boost::python::tuple)" (?set_devices@caffe@@YAXVtuple@python@boost@@@Z) [C:\projects\caffe\build\python\pycaffe.vcxproj]
C:\projects\caffe\build\lib\Release\_caffe.pyd : fatal error LNK1120: 1 unresolved externals
</code>

dzyong:
The second submission fixed some windows compile error. But it still have some other errors.

None of these errors are caused by the first submission : )

Issue #4786, opened 2016-09-29
ImportError: cannot import name BytesIO 


huyng:
Hi there the caffe/io.py module conflicts with python's native io module, which causes an error when trying to `import caffe` .

This is the traceback I get:

```
  File "/usr/local/lib/python2.7/site-packages/IPython/utils/openpy.py", line 9, in <module>
    import io
  File "/usr/local/src/caffe/python/caffe/io.py", line 2, in <module>
    import skimage.io
  File "/usr/local/lib/python2.7/site-packages/skimage/io/__init__.py", line 11, in <module>
    from ._io import *
  File "/usr/local/lib/python2.7/site-packages/skimage/io/_io.py", line 1, in <module>
    from io import BytesIO
ImportError: cannot import name BytesIO
```

I'd like to rename caffe/io.py to caffe/ioutils.py  in order to prevent this issue.

Is this something you would like to incorporate into this repo? If so, I can send in a PR

PR #6100, opened 2017-12-10
fix installation issues with makeconfig example and caffe python name…

XikunZhang:
# 2 installation issues 
1. `libhdf5-dev` is not included in the Makefile.config.example, and this will cause problems when building using this file. (I just include the typical path for Linux, and mac should have a similar one. )
2. the `io.py` file will cause conflict with the native python `io` module. So I rename that file and change the `import` statement in the `__init__.py` file and `pycaffe.py` file.


Noiredd:
This PR changes the name of `caffe.io` module into `caffe.caffe_io` which is IMO a breaking change (everyone would have to modify all of their depending scripts). Can you provide an example of existing collision of `caffe.io` with built-in `io`? It would be easier to fix a single bug rather than changing the way pycaffe is used.

XikunZhang:
The issue #4786 has the exact same error I encounter when I `import caffe` in my Python script.  The issue is that when other Python built-in modules need to import the Python built-in `io` module,  the `caffe.io` module is imported since it is also in the `PYTHONPATH` environment variable. The best way I can think of is renaming the caffe `io` module. If you have other ideas, I am happy to help! Without solving this conflict issue, I am not able to use Caffe in Python.

PR #6104, opened 2017-12-12
Check on solver's parameters to avoid unexplained core dump

### LINKED ISSUE #4946: bug report: floating point exception when stepsize 0
Hi everybody,

I would like to report a 'bug' (kind of) which cost me a day of work and can probably save time to other in the future:
When by mistake the stepsize is set to zero in the solver, caffe gives a "floating point exception" during solver.step(). I would suggest an assert during the reading of the solver.prototxt which rise an error in case stepsize is 0.

Thanks! ;)

YaYaB:
Add check values of gamma and stepsize to avoid unexplained core dump.

Fixes #4946


Noiredd:
Thanks for the fix!

PR #5149, opened 2017-01-02
bug fix: ext string shouldn't have '.'

yanchen036:
ext string contains '.', like ".jpg", however, encoding string doesn't have '.', like "jpg". 
When call substr(), I make the position variable p add 1 to exclude '.'

PR #6158, opened 2018-01-09
fix matchExt

makefile:
In implementation of ReadImageToDatum in io.cpp uses matchExt, however the ext postion is not correct.


Noiredd:
Already raised in #5149, which is slightly more thorough but requires a rebase and squash.

Noiredd:
Closed in favor of #5149. Thank you @makefile!

Issue #2307, opened 2015-04-12
Unable to make caffe on MacOS 10.9


deepsamal:
MacOS 10.9, Cuda 6.5, Opencv 3.0, Anaconda
In Opencv 2.4.11 I was facing similar problem, solution was including library opencv_imgcodecs. But this was available only in opencv 3.0. But even after installing opencv 3.0.. I am getting same error, although in a different part. Any help appreciated.

It might seem there is a problem with leveldb, but all the environment variables in leveldb have -stdlib=libstdc++. So i can't figure out the problem.

PROTOC src/caffe/proto/caffe.proto
CXX src/caffe/blob.cpp
CXX src/caffe/common.cpp
CXX src/caffe/data_transformer.cpp
CXX src/caffe/internal_thread.cpp
CXX src/caffe/layer_factory.cpp
CXX src/caffe/layers/absval_layer.cpp
CXX src/caffe/layers/accuracy_layer.cpp
CXX src/caffe/layers/argmax_layer.cpp
CXX src/caffe/layers/base_conv_layer.cpp
CXX src/caffe/layers/base_data_layer.cpp
CXX src/caffe/layers/bnll_layer.cpp
CXX src/caffe/layers/concat_layer.cpp
CXX src/caffe/layers/contrastive_loss_layer.cpp
CXX src/caffe/layers/conv_layer.cpp
CXX src/caffe/layers/cudnn_conv_layer.cpp
CXX src/caffe/layers/cudnn_pooling_layer.cpp
CXX src/caffe/layers/cudnn_relu_layer.cpp
CXX src/caffe/layers/cudnn_sigmoid_layer.cpp
CXX src/caffe/layers/cudnn_softmax_layer.cpp
CXX src/caffe/layers/cudnn_tanh_layer.cpp
CXX src/caffe/layers/data_layer.cpp
CXX src/caffe/layers/deconv_layer.cpp
CXX src/caffe/layers/dropout_layer.cpp
CXX src/caffe/layers/dummy_data_layer.cpp
CXX src/caffe/layers/eltwise_layer.cpp
CXX src/caffe/layers/euclidean_loss_layer.cpp
CXX src/caffe/layers/exp_layer.cpp
CXX src/caffe/layers/flatten_layer.cpp
CXX src/caffe/layers/hdf5_data_layer.cpp
CXX src/caffe/layers/hdf5_output_layer.cpp
CXX src/caffe/layers/hinge_loss_layer.cpp
[...]

PR #6174, opened 2018-01-15
Fix link issue when using Homebrew OpenCV on OS X.

qu1j0t3:
I'm suggesting this change to the config example because it resolves the problem reported in this thread (and also experienced by myself on OS X): https://github.com/BVLC/caffe/issues/2307#issuecomment-357552138

Let me know if I should raise a specific issue in addition to this PR.

### DIFF
--- a/Makefile.config.example
+++ b/Makefile.config.example
@@ -9,6 +9,7 @@
 
 # uncomment to disable IO dependencies and corresponding data layers
 # USE_OPENCV := 0
+OPENCV_VERSION := 3
 # USE_LEVELDB := 0
 # USE_LMDB := 0
 


Noiredd:
Thank you for the contribution.  
However, take a look just a [few lines further down](https://github.com/BVLC/caffe/blob/95d000153dc668bf86f24eb46e8982701fa7c59a/Makefile.config.example#L20-L21) the example config - there already is an option to inform Make to compile with OpenCV 3. All one needs to do is uncomment it.

toby5box:
Yes, you are right, and I missed it.

But maybe this can still be improved:
* Should it be uncommented by default, since (I think) OpenCV version is now most likely to be 3?
* Should it be moved closer to the option enabling OpenCV?

Noiredd:
I think most people are still on version 2, so it's better left commented.  
In my opinion it is also close enough to the `USE_OPENCV`, but if you feel that moving it above the `MDB_NOLOCK` block would improve clarity, and others agree with you by upvoting your PR - we can merge.

toby5box:
OpenCV 3 is 2.5 years old: https://opencv.org/releases.html

Yeah, it's probably too trivial to do anything urgently about. Probably just me being unobservant. At least the issue exists for people confused by the build error, and googling.

PR #6176, opened 2018-01-15
Simplify pip invocation.

qu1j0t3:


shelhamer:
Note that this is only now ok b.c. of more reasonable behavior in recent pip as reported here https://github.com/BVLC/caffe/pull/1950#issuecomment-119550080

Thanks @jqueguiner

PR #6210, opened 2018-02-02
Update Classifier and Detector to avoid deprecation warning

mzsanford:
A system I'm working on uses `caffe.Classifier` and continually logs warnings like those below. This PR changes the underlying `Net` call to match the new signature.

```
W0202 20:51:24.142218     1 _caffe.cpp:135] DEPRECATION WARNING - deprecated use of Python interface
W0202 20:51:24.142246     1 _caffe.cpp:136] Use this instead (with the named "weights" parameter):
W0202 20:51:24.142249     1 _caffe.cpp:138] Net('/path/to/deploy.prototxt', 1, weights='/path/to/model_name.caffemodel')
```


Noiredd:
Thanks for fixing the inconsistency, @mzsanford!

PR #6218, opened 2018-02-08
Documentation for solving build issues

TianxinTse:
Many people have suffered from the compilation.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -135,6 +135,60 @@ The basic steps are as follows:
 
 See [PR #1667](https://github.com/BVLC/caffe/pull/1667) for options and details.
 
+### Throuble shooting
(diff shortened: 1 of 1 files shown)


Noiredd:
Thank you for the contribution - it is a really good idea to gather solutions to most commonly encountered build issues. However, in order to make it really helpful, I think it should be made into a separate document which everyone could expand. For this reason I suggest that you make a [Wiki](https://github.com/BVLC/caffe/wiki) page instead, and link to it from other installation pages.  
What do you think?

TianxinTse:
@Noiredd Yeah, Wiki is a better idea. I have created one.

Noiredd:
That's great - I've added a foreword to it. More work to come piling up the answers, but it's a good start. Thank you @TianxinTse!

PR #6123, opened 2017-12-21
"weights" added to solver parameters, "snapshot_prefix" field default initialization

IlyaOvodov:
This pull request implements parts 1&2 of feature request #6110.

"weights" parameter is added to SolverParameters class. It has quite the same meaning as "--weights" command line parameter of caffe executable and is used if "--weights" command line parameter is not specified. It makes it possible to specify initial weights in a solver.prototxt file instead of command line and make it unnecessary to change command line or command file.

It is aimed to make it possible to localize all training parameters in a solver.prototxt file. See #6110 for more detailed motivation.

If a "--snapshot" command line parameter is specified, new "weights" parameter in a solver.prototxt file is ignored. So there is no need to modify this file to continue previously interrupted training process from a snapshot.

EDIT by @Noiredd: second important contribution of this PR is a default initialization of the "solver_prefix" parameter. Previously Caffe would fail on parsing the solver if a snapshot was requested but no prefix given.

PR #6229, opened 2018-02-12
Merge pull request #6123 from IlyaOvodov/master -> windows

IlyaOvodov:
"weights" added to solver parameters, "snapshot_prefix" field default initialization

### DIFF
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -98,7 +98,7 @@ message NetParameter {
 // NOTE
 // Update the next available ID when you add a new SolverParameter field.
 //
-// SolverParameter next available ID: 42 (last added: layer_wise_reduce)
+// SolverParameter next available ID: 43 (last added: weights)
 message SolverParameter {
   //////////////////////////////////////////////////////////////////////////////
   // Specifying the train and test networks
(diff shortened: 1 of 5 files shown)


willyd:
@IlyaOvodov is this commit different from what's in the master branch?

IlyaOvodov:
No, it's cherry-pick from master branch

Noiredd:
Wouldn't it be cleaner to not have this change pulled automatically during a standard `master`->`windows` merge rather than making a separate PR for it? I don't follow the windows branch too closely but it looks like this is the usual way of doing it, isn't it @willyd?

willyd:
@Noiredd You are right. This is what I intend to do. I will leave the PR open as a reminder.

The windows branch has been trailing behind master a little bit lately, this may explain @IlyaOvodov desire to include this commit right away and not wait for a complete update.

IlyaOvodov:
As last merge from master to windows was on 24.05.17, I didn't expect the next one soon. But as far as I'm using windows, it is inconvenient  to have branches unsynchronized, if I'm going to make any new improvements.

IlyaOvodov:
By the way, why not to merge these 2 branches  into 1?
I've compared their versions on 24.05.17 and found that there are not a lot of differences and 99% of them are like if(MSVC) somth new else() code from master branch endif()

willyd:
> As last merge from master to windows was on 24.05.17, I didn't expect the next one soon.

Sorry about that, as the only windows maintainer I have limited time to dedicate to caffe-windows. I am currently refactoring the windows build to leverage vcpkg (kind of apt-get for windows) for dependency management. Once this is done, I will update the windows branch to match master. I hope to get there before March.

willyd:
> By the way, why not to merge these 2 branches into 1?
> I've compared their versions on 24.05.17 and found that there are not a lot of differences and 99% of them are like if(MSVC) somth new else() code from master branch endif()

I agree. Once the build and CI are stable again we should consider this seriously.

PR #6209, opened 2018-02-02
[not ready] Pycaffe improvements from OpenCL branch

Noiredd:
I'm bringing @naibaf7's pycaffe improvements from OpenCL branch to Master. Some pretty cool features have been implemented, most importantly exposure of `SolverParameter` and possibility to reinstatiate Solver, which would address #2766, and even update its parameters on-the-fly (#2686).  
I also want to expose `Solver::ApplyUpdate` (similar to how it's been done in [this fork](https://github.com/dosovits/caffe-fr-chairs/commit/6421df7fa8586364b843a52fd8ced8a1e8e2a813#diff-9b973c3db5bcd89fc0a5e1977f7cd70aR306)) to make it easier (possible at all?) to train GANs, but if it'd be better to have this in a separate PR (or done in a whole another way) - let me know.

Unfortunately many of the changes are interleaved with the evolution of that branch itself (modifications of `MemoryDataLayer`, adoption of `int_tp` type, some multi-device stuff, etc.) so it's not possible to just *pull them over to master* - plenty of conflicts arise due to interdependencies.  
This is a WIP with a request for feedback or ideas, especially from you @naibaf7 as the original author of those changes. I have tracked the history of OpenCL's pycaffe, dissecting every commit since July 2015, cherry-picking the relevant ones and cleaning them from OpenCL-specific code. Here is my plan of attack for this (ticked the already copied commits):
- [x] 61a05f3 `Python update, Lint fixes.` and d82979b `Softmax fix for >4 shape dims.`  
A heavy modification of the `MemoryDataLayer` and other stuff which I do not want to incorporate here; however, it exposes `Solver::max_iter` and allows passing a layer index in `Net_SetInputArrays`, so I just took the relevant pieces as a new commit.
- [x] faf003a `Pycaffe check update.` and 5ebebfb `Pycaffe fixed.`  
**ignored**, follow-ups to the `MemoryDataLayer` and `int_tp` changes
- [x] 01139b0 `Fix for Malis, Softmax.` and 165d231 `Removed duplicate function.`  
**ignored**, irrelevant to Master but introduce some conflicts around `.def("step")`
- [x] 984745d `Expose SolverParameter to Python.`  
This introduces an issue that I'm not sure how severe is nor how to deal with. Since it registers a new binding to `SolverParameter`, Python interpreter will now output the following warning every time we `import caffe`:  
`
[...]

Issue #2686, opened 2015-07-03
Changing learning rate / momentum from Python interface


PiranjaF:
Many SGD schedules specify that the learning rate should be annealed when validation accuracy has somewhat converged. Similarly, momentum should be around 0.5 in the beginning (http://ufldl.stanford.edu/tutorial/supervised/OptimizationStochasticGradientDescent/) and then to 0.9 when learning is stabilized. 

Can these parameters be changed during training from the Python interface? Having this as an option seems like an obvious enhancement.

goodells:
I could really use this in my code as well.

shelhamer:
This is certainly worth doing and only requires minor coding to make the hyperparams for `SGDSolver` mutable and exposed through boost::python.

As a hack this can be done by running the solver through `gdb` -- not that I'm really suggesting that.

talda:
@shelhamer a much simpler hack would be to train with an initial set of parameters and than change the parameters and fine tune on each step of parameter change.

pgrossmann:
@talda Would you mind sharing an example of how to change any of the hyperparameters (e.g., the learning rate) using PyCaffe? I tried inspecting the solver objects, but I couldn't find the variable that stores that parameter. Thank you!

ravi-june:
For solution, see http://stackoverflow.com/questions/31823898/changing-the-solver-parameters-in-caffe-through-pycaffe

mitar:
+1 on allowing one to change the learning rate and other parameters.

hala3:
i want to finetune a pretrained model which trained with pycaffe  
are there any tutorial can help me please ??

PR #6238, opened 2018-02-14
Expose necessary interface to guide manual SGD process from Python

### LINKED ISSUE #3959: Python manual sgd
I am trying to implement the SGD functionality to update weights in python manually in caffe python instead of using solver.step() function. The goal is to match the weight updates after doing solver.step() and that by manually updating the weights.

The setup is as follows: Use MNIST data. Set the random seed in solver.prototxt as: random_seed: 52. Make sure momentum: 0.0, weight_decay: 0.0 and, base_lr: 0.01, lr_policy: "fixed". This is done so that, I can simply implement the SGD update equation (with out momentum, regularization etc.). The equation is simply: W_t+1 = W_t - mu \* W_t_diff

Following are the two tests:

**Test1:** Using pycaffe's forward() and backward() to calculate the forward propagation and backward propagation. For each layer that contain weights I do:

```
for k in weight_layer_idx:
    solver.net.layers[k].blobs[0].diff[...] *= lr # weights
    solver.net.layers[k].blobs[1].diff[...] *= lr # biases
```

Next, update the weight/biases as:

```
    solver.net.layers[k].blobs[0].data[...] -= solver.net.layers[k].blobs[0].diff
    solver.net.layers[k].blobs[1].data[...] -= solver.net.layers[k].blobs[1].diff
```

I run this for 5 iterations.

**Test2:** Run caffe's solver.step(5).

Now, what I expect is the two tests should yield exactly same weights after the two iterations.

I save the weights values after each of the above tests and calculate the norm difference between the weight vectors by the two tests, and I see that they are not bit-exact. Can some one spot something that I might be missing?

Following is the entire code for reference:

```
import caffe
caffe.set_device(0)
caffe.set_mode_gpu()
import numpy as np
niter = 5
solver = None
solver = caffe.SGDSolver('solver.prototxt')

# Automatic SGD: TEST2
solver.step(niter)
# save the weights to compare later
w_solver_step = copy(solver.net.layers[1].blobs[0].data.astype('float64'))
b_solver_step = copy(solver.net.layers[1].blobs[1].data.astype('float64'))

# Manual SGD: TEST1
solver = None
solver = caffe.SGDSolver('solver.prototxt')
lr = 0.01

# Get layer types
layer_types = []
for ll in solver.net.layers:
    layer_types.append(ll.type)

# Get the indices of layers that have weights in them
weight_layer_idx = [idx for idx,l in enumerate(layer_types) if 'Convolution' in l or 'InnerProduct' in l]

for it in range(1, niter+1):
    solver.net.forward()  # fprop
    solver.net.backward()  # bprop
    for k in weight_layer_idx:
        solver.net.layers[k].blobs[0].diff[...] *= lr
        solver.net.layers[k].blobs[1].diff[...] *= lr
        solver.net.layers[k].blobs[0].data[...] -= solver.net.layers[k].blobs[0].diff
        solver.net.layers[k].blobs[1].data[...] -= solver.net.layers[k].blobs[1].diff

# save the weights to compare later
w_fwdbwd_update = copy(solver.net.layers[1].blobs[0].data.astype('float64'))
b_fwdbwd_update = copy(solver.net.layers[1].blobs[1].data.astype('float64'))

# Compare
print "after iter", niter, ": weight diff: ", np.linalg.norm(w_solver_step - w_fwdbwd_update), "and bias diff:", np.linalg.norm(b_solver_step - b_fwdbwd_update)
```

The last line that compares the weights with the two tests produces:

after iter 5 : weight diff:  0.000203027766144 and bias diff: 1.78390789051e-05 

where as I expect this difference to be 0.0

Any ideas?

---

Also note, 
If I run these two tests for only 1 iteration, I get exactly matching weight vectors from all layers, but not the subsequent iterations.

mitar:
This fixes #3959 and #2686.

EDIT by @Noiredd: do not merge until #6209 is merged first (or else terrible conflicts will arise during `master` to `opencl` pull).

### DIFF
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -507,6 +507,7 @@ BOOST_PYTHON_MODULE(_caffe) {
     .def("restore", &Solver<Dtype>::Restore)
     .def("snapshot", &Solver<Dtype>::Snapshot)
     .def("share_weights", &share_weights)
+    .def("apply_update", &Solver<Dtype>::ApplyUpdate)
     .add_property("param", bp::make_function(&Solver<Dtype>::param,
               bp::return_value_policy<bp::copy_const_reference>()));
   BP_REGISTER_SHARED_PTR_TO_PYTHON(Solver<Dtype>);
--- a/python/caffe/_caffe.cpp
+++ b/python/caffe/_caffe.cpp
@@ -490,7 +490,9 @@ BOOST_PYTHON_MODULE(_caffe) {
   bp::class_<SolverParameter>("SolverParameter", bp::no_init)
     .add_property("max_iter", &SolverParameter::max_iter)
     .add_property("display", &SolverParameter::display)
-    .add_property("layer_wise_reduce", &SolverParameter::layer_wise_reduce);
+    .add_property("layer_wise_reduce", &SolverParameter::layer_wise_reduce)
+    .add_property("base_lr", &SolverParameter::base_lr,
+           &SolverParameter::set_base_lr);
   bp::class_<LayerParameter>("LayerParameter", bp::no_init);
 
   bp::class_<Solver<Dtype>, shared_ptr<Solver<Dtype> >, boost::noncopyable>(
@@ -509,25 +511,26 @@ BOOST_PYTHON_MODULE(_caffe) {
     .def("share_weights", &share_weights)
     .def("apply_update", &Solver<Dtype>::ApplyUpdate)
     .add_property("param", bp::make_function(&Solver<Dtype>::param,
-              bp::return_value_policy<bp::copy_const_reference>()));
+              bp::return_internal_reference<>()));
   BP_REGISTER_SHARED_PTR_TO_PYTHON(Solver<Dtype>);
 
   bp::class_<SGDSolver<Dtype>, bp::bases<Solver<Dtype> >,
     shared_ptr<SGDSolver<Dtype> >, boost::noncopyable>(
-        "SGDSolver", bp::init<string>());
-  bp::class_<NesterovSolver<Dtype>, bp::bases<Solver<Dtype> >,
+        "SGDSolver", bp::init<string>())
+        .add_property("lr", &SGDSolver<Dtype>::GetLearningRate);
+  bp::class_<NesterovSolver<Dtype>, bp::bases<SGDSolver<Dtype> >,
     shared_ptr<NesterovSolver<Dtype> >, boost::noncopyable>(
         "NesterovSolver", bp::init<string>());
-  bp::class_<AdaGradSolver<Dtype>, bp::bases<Solver<Dtype> >,
+  bp::class_<AdaGradSolver<Dtype>, bp::bases<SGDSolver<Dtype> >,
     shared_ptr<AdaGradSolver<Dtype> >, boost::noncopyable>(
         "AdaGradSolver", bp::init<string>());
-  bp::class_<RMSPropSolver<Dtype>, bp::bases<Solver<Dtype> >,
+  bp::class_<RMSPropSolver<Dtype>, bp::bases<SGDSolver<Dtype> >,
     shared_ptr<RMSPropSolver<Dtype> >, boost::noncopyable>(
         "RMSPropSolver", bp::init<string>());
-  bp::class_<AdaDeltaSolver<Dtype>, bp::bases<Solver<Dtype> >,
+  bp::class_<AdaDeltaSolver<Dtype>, bp::bases<SGDSolver<Dtype> >,
     shared_ptr<AdaDeltaSolver<Dtype> >, boost::noncopyable>(
         "AdaDeltaSolver", bp::init<string>());
-  bp::class_<AdamSolver<Dtype>, bp::bases<Solver<Dtype> >,
+  bp::class_<AdamSolver<Dtype>, bp::bases<SGDSolver<Dtype> >,
     shared_ptr<AdamSolver<Dtype> >, boost::noncopyable>(
         "AdamSolver", bp::init<string>());
 
--- a/src/caffe/solver.cpp
+++ b/src/caffe/solver.cpp
@@ -266,10 +266,6 @@ void Solver<Dtype>::Step(int iters) {
     }
     ApplyUpdate();
 
-    // Increment the internal iter_ counter -- its value should always indicate
-    // the number of times the weights have been updated.
-    ++iter_;
-
     SolverAction::Enum request = GetRequestedAction();
 
     // Save a snapshot if needed.
--- a/include/caffe/sgd_solvers.hpp
+++ b/include/caffe/sgd_solvers.hpp
@@ -23,10 +23,11 @@ class SGDSolver : public Solver<Dtype> {
 
   const vector<shared_ptr<Blob<Dtype> > >& history() { return history_; }
 
+  virtual void ApplyUpdate();
+
  protected:
   void PreSolve();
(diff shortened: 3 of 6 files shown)


python/caffe/_caffe.cpp
```diff
@@ -490,7 +490,8 @@ BOOST_PYTHON_MODULE(_caffe) {
   bp::class_<SolverParameter>("SolverParameter", bp::no_init)
     .add_property("max_iter", &SolverParameter::max_iter)
     .add_property("display", &SolverParameter::display)
-    .add_property("layer_wise_reduce", &SolverParameter::layer_wise_reduce);
+    .add_property("layer_wise_reduce", &SolverParameter::layer_wise_reduce)
+    .add_property("base_lr", &SolverParameter::base_lr, &SolverParameter::set_base_lr);
```
Noiredd: This change is included in #6209 which will be merged prior to any other changes to `_caffe.cpp`

python/caffe/_caffe.cpp
```diff
@@ -507,13 +508,15 @@ BOOST_PYTHON_MODULE(_caffe) {
...
     .add_property("param", bp::make_function(&Solver<Dtype>::param,
-              bp::return_value_policy<bp::copy_const_reference>()));
+              bp::return_internal_reference<>()));
   BP_REGISTER_SHARED_PTR_TO_PYTHON(Solver<Dtype>);
 
   bp::class_<SGDSolver<Dtype>, bp::bases<Solver<Dtype> >,
     shared_ptr<SGDSolver<Dtype> >, boost::noncopyable>(
-        "SGDSolver", bp::init<string>());
+        "SGDSolver", bp::init<string>())
+        .def("get_learning_rate", &SGDSolver<Dtype>::GetLearningRate);
```
Noiredd: What about other solvers, will they get this property only due to the inheritance?

src/caffe/solver.cpp
```diff
@@ -266,10 +266,6 @@ void Solver<Dtype>::Step(int iters) {
     }
     ApplyUpdate();
 
-    // Increment the internal iter_ counter -- its value should always indicate
-    // the number of times the weights have been updated.
-    ++iter_;
-
```
Noiredd: Why did this have to be moved to `SGDSolver`?

python/caffe/_caffe.cpp
mitar: Oho, looking at #6209 looks like everything in this pull request is done also there, no?

Noiredd: This change is included in #6209 which will be merged prior to any other changes to `_caffe.cpp`.

mitar:
Thanks for the quick code review!

python/caffe/_caffe.cpp
Noiredd: What about other solvers, will they get this property only due to the inheritance?

src/caffe/solver.cpp
Noiredd: Why did this have to be moved to SGDSolver?

Noiredd:
Please run `make lint` and fix the style issues and we will resume the review.  
Also, I'm not sure if you noticed but I accidentally removed your comment - please repost it if needed.

python/caffe/_caffe.cpp
mitar: Oho, looking at #6209 looks like everything in this pull request is done also there, no?

python/caffe/_caffe.cpp
mitar: Other solvers extending from `SGDSolver`? I think only them will get this, but this is OK, no? this is on purpose.

src/caffe/solver.cpp
mitar: Based on the comment: "its value should always indicate the number of times the weights have been updated."

Because now it is possible to update weights without having to call `step`, this has to be moved so that also calling directly `ApplyUpdate` increases it. I think that this is even cleaner from the design perspective. The code is next to the code which is updating weights. Not somewhere else.

mitar:
> Good addition, I was about to implement something like this myself but first we need to pull some Pycaffe changes from the OpenCL branch to synchronize the whole repository. For this reason I recommend holding this until #6209 is reviewed and merged.

Thanks. I can rebase after that is merged.

> Please run make lint and fix the style issues and we will resume the review.

I cannot run it, it says:

```
Makefile:6: *** Makefile.config not found. See Makefile.config.example..  Stop.
```

Also, I cannot see results from Travis CI run. Are they set as private?

python/caffe/_caffe.cpp
Noiredd: So if I do `a = caffe.AdamSolver("solver.prototxt")` in my pycaffe script, will I be able to call `a.get_learning_rate()` or not?

mitar:
Pushed linting fixes.

python/caffe/_caffe.cpp
mitar: To my understanding, the answer is yes, but I have not tested it.

Noiredd: Would be good to know this. Maybe consider implementing some [tests](https://github.com/BVLC/caffe/blob/master/python/caffe/test/test_solver.py) for this change? See #4342, it is somewhat similar to this (a bit more clumsy) but also had some very simple tests.

shelhamer:
Thanks @mitar! I have been meaning to make a PR like this myself for some time. I'll take a look at both #6209 and this soon.

src/caffe/solver.cpp
shelhamer: Ideally the base `Solver` would enforce the increment to `iter_`, but since all Caffe solvers are SGD solvers and other solvers that depart from this mold aren't expected I think the move is a pragmatic choice that avoids unnecessary interface code.

python/caffe/_caffe.cpp
shelhamer: No, this does not inherit correctly b.c. the `bp::bases` of the child `SGDSolver`s are set to `Solver` and not `SGDSolver`. Switching the `NesteroveSolver` and other children to have the `SGDSolver` base fixes the issue. I second incorporating the `apply_update()` test from c5b5b55845c20930026f1f563136247c41f58075, so could you cherry-pick it to give @nitnelave credit for it?

shelhamer:
@mitar I had time to do a pass so I updated this branch by rebasing on master, rewording for detail, and including a test for updating from pycaffe. I'm planning to merge once the tests pass to double-check my local results.

mitar:
Oh, thanks!

mitar:
What are plans for a release which would include this? Or do you have some daily build I could pip install without compiling the repo myself?

shelhamer:
The current state of affairs with master is that you have to compile it yourself.

> What are plans for a release which would include this?

There is no definite plan right now but I could imagine a release in early July to collect improvements since 1.0. I will bring it up with the core devs.

> Or do you have some daily build I could pip install without compiling the repo myself?

No, we do not have a daily build, although it would obviously be useful. I've never rigged up such a system myself, but perhaps a daily linux build (specifically for the Ubuntu LTS) would be feasible.

mitar:
> No, we do not have a daily build, although it would obviously be useful. I've never rigged up such a system myself, but perhaps a daily linux build (specifically for the Ubuntu LTS) would be feasible.

You could even push artifacts from Travis CI.

I think you also publish a PyPi package like other libraries do. Using [manylinux](https://github.com/pypa/manylinux) and if you need more space there for the package, [you can request it here](https://github.com/pypa/packaging-problems).

PR #6245, opened 2018-02-20
Fix failure to run on system without GPU in CPU mode

artyom-beilis:
When running caffe compiled with CUDNN in CPU mode on hardware without GPU it fails with errors like

     cudnn_conv_layer.cpp:52] Check failed: error == cudaSuccess (35 vs. 0)  CUDA driver version is insufficient for CUDA runtime version

Using Caffe default engine instead of CUDNN for CPU only mode fixes the issue

### DIFF
--- a/src/caffe/layer_factory.cpp
+++ b/src/caffe/layer_factory.cpp
@@ -50,7 +50,7 @@ shared_ptr<Layer<Dtype> > GetConvolutionLayer(
   if (engine == ConvolutionParameter_Engine_DEFAULT) {
     engine = ConvolutionParameter_Engine_CAFFE;
 #ifdef USE_CUDNN
-    if (!use_dilation) {
+    if (!use_dilation && Caffe::mode()==Caffe::GPU) {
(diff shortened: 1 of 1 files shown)


Noiredd:
Can you elaborate on the issue? I cannot even compile Caffe with `USE_CUDNN` and `CPU_ONLY` together.

artyom-beilis:
> Can you elaborate on the issue? I cannot even compile Caffe with USE_CUDNN and CPU_ONLY together

I have one version of caffe build with CUDNN support - such that on machines with GPU I can start it in GPU mode and without one in CPU mode and of course I call `set_mode(Caffe::CPU)` or GPU accordingly

However it fails to run in CPU mode despite I asked to do prediction on CPU because all CUDNN layers do some initialization _specific to gpu_ in LayerSetUp - something that caffe default layers don't do.

After this patch I can deploy caffe library on machine without GPU when I setup it in CPU mode.

artyom-beilis:
The major goal is to deploy save build of Caffe on different machines regardless they have nVidia GPU or not.

PR #6249, opened 2018-02-23
Remove legacy tools

### LINKED ISSUE #5543: [minor] Please remove useless stubs e.g. tools/train_net
These binary files to be built are useless, please consider a removal.

```
net_speed_benchmark
train_net
test_net
finetune_net
device_query
```

Get rid of these legacy stuff.

Noiredd:
It's been almost 4 years (since a916ef4), I think we can get rid of those?

Fixes #5543


naibaf7:
I agree :)

PR #6253, opened 2018-02-26
Selecting only a specified class for backprop using SoftmaxWithLoss

SOLARleisu:
1. Add an new variable for softmax_loss_layer to allow the user to give
different weight to different class ( or label).
  loss_param {
      only_care_label: 0
  }
2. Tested with examples/mnist/train_lenet.sh for both GPU
mode and CPU mode.
3. Use examples/mnist/lenet_train_test.prototxt as an example for
only_care_label.
4. Note: ignore_label gets the priority if only_care_label and
ignore_label were in conflict.

5. Update the .gitignore for some vim-plugin files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,12 @@ LOCK
 LOG*
 CURRENT
 MANIFEST-*
+
+#leisu
+cscope.files
+cscope.in.out
+cscope.out
+cscope.po.out
+filenametags
+tags
+
--- a/include/caffe/layers/softmax_loss_layer.hpp
+++ b/include/caffe/layers/softmax_loss_layer.hpp
@@ -119,6 +119,10 @@ class SoftmaxWithLossLayer : public LossLayer<Dtype> {
   bool has_ignore_label_;
   /// The label indicating that an instance should be ignored.
   int ignore_label_;
+  /// Whether to only care about an instance with certain label.
+  bool has_only_care_label;
+  /// The only label we care about in this loss-layer.
+  int only_care_label;
   /// How to normalize the output loss.
   LossParameter_NormalizationMode normalization_;
 
--- a/src/caffe/proto/caffe.proto
+++ b/src/caffe/proto/caffe.proto
@@ -448,6 +448,7 @@ message TransformationParameter {
 message LossParameter {
   // If specified, ignore instances with the given label.
   optional int32 ignore_label = 1;
+  optional int32 only_care_label = 4;
   // How to normalize the loss for loss layers that aggregate across batches,
   // spatial dimensions, or other dimensions.  Currently only implemented in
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -97,3 +97,12 @@ LOCK
...
 MANIFEST-*
+
+#leisu
+cscope.files
+cscope.in.out
+cscope.out
+cscope.po.out
+filenametags
+tags
+
```
Noiredd: Please do not change the GIT settings unless your PR is strictly about that.

include/caffe/layers/softmax_loss_layer.hpp
```diff
@@ -119,6 +119,10 @@ class SoftmaxWithLossLayer : public LossLayer<Dtype> {
   bool has_ignore_label_;
   /// The label indicating that an instance should be ignored.
   int ignore_label_;
+  /// Whether to only care about an instance with certain label.
+  bool has_only_care_label;
+  /// The only label we care about in this loss-layer.
+  int only_care_label;
```
Noiredd: Class member variables should end with an underscore. Side note: why doesn't `lint` detect this?

Also, if I may suggest, a better name for this field would be `select_label`, or `specify_label` - to make it extra clear that this means: _select just these instances for the loss calculation_. Or maybe someone has a better idea?

src/caffe/proto/caffe.proto
```diff
@@ -448,6 +448,7 @@ message TransformationParameter {
 message LossParameter {
   // If specified, ignore instances with the given label.
   optional int32 ignore_label = 1;
+  optional int32 only_care_label = 4;
```
Noiredd: Same as above: let's consider `select_label` or `specify_label` instead. Also, a word of a comment above would be nice.

SOLARleisu: Thank you very much for your advice. I will modify them.
I am sorry for being late.

include/caffe/layers/softmax_loss_layer.hpp
SOLARleisu: thank you for you side note.

SOLARleisu:
I will make a new request name as "Selecting only a specified class for backprop using SoftmaxWithLoss #6253"

src/caffe/proto/caffe.proto
SOLARleisu: I will make a new request name as "Selecting only a specified class for backprop using SoftmaxWithLoss #6253"

include/caffe/layers/softmax_loss_layer.hpp
SOLARleisu: Thank you for you advice. I will make a new request named "Enhance SoftmaxWithLoss by adding select_label. #6324".
Please review the new one.

Issue #6268, opened 2018-03-01
ClipGradients cause  CUBLAS_STATUS_MAPPING_ERROR


twmht:
I am trying to do a language model with LSTM.

the network architecture is from https://github.com/jeffdonahue/caffe/blob/recurrent/examples/coco_caption/lstm_language_model.prototxt

the solver is from https://github.com/jeffdonahue/caffe/blob/recurrent/examples/coco_caption/lstm_lm_solver.prototxt

However,  the errors happened when calling ClipGradients()

```
F0301 23:34:49.134544 15611 math_functions.cu:130] Check failed: status == CUBLAS_STATUS_SUCCESS (11 vs. 0)  CUBLAS_STATUS_MAPPING_ERROR
*** Check failure stack trace: ***
    @     0x7ff6d6b055cd  google::LogMessage::Fail()
    @     0x7ff6d6b07433  google::LogMessage::SendToLog()
    @     0x7ff6d6b0515b  google::LogMessage::Flush()
    @     0x7ff6d6b07e1e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7ff6d75a71da  caffe::caffe_gpu_dot<>()
    @     0x7ff6d73dfffe  caffe::Blob<>::sumsq_diff()
    @     0x7ff6d7343c3a  caffe::SGDSolver<>::ClipGradients()
    @     0x7ff6d73499ec  caffe::SGDSolver<>::ApplyUpdate()
    @     0x7ff6d75926c6  caffe::Solver<>::Step()
    @     0x7ff6d75934e3  caffe::Solver<>::Solve()
    @           0x40d2c5  train()
    @           0x40810d  main
    @     0x7ff6d507e830  __libc_start_main
    @           0x408f69  _start
    @              (nil)  (unknown)

```

Any idea?

Noiredd:
CUDA/cuDNN/driver version, hardware, OS?

twmht:
Sorry for missing that, Here is the software/hardware info

cuda: 8.0
cudnn: 6.0
os: ubuntu 16.04

Noiredd:
Does your build pass the tests? What is the output of `make runtest`?

Noiredd:
Well, I've had a similar mapping error when trying to run a net with cuDNN backend on a card that didn't support it. But it [looks like](https://developer.nvidia.com/cuda-gpus) the 970M should be able to run it. Did you compile with `USE_CUDNN := 1`?

[...]

PR #6277, opened 2018-03-06
check embed index in debug mode

twmht:
checking embed index should be enabled in non-debug mode.

### DIFF
--- a/src/caffe/layers/embed_layer.cu
+++ b/src/caffe/layers/embed_layer.cu
@@ -15,6 +15,11 @@ __global__ void EmbedForward(const int nthreads, const Dtype* bottom_data,
     const int n = top_index / N;
     const int d = top_index % N;
     const int index = static_cast<int>(bottom_data[n]);
+    #ifdef DEBUG
+        assert(index >= 0);
+        assert(index < K);
+        assert(static_cast<Dtype>(index) == bottom_data[n]);
+    #endif
     const int weight_index = index * N + d;
     top_data[top_index] = weight[weight_index];
   }


Noiredd:
I will copy your findings from the issue #6268 to start the conversation:

> when input contains a negative padding value, the kernel function in the embed layer may have undefined behavior.
it's hard to debug, since embed layer in gpu mode does not check the valid values for input. However, it does not fail on the embed layer but on other part of cuda code (like cudaCheckEvent in the Timer).
in cpu mode, it only checks the valid values in debug mode.

However, my opinion is that this is what the debug mode is for. By default (in release) we focus on performance and do not make any checks during propagation. In case of problems the scenario is to fallback to the debug mode and check what's going on - so in a way, this PR breaks this workflow.

twmht:
@Noiredd 

I see.

twmht:
@Noiredd 

But why not add DCHECK utility in kernel function of Embed layer?

If I fall back to Debug Mode, it's still hard to debug. Since it would not fail on embed layer but on the other part of cuda API.

Noiredd:
Debugging or safeguarding GPU code is not straightforward: kernels (GPU functions) cannot execute host (CPU) code, so any DCHECK placed there will fail the compilation. In situations like this, I think one is supposed to not only fall back to debug but also CPU mode.  
Perhaps a check could be added behind `#ifdef DEBUG`, for example by running a special debug kernel before the actual `EmbedForward` kernel?

I think this PR should be open, if only to discuss potential solutions (and hear other's opinions).

twmht:
@Noiredd 

Adding `#ifdef DEBUG` is a good idea.

I may update this PR in a few days.

thanks.

twmht:
@Noiredd 

I have updated this PR, and it works in debug mode.

If I have negative index or other unaccepted index, then the outputs can correctly tell me what is wrong, this is the same behavior when running in cpu mode.

Here is the output

```
src/caffe/layers/embed_layer.cu:19: void caffe::EmbedForward(int, const Dtype *, const Dtype *, int, int, int, Dtype *) [with Dtype = float]: block: [1765,0,0], thread: [251,0,0] Assertion `index >= 0` failed.
src/caffe/layers/embed_layer.cu:19: void caffe::EmbedForward(int, const Dtype *, const Dtype *, int, int, int, Dtype *) [with Dtype = float]: block: [1765,0,0], thread: [252,0,0] Assertion `index >= 0` failed.
src/caffe/layers/embed_layer.cu:19: void caffe::EmbedForward(int, const Dtype *, const Dtype *, int, int, int, Dtype *) [with Dtype = float]: block: [1765,0,0], thread: [253,0,0] Assertion `index >= 0` failed.
src/caffe/layers/embed_layer.cu:19: void caffe::EmbedForward(int, const Dtype *, const Dtype *, int, int, int, Dtype *) [with Dtype = float]: block: [1765,0,0], thread: [254,0,0] Assertion `index >= 0` failed.
src/caffe/layers/embed_layer.cu:19: void caffe::EmbedForward(int, const Dtype *, const Dtype *, int, int, int, Dtype *) [with Dtype = float]: block: [1765,0,0], thread: [255,0,0] Assertion `index >= 0` failed.
F0323 22:00:47.647169 14711 syncedmem.hpp:39] Check failed: error == cudaSuccess (59 vs. 0)  device-side assert triggered
    @     0x7f79ce3045cd  google::LogMessage::Fail()
    @     0x7f79ce306433  google::LogMessage::SendToLog()
    @     0x7f79ce30415b  google::LogMessage::Flush()
    @     0x7f79ce306e1e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7f79cebc6c24  caffe::CaffeFreeHost()
    @     0x7f79cebc5da0  caffe::SyncedMemory::~SyncedMemory()
    @     0x7f79ceadf609  boost::checked_delete<>()
    @     0x7f79ceadf9f4  boost::detail::sp_counted_impl_p<>::dispose()
    @           0x41ff4a  boost::detail::sp_counted_base::release()
    @           0x41ffdd  boost::detail::shared_count::~shared_count()
    @     0x7f79ceaddd5e  boost::shared_ptr<>::~shared_ptr()
    @     0x7f79ceade262  boost::shared_ptr<>::operator=()
    @     0x7f79cead1dc6  caffe::Blob<>::ShareData()
    @     0x7f79cec2f82a  caffe::Net<>::ShareWeights()
    @     0x7f79cec6e413  caffe::RecurrentLayer<>::Forward_gpu()
    @           0x422606  caffe::Layer<>::Forward()
    @     0x7f79cec2ebfb  caffe::Net<>::ForwardFromTo()
    @     0x7f79cec2e863  caffe::Net<>::Forward()
    @     0x7f79cec21332  caffe::Solver<>::Test()
    @     0x7f79cec2103e  caffe::Solver<>::TestAll()
    @     0x7f79cec1f916  caffe::Solver<>::Step()
    @     0x7f79cec1f407  caffe::Solver<>::Solve()
    @           0x41cba9  train()
    @           0x41ef31  main
    @     0x7f79cca9a830  __libc_start_main
    @           0x41b4b9  _start

```

Noiredd:
This is a reasonable solution -- thank you @twmht

PR #6278, opened 2018-03-06
1D blob handling in MSRA/Xavier fillers + new filler tests

### LINKED ISSUE #6275: XavierFilller and MSRAFiller fail to fill a bias blob
### Issue summary
XavierFilller and MSRAFiller doesn't work in current master, the error info as below:
```
I0305 17:03:08.437372  6693 net.cpp:88] Creating Layer conv1
I0305 17:03:08.437381  6693 net.cpp:410] conv1 <- data
I0305 17:03:08.437389  6693 net.cpp:384] conv1 -> conv1
F0305 17:03:08.437960  6693 blob.hpp:122] Check failed: axis_index < num_axes() (1 vs. 1) axis 1 out of range for 1-D Blob with shape 64 (64)
```

### Steps to reproduce
run a caffe prototxt with Convolution, which para filler use Xavier or MSRA.
I ran a Resnet50 prototxt. Part as follow:
```
layer {
	bottom: "data"
	top: "conv1"
	name: "conv1"
	type: "Convolution"
	convolution_param {
		num_output: 64
		kernel_size: 7
		pad: 3
		stride: 2
                weight_filler {
                  type: "msra"
                }
                bias_filler {
                  type: "msra"
                }
	}
}
```

Noiredd:
#6240 changed fillers to use `Blob::shape` instead of legacy accessors. However, the legacy functions returned 1 when the requested dimension didn't exist (e.g. `Blob::width` on a 1D blob). Unfortunately, some fillers (Xavier, MSRA) relied on this fact when computing `fan_out`. Since `shape` will fail a `CHECK` when a dimension is exceeded, issue #6275 happened (attempt to fill a bias blob with MSRA will cause an error).  
This PR adds the necessary checks to the filler code - now a filler will check whether a specified dimension can be requested before calling `shape`.

We could also consider upgrading tests for fillers - currently they are only ran on 4D blobs. Tests for other dimensions might, IMHO, skip the mathematical correctness checks and just make sure we can fill other things (perhaps with a `LOG` on Xavier and MSRA that they were designed for 4D blobs).

Fixes #6275

### DIFF
--- a/include/caffe/filler.hpp
+++ b/include/caffe/filler.hpp
@@ -148,8 +148,10 @@ class XavierFiller : public Filler<Dtype> {
   virtual void Fill(Blob<Dtype>* blob) {
     CHECK(blob->count());
     int fan_in = blob->count() / blob->shape(0);
-    // Compatible for ND Convolution
-    int fan_out = blob->count() / blob->shape(1);
+    // Compatibility with ND blobs
+    int fan_out = blob->num_axes() > 1 ?
+                  blob->count() / blob->shape(1) :
+                  blob->count();
     Dtype n = fan_in;  // default to fan_in
     if (this->filler_param_.variance_norm() ==
         FillerParameter_VarianceNorm_AVERAGE) {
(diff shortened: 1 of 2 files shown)


Noiredd:
I refactored the filler tests - now we can run them on different shapes of blobs more easily. The test suite has been expanded to make use of that, so we now check 1D as well as ND blobs. Those new tests would fail on an issue like #6275.

PR #6282, opened 2018-03-07
PoolingLayer customizable output shape rounding mode

### LINKED ISSUE #1318: Should pooling regions be identical to convolution regions?

Noiredd:
Resurrection of #3057 which allowed the user to choose how is the pooling output shape calculated, as there is confusion between how does the PoolingLayer do it and how does ConvolutionLayer (see @jeffdonahue's [note](https://github.com/BVLC/caffe/pull/473#issuecomment-45386156) on this).

This implementation incorporates @seanbell's [idea](https://github.com/BVLC/caffe/pull/3057#issuecomment-139284546) to have the rounding mode as an enum.

Fixes #1318
Closes #3057

### DIFF
--- a/include/caffe/layers/pooling_layer.hpp
+++ b/include/caffe/layers/pooling_layer.hpp
@@ -51,6 +51,7 @@ class PoolingLayer : public Layer<Dtype> {
   int height_, width_;
   int pooled_height_, pooled_width_;
   bool global_pooling_;
+  PoolingParameter_RoundMode round_mode_;
   Blob<Dtype> rand_idx_;
   Blob<int> max_idx_;
 };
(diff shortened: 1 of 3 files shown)


shaibagon:
@Noiredd thank you for resurrecting #3057, what keeps you from merging this PR?

wuz93308:
+1 for merging this!

Noiredd:
Sorry for a long delay, I've been tied up with work. I'll try to catch up with PRs and issues in the following days.

PR #6303, opened 2018-03-20
[AVE POOL ISSUE] New thread of #6296, Add an optional switch to support dynamic pool_size in AVE pooling

oscarriddle:
Because the commits on old thread #6296 got messed up, I try to create this new thread based on a clean branch. If @Noiredd could help close that thread, it will be very helpful.

1) In this fix, I add an optional boolean switch named "dynamic_ave_pool_size" in the caffe.proto, and default=false. 
2) Then pass the handle of layer_param_.pooling_param() into Forward_cpu() and Backward_cpu() functions in pooling_layer.cpp, and the AvePoolForward() function in pooling_layer.cu (The AvePoolBackward() seems correct already).
3) If the dynamic_ave_pool_size==true, the variable "pool_size" would be updated immediately after the 4 coordinates been updated. 

I just tested the Forward_cpu() version, it seems good. I will test the GPU forward version later on, but need some help on the Backward_cpu().

I'm not quite familiar with travis, I will try to learn the scripts of travis under the repo and try to use it on my own travis account later.

Thanks,
Xiaolun Cao

### DIFF
--- a/src/caffe/layers/pooling_layer.cpp
+++ b/src/caffe/layers/pooling_layer.cpp
@@ -127,6 +127,7 @@ void PoolingLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
 template <typename Dtype>
 void PoolingLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
+  PoolingParameter pool_param = this->layer_param_.pooling_param();
   const Dtype* bottom_data = bottom[0]->cpu_data();
   Dtype* top_data = top[0]->mutable_cpu_data();
   const int top_count = top[0]->count();
@@ -203,6 +204,9 @@ void PoolingLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
             wstart = max(wstart, 0);
             hend = min(hend, height_);
             wend = min(wend, width_);
+	    if(pool_param.dynamic_ave_pool_size()==true) {
+	      pool_size = (hend - hstart) * (wend - wstart);
+	    }
             for (int h = hstart; h < hend; ++h) {
               for (int w = wstart; w < wend; ++w) {
                 top_data[ph * pooled_width_ + pw] +=
@@ -229,6 +233,7 @@ void PoolingLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
 template <typename Dtype>
 void PoolingLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
       const vector<bool>& propagate_down, const vector<Blob<Dtype>*>& bottom) {
+  PoolingParameter pool_param = this->layer_param_.pooling_param();
   if (!propagate_down[0]) {
     return;
   }
@@ -284,6 +289,9 @@ void PoolingLayer<Dtype>::Backward_cpu(const vector<Blob<Dtype>*>& top,
             wstart = max(wstart, 0);
             hend = min(hend, height_);
             wend = min(wend, width_);
+	    if(pool_param.dynamic_ave_pool_size()==true) {
+	      pool_size = (hend - hstart) * (wend - wstart);
+	    }
             for (int h = hstart; h < hend; ++h) {
               for (int w = wstart; w < wend; ++w) {
                 bottom_diff[h * width_ + w] +=
--- a/src/caffe/layers/pooling_layer.cu
+++ b/src/caffe/layers/pooling_layer.cu
@@ -53,6 +53,7 @@ __global__ void AvePoolForward(const int nthreads,
     const int pooled_width, const int kernel_h, const int kernel_w,
     const int stride_h, const int stride_w, const int pad_h, const int pad_w,
     Dtype* const top_data) {
+    PoolingParameter pool_param = this->layer_param_.pooling_param();
   CUDA_KERNEL_LOOP(index, nthreads) {
     const int pw = index % pooled_width;
     const int ph = (index / pooled_width) % pooled_height;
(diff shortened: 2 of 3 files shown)


src/caffe/layers/pooling_layer.cpp
```diff
@@ -127,6 +127,7 @@ void PoolingLayer<Dtype>::Reshape(const vector<Blob<Dtype>*>& bottom,
 template <typename Dtype>
 void PoolingLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
       const vector<Blob<Dtype>*>& top) {
+  PoolingParameter pool_param = this->layer_param_.pooling_param();
```
Noiredd: How about, instead of reading the param every time, just add a class member variable to hold the value of the flag, and just set it once during `LayerSetUp`?

src/caffe/layers/pooling_layer.cpp
```diff
@@ -203,6 +204,9 @@ void PoolingLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
             wstart = max(wstart, 0);
             hend = min(hend, height_);
             wend = min(wend, width_);
+            if (pool_param.dynamic_ave_pool_size() == true) {
```
Noiredd: `if (pool_param.dynamic_ave_pool_size())` will suffice. And if you decide to follow my advice from the comment above, it will get even simpler (`if (member_var_)`).

src/caffe/layers/pooling_layer.cu
```diff
@@ -62,11 +62,14 @@ __global__ void AvePoolForward(const int nthreads,
     int wstart = pw * stride_w - pad_w;
     int hend = min(hstart + kernel_h, height + pad_h);
     int wend = min(wstart + kernel_w, width + pad_w);
-    const int pool_size = (hend - hstart) * (wend - wstart);
+    int pool_size = (hend - hstart) * (wend - wstart);
     hstart = max(hstart, 0);
     wstart = max(wstart, 0);
     hend = min(hend, height);
     wend = min(wend, width);
+    if (dynamic_ave_pool_size == true) {
```
Noiredd: Like above: `if (dynamic_ave_pool_size)` would be cleaner.

src/caffe/proto/caffe.proto
```diff
@@ -935,6 +935,10 @@ message PoolingParameter {
   // If global_pooling then it will pool over the size of the bottom by doing
   // kernel_h = bottom->height and kernel_w = bottom->width
   optional bool global_pooling = 12 [default = false];
+  // If dynamic_ave_pool_size is set to true, every iteration of AVE pooling will only
+  // concerns valid elements: (sum of elements within kernel)/(valid elements num)
+  // Otherwise, invalid padded elements will always be counted in the denominator
+  optional bool dynamic_ave_pool_size = 13 [default = false];
```
Noiredd: My last comment and the most opinion-based: is this the best name for this flag? How about `valid_ave_pooling`?  
Also, while it's good that you've explained what this does, I think that the comment should make it more clear that _valid pooling_ ignores the padding elements in the calculation.

oscarriddle:
Hi Noiredd, 

Thanks for help reviewing my fix. Now the code has been updated accordingly. 
However, I'm not quite sure about why the version 1fddd65 can pass the travis CI, but after I modified several comment lines, and new version 14f04c9 and ad291f5 become failed. I'm still trying to address the issue.
The new fix is test passed under my project. I think I can help write a caffe test for this if you think that would be much better.

Thanks,
Xiaolun Cao

oscarriddle:
`E: Failed to fetch http://dl.bintray.com/apache/cassandra/dists/39x/main/binary-amd64/Packages  502  Bad Gateway [IP: 52.10.151.113 80]
E: Failed to fetch http://dl.bintray.com/apache/cassandra/dists/39x/main/binary-i386/Packages  502  Bad Gateway [IP: 52.10.151.113 80]
E: Some index files failed to download. They have been ignored, or old ones used instead.`

Seems something happened to Travis CI server, perhaps I need to try again later.

Noiredd:
Thanks for the fixes. One other thing: now that you use a member variable, we don't need to declare `PoolingParameter pool_param` in forward/backward methods.

As I said, this passes tests because it is actually never tested - every time any test case instantiates a Pooling Layer, it does so with its default setting, which is without `valid_ave_pooling`. For this reason this needs an explicit test for this particular functionality.

Those random Travis errors are obviously not related to the code, so let's not worry about them for now. If need be, I will restart the tests.

oscarriddle:
The param declaration within Forward/Backward are removed now, and new ver. passed the travis.

Yes, I suggest this needs specific tests, and I'm looking into the existed tests and learning to create one. Besides, I think this feature shall be both tested under the CPU and GPU backend.

Thanks,

oscarriddle:
I created an explicit test named "TestForwardAveValid", it will turn on the member variable valid_ave_pooling and compare the outputs. 

I noticed the comments in the cuDNN version tests, says cuDNN doesn't support padding. So I'm not quite sure about the test method in cuDNN version. 

If you would help take a review, it'll be much appreciated.

Thanks,

Noiredd:
>cuDNN doesn't support padding

Must be some old information, I'm quite sure we can use padded pooling with cuDNN.  
I think we should test the padded version too, as without it there is no difference between valid/invalid pooling.

Another thing is that the blob that you're using is quite trivial - filled with a constant value. Are you sure this will be a solid test?  
Also, what do you think about testing the backward pass on this as well?

oscarriddle:
Sorry, I was busy working on other projects afterwards.

The test blob that used in this test is imitating the existed average pool test, which is "TestForwardAve".

I'm not 100% sure about the meaning of gradient test that in test_pooling_layer.cpp, like "TestGradientAve" corresponds to "TestForwardAve". 

I think my previous code modification shall be safe enough, however add a gradient test would also be a good idea. Would you help check the actual mechanism and sufficiency of "TestGradientAve". 

Thanks,

PR #6306, opened 2018-03-20
add name scope and arg scope for netspec

twmht:
add name scope and arg scope like tensorflow slim (https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/slim).

tensorflow slim has defined `name_scope` and `arg_scope`, this can be done with a similar way in caffe.

and both `name_scope` and `arg_scope`  help organize the network when using netspec.

with `name_scope`, every layer name is renamed to `scope_name / layer_name`, you can have identical layer's names but with different scope names. you can also define the separator you want, for example,  [resent](https://github.com/KaimingHe/deep-residual-networks/blob/master/prototxt/ResNet-50-deploy.prototxt#L179) uses `_` as the separator. 

```python
with name_scope('conv3_1'):
# this layer name would be conv3_1/conv
    L.Convolution(bottom, name='conv', pad_h=3, pad_w=3, kernel_h=7, kernel_w=7, stride_h=2, stride_w=2, num_output=16, bias_term=False)

with name_scope('conv4_1'):
# this layer name would be conv4_1/conv
    L.Convolution(bottom, name='conv', pad_h=3, pad_w=3, kernel_h=7, kernel_w=7, stride_h=2, stride_w=2, num_output=16, bias_term=False)

with name_scope('conv5', sep='_'):
# this layer name would be conv5_conv
    L.Convolution(bottom, name='conv', pad_h=3, pad_w=3, kernel_h=7, kernel_w=7, stride_h=2, stride_w=2, num_output=16, bias_term=False)
```

with `arg_scope`, it helps a lot to define common configs for each layer within a scope.

for example, in batch normalization, we always need to define the learning rate for each parameter to zero.

```python
L.BatchNorm(bottom, param=[dict(lr_mult=0),dict(lr_mult=0),dict(lr_mult=0)], in_place=True)
```

it's annoying if you have to define many of these.

Instead, with the help from `arg_scope`, you have less code to define the layers.

```python
with arg_scope(['BatchNorm'], param=[dict(lr_mult=0),dict(lr_mult=0),dict(lr_mult=0)], in_place=True):
    with arg_scope(['Scale'], param=[dict(lr_mult=0.1, decay_mult=0), dict(lr_mult=0.1, decay_mult=0)], bias_term=True, in_place=True):
        with arg_scope(['Convolution'], param=[dict(lr_mult=0.1, decay_mult=0.1)]):
# arg_scope works like defalut configs
# you can override default configs if you want
    L.BatchNorm(bottom, name="bn", in_place=False)
```

### DIFF
--- a/python/caffe/__init__.py
+++ b/python/caffe/__init__.py
@@ -5,4 +5,4 @@ from .proto.caffe_pb2 import TRAIN, TEST
 from .classifier import Classifier
 from .detector import Detector
 from . import io
-from .net_spec import layers, params, NetSpec, to_proto
+from .net_spec import layers, params, NetSpec, to_proto, name_scope, arg_scope
--- a/python/caffe/net_spec.py
+++ b/python/caffe/net_spec.py
@@ -302,21 +302,14 @@ class arg_scope(object):
         for layer_type in layer_types:
             self.layer_fns.append(getattr(layers, layer_type))
         self.layer_types = layer_types
-        self.check_args(kwargs)
         self.args = kwargs
-        if 'param' in kwargs:
-            assert len(self.layer_fns) == 1
-
-    def check_args(self, kwargs):
-        for layer_type in self.layer_types:
-            for key, value in six.iteritems(kwargs):
-                if key in ('param', 'in_place'):
-                    continue
-                getattr(getattr(caffe_pb2, layer_type + 'Parameter')(), key)
+        self.old = {}
 
     def __enter__(self):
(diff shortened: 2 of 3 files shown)


naibaf7:
@Noiredd This certainly looks worth investigating.

twmht:
I am writing the test case for `name_scope` and `arg_scope`. And this would help review more quickly.

Noiredd:
Yes it does. I'll take a look soon.

PR #6324, opened 2018-04-02
Enhance SoftmaxWithLoss by adding select_label.

SOLARleisu:
Selecting only a specified class for backprop using SoftmaxWithLoss.

1. Add an new variable for softmax_loss_layer to allow the user to give
   different weights to different classes (or labels).
   use select_label like:
   loss_param {
     select_label: 0
     }
2. Tested with examples/mnist/train_lenet.sh for both GPU mode and CPU
   mode.
3. Use examples/mnist/lenet_train_test.prototxt as an example for
   select_label (has been commented out).
4. Note: ignore_label gets the priority if select_label and ignore_label
   were in conflict.

### DIFF
--- a/examples/mnist/lenet_train_test.prototxt
+++ b/examples/mnist/lenet_train_test.prototxt
@@ -166,3 +166,28 @@ layer {
   bottom: "label"
   top: "loss"
 }
+# Test code for select_label
+#layer {
+#  name: "loss1"
+#  type: "SoftmaxWithLoss"
+#  bottom: "ip2"
+#  bottom: "label"
+#  top: "loss1"
+#  loss_param {
+#      ignore_label: 1
+#      #select_label: 1
+#  }
+#  loss_weight: 1
+#}
+#
+#layer {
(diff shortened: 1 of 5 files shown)


SOLARleisu:
@Noiredd 
Thank you for you advice for "Selecting only a specified class for backprop using SoftmaxWithLoss #6253".
I push this new request for the same work.

SOLARleisu:
@Noiredd 
I finally choose "select_label_" in the c++ code and use "select_label" in protobuffer file.

Noiredd:
This looks good so far. Have you considered adding some tests to this feature? See [TestForwardIgnoreLabel](https://github.com/BVLC/caffe/blob/300f43f3ae6347ac8e01093f9a57ee99e551ed74/src/caffe/test/test_softmax_with_loss_layer.cpp#L64) for reference.

SOLARleisu:
@Noiredd 
Ok. I will give the tests related code. Your advice is really useful.
Thank you very much.

PR #4914, opened 2016-10-28
Add Github issue template to curb misuse.

williford:
This merge adds a file ".github/ISSUE_TEMPLATE.md" which will appear in the text box when someone starts to create an issue. This is an effort to reduce the issues that do not follow the guidelines for contributing. You can see what this looks like by going to:
https://github.com/williford/segmentation-caffe/issues/new

Unfortunately, Github doesn't allow us to change the message "Please review the guidelines for contributing to this repository" in order to make the guidelines more apparent (ideally, I think this should show the guidelines).

The file could also be placed in the root directory, but then clutters that directory. I'm happy to modify the message as desired.

For information on Github issue templates, see:
https://github.com/blog/2111-issue-and-pull-request-templates

PR #6327, opened 2018-04-03
Revised guidelines for GitHub issues

Noiredd:
Inspired by @williford's idea for a [submission checklist](https://github.com/BVLC/caffe/pull/4914#issuecomment-256912734), I propose the following changes to ISSUE_TEMPLATE and CONTRIBUTING:
* Default template for issues will contain a checklist with issue requirements; the user's responsibility will be to make sure every position on the list is ticked before they submit the issue. Hopefully this will result in issues of higher quality and easier to solve.
* The list of requirements for issues is moved entirely to contributing guidelines (CONTRIBUTING). The issue template will only redirect the user to this file, while the first position on the checklist requires them to read it.
* One new requirement is that issues do not contain overly long pieces of code, config files, or logs; these should be attached instead. It is somewhat opinion-based, as *for me* it is uncomfortable to scroll through a long post (I prefer to open the relevant files on another monitor), so feel free to discuss that.

I intended the checklist to be intuitive more than specific, so that we don't have to start with a strict enforcement (something like "make sure to go through the following list as issues without all the ticks will be dismissed"), so any suggestions about wording are welcome.

Another thing I consider in addition to the above is a new header in the template: "Tried solutions". Some issues are similar to others, and this would at least (attempt to) prompt users to do some research, hopefully making the work easier for us.

Note to self: add the requirement to include which fork and version of Caffe is being used.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,19 +1,31 @@
-Please use the [caffe-users list](https://groups.google.com/forum/#!forum/caffe-users) for usage, installation, or modeling questions, or other requests for help.
-_Do not post such requests to Issues._ Doing so interferes with the development of Caffe.
+## Important - read before submitting
 
-Please read the [guidelines for contributing](https://github.com/BVLC/caffe/blob/master/CONTRIBUTING.md) before submitting this issue.
+*Please read the [guidelines for contributing](https://github.com/BVLC/caffe/blob/master/CONTRIBUTING.md) before submitting this issue!*
+
+*Please do not post installation, build, usage, or modeling questions, or other requests for help to Issues.*
+Use the [caffe-users list](https://groups.google.com/forum/#!forum/caffe-users) instead.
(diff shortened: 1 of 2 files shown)


Noiredd:
Side note - the filler tests seem to randomly fail. I swear this PR doesn't touch them ;)

`[  FAILED  ] GaussianFillerTest/1.TestFill2D, where TypeParam = double`

Maybe the tolerance should be upped a notch or two.

Noiredd:
I've added some additional guidelines for user list posts and attempted to consolidate links to places where users can look for solutions to their problems. Hopefully this will make it easier for users to make use of the existing knowledge, and increase the quality of questions on the users list as well.  
The "tried solutions" check for issue template was also added.

Let me know your opinions, or approval if this is good to merge as is.

shelhamer:
> Note to self: add the requirement to include which fork and version of Caffe is being used.

This is worth including since it is sometimes overlooked even though it is usually obvious to diagnose as the poster will reference a layer/solver/etc. that does not exist in BVLC/caffe:master. Thanks for adding a note about this in the guidelines.

PR #6330, opened 2018-04-04
[BugFix for OpenCL branch] Init two Caffe class at the same time will cause segment fault crash

dzyong:
Suppose there are two threads init Caffe class at the same time, the first thread will init the global caffe instance, but if the second thread init Caffe before global caffe init finish, it will use a NULL global caffe instance to init itself and cause a segment fault crash.

### DIFF
--- a/src/caffe/common.cpp
+++ b/src/caffe/common.cpp
@@ -96,7 +96,7 @@ std::vector< shared_ptr<device> > Caffe::devices_;
 
 Caffe& Caffe::Get() {
   instance_mutex_.lock();
-  if (NULL == global_instance_) {
+  if (global_instance_ == nullptr) {
     // The first call must be single threaded
     // and defines the global instance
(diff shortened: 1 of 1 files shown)


naibaf7:
Are you sure sleep is a good solution here? I haven't observed the issue yet, but wouldn't a locking mechanism (if required) work better?

dzyong:
Locking mechanism and sleep are both OK here. But I think sleep is simpler.

dzyong:
Hi @naibaf7 , this code does exist a bug. Do you think if it is a good idea to use mutex here?

If so, I would like to use mutex to replace the original `std::atomic<bool> first`, they have something in common.

dzyong:
OK, I will fix it before this Sunday.

dzyong:
Hi @naibaf7 , I uploaded the new solution. Please help me to review if this one is OK, thanks.

PR #6338, opened 2018-04-09
Fix cmake < v3.7 compatibility in Cuda.cmake

vrichter:
string(COMPARE GREATER_EQUAL is a cmake-3.7+ feature see 37e4289024d80632e2c721e865c14be63aab9d8e


Noiredd:
One thing I didn't notice then. Variable name `iscudanewerthan90` is actually misleading, since it is set to true when CUDA is actually *older* than 9.0, and false if it's newer. Would you mind changing the name?

Noiredd:
Thank you!

PR #6364, opened 2018-04-23
fix bug when adding the extra top blob in SoftmaxWithLoss

twmht:
When adding extra top for computing softmax prob, it will throw the exceptions.

```
F0423 15:09:26.128675 31870 layer.hpp:392] Check failed: top.size() == num_loss_weights (1 vs. 2) loss_weight must be unspecified or specified once per top blob.
```

this PR solves this issue by removing the weight in the internal softmax layer.


TonightGo:
I got the similar problem when I add a top blob on the multibox-loss layer in caffe-ssd. 
"Check failed: top.size() == num_loss_weights (2 vs. 1) loss_weight must be unspecified or specified once per top blob."

PR #6371, opened 2018-04-26
COMP: Fix using find_package(Caffe) missed build tree include path

hjmjohnson:
When building caffe outside of the source tree with cmake
(i.e. mkdir caffe-bld; cd caffe-bld; cmake ../caffe ) the
resulting IMPORTED target was missing the include path
caffe-bld/include/caffe/proto/caffe.pb.h

which prevented '#include "caffe/proto/caffe.pb.h"'
from being found when building tools that depend on caffe.


hjmjohnson:
@pluskid, @sergeyk, @Franck-Dernoncourt , @darxriggs  Is there a problem with this PR?  Can it be included? 

If not, please close.

hjmjohnson:
This project seems unresponsive to accepting changes.

PR #6399, opened 2018-05-15
concat empty blob

twmht:
It's necessary that Concat layers should be able to deal with empty blobs where the `concat_axis` is zero.

For example, when concat `blob(2, 5)` and `blob(0, 5)`, it should produce `blob(2,5)`. And it should never back-propagate the empty blob.

it's possible that the input blob is empty, for example, when implementing FPN (https://arxiv.org/abs/1612.03144), some of the roi pooling layers may produce empty rois. And caffe should be able to concat all of the rois to feed into the next fully connected layers.


makefile:
Maybe add an option like `allow_empty` and set the default value to false will be better.

twmht:
@makefile 

thanks for providing great advice.

I will update this PR soon.

PR #6490, opened 2018-08-04
Silence deprecation warn

henczati:
Handle CMP0022: INTERFACE_LINK_LIBRARIES vs LINK_INTERFACE_LIBRARIES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,7 @@
 cmake_minimum_required(VERSION 2.8.7)
+if(POLICY CMP0022)
+  cmake_policy(SET CMP0022 NEW)
+endif()
 if(POLICY CMP0046)
   cmake_policy(SET CMP0046 NEW)
 endif()


Noiredd:
This policy only silences the warning and this is some approach, but maybe we could figure out why does it even come up? The warning says `Target "caffe" has an INTERFACE_LINK_LIBRARIES property which differs from its LINK_INTERFACE_LIBRARIES properties.` I checked and the difference is about `libprotobuf.so`.  
INTERFACE_LINK_LIBRARIES:  
`$<$<NOT:$<CONFIG:DEBUG>>:/usr/local/lib/libprotobuf.so>`  
`$<$<CONFIG:DEBUG>:/usr/local/lib/libprotobuf.so>`  
LINK_INTERFACE_LIBRARIES:
`/usr/local/lib/libprotobuf.so`

Why does the difference arise in the first place?

henczati:
Don't have time to attend to this now, but my impression at the time was that either both variables were kept for backward compatibility and the old one was improperly maintained or the old one was left in the code by accident.

PR #6499, opened 2018-08-14
python: Set gpu device id before setting gpu mode

xerus:
Otherwise caffe allocates some memory on GPU#0


naibaf7:
This might be a problem on the OpenCL branch, where the order is required to be GPU mode first, select device second.

Issue #6294, opened 2018-03-15
Add a Padding layer to Caffe


twerdster:
We have developed a padding layer for Caffe similar in functionality to Keras, CoreML and others. The implementation is currently for GPU (Cuda) and CPU and has been working without issue for around a year. Is this likely to be accepted as a PR? If it is then we can make the effort to add the necessary layer tests.

This is what the proto message looks like.

`message PaddingParameter {
  enum PadMethod {
    CONSTANT = 0; // constant border fill, can choose what value
    REPEAT = 1;   // repeat the values along the border (i.e. CLAMP)
    CLAMP = 2;    // same as REPEAT
    REFLECT = 3;  // mirror about the edge
    WRAP = 4;     // wrap the values around from the other side
  }
  optional PadMethod pad_method = 1 [default = REPEAT]; // The padding method
  // Explicit padding for every side is given
  // For reference: top left corner is the origin
  optional uint32 pad_l = 2 [default = 0]; // number of columns to add to the left
  optional uint32 pad_r = 3 [default = 0]; // number of columns to add to the right
  optional uint32 pad_t = 4 [default = 0]; // number of rows to add to the top 
  optional uint32 pad_b = 5 [default = 0]; // number of rows to add to the bottom
  optional float pad_value = 6 [default = 0]; // if pad_method is CONSTANT then use this value

  enum Engine {
    DEFAULT = 0;
    CAFFE = 1;
  }
  optional Engine engine = 7 [default = DEFAULT];
}`

twmht:
@twerdster 

when do we need padding layer?

twerdster:
Style transfer filters such as [fast-neural-style](https://github.com/jcjohnson/fast-neural-style/blob/master/fast_neural_style/models.lua)  can make use of it.

Also, the use of symmetric zero padding as the only option available seems to be an unnecessary limitation. Many image processing algorithms require non-zero boundary conditions. 

Here are paddings provided in some other frameworks:
[Torch](https://github.com/torch/nn/blob/master/doc/convolution.md#nn.SpatialZeroPadding)
[Tensorflow](https://www.tensorflow.org/api_docs/python/tf/pad) 
[Keras](https://keras.io/layers/convolutional/#conv2d)
[CoreML](https://apple.github.io/coremltools/coremlspecification/sections/NeuralNetwork.html#validpadding)
[PyTorch](http://pytorch.org/docs/master/_modules/torch/nn/modules/padding.html)

twmht:
[...]

PR #6506, opened 2018-08-21
Pad layer

17cspence:
Pad layer does what it says: Adds padding to the boundaries of an image. Options allow for zero padding (what conv layers effectively do), replicating the pixel on the edge, and two kinds of reflective padding.


17cspence:
Addresses issue #6294, though I hadn't noticed it when I submitted this request.

HangJie720:
@17cspence, Thank you for your code, I find performance of PadForwardPadZero is not efficient, I try to use it to training resnet with cuda(with kernel config is <<<16, 512>>>), profile result is shown as follows, 


I suggest that If PadForward and PadForwardPadZero kernel function can be merged avoid wasting thread resource, or more efficient pad implementation?

PR #6508, opened 2018-08-22
Implement ResizeBilinear layer in caffe(equal to tf.image.resize_bilinear operation)

ujsyehao:
Hi, when I convert mobilenet/mobilenet v2 tensorflow model to caffe model, it has the error that caffe does not support the Resize_Bilinear operation, so I implement it and test it successfully. 
However, the ResizeBilinear layer now not supports for backpropagation, I will implement it if anyone thinks it is necessary.


twmht:
it's much better if you provide the backward pass.

ujsyehao:
Okay, @twmht 
Can it be accepted as a PR?

twmht:
Not sure. I have no authority to accept. But I think the original way in Deconvolution is too slow.

PR #6513, opened 2018-08-26
add CUDA block limitation for CAFFE_GET_BLOCKS

wkcn:
There is a limitation for the number of CUDA blocks `KERNEL<<<CAFFE_GET_BLOCKS(N), CAFFE_CUDA_NUM_THREADS>>>`.
In the previous implementation, when N is extremely larger than the limitation (e.g. N > 65535 * 512), the kernel will not launch.

So I add a limitation for the number of CUDA block,

### DIFF
--- a/include/caffe/util/device_alternate.hpp
+++ b/include/caffe/util/device_alternate.hpp
@@ -83,10 +83,13 @@ const char* curandGetErrorString(curandStatus_t error);
 
 // CUDA: use 512 threads per block
 const int CAFFE_CUDA_NUM_THREADS = 512;
+// CUDA: The maximum number of CUDA Block x
+const int CAFFE_CUDA_MAX_BLOCK_NUM = 65535;
 
 // CUDA: number of blocks for threads.
 inline int CAFFE_GET_BLOCKS(const int N) {
-  return (N + CAFFE_CUDA_NUM_THREADS - 1) / CAFFE_CUDA_NUM_THREADS;
+  const int num_blocks = (N + CAFFE_CUDA_NUM_THREADS - 1) / CAFFE_CUDA_NUM_THREADS;
+  return num_blocks < CAFFE_CUDA_MAX_BLOCK_NUM ? num_blocks : CAFFE_CUDA_MAX_BLOCK_NUM;
 }
 
 }  // namespace caffe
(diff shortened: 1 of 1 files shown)


Noiredd:
I don't think this is the right way to handle this problem. AFAIR, forward and backward functions compute the number of kernels to launch basing on the blob size. If we put `std::min` in `CAFFE_GET_BLOCKS` two things might happen: either a sufficient number of kernels will be launched, *or* the maximum allowed number. Now correct me if I'm wrong, but won't this mean that for a big enough blob, this will cause silently ignoring a portion of it?  
Personally, I would prefer the framework to give me an explicit message that something can't be done, rather than execute *as much as it can* and silently fail, leaving some data unprocessed.

wkcn:
Thanks.
If we put `std::min` in `CAFFE_GET_BLOCKS`, it will not leave some data unprocesses.
Here is the implementation of [CUDA_KERNEL_LOOP](https://github.com/BVLC/caffe/blob/master/include/caffe/util/device_alternate.hpp#L70).
```c++
#define CUDA_KERNEL_LOOP(i, n) \
  for (int i = blockIdx.x * blockDim.x + threadIdx.x; \
       i < (n); \
       i += blockDim.x * gridDim.x)
```
`blockDim.x * gridDim.x` is equal to the number of kernels, I think all data will be processed even though the number of kernels reaches the maximum allowed number.

The following projects all limit the maximum allowed number of CUDA kernels.
- [Caffe2 core/common_gpu.h#L270](https://github.com/pytorch/pytorch/blob/master/caffe2/core/common_gpu.h#L270)
```c++
inline int CAFFE_GET_BLOCKS(const int N) {
  return std::max(
      std::min(
          (N + CAFFE_CUDA_NUM_THREADS - 1) / CAFFE_CUDA_NUM_THREADS,
          CAFFE_MAXIMUM_NUM_BLOCKS),
      // Use at least 1 block, since CUDA does not allow empty block
      1);
}
```

- [MXNet src/operator/mxnet_op.h#L75](https://github.com/apache/incubator-mxnet/blob/master/src/operator/mxnet_op.h#L75)
```c++
inline int cuda_get_num_blocks(const int N) {
  using namespace mshadow::cuda;
  return std::min(kMaxGridNum, (N + kBaseThreadNum - 1) / kBaseThreadNum);
}
```

- [Tensorflow core/util/cuda_launch_config.h#L134](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/cuda_launch_config.h#L134)
```c++
  const int block_count =
      std::min(DivUp(physical_thread_count, thread_per_block),
               d.getNumCudaMultiProcessors());
```

However, PyTorch doesn't have the limitation.
I think it's better to implement the limitation like Tensorflow.

Issue #5729, opened 2017-06-30
cuDNN bug in Caffe with "group" in conv layer (misaligned address)


svobora:
### Issue summary

Using "group" parameter in any convolution layer, with CUDNN, I get "misaligned address" error when the training phase starts. The (first?) test phase is not affected. The error disappears when I build caffe with CUDA but without CUDNN. However such a training is 2x slower...

### Steps to reproduce

Checkout repo, build with CUDNN, use "group" parameter of Convolution layer in some net and run training.   

### Your system configuration
Operating system: Ubuntu 16.04
Compiler: gcc 5.4
CUDA version (if applicable): 8
CUDNN version (if applicable): 5.1
BLAS: open

deepali-c:
Could you please post the error log?

svobora:
F0630 15:37:53.939421 12138 benchmark.cpp:92] Check failed: error == cudaSuccess (74 vs. 0)  misaligned address
*** Check failure stack trace: ***
F0630 15:37:53.939426 12256 math_functions.cu:79] Check failed: error == cudaSuccess (74 vs. 0)  misaligned address
*** Check failure stack trace: ***
    @     0x7f71ff1c55cd  google::LogMessage::Fail()
    @     0x7f71ff1c55cd  google::LogMessage::Fail()
    @     0x7f71ff1c7433  google::LogMessage::SendToLog()
    @     0x7f71ff1c7433  google::LogMessage::SendToLog()
    @     0x7f71ff1c515b  google::LogMessage::Flush()
    @     0x7f71ff1c515b  google::LogMessage::Flush()
    @     0x7f71ff1c7e1e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7f71ff1c7e1e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7f71ff8101da  caffe::Timer::MilliSeconds()
    @     0x7f71ff9bdc0a  caffe::caffe_gpu_memcpy()
    @     0x7f71ff82eb1d  caffe::SyncedMemory::mutable_cpu_data()
    @     0x7f71ff80f73a  caffe::Timer::Seconds()
    @     0x7f71ff8161f2  caffe::Blob<>::mutable_cpu_data()
    @     0x7f71ff98e85d  caffe::Solver<>::Step()
    @     0x7f71ff916144  caffe::ImageDataLayer<>::load_batch()
    @     0x7f71ff98f26a  caffe::Solver<>::Solve()
[...]

PR #6548, opened 2018-09-25
Fixed alingment issue "cudaSuccess (74 vs. 0) misaligned address"

kostinalexey:
In some situations it is possible to have max_workspace to be not multiple of 16. This will lead to at least workspace[1] to be unaligned. Resulting later in "cudaSuccess (74 vs. 0) misaligned address" error.

I am not sure if we must have alignment to 16 bytes (in my case the fix works with alignment to 8 bytes). But I believe using extra 8 bytes in this rare case to be a bit future proof is justified.


kostinalexey:
This will also fix issue #5729

PR #6555, opened 2018-10-05
Add FLAG -std=c++11 in CMakeLists.txt makes cmake work

kklt007:


deepankar1994:
I second this. I required this as well for building via CMake.

PR #6590, opened 2018-10-31
dataframe sort api is not available

zhcf:
dataframe sort api is not available, should use the sort_values to replace. Previously, sort and sort_values are all supported by dataframe, but in the recent several versions, sort is not provided anymore.


zhcf:
Do I need unittest for my code change?

Issue #6600, opened 2018-11-06
Caffe Python 3 can dump core when processing NCCL uid


dncliss:
Support for NCCL with Python3 was added in part via https://github.com/BVLC/caffe/pull/5527
However, there is a line which 
```  // Since boost 1.53 boost.python will convert str and bytes
  // to std::string but will convert std::string to str. Here we
  // force a bytes object to be returned. When this object
  // is passed back to the NCCL constructor boost.python will
  // correctly convert the bytes to std::string automatically
  PyObject* py_uid = PyBytes_FromString(uid.c_str());
return bp::object(bp::handle<>(py_uid));
```

The use of `PyBytes_FromString` can cause an error down the line later if there is a null within the uid.c_str(), such that it ends up truncating the string at that point and assigning it (the truncation) to py_uid.
In my case, I've seen this result in a core dump.
This can be corrected by setting the UID to explicitly use the full defined length of the uid by using `PyBytes_FromStringAndSize`.  In fact, I've tested the following as a fix and it resolves the issue:
`PyObject* py_uid = PyBytes_FromStringAndSize(uid.data(), uid.size());`

*Operating system: in my case was Ubuntu 18.04, but it's not an OS-specific issue
*Python version:  Python 3.6   (code issue is specific to Python 3).

PR #6601, opened 2018-11-06
switch UID conversion from PyBytes_FromString to PyBytes_FromStringAn…

dncliss:
Proposed patch for https://github.com/BVLC/caffe/issues/6600

A one-liner.
This addresses the case of a null-truncated UID.


dncliss:
I realize the contribution guide indicates that any PR "must include tests"; the actual incidence of demonstrating the failure scenario with vs. without this patch is not conducive to any direct unit or functional test that I can effectively drum up, unfortunately.  I've observed this patch in action in a Ubuntu 18.04 environment running the provided default 'train.py' script which completed successfully under the environment where it had failed prior to the fix.  Hopefully, the explanation of the issue coupled with code inspection makes it reasonable enough for review.

Issue #6623, opened 2018-11-26
Windows support - Suggestion


danielgindi:
### Issue summary
Windows support currently is based on a different branch, which has to be synchronized all the time.
The problem with that, aside from synchronizing the code, is that it is not really cross platform. If you're using Caffe as a submodule in a crossplatform project - you would have to have two copies of the submodule, one on `master` and one on `windows`.

What I'm suggesting is making `master` cross platform. Compilable on windows too.
If someone needs Visual Studio projects, that's a different issue and could be generated from the cmake or manually constructed (by just dragging in the code)... It's very easy when pulling dependencies through `Conan` package manager.

I actually have the pieces of code for making it cross platform ready. These do not require almost any pollution of the code with macros. They basically provide tiny polyfills for POSIX code.
I would like the maintainers of this project to respond, whether they would like to merge such a PR, or maybe they have a fundamental objection and want to keep it separate.

### Issue checklist

- [X] read the guidelines and removed the first paragraph
- [X] written a short summary and detailed steps to reproduce
- [X] explained how solutions to related problems failed (tick if found none)
- [ ] filled system configuration
- [ ] attached relevant logs/config files (tick if not applicable)

willyd:
@danielgindi Thank you for your suggestion. I think this is inline with @naibaf7 suggestion in https://github.com/BVLC/caffe/pull/6427.

I have way less time to maintain the windows branch than I used to have. Having the windows branch merged in master would relieve me from having to merge master into the windows branch and make everyone happy.

I have some unmerged changes that change the dependency management from hand-built dependencies to dependencies built with vcpkg. Conan is probably a good choice too. I also have some scripts to create conda packages for those who want to use caffe without building it.

I'd gladly accept your help to make windows support part of the master branch.

danielgindi:
It's great to see that you are aboard with me on this :-)

Okay so I have a preliminary PR - which is the core of the request - already ready and set. It's #6624.
It takes a different approach than the `windows` branch, in that it does *not* require you to "maintain" it, you don't have to put `_WIN32` or `_MSC_VER` macros everywhere and god forbid if you forgot them. It has a **very** small set of compatibility headers to match unix/linux/windows behavior on some platform-dependent features.

What it includes is the actual possibility to compile with MSVC.
At this point - you have to make adjustments to your CMake environment, or just drag all Caffe files into a fresh MSVC project - and it compiles perfectly.

[...]

PR #6624, opened 2018-11-26
Embedded support for compilation on Windows

danielgindi:
Relates to #6623 

Here I have embedded very simple and light additions/modifications to the code for being able to compile on Windows.

This does not include orchestration of dependencies, protobuf etc. As we do it with other tools, mainly Conan.
But as long as you drag all the sources to an MSVC project, and set the include folders and required preprocessor flags - you're good.

It works, tested on a heavily tested project.

This might also be the cornerstone for a crossplatform Conan package.


danielgindi:
Did some modifications to satisfy the very unstylish linter

Issue #6530, opened 2018-09-09
error "OpenCV 4.x+ requires enabled C++11 support"


zhangkom:
when make all, the error happened as following:

/usr/local/include/opencv2/core/cvdef.h:440:4: error: "OpenCV 4.x+ requires enabled C++11 support"
#  error "OpenCV 4.x+ requires enabled C++11 support"

Laser-Cat:
change Makefile line 420(if you are using the newest caffe)
CXXFLAGS += -MMD -MP
to 
CXXFLAGS += -MMD -MP -std=c++11

PR #6625, opened 2018-11-26
Fix build with OpenCV 4.0

dbermond:


comicchang:
please also change `CV_LOAD_IMAGE_COLOR` to `IMREAD_COLOR` and `CV_LOAD_IMAGE_GRAYSCALE` to `IMREAD_GRAYSCALE` in these files:

https://github.com/BVLC/caffe/blob/2a1c552b66f026c7508d390b526f2495ed3be594/src/caffe/layers/window_data_layer.cpp#L293
https://github.com/BVLC/caffe/blob/2a1c552b66f026c7508d390b526f2495ed3be594/src/caffe/test/test_io.cpp#L23
... and many more

ref: https://github.com/opencv/opencv/blob/a4ab60920f4c09a20b18052ec7e54561d2eb42e5/modules/imgcodecs/include/opencv2/imgcodecs/legacy/constants_c.h#L8

xerus:
Thanks, it fixed the build issue for me

jinfagang:
@dbbert @xerus  Seems not all opencv4 can be found for example I built from source install make install:

```
 pkg-config opencv4 --libs
Package opencv4 was not found in the pkg-config search path.
Perhaps you should add the directory containing `opencv4.pc'
to the PKG_CONFIG_PATH environment variable
No package 'opencv4' found

```

xerus:
that Makefile part should be removed from the pull request

jinfagang:
@xerus I can not built with opencv4, I am using CMake, it can found opencv4, but just can not link it . always undefined cv::xx  error. My other simple programs which using opencv4 just work fine

aPonza:
After applying this PR I was still facing `error: #error "OpenCV 4.x+ requires enabled C++11 support"` which I solved [adding CXXFLAG -std=c++11](https://github.com/BVLC/caffe/issues/6530#issuecomment-419715086) which was still missing (guessing for compatibility). This is probably due to me using OpenCV 4.0.1

edrozenberg:
It would be great if this PR could be incorporated into the master b/c everyone running OCV 4 (increasingly, everybody) is going to run into the exact same issue and they'll all start hitting their heads against the wall in synchronicity.

PR #6675, opened 2019-01-21
Improve Vector Performance - Save Reallocation Cost

shahzadlone:
We know the number of elements pushed into the vector will be (1 + bottom[0]->shape().size()). 
So reserve it, to save on reallocation costs.


shahzadlone:
Is there anything from my side I need to provide more?

PR #6728, opened 2019-03-20
Fixed build_win.cmd for newer versions VS, Python

GermanAizek:


Coderx7:
For python3 you have to have all the prerequisites built for python3.6+ 
So in order to add support for python3.6+, you must have the built libraries for python3.6+. 
there are also other files (e.g download_prebuilt_dependencies.py) that need to be changed to address the changes. 
You can find a sample from [my personal repo](https://github.com/Coderx7/Caffe_1.0_Windows/releases) that supports python3.6 and do that for python 3.7

PR #6775, opened 2019-05-28
Fixed running Caffe build with CUDA and CUDNN on hardware without GPU

artyom-beilis:
There is a selection to run caffe with CPU or GPU mode in the API.

It makes it very useful to deploy the library on different hardware allowing user to enable GPU support if available by simple setting CPU or GPU caffe mode.

However in the current version if the library is compiled with CUDNN it would fail to run if Caffe::mode(CPU) is selected. This patch fixes this issue allowing to select GPU or CPU in runtime without requiring the library to be compiled in CPU only mode.

The root cause of the issue is that CUDNN layers allocate different GPU resources on start even if they run in CPU mode. This fix disables CUDNN based layer and switches them to native Caffe layers when working in CPU mode.


artyom-beilis:
Updated pool request for test fix

artyom-beilis:
Hello, are there any chances for this branch to be merged?

PR #6776, opened 2019-05-29
Data Layer/Transform performance improvement for high number of channels

artyom-beilis:
If the data fetched from LMDB has very high number of channels the transformation becomes ineffective and sometimes bottlenecks the training.

I implemented a fix that unrolls transformation loop in case no cropping is performed and no mean file/values are required. It significantly improves the performance of data access when the number of channels is very high.


artyom-beilis:
Fixed style

PR #6835, opened 2019-09-15
fix glog finder

pidhii:
I couldn't build with 'glog' being installed non-systemwide. It seems that you aim that a user can supply GLOG_ROOT_DIR in such case, but you missed an "include"-suffix to resolve headers.

* I built on Linux only


albertjin:
I met similar issue. However, I eventually realized that the root cause is that `HDF5_BUILD_CPP_LIB=ON` is not set while configuring `hdf5`. Adding `PATH_SUFFIXES include` does not help because this is just a hint to narrow down the scope of search.

PR #6855, opened 2019-11-04
cmake 3.12.4 Python2/3 & OpenCV4 

petersktang:
updated CMake build to use FindPython and c++11, changed code to use cv::IMREAD_COLOR for OpenCV4


petersktang:
rework

PR #6921, opened 2020-03-05
Remove unused codes

t-ubukata:
size_, pre_pad_, alpha_, beta_, k_ is defined in the base LRNLayer class.


JamesAshwood07:
Looks good to me

PR #6997, opened 2021-03-26
Cuda.cmake modified for new cudnn version

jiapei100:
cudnn version info is now all stored in cudnn_version.h


naiduvenkatadeepak-ctrl:
Please I have 1000 followers cheyandi sir

Issue #6391, opened 2018-05-08
Incorrect checks in Blob::offset


csukuangfj:
## Code
 - branch: master

## Description
See the code
https://github.com/BVLC/caffe/blob/864520713a4c5ffae7382ced5d34e4cadc608473/include/caffe/blob.hpp#L153-L164

 - `n` should be in the range `[0, num())`
 - `c` is in `[0, channels())`
 - `h` is in `[0, height())`
 - `w` is in `[0,width()))`

Chessing234:
Opened #7125 — the 4D `offset(n,...)` overload should use `CHECK_LT(n, num())` like the vector overload, so `n == num()` can't index past the buffer.

PR #7107, opened 2026-04-20
Fix Blob::offset lower-bound checks guarding the dim instead of the index

Chessing234:
Issue #6391 (complements #7102, which fixes the upper bounds).

## Bug

In `include/caffe/blob.hpp`, the four-argument `Blob::offset(n, c, h, w)` writes its lower-bound guards as:

```cpp
CHECK_GE(n, 0);
CHECK_LE(n, num());
CHECK_GE(channels(), 0);
CHECK_LE(c, channels());
CHECK_GE(height(), 0);
CHECK_LE(h, height());
CHECK_GE(width(), 0);
CHECK_LE(w, width());
```

Only the first `CHECK_GE(n, 0)` actually checks the index. The next three `CHECK_GE` calls check the **dimensions of the blob** (`channels()`, `height()`, `width()`), not the indices `c`, `h`, `w`. For any constructed blob those dims are structurally non-negative, so the asserts are vacuous — a negative `c`/`h`/`w` is accepted and then fed into

```cpp
return ((n * channels() + c) * height() + h) * width() + w;
```

which addresses memory outside the buffer.

## Root cause

Almost certainly a copy-paste slip when this legacy-accessor overload was written. The sibling vector overload immediately below shows the intended pattern:

```cpp
CHECK_GE(indices[i], 0);
CHECK_LT(indices[i], shape(i));
```

— guarding the index, not the dimension.

## Fix

Point the three `CHECK_GE` calls at `c`, `h`, `w` (the actual indices) so the legacy overload enforces the same `[0, dim)` contract as the vector overload:

```diff
-    CHECK_GE(channels(), 0);
+    CHECK_GE(c, 0);
     CHECK_LE(c, channels());
-    CHECK_GE(height(), 0);
+    CHECK_GE(h, 0);
     CHECK_LE(h, height());
-    CHECK_GE(width(), 0);
+    CHECK_GE(w, 0);
     CHECK_LE(w, width());
```

Scope is intentionally limited to the lower-bound variable mistake. Upper-bound hardening (`CHECK_LE` → `CHECK_LT`) is handled separately by #7102; the two PRs touch disjoint lines in the same function and can land in either order without conflict.


dragonQian:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

